Issue #40, opened 2021-05-16
Add support for Linux, maybe publish on Flathub


GabMus:
Hi, I just tried Finamp and I gotta say it's a pretty great way to listen to Jellyfin music. I was wondering if you wanted to add support for Linux, since this is built with Flutter. This way we could use it on our desktop PCs as well as on the new Linux phones.

Publishing on Flathub would be an added value, you can look at [Fluffy chat](https://flathub.org/apps/details/im.fluffychat.Fluffychat) as an example for another Flutter app that has been published on Flathub.

UnicornsOnLSD:
I'd love to create a Linux version, but not all Flutter packages run on all platforms. Many use platform code to use native Android/iOS stuff. Here are all of the packages that are currently incompatible with Linux:

* device_info (Used to give Jellyfin device info)
* package_info (Used to give Jellyfin device info)
* cached_network_image (Image widget that automatically caches image to save bandwidth)
* just_audio/audio_service/audio_session (The packages that handle audio playback)
* flutter_downloader (Handles downloads)
* file_picker (Used to select custom directories)
* permission_handler (Used to request storage permissions for external directories)

The only packages that are really essential there are the audio packages. I could probably create a desktop interface for device_info and package_info. On desktop, caching images isn't so important, so I could use the normal image widget on desktop. Downloading songs isn't so important on desktop, which would remove the need for file_picker too. permission_handler won't be needed on desktop (it might on macOS, but this issue is about Linux). Of course, this still means that the app won't work on desktop in its current state, unless you want a music player that doesn't play music. The audio packages are actually supported on macOS, so I might work on some desktop stuff for a potential macOS release.

GabMus:
As I mentioned in the PR, I'm not a dart/Flutter developer, but maybe one could re-implement those interfaces to work on Linux? Or possibly request Linux support for the original libraries?

UnicornsOnLSD:
Windows/Linux support is being worked on for just_audio. The issue can be viewed at https://github.com/ryanheise/just_audio/issues/103. They're working on using VLC for the backend. For audio_service, a "default audio service" was added since Desktop doesn't rely on platform stuff for background playback (https://github.com/ryanheise/audio_service/issues/609/https://github.com/ryanheise/audio_service/pull/632). audio_session may not be required, since that package is meant for configuring the audio player for pausing during calls and stuff.

[...]

PR #41, opened 2021-05-16
[WIP] Linux+Flatpak build

GabMus:
Related to #40, I added the first files to make this buildable on Linux using Flatpak. Most of what I added has been copied from Fluffychat since they already have such a setup in place.

To build currently you need to run `./scripts/build_linux.sh`, then build the Flatpak (using gnome-builder is the easier way).

Currently the app launches to an exception: `MissingPluginException(No implementation found for method initialize on channel vn.hunghd/downloader)`. I'm not a Flutter or dart developer, so I'm not sure about how to move on from here.

### DIFF
--- a/dist/flatpak/com.unicornsonlsd.finamp.desktop
+++ b/dist/flatpak/com.unicornsonlsd.finamp.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Name=Finamp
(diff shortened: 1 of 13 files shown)


UnicornsOnLSD:
Thanks for the PR, I explained in #40 why a Linux release won't work right now. The MissingPluginException you're getting is flutter_downloader failing to initialise because there isn't any Linux platform code for it.

PR #55, opened 2021-06-24
Added Customization Options to Hide Tabs

arunk140:
Hey - 

Added a Customization Section in the Settings Page - with Toggles to Hide/Show Tabs.

It's not a good implementation - I am not sure on how to reload a Widget after changing settings.
The app need to be restarted in order to see the changes.

I hope you can help fix this? I'll keep experimenting!

Great app by the way!

** Also added the Finamp Icon in the Drawer Menu


UnicornsOnLSD:
Looks good! Merging now, I'll implement reloading after changing the settings. For stuff like that, I just use Hive listenables. It's probably not the most efficient way of doing things, but it works.

PR #115, opened 2021-08-24
Add pressable queue items

### LINKED ISSUE #74: An ability to navigate a play queue by pressing queue items.
I think this feature is present in standard Jellyfin app and can be used for reference.

Pressing a song in a play queue starts a playback of this song, and after it ends, the next song (after selected one) in queue starts. This way the user can navigate the queue.

tabjy:
This is a rather trivial fix that closes #74. Let me what you think. :)

Note: I only tested this patch on Android, but I didn't touch any platform-dependent code so iOS *should* work too.


UnicornsOnLSD:
Looks great, thanks :D

Issue #105, opened 2021-08-16
Finamp always shows the album artist instead of the track artist


laurinneff:
I think what Jellyfin Web does is the best solution: on the album page it shows the album artist, and it shows the track artists for tracks where it's different from the album artist. On the now playing page, it always shows the track artist

PR #116, opened 2021-08-25
Display all track artists in various places

### LINKED ISSUE #104: Songs with multiple artists only show the first artist
In Jellyfin you can set multiple artists by separating them with a semicolon. Jellyfin Web shows them separated with a slash, and Finamp only shows the first artist

rom4nik:
Fixes #104 and #105. From my testing it appears that Jellyfin always returns a non-empty `artists` list, but I left `albumArtist` as a fallback.

|Album screen|Notification|
|---|---|
|album|notification|
|Now playing|Queue|
|now playing|queue|

### DIFF
--- a/lib/components/AlbumScreen/SongListTile.dart
+++ b/lib/components/AlbumScreen/SongListTile.dart
@@ -82,7 +82,7 @@ class _SongListTileState extends State<SongListTile> {
         },
       ),
       subtitle: Text(widget.isSong
-          ? processArtist(mutableItem.albumArtist)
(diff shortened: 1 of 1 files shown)


Terrance:
I've just started using Finamp and have also bumped into this -- any chance we can get an updated release with this change in?  As it stands, having "Various Artists" in the notification's artist field makes client-side scrobbling to services like Last.fm effectively unusable.

UnicornsOnLSD:
I'm going to put out a new release very soon, I'm working on one of the last major features of the new update now (downloading images)

PR #127, opened 2021-09-11
Add track numbers on album screen

### LINKED ISSUE #80: [Feature Request] Show disc and track number when viewing albums
It would be nice if Finamp would show the disc and track numbers when viewing an album. Right now you are presented a list of all the tracks, but disc and track numbers are not visible. With multi disc albums this can quickly become a little chaotic. For example I want to spin disc 2, now I need to manually look up which track is the start of disc 2. With box sets this becomes a chore very quickly.

You could take the approach of the Jellyfin client (a header Disc 1, then the tracks, header Disc 2, then the tracks, and so on) or - which would be a little cleaner and I think I'd prefer it - add this info before the song titles. Right now the tracklist is composed of [album art thumbnail] [track title]. This could be changed to [album art thumbnail] [disc number] [track number] [track title] in some sort of aesthetically pleasing way. The track titles are in a bold white font, perhaps disc and track number in non-bold light grey would be nice.

ps. Thank you for your work on this great app! I am very happy user.

rom4nik:
With disc numbers (PR hopefully coming soon) this will fix #80.
"Cry" is the first track on second disc of album shown below.

preview

### DIFF
--- a/lib/components/AlbumScreen/SongListTile.dart
+++ b/lib/components/AlbumScreen/SongListTile.dart
@@ -72,7 +72,12 @@ class _SongListTileState extends State<SongListTile> {
           return RichText(
             text: TextSpan(
               children: [
-                if (mutableItem.indexNumber != null && !widget.isSong) TextSpan(
+                // third condition checks if the item is viewed from its album (instead of e.g. a playlist)
+                // same horrible check as in canGoToAlbum in GestureDetector below
+                if (mutableItem.indexNumber != null
+                    && !widget.isSong
+                    && mutableItem.parentId == widget.parentId)
+                  TextSpan(
                     text: mutableItem.indexNumber.toString() + ". ",
                     style: TextStyle(color: Theme.of(context).disabledColor)
                 ),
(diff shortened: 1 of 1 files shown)


UnicornsOnLSD:
Thanks for this! Could this be implemented as a RichText instead of a Row? RichText should play better with stuff like screen readers.

rom4nik:
Done. In the meanwhile I noticed track numbers appear also in playlists - I'll try to fix that.

rom4nik:
Borrowed the "horrible check" from canGoToAlbum located below, not ideal but it works. Let me know what you think.

rom4nik:
Nope, it doesn't work, great testing on my behalf lol. In fact, `mutableItem.parentId == widget.parentId` appears to be true not only in playlists, but also in multi-disc albums. This means it also leaves "Go To Album" menu option enabled there, which - when pressed - goes to an album screen, but showing tracks only on the *selected* CD.
|Go To Album|CD|
|---|---|
|go to album|cd|

Should I create a separate issue?

UnicornsOnLSD:
Could `mutableAlbum.albumId` be used instead of `mutableAlbum.parentId`?

rom4nik:
That helped, thanks!

UnicornsOnLSD:
Looks good, merged :)

PR #129, opened 2021-09-14
Add disc number headers on album screen

rom4nik:
Disc numbers requested in #80 (first suggested style).
This PR add disc numbers as headers between songs. If Jellyfin doesn't return `parentIndexNumber` (assumed to be the disc number: https://github.com/jellyfin/jellyfin-web/blob/38f486339f191af50215532cb86679f53efd0aec/src/controllers/itemDetails/index.js#L1422), the headers won't be displayed.

I have also increased the song title font size a bit - 04912eef0712391e13701d20fc1bd292d600be3c from #127 has decreased it and I haven't noticed it until now.


UnicornsOnLSD:
Looks good, maybe something like a sticky header (https://pub.dev/packages/flutter_sticky_header) could look even better? AlbumScreenContent.dart already uses a CustomScrollView so it shouldn't be too hard to try :)

PR #138, opened 2021-10-09
Add sticky disc number headers on album screen

### LINKED ISSUE #80: [Feature Request] Show disc and track number when viewing albums

rom4nik:
Another PR to try and completely fix #80. I'm not sure if this is the best approach, therefore I'm submitting it (with minor issues) as a draft - let me know what you think.

https://user-images.githubusercontent.com/46846000/136673440-d5f55595-1c3b-4daa-bc93-7fa16fa4663a.mp4

A few more things:
- TODOs in code
- Now playing track isn't highlighted anymore - I'll try to fix that if code overall looks OK.
- Some single disc albums have disc numbers in tracks' metadata, some don't. It was a bit confusing to sometimes see the header and sometimes don't, so I opted to hide the headers for all single disc albums. Should they perhaps be shown for all albums instead (assuming disc number = 1 for those that don't have it in metadata)? FWIW Jellyfin web client shows the headers always when the numbers are present.
- `SliverList` is pretty much the same inside and outside `SliverStickyHeader`, that seems like duplicate code to me. I suppose it could be easily fixed by forcing all albums to have at least one disc header and removing `SliverList` outside.

### DIFF
--- a/lib/components/AlbumScreen/SongListTile.dart
+++ b/lib/components/AlbumScreen/SongListTile.dart
@@ -86,7 +86,7 @@ class _SongListTileState extends State<SongListTile> {
                 style: TextStyle(
                   color:
                       snapshot.data?.extras?["itemJson"]["Id"] == mutableItem.id &&
-                              snapshot.data?.extras?["itemJson"]["ParentId"] ==
+                              snapshot.data?.extras?["itemJson"]["AlbumId"] ==
                                   widget.parentId
                           ? Theme.of(context).colorScheme.secondary
                           : null,


UnicornsOnLSD:
Code looks good! IMO it doesn't really make sense to show headers when there is only one disc. You could probably extract the SliverList into its own widget for better reusability and so that it isn't duplicated code.

PR #142, opened 2021-10-28
Add "play next" menu option for songs

### LINKED ISSUE #92: [Feature Request] Add an 'Up Next' Queue
In Spotify, when you play from a playlist, album, etc, it begins "Playing From xyz". Then, if you swipe on a song to add it to the queue it adds it to the "Up Next" section. Anything in Up Next gets played before the next item in Playing From.

I'd like to see this Up Next queue added to Finamp, so I can have a playlist running but queue songs to play next as I please.  Perhaps have a swipe left add to Up Next and a swipe right add to the bottom of the queue? (currently both directions add to the bottom of the queue)

rom4nik:
This aims to fix #92, let me know if code looks okay.

https://user-images.githubusercontent.com/46846000/139176753-91bd8e2b-8df8-40eb-8d28-38474d4920d4.mp4

Some things to consider:
- Adding a song this way to a shuffled queue causes the new song to also get shuffled. I don't think that's expected behavior (doesn't happen in Jellyfin web client), so I'm submitting this PR as a draft until I figure out a way to fix it.
- Personally I think it would look better if the menu option had some different icon (Jellyfin web reuses their "Add to play queue" icon), but I haven't found anything better in Material icons.
- Jellyfin web client hides both "Add to play queue" and "Play next" when no media is playing. Would it make sense to do the same for Finamp?
- Maybe a gesture would be handy? There's already swipe left and right to add song to queue, maybe one of those directions could trigger a "play next" instead?


Chaphasilor:
> This approach would be harder to make since it'd need two queues in the player, but it could probably be done by adding something like a "isPlayNext" value to the MediaItem and sorting them in the queue widget. That solution is a bit jank but would be way easier to make. Either way, your solution works and is perfectly fine for release.

The huge benefit of using a separate queue is that his queue could be "immune" to shuffling, so that all songs are always played in the same order in which they were added. This could be challenging when using just a boolean property.  

Anyway, I *really* need that "Play Next" button 😅

rom4nik:
A few rambling thoughts:
- AudioServiceShuffleMode has third option called "group", no idea how to use it so far but maybe it could be helpful
- I could create a custom shuffleOrder to affect only MediaItems with an isPlayNext property, and use one common queue as it is now
  - put the playnext songs on top of queue, shuffle the rest and put it below them, in GUI check if first song in whole list has isPlayNext, if so make a visual distinction that there are two "types" of queue now
  - however, that implies that the answer to question below is true
- Should queue items be removed as they're played? Should that affect both regular and play next queues, or only the latter?
  - I don't see how the latter could work and look well, e.g.:
    - play an album
    - add 2 songs to play next
    - those play next songs are shown somewhere recently played and next album song
    - add another song to play next
    - where should it appear? ugh, it's a mess imo, if those are going be 2 separate queues and the play next queue is supposed to not be affected by shuffle, at the very least I'd suggest removing songs from it as they finish playing
  - it'd be good to make this optional anyway
  - for reference, web client simply inserts the songs after current song queue, and they behave just like that - when you toggle shuffle they get shuffled too, but if you add a song to play next while queue is shuffled already, it's added after current next; however, when you have shuffled queue, add songs to play next and disable shuffle, the newly added songs disappear from queue
- I could probably add "abstract" regular and play next queues somewhere in MusicPlayerBackgroudTask, which would then have to be combined before updating _queue and actual queue going to _player
  - but as far as I understand shuffling happens after queue is updated, so that's pointless, unless we have custom shuffleOrder, but then I don't see any benefit over just specially tagged MediaItems in single queue
  - in audio_service example it's hinted that multiple standalone queues should be implemented by creating multiple AudioHandlers: https://github.com/ryanheise/audio_service/blob/minor/audio_service/example/lib/example_multiple_handlers.dart - too much effort imo, also those queues aren't exactly meant to be standalone

Whatever we settle on, I hope it won't cause problems with persisting the whole queue across app restarts, since that's something I'd like to try implementing next.

Chaphasilor:
> * where should it appear? ugh, it's a mess imo, if those are going be 2 separate queues and the play next queue is supposed to not be affected by shuffle, at the very least I'd suggest removing songs from it as they finish playing

As unintuitive as it might seem, adding a *third* "queue" might solve this issue. So that there is the Play Next queue (always played in order), the regular queue (can be shuffled) and a Recently Played "queue". Just move any item that is played to the bottom of the Recently Played queue and always show it above the currently playing track. Then show the Play Next queue (if it isn't empty), and the regular queue below that.

Mockup:

image

This should make it pretty clear what is going on I think

rom4nik:
That does look good and seem convenient, hopefully I'll figure out how to cleverly implement this in MusicPlayerBackgroundTask without making a mess (one idea that I had is giving player a single-item queue at a time, and handling all 3 queues on our own, but that'd probably be even more complex and cause other issues).

Also, the increasing complexity reminded me that in my Finamp to-do list there's also adding cast receiver/controller functionality. I'm not familiar with how the Jellyfin's builtin casting works across various clients, but I'm concerned that if we put custom queueing in Finamp, we'd have to replicate it when Finamp is controlled by another client, or when it controls e.g. Jellyfin Media Player.

jmshrv:
I like the look of that, would items added to next up go into recently played? In Spotify, once an add-to-queue song is played its forgotten, so going to the last track would go to the last track on the queue.

Using a property like `isPlayNext` is probably the simplest way of doing stuff, it lets us keep one queue in just_audio, I'm not sure if you can concatenate queues but if you can that would probably be easier to understand

Chaphasilor:
> would items added to next up go into recently played?

I'd prefer if they would show up there, yes. After all, the item *has recently been played*. I find Spotify's implementation unintuitive, I've lost quite a few song that way, because I accidentally skipped to the next song and couldn't go back...

> Using a property like isPlayNext is probably the simplest way of doing stuff, it lets us keep one queue in just_audio, I'm not sure if you can concatenate queues but if you can that would probably be easier to understand

Wouldn't it be time for some abstraction at this point? Like implementing a new class that has some methods, like `void addToNextUp()`, `void addToQueue()`, `Track nextTrack()`, `Track previousTrack()`, `Track peekNextTrack()`. Calling `nextTrack()` or `previousTrack()` would skip to the next (previous) track, popping it from the Next Up queue (if it isn't empty, otherwise from the regular queue), returning an object describing the track and adding the track which was playing before to the Recently Played queue.  
While this *does* require a small rewrite of the queuing system I assume, it would make things easier down the road, and could also be implemented as an abstract class with different implementations, so that @rom4nik's cast controller should also be feasible.

jmshrv:
> I'd prefer if they would show up there, yes. After all, the item has recently been played. I find Spotify's implementation unintuitive, I've lost quite a few song that way, because I accidentally skipped to the next song and couldn't go back...

Yeah that makes sense, although what should happen when a queue ends? Finamp currently sets the queue to the start, which is pretty standard for music apps. I guess we could remove play next songs when the queue ends.

> Wouldn't it be time for some abstraction at this point?

Thats a very good idea lol, I should really do things properly sometimes 🙃

just_audio is a very well designed package, so I'd imagine we can extend the current queue logic.

Chaphasilor:
> what should happen when a queue ends? Finamp currently sets the queue to the start, which is pretty standard for music apps. I guess we could remove play next songs when the queue ends.

that's a good point, I don't really listen like this but I see the problem. we could add a boolean property for that, like `removeWhenLooping`. This way, if "loop all" is enabled (the behavior you described with starting the queue over), these songs could be filtered out. And the "recently played" queue should not be cleared by this, but instead either have a maximum length or get cleared with each cold start. You don't want to loose your playback history just because the album is looping, right?

rom4nik:
Another attempt at this is happening in #481, I'm closing my PR since I'm unlikely to work on it anytime soon. :v

PR #143, opened 2021-10-31
Add album name on player screen

rom4nik:
Another small addition. It might seem a bit pointless because there's album art already (hopefully making clear which album the song belongs to), but I was thinking about adding an `onTap` callback later, showing the album screen when tapped. 
Same could be done with artist name, but I'm not sure how this would work when there are multiple artists, some of whom don't have their entry in Artists tab (Album Artists, according to JF web client).

Screenshot_20211031-204829_Finamp

### DIFF
--- a/lib/components/PlayerScreen/SongName.dart
+++ b/lib/components/PlayerScreen/SongName.dart
@@ -4,7 +4,7 @@ import 'package:get_it/get_it.dart';
 
 import '../../services/MusicPlayerBackgroundTask.dart';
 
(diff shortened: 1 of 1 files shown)


UnicornsOnLSD:
This looks good, and I like the onTap idea. Do you want me to merge it now or do you want to add the onTap callback first?

rom4nik:
I wouldn't mind having this merged as is, I'll just try to get onTap working in another PR.

PR #148, opened 2021-11-22
Add library selection button in settings

rom4nik:
This PR allows user to select libraries after initial setup without wiping app data. No further explanation needed, here's a video anyway:

https://user-images.githubusercontent.com/46846000/142790308-d952537e-05de-4458-889f-237b949c6bdc.mp4


UnicornsOnLSD:
Thanks, just made a few changes to actually disable the button when offline mode is enabled, and to make the caption a bit more clear as to why it was disabled. I also applied this to the logout button since it also shouldn't be available when offline.

PR #149, opened 2021-11-22
Skip music library selection if there's only one available during setup

### LINKED ISSUE #42: [Feature Request] Skip library selection when there's only one choice

rom4nik:
Fixes #42.
This might cause issues when user has one `music` library and one or more for audiobooks (of type `books` afaik), as the audiobook library wouldn't get selected. I see three solutions:
- Ignore this, as there's no explicit support for audiobooks in Finamp anyway (yet).
- Start with music library only and make user aware of selector in settings (#148). Maybe mention it in a toast?
- Disable selector skip if any library with type `books` exists.

### DIFF
--- a/lib/screens/ViewSelector.dart
+++ b/lib/screens/ViewSelector.dart
@@ -31,23 +31,7 @@ class _ViewSelectorState extends State<ViewSelector> {
       ),
       floatingActionButton: FloatingActionButton(
         child: const Icon(Icons.check),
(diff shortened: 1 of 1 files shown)


UnicornsOnLSD:
Lookes good, merged!

About the audiobook issue, its probably best to ignore it for now since Finamp doesn't support books just yet. When support is added, we can make a case to show/enable audiobook libraries by default.

PR #192, opened 2022-02-21
Fixed bug where url for images would break if Jellyfin is published w…

nboumalham:
tldr; fix image downloads from urls with a path. Example :  https://www.url.com/jellyfin

If a jellyfin instance is published on a url with a path (example www.website.com/media). images won't download using the current JellyfinAPiData.

I added some code that will get the pathSegments of the parsedBaseUrl and then append the path segments to retrieve the image.

My dart skills are not great. So if you want to disregard this request and fix it yourself writing differently, I won't mind.

(PS:  Full disclosure, I didn't test what happens if the url path segments are empty in the baseUrl)

### DIFF
--- a/lib/services/JellyfinApiData.dart
+++ b/lib/services/JellyfinApiData.dart
@@ -401,17 +401,18 @@ class JellyfinApiData {
 
     if (imageId != null) {
       final parsedBaseUrl = Uri.parse(currentUser!.baseUrl);
-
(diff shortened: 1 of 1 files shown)


nboumalham:
I found the same bug when building the songUri. I also fixed it but don't want to flood you with random pull requests. If this gets validated I'll push the other one.

PR #198, opened 2022-03-07
Add playback progress bar on NowPlayingBar

rom4nik:
Tiny UX thing. I've also added 2 comments in `PlayerScreen/ProgressSlider.dart` to avoid mistaking one slider for another.

https://user-images.githubusercontent.com/46846000/157125480-3d968625-d7b3-412d-a0a9-6fe4ef0eeed3.mp4

### DIFF
--- a/lib/components/NowPlayingBar.dart
+++ b/lib/components/NowPlayingBar.dart
@@ -8,6 +8,7 @@ import '../services/progressStateStream.dart';
 import '../services/FinampSettingsHelper.dart';
 import '../services/processArtist.dart';
 import '../services/MusicPlayerBackgroundTask.dart';
+import 'PlayerScreen/ProgressSlider.dart';
 
 class NowPlayingBar extends StatelessWidget {
   const NowPlayingBar({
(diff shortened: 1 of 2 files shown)


UnicornsOnLSD:
Thanks for this! Just a few things that could be improved:

* There's a lot of code reuse between the progress bar on the NowPlayingBar and the ProgressSlider. Maybe it would make sense to add some parameters so that the ProgressSlider could be reused? Like, pass in a "scrolling enabled" and "thumb visible", both of which are true by default.
* You can kind of see that the slider isn't totally flat on the sides, and IMO it would look better flat. This could be another parameter for the ProgressSlider.
Screenshot 2022-03-07 at 23 54 48
* The playback progress follows the listtile when it is being dismissed, and the slider sort of overhangs the NowPlayingBar. This could probably be fixed by using a Stack with the progress bar and the dismissible.

Hopefully I find some time to finish 0.6, there's quite a bit of work that you've done that hasn't made it out yet. Thanks again :)

Issue #87, opened 2021-07-21
Despite synchronising playback to the server, songs are not scrobbled to last.fm


el-tiuri:
Now, I have no clue if the issue here is with FinAmp or the last.fm plugin. Either way, as the title says, in the admin dashboard I can verify that it is reporting my account playing the song I’m actually playing. However, no music is ever scrobbled despite the last.fm being configured for my account.

It just works with the main Jellyfin app, so I’m wondering if FinAmp might report the tracks being played in a slightly different way or something.

UnicornsOnLSD:
Finamp probably isn't doing the full track reporting properly, I'll look into it.

UnicornsOnLSD:
It was working fine for me, does last.fm work when you play songs via the Jellyfin web interface?

vdatanet:
Yes, when I play using Jellyfin web interface, it scrobbles to last.fm. FinAmp it doesn't, but used to work. I'd swear it worked last week.

el-tiuri:
Same for me. Scrobbling works when using the web interface or the Jellyfin app, but not when using FinAmp.

vdatanet:
Even the songs don’t appear as recently played on Jellyfin’s music recommendations.

deafmute1:
+1 on this, neither the listenbrainz nor the lastfm scrobbler plugins work for me via finamp. Scrobbling  does work for the following clients for me: 
- mopidy
- web UI 
- official desktop client.

Just started using finamp, so can't speak to it being new or old issue.

gerardlouw:
In https://github.com/jesseward/jellyfin-plugin-lastfm/issues/27#issuecomment-744031810 it is mentioned that the plugin expects a `PlaybackStartEvent` and `PlaybackStopEvent` pair to be received by the Jellyfin server for it to register a scrobble. I did a quick code search here and that not find any direct references to these, so I believe that may be what is required for a fix.

UnicornsOnLSD:
Should be fixed in 7d56bdd, I'll close this when someone else can verify that it works

vdatanet:
Hi,

In 0.5.1 scrobbling to last.fm is fixed and played songs are correctly registered in recently played songs in Jellyfin. Thanks!

gerardlouw:
[...]

PR #200, opened 2022-03-13
:bug: Fix playback stop events using an incorrect position

Maxr1998:
As outlined in [my recent comment](https://github.com/UnicornsOnLSD/finamp/issues/87#issuecomment-1064319704) in #87, stop events submitted to the server seem to use the playback progress of the following song that's already playing, which causes scrobbling plugins to drop the listen.

Simply using the song's duration for the stop event works around the issue. Unfortunately, this isn't without problems either, since it'll probably submit a complete listen when skipping or changing a track. Due to my limited experience with the app's architecture, I'm not sure how to fix that.

This PR mainly serves as a first attempt, hoping you or others can develop a better solution.

Note that [the other stop event](https://github.com/Maxr1998/finamp/blob/ea1582e28379b8a9661da080f9affa62e35bd229/lib/services/MusicPlayerBackgroundTask.dart#L143) in the playback task was deliberately not changed, since the issue shouldn't occur here and no played item is supplied in that method.

### DIFF
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -100,6 +100,7 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
           final playbackData = generatePlaybackProgressInfo(
             item: _previousItem,
             includeNowPlayingQueue: true,
+            isStopEvent: true,
           );
 
           if (playbackData != null) {
@@ -373,6 +374,7 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
   PlaybackProgressInfo? generatePlaybackProgressInfo({
     MediaItem? item,
     required bool includeNowPlayingQueue,
+    bool isStopEvent = false,
   }) {
     if (_queueAudioSource.length == 0 && item == null) {
       // This function relies on _queue having items, so we return null if it's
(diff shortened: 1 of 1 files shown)


UnicornsOnLSD:
Thanks for looking into this and finding the root issue! I'll have a look at seeing if we can ignore skips and stuff, I don't blame you for being able to do it yourself, the app's architecture is a bit over the place lol

Maxr1998:
Thanks! I've been testing my change the last few days, and it definitely has a lot of issues. Sometimes re-scrobbles the previous track, scrobbles the first track again when the queue ended, and goes completely crazy when modifying the queue.
Thanks in advance for taking the time to look into it and fix it properly 👍

Maxr1998:
Sorry for nagging, but was there any update on this issue? 😅

jmshrv:
I was wondering if it was ready, I'll test it and merge if it works as intended :)

jmshrv:
Looks like it works as intended, Is this ready for merging then?

Maxr1998:
Well, it causes some invalid scrobbles on its own, especially when moving queue items, skipping incomplete tracks, etc. 

I could make it work for myself in the last few months, but I think for others there should be a better approach..

I don't know a lot about the player library and it's callback handling, so I'm not sure how I could contribute to this.

Chaphasilor:
> Basically, if isStopEvent is true, a full playback will be reported, even though stopping at the start isn't actually a full playback.

Jellyfin's Playback Reporting plugin shouldn't have a problem with that, because it just tracks durations. So I'm guessing only last.fm is problematic? If so, it would be nice if this could be merged, and possibly improved by successive PR.

Y0ngg4n:
@jmshrv  Can this get merged?

Maxr1998:
I think it should be good enough to merge this now, I'll probably try to further improve this later.

jmshrv:
Merged, finally :upside_down_face:

Maxr1998:
Awesome, thanks a ton!

Issue #61, opened 2021-07-11
Shuffle starts on first song


drazil100:
When opening a playlist or album and pressing the big shuffle button at the top it always starts by playing the first song on the list. It would be great if it started on a random song.

UnicornsOnLSD:
This seems to only happen on iOS, so it's probably something to do with the audio package I'm using. I'll look into it.

UnicornsOnLSD:
Should be fixed in 4814a88, I'll test it on iOS in a bit

UnicornsOnLSD:
As I said in #86, 4814a88 didn't work so this issue still isn't fixed.

NotFluffy:
Still happening on android app version 0.5.1

MagicRB:
Yep

drazil100:
Still an issue on iOS at least in version 0.6.0.

If it's not easy/possible to get the tools you are using to generate a random queue that doesn't start with the top song of the album/playlist, you could have it pick a random song to start playing and then enable shuffle mode after the fact like you do when you enable it from the now playing screen. May not be the best or most ideal solution but the now playing screen shuffle button works perfectly fine.

foss-:
Persisting with 0.6.1 on Android and iOS.

UnicornsOnLSD:
This is dependant on https://github.com/ryanheise/just_audio/issues/293

drazil100:
I still suggest as a workaround to manually randomize what the initial index is before entering shuffle mode. It's a hacky workaround but its better than having to hear the same first song repeatedly every time you shuffle a playlist. Also given it's been almost a year since you posted on that issue I don't expect it to be fixed on their end anytime soon.

UnicornsOnLSD:
A better solution would be to skip once, since then you're only missing one song. It's still not a great solution, but it works for now. I could also something like this:

1. Put the first song in twice
2. Skip once
3. Remove the song in the first position

The code for that would look bad but it should be pretty transparent to the user

drazil100:
I don't see a reason to remove it. If you have repeat all enabled it would just become the last song in the shuffle instead of the first. 

I am personally on board with this solution for my use case however for shorter playlists (especially if you drop the first song) I think some people might take issue.

drazil100:
I'll get back to you as soon as it hits my phone. 👍

drazil100:
No luck. Still starts on the first song.

seniorm0ment:
[...]

PR #229, opened 2022-05-09
When shuffle is pressed, shuffling items list before sending it to audioService 

faizzzzz:
This should fix this issue: https://github.com/UnicornsOnLSD/finamp/issues/61

Seems like the audio service only shuffles next songs that are in the queue and not existing ones. Shuffling the list before hand ensures that the first song gets shuffled as well.


UnicornsOnLSD:
Thanks for the PR, while this technically solves the problem, I don't think this is an acceptable solution because if the user disables shuffling, the resulting queue will still be shuffled. This is really an issue with the audio package I'm using, so we'll have to wait for a fix there (or submit a fix ourselves lol). I'll post the relevant issue in #61. Thanks again for your input, I really should have fixed this by now :upside_down_face:

faizzzzz:
Got it, thanks for your input.

PR #232, opened 2022-05-13
Matched sorting behaviour to the one that is in the jellyfin web client.

faizzzzz:
Web Client behavior:
Albums - https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/music/musicalbums.js#L243
Songs - https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/music/songs.js#L157

The Web Client does not have sorting for Artists, Playlists or Genres. The current finamp app provides sorting for all categories.
The Web Client also has different sort options for Albums and Songs. The current finamp app has the same sorting options.
If you want, I can make the behavior exactly as the one in the web client. Let me know in the comments and I'll update this PR accordingly.

### DIFF
--- a/lib/components/LayoutSettingsScreen/ContentGridViewCrossAxisCountListTile.dart
+++ b/lib/components/LayoutSettingsScreen/ContentGridViewCrossAxisCountListTile.dart
@@ -4,11 +4,8 @@ import '../../services/FinampSettingsHelper.dart';
 
 enum ContentGridViewCrossAxisCountType {
   portrait,
-  landscape,
-}
+  landscape;
 
-extension ContentGridViewCrossAxisCountTypeExtension
-    on ContentGridViewCrossAxisCountType {
   /// Human-readable version of the [ContentGridViewCrossAxisCountType]. For
   /// example, toString() on [ContentGridViewCrossAxisCountType.portrait],
   /// toString() would return "ContentGridViewCrossAxisCountType.portrait". With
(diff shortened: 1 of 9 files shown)


UnicornsOnLSD:
This looks great, in my opinion having sorting on playlists is useful, although sorting doesn't seem to work on genres so it may make sense to disable the sort buttons when on the album tab (this is out of scope for this PR though since it'll probably be a bit of a pain to implement, I really need to rewrite MusicScreen and the tab view because it's a mess)

I'm going to move the other enums to Dart 2.18's new thing where you don't need to use extensions, so I'll do that to this PR and then merge it.

faizzzzz:
Ok, sounds good. Let me know if I can help with that. 
By 
> move the other enums
Do you mean, enums other than `SortBy`?

faizzzzz:
ok, I'll check it out and see what I can do.

faizzzzz:
we could also merge this PR and address all the enums in a separate PR

faizzzzz:
I've implemented enhanced enums. Let me know what you think.
Mostly used these as references: 
1. https://dart.dev/guides/language/language-tour#enumerated-types
2. https://medium.com/dartlang/dart-2-17-b216bfc80c5d

We could also deprecate `humanReadableName` and use the overrideable `toString` method provided by enhanced enums.

UnicornsOnLSD:
That's great, I agree that we should replace `humanReadableName` with a `toString()` override

faizzzzz:
Replaced `humanReadableName` with the `toString` method.

UnicornsOnLSD (approved):
LGTM, merging

PR #264, opened 2022-07-06
Tap artist or album name to go to its page in library

rom4nik:
https://user-images.githubusercontent.com/46846000/177649467-d3f34418-1189-4e74-a389-2ec3562d4da8.mp4

Some issues:
- There's no visual indicator that those spots can now be tapped to navigate to artist/album page.
- Janky handling of TextSpans for artists.
- Artist of a song that isn't albumArtist on any album will have empty artist page (last few seconds of video). Web client shows an "appears on" section for such not-album artists, I'm thinking about adding this feature here too:

obraz

### DIFF
--- a/lib/components/AlbumScreen/ItemInfo.dart
+++ b/lib/components/AlbumScreen/ItemInfo.dart
@@ -1,6 +1,9 @@
 import 'package:flutter/material.dart';
+import 'package:get_it/get_it.dart';
 
 import '../../models/JellyfinModels.dart';
+import '../../screens/ArtistScreen.dart';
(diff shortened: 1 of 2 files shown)


UnicornsOnLSD:
This is good, just pushed a fix to stop a null error from occuring when opening the player screen as `mediaItem` is breifly null.

* No visual indicator - I wouldn't worry about it. Other music apps do the same, and making it look like a button would look odd.
* Janky handling of TextSpans - I can't immediately think of anything better, your solution works
* That would be a good feature to add :)

PR #272, opened 2022-07-13
Use last known playback position for ProgressState while paused

rom4nik:
If you pause playback and open player screen or go to album/artist, causing fresh NowPlayingBar to be created, the ProgressState won't emit anything since AudioService.position doesn't emit position unless playback is in progress. This causes the player to have 00:00 position & duration and grayed out slider, and NPB doesn't appear at all (because snapshot.hasData is false).

This PR reuses last known playback position from audioHandler until AudioService emits new position after unpausing, fixing both issues.


jmshrv:
Didn't realise that you made this PR, thanks! Been meaning to look in to this for a while (think it came in audio_service v18)

PR #273, opened 2022-07-13
Add toggleable blurred coverart player background

rom4nik:
A bit of eye candy, disabled by default. 2nd row has dark mode, 3rd row has light mode.

Bright cover|Dark cover|Colorful cover
---|---|---
Screenshot_20220714-011145_Finamp|Screenshot_20220714-011215_Finamp|Screenshot_20220714-011307_Finamp
Screenshot_20220714-011108_Finamp|Screenshot_20220714-011229_Finamp|Screenshot_20220714-011254_Finamp

Some issues:
- occasional low contrast, especially with the gray buffer status slider
- enabling this feature isn't saved to disk, tried to follow https://github.com/hivedb/docs/blob/master/custom-objects/generate_adapter.md but got the white screen about broken config
- bottom bar would probably look better transparent, which additionally would fix contrasting with light theme throughout app

### DIFF
--- a/lib/components/LayoutSettingsScreen/show_cover_as_player_background_selector.dart
+++ b/lib/components/LayoutSettingsScreen/show_cover_as_player_background_selector.dart
@@ -4,8 +4,8 @@ import 'package:hive/hive.dart';
 import '../../models/finamp_models.dart';
 import '../../services/finamp_settings_helper.dart';
 
-class ShowCoverOnPlayerBackgroundSelector extends StatelessWidget {
-  const ShowCoverOnPlayerBackgroundSelector({Key? key}) : super(key: key);
+class ShowCoverAsPlayerBackgroundSelector extends StatelessWidget {
+  const ShowCoverAsPlayerBackgroundSelector({Key? key}) : super(key: key);
 
   @override
   Widget build(BuildContext context) {
@@ -16,9 +16,9 @@ class ShowCoverOnPlayerBackgroundSelector extends StatelessWidget {
           title: const Text("Show Blurred Cover as Player Background"),
           subtitle: const Text(
               "Whether or not to use blurred cover art as background on player screen."),
-          value: FinampSettingsHelper.finampSettings.showCoverPlayerBackground,
+          value: FinampSettingsHelper.finampSettings.showCoverAsPlayerBackground,
           onChanged: (value) =>
-              FinampSettingsHelper.setShowAlbumArtPlayerBackground(value),
+              FinampSettingsHelper.setShowCoverAsPlayerBackground(value),
         );
       },
     );
--- a/lib/components/LayoutSettingsScreen/show_cover_on_player_background_selector.dart
+++ b/lib/components/LayoutSettingsScreen/show_cover_on_player_background_selector.dart
@@ -1,0 +1,26 @@
+import 'package:flutter/material.dart';
+import 'package:hive/hive.dart';
+
+import '../../models/finamp_models.dart';
+import '../../services/finamp_settings_helper.dart';
+
+class ShowCoverOnPlayerBackgroundSelector extends StatelessWidget {
+  const ShowCoverOnPlayerBackgroundSelector({Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<Box<FinampSettings>>(
+      valueListenable: FinampSettingsHelper.finampSettingsListener,
+      builder: (_, box, __) {
+        return SwitchListTile(
+          title: const Text("Show Blurred Cover as Player Background"),
+          subtitle: const Text(
+              "Whether or not to use blurred cover art as background on player screen."),
+          value: FinampSettingsHelper.finampSettings.showCoverPlayerBackground,
+          onChanged: (value) =>
+              FinampSettingsHelper.setShowAlbumArtPlayerBackground(value),
+        );
+      },
+    );
+  }
+}
--- a/lib/models/finamp_models.dart
+++ b/lib/models/finamp_models.dart
@@ -66,7 +66,7 @@ class FinampSettings {
     this.showTextOnGridView = _showTextOnGridView,
     this.sleepTimerSeconds = _sleepTimerSeconds,
     required this.downloadLocationsMap,
-    this.showCoverPlayerBackground = false,
+    this.showCoverAsPlayerBackground = false,
   });
 
   @HiveField(0)
@@ -156,7 +156,7 @@ class FinampSettings {
 
   /// Whether or not to use blurred cover art as background on player screen.
   @HiveField(16)
-  bool showCoverPlayerBackground;
+  bool showCoverAsPlayerBackground;
 }
 
 /// Custom storage locations for storing music.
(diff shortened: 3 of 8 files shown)


UnicornsOnLSD:
This looks really nice, much more interesting than the boring black/white background!

The settings likely aren't working because there Hive wants a default value for `showCoverAsPlayerBackground`. I have it layed out so that there are a load of constants above the `FinampSettings` so that we can easily use the same value in the constructor and for Hive's `defaultValue` field. https://github.com/rom4nik/finamp/blob/038ab93c1b336791ff285cd5e14a96d03d45fabe/lib/models/finamp_models.dart#L36

I also noticed some weird behaviour when closing the player screen on iOS, I'll check to see if the same thing happens on Android.

The low contrast probably isn't too much of an issue - we could change some text colours if it makes sense. I did try messing about with blurred images really early on in Finamp's development (The widget still exists at `lib/components/blurred_image` but I'm pretty sure the method it uses is inefficient) and I think I stopped because of contrast issues. I was putting them on the flexible space bar on the album screen though, not the player screen.

rom4nik:
Thanks for pointing out what was missing, there were *so many* places to adjust just to get a toggle in settings lol. I've decided to make this feature turned on by default, improves discoverability of it a bit, and it's easy to permanently turn off now. I haven't observed any weird behaviors on Android.

UnicornsOnLSD:
Yeah, FinampSettings is a bit cluttered lol. As far as I know there aren't any prebuilt solutions for settings, so I just threw something together.

I've reimplemented the blurred background with a BlurHash, which should be faster and doesn't have the weird behaviour on iOS. It looks like flutter_blurhash has gotten better since I last looked (it now calculates the blur on another thread, before it caused really bad frame drops), so maybe I'll make AlbumImage use BlurHashes while the image is being fetched.

One tiny improvement that would be nice would be fading between backgrounds when a track changes. AlbumImage gets away by fading in new images, but in its current state the background changes abruptly.

rom4nik:
Thanks for working on this PR lol. I forgot that other clients use blurhashes already, and initially thought it would be a great idea to reuse them, but right away I noticed that the blurhash background can sometimes look awful:

Screenshot_20220727-205235_Finamp

I suppose lower "source resolution" of blurhashes is to blame here. Using full res images for blur avoided the white color spill, as seen in original screenshot in this PR.

rom4nik:
Another quick comparison (hot reload in Flutter rulz), take notice especially of the horrible black/green spill present in blurhash:

Full res|Blurhash
---|---
Screenshot_20220727-210304_Finamp|Screenshot_20220727-210324_Finamp

UnicornsOnLSD:
Yeah, blurring the images in Flutter does look better but the iOS issue makes it unfeasible. I'll record a video in a minute of the issue happening.

UnicornsOnLSD:
Here's what's happening on iOS. Looks like it's stretching because the image isn't the right aspect ratio or something

https://user-images.githubusercontent.com/44349936/181355247-4eb9e3a2-b69e-4fd5-a9c6-aabafdca70ce.mp4

UnicornsOnLSD:
Looks like it may be related to this: https://github.com/flutter/flutter/issues/31706#issuecomment-522697161. It would also explain why the filter applies under the AppBar without `extendBodyBehindAppBar` being true in the Scaffold.

UnicornsOnLSD:
If we do add back an ImageFilter based solution later, it may be a good idea to extract AlbumImage into an image provider so that we don't have to grab the same image twice from Jellyfin. It would also be pretty satisfying watching both images fade in at exactly the same time 🙃

rom4nik:
Unless it's video compression causing the effect, it seems like the blurred image in background keeps moving as you're swiping the player screen away. That made me guess it's just square album image being scaled up to fit height and not having left and right sides cropped. It doesn't manifest itself on Android (no horizontal gesture), but on iOS it's just being dragged away along with the other items on player screen until entire screen is removed.

Maybe just cropping it to screen aspect ratio would help?

UnicornsOnLSD:
I tested with a Container that was set to the screen size and it still happened

rom4nik:
Won't container expand to fit child inside anyway? Maybe try setting blur value to 0 and check if it's indeed whole image being dragged, or if it's something else.

UnicornsOnLSD:
Here's a recording of a Container wrapped in a 100x100 sized box (with the ImageFiltered highlighted with devtools) and what it looks like with a blur of 0.

https://user-images.githubusercontent.com/44349936/181363284-f12d194a-648b-4b8d-8e4c-644d47254584.mp4

https://user-images.githubusercontent.com/44349936/181363293-cdefe770-2041-46d4-8081-643d642dc5f6.mp4

Also Hackintoshes are fun, highly recommended if you're bored and have compatible hardware :)

UnicornsOnLSD:
I've always wanted to make NowPlayingBar properly slide in with the player screen as an animation, which would in theory get past the iOS transition issue. [Sweyer](https://github.com/nt4f04uNd/sweyer/) has the look I want, and I did try to implement it ([even asked how they did it lol](https://github.com/nt4f04uNd/sweyer/issues/17)), but I never managed to do it since it's quite hard to implement. I could have a look into it again though.

rom4nik:
Another way to get some colors from cover art on background is [taking average color from blurhash](https://github.com/fluttercommunity/flutter_blurhash/issues/5#issuecomment-915746685). That won't cause any artifacts like on *Gravity* album cover. Ignore artifacts caused by mpv-webm encoding on videos below.

https://user-images.githubusercontent.com/46846000/185769698-1bf95108-e5fd-4252-a96d-1981ff09612e.mp4

https://user-images.githubusercontent.com/46846000/185769706-f4ff2c27-5034-48bc-93a1-f505af4dd149.mp4

To make it a bit fancier we could do some gradient using avg colors from blurhash corners or edges (a bit like Plexamp?), but I haven't figured that out yet.

Chaphasilor:
@rom4nik I'm currently investigating how to extract colors from the album covers and blurhashes. The problem with using the average color is that it can end up looking pretty dull, at least when using it as an accent color and not a background. But maybe the average color can be used as a background for thr blurhash in order to make it stand out less. I'll play around with it :)

jmshrv:
Turns out you just need to add a ClipRect to stop the blur from overflowing... We'll go back to generating the blur dynamically because it looks much nicer.

Screenshot 2022-09-01 at 13 57 54

jmshrv:
Note that it would probably be a good idea for us to abstract out `AlbumImage` a bit since aspect ratio and clipping is being applied to that blur for no reason

jmshrv:
This should be ready now - I've reworked `AlbumImage` to make it easier to grab a raw `AlbumImage` without the padding and stuff, and I've made `AlbumImageProvider`, which has a function to get an image provider for a Jellyfin item (either a `NetworkImage` or a `FileImage` depending on if its downloaded or not. The only real difference I've seen is that we have to provide a key to `AlbumImage` items that change, such as the one on the player screen. Not sure if that's a bug on my part or Flutter doing what it's meant to do.

jmshrv:
Key issue has been fixed, had to make sure to update the future in `BareAlbumImage` when we pass a different value

PR #284, opened 2022-07-20
Show all/more album artists on album screen and tap them to go their page

### LINKED ISSUE #280: [Feature Request] Click on Artist Name from Album View
Hi!

I often find myself trying to click on the artist name when viewing an album to try to view all albums by that artist.

For instance, if I'm looking at the album page for Depeche Mode's "Construction Time Again" (i.e. the list of tracks), it lists the basic info of the album at the top (artwork, artist, year, length). It'd be super helpful to be able to click on the artist name and see all of Depeche Mode's albums instead of having to exit that view, go to artists, find Depeche Mode and click on them from there.

I hope that explanation makes sense! My apologies if this has already been a requested feature; I wasn't able to find it if so.

Thanks again!!

rom4nik:
Fixes #280. Extracted code generating `separatedArtistTextSpans` on player screen and reused it on album page for tappable artists.
Screenshot_20220721-013035_Finamp


Y0ngg4n:
@rom4nik can you resolve the conflicts?

Chaphasilor:
fixed the conflict, but can't really test it. How do you specify multiple artists for an album? In the web UI I can only add multiple album artists to a song, but when editing the metadata of an album there's no field to set the album artists...  
Am I missing something?

rom4nik:
https://github.com/jellyfin/jellyfin/blob/099632f37e559c1aad4944fb328600a242679988/MediaBrowser.Providers/Music/AlbumMetadataService.cs#L89

Perhaps it would be enough to edit some song's albumArtists, lock it, then force rescan/metadata refresh on the whole album? If not, then I suppose editing tags in song file then reimporting would help.

Chaphasilor:
It's not possible to lock the album artist field, it just gets overwritten on rescan 🙃
But I managed to copy a file, update the tags and then scan it like you suggested. (Side note: Jellyfin doesn't extract artists from the "contributing artists" tag?).

Everything seems to be working fine, but I noticed that in the album list only one of the artists is shown. While that behavior is identical to Jellyfin web, I believe it would be better to show all artists in the album list as well. I'll go ahead and implement that change...

Chaphasilor:
If anyone could take a quick look at this to confirm that it works and I didn't break anything else that would be great!  
Otherwise I'll merge this in tomorrow :)

Chaphasilor:
Merged, thanks a ton! 😁

Y0ngg4n:
@Chaphasilor you are a genius :)

PR #287, opened 2022-07-23
Show song artists on album screen (and more)

### LINKED ISSUE #265: Feature Request: Show Track Artist next to Play Time in Album View
Artists might be different from the Album Artist printed at the top. I request adding this information to the album-view screen, my recommendation for the location is next to the Play Time.

A general Idea would be to add developer settings, where you can turn on/off such behavior

Screenshot_20220707-153149_Finamp

(Thumbs up to Finamp! It gets better with every release! <3 )

rom4nik:
Fixes #265. By default song artists are hidden if their set matches set of album's artists, but it's configurable (and persists across app restarts, thanks for hint in the other PR about that!).

As discussed in that issue, this PR also reorganizes SongListTile a bit and replaces stars in favorite buttons with hearts. I also added the now a bit more portable favorite button to artist, album and playlist screens.

Album screen|Settings
---|---
Screenshot_20220723-031707_Finamp|Screenshot_20220723-024024_Finamp

### DIFF
--- a/lib/components/AlbumScreen/album_screen_content.dart
+++ b/lib/components/AlbumScreen/album_screen_content.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_sticky_header/flutter_sticky_header.dart';
 
@@ -73,13 +74,13 @@ class AlbumScreenContent extends StatelessWidget {
                 sliver: _SongsSliverList(
                     childrenForList: childrenOfThisDisc,
                     childrenForQueue: children,
-                    parentId: parent.id),
+                    parent: parent),
               )
           else
             _SongsSliverList(
                 childrenForList: children,
                 childrenForQueue: children,
-                parentId: parent.id),
+                parent: parent),
         ],
       ),
     );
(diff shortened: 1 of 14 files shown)


UnicornsOnLSD:
Just pushed a commit that puts the download indicator in a WidgetSpan and aligns it to the text. This looks a lot better IMO, do you agree?

Simulator Screen Shot - iPhone 13 - 2022-07-24 at 13 28 46

UnicornsOnLSD:
The alignment is still a bit off, but I could have a look in a bit

rom4nik:
It's probably because SVG of the icon itself has padding:

obraz

obraz

Also the indicator looks a bit tiny now imo, but at least it's indeed nicely aligned to text in the same line.

UnicornsOnLSD:
Yeah, it is a bit small. We could just put it back at the side. The Material icons don't really have a good icon for this use case. `download_for_offline` is pretty close to what Spotify has, but it doesn't look great in Finamp due to how small the download is.

rom4nik:
Doesn't look much smaller than in Spotify imo, I'd just suggest aligning it to top, since song length won't have characters like `y` next to the indicator. Size will depend on DPI anyway.

obraz

UnicornsOnLSD:
It's pretty much perfect with the offset you had :)

Screenshot 2022-07-25 at 16 35 07

rom4nik:
Looks good to me too! The bottom bar of the icon is a bit below text, but I didn't notice it until I zoomed in, so that's fine by me.

IMG_20220725_193150

UnicornsOnLSD:
That alignment is probably the best we can get, the code looks good so I'm going to merge this

PR #300, opened 2022-08-11
Fix retrieval of Android ID

hannesbraun:
With device_info_plus_platform_interface 2.6.0 (which is a nested dependency of device_info_plus), getting the Android ID is no longer possible. This functionality is now available through the plugin "android_id".


UnicornsOnLSD:
Thanks :)

PR #301, opened 2022-08-14
Only show StopForegroundSelector on Android

hannesbraun:
As this setting is only used for Android, there's no need to display it on other platforms.


jmshrv:
Thanks :)

PR #308, opened 2022-08-21
Fix missing album images in queue list

rom4nik:
Probably some regression happened at some point.

obraz


jmshrv:
Thanks, been meaning to fix this for ages

Issue #160, opened 2021-12-20
Allow deleting local files from "Downloads" view


mwip:
First of all thanks for your awesome work on this. :+1: love the app.

I think it would be a nice feature to be able to delete local files from the Downloads view. 

**current behavior**  
When I want to delete an album that I downloaded, say to free up space, I need to go to the Downloads view and check which albums are local, then go to the Album view and delete the local copy. I know it is also possible to do this a bit easier through "offline mode", but overall I think it is a bit unintuitive. 

**desired behavior**  
It would be nice if the Downloads view would add some possibility to delete downloaded files. Maybe through a " long-press-context-menu" (for lack of a better word, sorry for the probably incorrect term)? This menu could also be used for additional context functionality such as sharing a media file etc...

UnicornsOnLSD:
I've been meaning to do this for a while, and there's probably some old issue that I can't find. I think at one point I was going to do it before initial release lol

HumerousTwist:
Adding a vote to this as well.  Potentially a swipe to delete with confirmation popup?

Issue #213, opened 2022-04-10
New Download Package


jmshrv:
Currently, Finamp uses flutter_downloader to handle downloading songs and images. This package isn't the greatest, and has been the root cause of issues such as janky download indicators and the app freezing when downloading a lot of songs at once. The main focus of 0.7 will be creating a new plugin which would make managing downloads much easier. This could include stuff like adding streams to listen to download progress, and the ability to handle URL download paths for better custom path support. This package will also have a generic Dart fallback for Windows/Linux, as currently flutter_downloader only supports iOS and Android (the iOS code should run on macOS, but I'd rather not inherit any code from flutter_downloader).

drazil100:
Honestly this can't come soon enough. The current solution is extremely buggy. The inclusion of thumbnail downloading really confused my install and I ended up deleting it and redownloading it to start completely fresh. I am currently in the process of redownloading all of my music and it has hung multiple times with where there queue was 4 or there were 2 running but half of multiple albums were not downloaded and the enqueued/running numbers did not change through multiple attempts to close and reopen the app.

axelsimon:
Glad to see this is already reported in some way. Could the issue be retitled with something more user-oriented, like "Current download system provides no information and can create a mess" or something like that? :)

I, like others i expect, have had a mess created by trying to download a single album, lose connectivity, and end up with a "failed" download with no way to check, clean-up or retry.

UnicornsOnLSD:
I don't want to be harsh on flutter_downloader, it works for it's intended purpose - Finamp has just outgrown that and needs proper dependency management built-in.

In Finamp, you can delete failed albums and they should clean themselves up. At some point I'm going to add easier to access delete buttons and a retry button to the download screen.

axelsimon:
Still, a good start! Good to hear :slightly_smiling_face:

UnicornsOnLSD:
From @rodcramp in #62 - we will want to model genres in this plugin, which can be implemented by making a genre a group. To follow how Finamp works, genres will have albums as their children.

MrPotatoBobx:
Will transcoded downloads be included in the release?

Chaphasilor:
> Will transcoded downloads be included in the release?

It's probably unrelated, but they most likely won't be included *before* this is done \^\^

jmshrv:
[...]

PR #331, opened 2022-09-22
Add deletion of albums and songs

Y0ngg4n:
This PR Adds simple deletion of already downloaded albums and songs


jmshrv:
Thanks for this :)

I've just added a push that makes items actually disappear when deleting, but there are a few issues before this is merged:

* The rest of the downloads screen doesn't update (such as the overview at the top, file sizes etc
* When deleting an album in offline mode, the album still shows in the album screen, which causes an error when opened as the album no longer exists

Both could be solved by doing some state management with Riverpod. For the 2nd one, this would mean hooking up the album list to use a central source of albums from Riverpod, which would be a pretty big task. I don't mind doing this, as it would also help with other issues like deleting albums in offline mode from the album screen (which I've disabled for the same reason).

Y0ngg4n:
Hi thanks for your review. Sure would be nice if i could get some help from you and learn fron your project 🙂

Yes i have already seen that the list does not update caused by statelesswidget. But do not know how you handle states via hive. Quick side question wouldnt it be better to use isar?

jmshrv:
Finamp gets the list of albums with the `JellyfinApiHelper` class, which simply returns once without any sort of live updating. Currently, downloaded songs are stored in Hive databases, as at the time Isar was still in development. #213 will replace this janky system of Hive databases with an Isar database which should make stuff much easier. For now, we can either:

* Add a provider that handles getting the albums and allows different bits of the app to modify the list
* Rework `JellyfinApiHelper` to return a listener on the database. I'd like to more cleanly separate online and offline mode, so maybe I'll consider this then.

jmshrv:
Fixes #160

Y0ngg4n:
Ok thank you for the explanation. So what do we have to do with this PR now?

jmshrv:
I'll probably go with the stream option as it would be much easier than using providers, although I might change my mind mid-implementation as it would cause a lot of useless StreamBuilders in the code.

Y0ngg4n:
Ok sounds good

arthurlutz:
Looking forward to this feature. Do you need help testing it out ?

Y0ngg4n:
No i have to first implement the missing parts. But i am currently short on time.

Y0ngg4n:
@jmshrv i have tested it now and it seems to work just fine. anything we need to do here or can we just merge?

jmshrv:
I'll merge this for now, the other improvements can come later

Y0ngg4n:
@jmshrv thank you :)

Y0ngg4n:
@jmshrv could we push out a new release?

jmshrv:
Yeah, I'll make a new one as discussed in #371

PR #333, opened 2022-09-23
Add music visualizer on currently playing song

Y0ngg4n:
Add this https://pub.dev/packages/mini_music_visualizer to the currently playing song


jmshrv:
I feel like this is a bit redundant as the song list tile already changes the title colour for the currently playing song

Y0ngg4n:
I dont think so because the animation is a movement, so it is easier for the eyes to spot the title. I personally had it often that i did not spot the song.

Y0ngg4n:
ok so how to proceed with this? Close it?

jmshrv:
It could be merged, I just need to have a look at the code since there may be a more efficient way of displaying it (I'm pretty sure SongListTile already has a StreamBuilder for the song title that we could expand)

espector:
This looks like a nice visualization. Vote to add it :)

Chaphasilor:
I'm also in favor of the change, although I'll probably end up putting it on top of the album cover in the song list tile to save some space...

jmshrv:
Moved it about a bit, enabling/disabling the visualiser is now in SongListTIle. Also fixed merge conflicts and stuff. Only thing to do now is to hide the visualiser when the song is paused.

Issue #177, opened 2022-01-15
Music is playing but there is no audio


MagentaPuppy:
iphone 8, ios 13

Maxplosion:
logs from finamp? 
does it work in on other clients and if yes which?
jellyfin server logs? 
are you transcoding?
see a pattern here...?

MagentaPuppy:
**Finamp logs:** https://paste.gg/p/anonymous/c238f1a4142447909d5abae177e9166e
**Finamp audio logs:** https://paste.gg/p/anonymous/4ab3e8c877bb41bdbcdb6dd737de7c0e
**Jellyfin logs:** https://paste.gg/p/anonymous/3c221e7311fe4066a5346fe36e81b27a

It works on Jellyfin web client and ios client, i don't have and android phone to try Finamp on

not transcoding, audio plays if i turn it on

UnicornsOnLSD:
What codec is your audio? OPUS and Vorbis aren't supported yet on iOS as iOS doesn't natively support those formats (it's dumb I know)

MikePadge:
@UnicornsOnLSD thanks for your work on this app, I love it

I'm not super familiar with Flutter, but this might be a solution
https://github.com/EPNW/opus_flutter/tree/master/opus_flutter

UnicornsOnLSD:
That's a really interesting looking package, I'll do some testing with it :)

MagentaPuppy:
> What codec is your audio? OPUS and Vorbis aren't supported yet on iOS as iOS doesn't natively support those formats (it's dumb I know)

Not even transcoding it?? Also I’m kinda having this issue again, but now it only shows songs I’ve hearted on jellyfin web/iOS app, and yeah nothing opus is playing, annoying given that the vast majority of my music is opus

UnicornsOnLSD:
Transcoding should work since Jellyfin will then serve it as AAC, which is supported. As I said, this is an iOS issue so it isn't likely to get fixed in the future. I believe the libopus package that was sent here was for when you want to encode Opus audio in your app, so it won't be useful here.

As for only showing favourite content, you can disable the favourite view by clicking on the star in the top right.

UnicornsOnLSD:
Just noting here a comment I left on a duplicate issue - https://github.com/UnicornsOnLSD/finamp/issues/236#issuecomment-1174953175

[...]

PR #334, opened 2022-09-23
Adds a shortcut to system equalizer. (Only works on android)

Y0ngg4n:
Added a check if system is android

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -4,6 +4,7 @@
   <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
   <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
   <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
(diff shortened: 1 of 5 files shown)


jmshrv:
This isn't too useful imo, the user can get to their equaliser in their system settings. If I ever get around to #177, we could likely implement an equaliser in Finamp that will support all platforms

Y0ngg4n:
Most People dont know about the system Equalizer. I dont know how we could implement it for all platforms.

jmshrv:
If my plans to use GStreamer as a player backend work, we could implement it in the player: https://gstreamer.freedesktop.org/documentation/equalizer/index.html

Another thing to note is that not all Android skins have a system equaliser, and from quick look at the Equalizer package it looks like there isn't a way to determine this.

Y0ngg4n:
ah ok thank you :)

Y0ngg4n:
But i guess there is no gstreamer package we can use so we have to implement it our selves right?

jmshrv:
Yeah, my initial idea is to make a just_audio implementation that uses GStreamer. It's pretty far off, but it'll be really nice once it's done.

PR #335, opened 2022-09-23
Add minutes description to sleep timer

Y0ngg4n:
This just replaces the doubled headline with minutes instead


Y0ngg4n:
Thank you 😊

PR #345, opened 2022-10-06
Add Gesture to swipe up to open the player

Y0ngg4n:
This adds a simple gesture to swipe up to open the player.


Chaphasilor:
This is only triggered if the swipe starts on the mini player, right? Otherwise it would lead to problems when scrolling...

Chaphasilor:
Awesome! Yeah that asymmetry has been bugging me from the start! Thanks for fixing it :)

jmshrv:
Looks good, merging :)

Y0ngg4n:
Thank you 🙂

Issue #344, opened 2022-10-05
Cant play downloaded items in playlist if playlist contains download errors


Y0ngg4n:
Expected: Ignore the failed downloads and play only the downloaded.

How it is currently:
You cant play any song of the playlist. Neither via Shuffle, Play or by tapping on the songs.

Chaphasilor:
Duplicate of #252

Issue #252, opened 2022-06-12
Can't play albums and playlists with download errors


abstractionmage:
Songs with download errors in them still show up alonglide regular songs in playlists and albums (see image below). In offline mode, upon trying to play (or shuffle play) downloaded albums or playlists containing both properly downloaded songs and songs with download errors, nothing happens. I would expect the app to still play the albums or playlists and skip any songs with download errors.

In combination with #251 this issue renders the app unusable for me unfortunately, since I can't play most of my playlists offline.

![Screenshot_20220612-113058](https://user-images.githubusercontent.com/37195767/173227037-ae319468-eefb-4585-ac1c-5118070a4287.jpg)

jeremyswright:
I'm seeing this same behaviour.

jmshrv:
Fixed in #346

PR #346, opened 2022-10-06
Fix play with download errors

Y0ngg4n:
This fixes the issues #344 and #252

You can now play songs in playlists with download Errors


Y0ngg4n:
Thank you 🙂

PR #361, opened 2022-11-17
Fix queue modification issues

Maxr1998:
By only keeping one queue for the AudioSources, using the built-in move method and generating the UI queue on the fly, issues when moving items in the queue should be a thing of the past.


jmshrv:
This is great, merged!

Maxr1998:
Thanks!

PR #380, opened 2023-01-09
Add setting to disable gestures

BhasherBEL:
Add a new setting "Disable gestures" to disable horizontals and verticals gestures. Useful to prevent unwanted swap when using gesture navigation.


jmshrv:
Thank you!

PR #347, opened 2022-10-06
WIP: Delete Items from Playlist

Y0ngg4n:
@jmshrv maybe you could review it. 
i have created the Api Request, and it sends the API Request to the Jellyfin Server.
I get a 204 back but the item does not get removed from the Playlist. Am i doing something wrong?
```
--> DELETE https://media.example.com/Playlists/bba88a58b8687fe6d8b0c280cb7acd9b/Items?entryIds=ac8bb27d9796804cc40e5a5503f1bea8
I/flutter (15961): [Chopper/INFO] 2022-10-06 18:44:41.959262: Content-Type: application/json
I/flutter (15961): [Chopper/INFO] 2022-10-06 18:44:41.959753: X-Emby-Authorization: MediaBrowser UserId="USERID", Client="Finamp", Device="Bla", DeviceId="BLABLUP", Version="0.6.8"
I/flutter (15961): [Chopper/INFO] 2022-10-06 18:44:41.960282: X-Emby-Token: TOKEN
I/flutter (15961): [Chopper/INFO] 2022-10-06 18:44:41.961047: --> END DELETE
I/flutter (15961): [Chopper/INFO] 2022-10-06 18:44:42.768490: <-- 204 https://media.example.com/Playlists/bba88a58b8687fe6d8b0c280cb7acd9b/Items?entryIds=ac8bb27d9796804cc40e5a5503f1bea8

```

Issue #340, opened 2022-10-01
Can't remove songs from playlists??


MisterSirCode:
When I added some songs to my playlist, I accidentally added one I didn't want in it, so I tried to remove it, but theres no button anywhere I can see to remove it.

Swiping just adds to the queue which is strange, most apps use that as the delete function, you swipe and it shows a red delete button. Otherwise, holding down on the song just shows the normal stuff, I can't remove it from the playlist.

MisterSirCode:
On a side note... why are there so many open issues and PRs...

Chaphasilor:
> On a side note... why are there so many open issues and PRs... 

Because it's mostly just one person making the app and there are a lot of users with different expectations :)

So if you happen to know some Flutter, that would be amazing!

jmshrv:
Removing songs from playlists has been something that I was meant to add for a while, it shouldn't be too hard to implement™

As @Chaphasilor said, I'm only one person doing this in my free time. I've been meaning to look through all the issues to clean up duplicates and issues that have been completed, and @Y0ngg4n recently made quite a few PRs that I'm getting through slowly. Feel free to help with getting some of those issues closed though ;)

jmshrv:
Implemented in #382

PR #382, opened 2023-01-09
Remove from playlist

BhasherBEL:
This implement the "remove from playlist" button, fixing #340 and most of #347.
But I have two concerns:
 - When a music is delete, you need to update the playlist content to see it visually removed. Maybe you have an idea of how to update the list ?
 - When the last music is delete, you get an empty playlist. Trying to open a empty playing leads to the crash of the application. Even if an empty playlist isn't usually desirable, it could be sometime temporary. It could be useful to be able to display them ?

If you have any other concerns, feel free to ask fixes.


jmshrv:
This looks good so far, is it intended to supercede #347?

* For getting rid of the song list tile, the simplest solution would probably be a callback function. I suggested using a provider in #347, but that is a pretty complex solution that is arguably less clean. It'd look something like `onDelete: () => setState({ items.removeAt(index) })`.
* There's probably some null check error happening, should be a simple fix

BhasherBEL:
Yes, it's a reimplementation of #347.
 
I tried something like you suggest but sadly after 2 seconds (animation duration) the item come back and remove the last one. Here is a first implementation. I will try to fix the problem completely and also the empty playlist crash when I have time.

BhasherBEL:
This should fix the problem. But I had to replace all `mutableItem` by `wiget.item` and delete the `late` item in `components/AlbumScreen/song_list_tile.dart`. The comment said it was used to change favourite property, but it seems to still work even with my fix.

BhasherBEL:
This should fix the problems. The first commit is about visually remove items and the second prevents the app to crash when opening an empty playlist.

For the first one, I had to replace all `mutableItem` by `wiget.item` and delete the late item in `components/AlbumScreen/song_list_tile.dart`. The comment said it was used to change favourite property, but it seems to still work even with my fix.

Feel free to suggest any other change.

BhasherBEL:
@jmshrv ?

jmshrv:
Sorry, missed this in my emails :upside_down_face: 

The code looks good, I'll do a bit more testing in a bit (since this is quite a big PR) but from what I can see it should work well :)

jmshrv:
Just did a few other things (described in the commit messages), merging now. Thanks for implementing this!

PR #403, opened 2023-02-21
Trim searchTerm before searching

Maxr1998:
When using the keyboard's autocomplete/suggestions while searching, a space is usually appended to the search term. Because this could influence the returned search results, it makes sense to trim the search term before sending it to the server.


jmshrv:
Thank you!

PR #404, opened 2023-02-21
:bug: Fix crash in ErrorApp

Maxr1998:
Currently, the error app crashes due to `AppLocalization.of(context)` returning `null`.
This is apparently [a bug in Flutter](https://stackoverflow.com/a/72754385/2475176), and worked around by adding another widget wrapping the Scaffold.


jmshrv:
That's a very strange edge case, thanks for somehow finding a fix :)

Maxr1998:
Sure thing, I noticed it when introducing some other bugs while working on something else. Glad it was this easily fixable 😄

PR #405, opened 2023-02-22
:sparkles: Change sort options to apply per tab

### LINKED ISSUE #372: Allow different sorting for different tabs
Currently, the sort setting is synchronized between the tabs, so sorting albums by date added also sorts artists by their library addition date.

Preferably, the sort options would be independent. If that is desired, I can look into implementing this in a PR.

Maxr1998:
Closes #372.


jmshrv:
Nice implementation, thanks!

PR #406, opened 2023-02-22
:bug: Fix vertical padding in artist/genre lists

### LINKED ISSUE #183: [Bug] Inconsistent List Gaps
The gaps between albums, artists, tracks, etc. tabs are all different from each other. For example graps between items in the list om albums is bigger where artists tab has zero gaps and art is touching.

Maxr1998:
Fixes #183.


jmshrv:
I'd rather have some actual data in the subtitle of the artists so that Flutter can correctly handle the spaces for us, but this is a good stopgap solution for now :)

Maxr1998:
Yup, it should be easy to change when the time comes.

PR #418, opened 2023-03-09
:eyes: Download original images instead of always converting them to png

Maxr1998:
Reasoning: for some cover arts, storing PNGs takes up a lot of space (up to 20 MB per image as I've seen) but due to the conversion, there is still a noticeable quality loss like artifacting or patterns from the conversion. Just downloading and storing the original file is thus preferable.

### DIFF
--- a/lib/services/downloads_helper.dart
+++ b/lib/services/downloads_helper.dart
@@ -880,13 +880,14 @@ class DownloadsHelper {
 
     final imageUrl = _jellyfinApiData.getImageUrl(
       item: item,
-      quality: 100,
-      format: "png",
+      // Download original file
+      quality: null,
+      format: null,
     );
     final tokenHeader = _jellyfinApiData.getTokenHeader();
     final relativePath =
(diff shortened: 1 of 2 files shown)


lib/services/downloads_helper.dart
```diff
@@ -880,13 +880,14 @@ class DownloadsHelper {
...
-      format: "png",
+      // Download original file
+      quality: null,
+      format: null,
     );
     final tokenHeader = _jellyfinApiData.getTokenHeader();
     final relativePath =
         path_helper.relative(downloadDir.path, from: downloadLocation.path);
-    final fileName = "${item.imageId}.png";
+    final fileName = item.imageId;
```
jmshrv, line 890: I'm surprised this works lol

Maxr1998: Both are a string, so it's a bit simpler than the pattern. The extension doesn't really matter anyways.

jmshrv:
Thanks :)

PR #434, opened 2023-04-14
Fix flutter_downloader and DownloadCallback

Y0ngg4n:
@jmshrv it seems like flutter_downloader switched the repository and the DownloadCallback works with an int now.
I hope this fixes the issues.

### DIFF
--- a/lib/components/DownloadsScreen/current_downloads_list.dart
+++ b/lib/components/DownloadsScreen/current_downloads_list.dart
@@ -41,7 +41,7 @@ class _CurrentDownloadsListState extends State<CurrentDownloadsList> {
   }
 
   static void downloadCallback(
-      String id, DownloadTaskStatus status, int progress) {
+      String id, int status, int progress) {
     final SendPort? send =
         IsolateNameServer.lookupPortByName('downloader_send_port');
     send?.send([id, status, progress]);
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -447,7 +447,7 @@ class ErrorScreen extends StatelessWidget {
 }
 
 class _DummyCallback {
-  static void callback(String id, DownloadTaskStatus status, int progress) {
+  static void callback(String id, int status, int progress) {
     // Add the event to the DownloadUpdateStream instance.
     final SendPort? send =
         IsolateNameServer.lookupPortByName('downloader_send_port');
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -5,18 +5,18 @@ packages:
     dependency: transitive
     description:
       name: _fe_analyzer_shared
-      sha256: "0c80aeab9bc807ab10022cd3b2f4cf2ecdf231949dc1ddd9442406a003f19201"
+      sha256: a36ec4843dc30ea6bf652bf25e3448db6c5e8bcf4aa55f063a5d1dad216d8214
       url: "https://pub.dev"
     source: hosted
-    version: "52.0.0"
+    version: "58.0.0"
   analyzer:
     dependency: transitive
     description:
       name: analyzer
-      sha256: cd8ee83568a77f3ae6b913a36093a1c9b1264e7cb7f834d9ddd2311dade9c1f4
+      sha256: cc4242565347e98424ce9945c819c192ec0838cb9d1f6aa4a97cc96becbc5b27
       url: "https://pub.dev"
     source: hosted
-    version: "5.4.0"
+    version: "5.10.0"
   android_id:
     dependency: "direct main"
     description:
@@ -29,18 +29,18 @@ packages:
     dependency: transitive
     description:
       name: archive
-      sha256: ed7cc591a948744994714375caf9a2ce89e1d82e8243997c8a2994d57181c212
+      sha256: "0c8368c9b3f0abbc193b9d6133649a614204b528982bebc7026372d61677ce3a"
       url: "https://pub.dev"
     source: hosted
-    version: "3.3.5"
+    version: "3.3.7"
   args:
     dependency: transitive
     description:
       name: args
-      sha256: "139d809800a412ebb26a3892da228b2d0ba36f0ef5d9a82166e5e52ec8d61611"
+      sha256: "4cab82a83ffef80b262ddedf47a0a8e56ee6fbf7fe21e6e768b02792034dd440"
       url: "https://pub.dev"
     source: hosted
-    version: "2.3.2"
+    version: "2.4.0"
   async:
     dependency: transitive
     description:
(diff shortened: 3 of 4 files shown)


jmshrv:
The download callback stuff should be fixed in main, I need to rebase redesign but once that's done it should fix the callback.

As for the design, I like the idea of a modal sheet. The current layout (in main) is getting a bit too big due to the amount of options

Y0ngg4n:
@jmshrv yes was just not able to deploy without fixing it.

Ok nice. Wasn´t that easy to implement it :)

Chaphasilor:
So should we go for the slide-up modal instead of the pop-over? I could adapt the current design to this style, but I'm not sure if we would gain a lot. The menu on the player page, with the added playback controls and (possibly) share button takes up a lot of space anyway.

Also, @Y0ngg4n does this also convert the menu on the Now Playing view, or only in the song list? Because the menu for the Now Playing view is the important one for the beta :)

Y0ngg4n:
@Chaphasilor i am not a big friend of modal dialogs. so in my opinion the bottom sheet is better. Currently this is only for the menu in the list view. but it is no big deal to implement it for the player view too. I am currently working on the queue but the bezier curve button at the bottom of your design is driving me crazy XD

Chaphasilor:
okay, no worries. You can leave out the curved shape for now, there were a lot of people who didn't really like it. Just add the button, and I'll think about it some more ^^

Y0ngg4n:
@Chaphasilor ok so yesterday evening i have worked on implementing the menu to the player screen. It is more difficult than i thougt. I had to rewrite the state managment in the playerscreen. Currently it works, but it does not update the screen. So i think there is some callback currently not working. Have to check.

Y0ngg4n:
@jmshrv maybe could you check this PR i am a little bit stuck why the player screen does not get updated

jmshrv:
I'll look into it in a bit, I've been very busy recently so I haven't kept up on Finamp issues

jmshrv:
Had a quick look through the code, it's mostly good, just a few things:

* The song info header seems to be rebuilt every frame
* The `imageProviderCallback` in `_SongInfo` seems to be copied from the player screen, which sets the providers that should be for the player screen only. Since we only need the colour for the header itself, no provider stuff should be required
* Your widgets appear to be backwards (properties and constructors after build). Minor issue, just threw me off a bit lol

Y0ngg4n:
@jmshrv will check it the next weeks

Y0ngg4n:
Ok so as i was to dumb to commit last time.... I had to rewrite everything again.
@jmshrv i have seen you changed the color generation. currently when you long press on the song_list_tile, the color stays white. i couldnt figure out how to initialize the imageprovider.
Maybe you can check it. I added a todo.

jmshrv:
Ah yeah that code is temporary, I've submitted a PR to `palette_generator` to do the multithreading within the package (https://github.com/flutter/packages/pull/4132), so that mess can get removed from `song_info.dart`.

I'll have a look at your other questions at some point (as well as getting that PR merged). I'm on holiday right now so I don't have much time to be doing Finamp stuff, but thanks for being active on the project today :)

Y0ngg4n:
Now it looks like this:


https://github.com/jmshrv/finamp/assets/30302975/795f1cf2-a222-4c0d-a3cd-2d425e98cf2e

Chaphasilor:
Nice, that's a good baseline to continue development from I think. On the player page I'm not a huge fan of another bottom sheet, because that's already used for the queue and would result in confusing interactions (does swiping up open the queue or the menu?). Right now there's also an unnecessary swipe up that's needed in order to show all options.  
I was thinking about using the modal, but collapsing away options that aren't frequently used (customizable).   

But I'm interested to hear why you're not a fan of modals dialogs? What are the issues with them? Maybe I'm missing something that makes bottom sheets (or something similar) more suitable :)

Y0ngg4n:
@Chaphasilor yes i know i would like a popup button menu more too. It was just easier to implement it this way. If it works i will change it to popupbutton menu.

Chaphasilor:
Okay, nice. Not sure if I ever posted this, but the popover menu would also show the selected item:

image

(this is for an album, but it would be almost the exact same for a song :))

And we could think about adding swipe up/down gestures for closing the menu, in addition to the X-button or clicking outside the modal

jmshrv:
This should be mergable now, also changed it to just use the dominant colour directly (because the player screen stuff is intended for getting a good colour on top of a background)

jmshrv:
Screenshot showing the new menu with Infinity on High

Is there a reason for this arrow?

Y0ngg4n:
@jmshrv the arrow is to show, that it is scrollable and that there are more options that are not displayed. because it can happen that inexperienced users never see the options that are hidden behind the scroll.

jmshrv:
Ah ok, a lot of apps get that wrong lol. A more "elegant" solution is to make it so that half of the next item is out of view, so that it's obvious that it can be scrolled. Apple does a very good job of this in their apps:

Apple Maps screenshot, showing an option half-hidden

It's just about working in this PR, I'll test on some other devices.

Y0ngg4n:
@jmshrv yes i know i tried it, but i did not get it to work. the problem here is that the buttons have no boundaries, if they would be elevated or something it would be far more remarkable that there is something cut of.

Chaphasilor:
@Y0ngg4n could you resolve the conflicts so we can get this merged? Would like to bring this up to speed for the beta :D

Y0ngg4n:
@Chaphasilor i can try

Y0ngg4n:
@Chaphasilor i hope i didn´t messed the merge up

Chaphasilor:
I'll take a look at it later. It's just some types that are wrong or missing parameters, not big deal hopefully. But thanks for doing the heavy lifting here! :D

Y0ngg4n:
@Chaphasilor i can try to fix it

Chaphasilor:
Managed to fix it. There were some changes that got lost while merging and I tried my best to restore those :)  
I'll do some more testing and try to merge it asap!

Y0ngg4n:
@Chaphasilor you are the best :)

Chaphasilor:
Only thing missing now is adding some extra buttons when the menu is shown on the player screen, otherwise we lose the sleep timer :)  
And the menu looks a bit different now...

Chaphasilor:
Okay, this should be done now :)  
@Y0ngg4n can you take a brief look at it to confirm that everything is working?

Y0ngg4n:
@Chaphasilor looks good should work :)

Chaphasilor:
Alright, it's merged. Thanks for all your work!

Y0ngg4n:
@Chaphasilor i have to thank you ;)

PR #446, opened 2023-05-13
minor player screen changes

Chaphasilor:
Some minor QoL improvements. Should probably rebase this to the regular redesign branch, as the changes are not related to the colors...

Also, the way the queue panel is created isn't ideal, it's basically just code repetition. I tried programmatically pressing the queue button instead, but couldn't get it to work. Would love to have some kind of method for that, which we can pass the current context and then open the panel through a simple function call!

### DIFF
--- a/lib/components/PlayerScreen/control_area.dart
+++ b/lib/components/PlayerScreen/control_area.dart
@@ -22,7 +22,7 @@ class ControlArea extends ConsumerWidget {
             padding: EdgeInsets.symmetric(horizontal: 16),
             child: ProgressSlider(),
           ),
-          Padding(padding: EdgeInsets.symmetric(vertical: 12)),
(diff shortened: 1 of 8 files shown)


jmshrv:
These changes look good :)

We could add a function that calls `showModalBottomSheet` for us with the relevant settings. I'll merge this when it's rebased to `redesign` if you're happy with it

Chaphasilor:
hmm, you can't rename a branch that's being referenced by a PR or the PR will be closed. TIL :D  
should be properly rebased now :)

Chaphasilor:
Oh and if you could add that method you mentioned that would be great, I'm not totally sure yet how to handle the context...

jmshrv:
Added the method, is this ready to merge?

Chaphasilor:
Yes, looks good! 😁  
Thanks!

PR #448, opened 2023-05-18
Sort artist albums by PremiereDate instead of ProductionYear

Maxr1998:
If an artist has multiple albums in one year, albums may not necessarily follow the release order. Sorting by PremiereDate instead fixes that. If no PremiereDate is set on the item, the server will automatically fall back to ProductionYear.


jmshrv:
Thanks, merged :)

PR #460, opened 2023-06-18
Sync downloads

Y0ngg4n:
@jmshrv This adds a new button to sync the current downloads.
Example: 
You download a playlist via finamp.
You add a track to this playlist via jellyfin.
Current situation: you have to delete the playlist and download it again.
With the new button: It checks for not downloaded files and downloads them too.

### DIFF
--- a/lib/components/DownloadsScreen/sync_downloaded_playlists.dart
+++ b/lib/components/DownloadsScreen/sync_downloaded_playlists.dart
@@ -25,19 +25,23 @@ class _SyncDownloadedPlaylistsButtonState
     _syncLogger.info("Syncing downloaded playlists");
     List<DownloadedParent> parents = downloadsHelper.downloadedParents.toList();
     List<DownloadedSong> songs = downloadsHelper.downloadedItems.toList();
+
     for (DownloadedParent parent in parents) {
-      List<BaseItemDto> children = parent.downloadedChildren.values.toList();
+      Set<String> children =
+          parent.downloadedChildren.values.map((e) => e.id).toSet();
       DownloadedSong firstSong = songs.first;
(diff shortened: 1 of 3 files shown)


Y0ngg4n:
@Chaphasilor no currently it will just download the new ones.

Chaphasilor:
Okay I think that makes sense. Otherwise we would need additional checks (is the song downloaded through a different playlist and shouldn't be removed?).  
At some point there could be a "cleanup" action that does exactly that and deletes all songs that aren't part of a downloaded playlist, album, or artists anymore :)

Y0ngg4n:
@Chaphasilor yes of cause would be nice to have that too. Maybe we can extend this functionality later. but i would merge it like that first. and you can still just remove the file manually if it distracts you.

manf0001:
awesome,  I was just thinking about something like this.   As I add songs or re-arrange songs on my playlists, and always found it a bit of a pain to remove, then redownload.   THis is a very welcome feature.  Thank you.  :-)

Y0ngg4n:
maybe @jmshrv can review it so we can merge this.

CameronCarranza:
Just came searching for this, hope it gets reviewed soon definitely a must-have!

NoSadBeHappy:
YAY!! Hope this gets added soon.

JPinSPACE:
I would absolutely love to see this merged. I do think that eventually the "sync" button should also remove songs that are no longer in the playlist otherwise it would be pretty confusing as a user. Also, I do occasionally remove songs that I've gotten sick of.

But this is a huge improvement that would make a real difference for me!

jmshrv:
Merged :)

Issue #221, opened 2022-04-21
Add (etc) album to queue


quite:
Just like we can add a song to the queue (or replace queue/add to playlist), it would be great to be able to do operations on album in a similar manner.

Long-pressing album in album list (and in list of albums by artist) should probably show these operations. And as well somewhere in the album screen. Long-press on header at the top? Or on the Play-button?

The operations should probably be Add To Queue, Replace Queue (well that would be the same as "Play" in album context, I guess), Add To Playlist, and maybe also Goto Artist.

UnicornsOnLSD:
That would be a pretty useful feature, I'll add it :)

BobcatNoah:
I'm still new to github, but shouldn't this issue be closed because it's been implemented?

Chaphasilor:
Once #484 is merged, this will be implemented. The options shown on the album screen or when long-pressing an album in a list will be (in addition to the current ones): "Play next", "Add to Next Up", "Shuffle next", "Shuffle to Next Up". Adding a simple "Play" and "Shuffle" option to the long-press menu would also be possible, but this wasn't available before and therefore hasn't been changed.  
Adding an album to a playlist is not handled yet, and there is a separate issue for this (#497). 

Also @BobcatNoah you are right, although the previous PR targeted the legacy design and this new PR is a more extensive change targeting the same issue, that why this issue hasn't been closed yet. It will be closed once #484 has been merged :)

kuflierl:
Alright, I think that issue has been mostly solved. I am still looking forward to a feature to add the queue to a playlist/make the queue a playlist

Chaphasilor:
@kuflierl that's a different feature, but you could create a separate issue for it!  

I'll close this issue, since all features except "add to playlist" are implemented in the redesign, and for that there's a separate issue: #497

kuflierl:
Alright

PR #481, opened 2023-08-07
Add Play Next & Add to Queue for songs and albums.

ray-kast:
I've seen that the UI redesign project intends to cover this, but I thought I'd offer this as a quality-of-life backport for the current UI until that gets released.  This partially covers #221, and maybe covers #92 depending on opinion.  Apologies for any style errors (and for the faux pas of swinging in out of the blue with a PR), my Dart is pretty rusty.

This has been tested on iOS Simulator and an iPhone 11 running iOS 16.6.

Open questions and future work:
- There is no i18n for this, I'm monolingual
- The icon is confusing but `Icons` doesn't really offer anything better than `queue_music`
- There are no toolbar items in the album screen for queuing, seemed like too big of a UX question for the scope of this PR
- Shuffle is horribly broken ([FIXED](https://github.com/jmshrv/finamp/pull/481#issuecomment-1667626056))
  - ~~The existing Add to Queue behaves badly with shuffle enabled (is there an open issue for this?), and the added ones behave worse~~ see linked comment
  - What is the expected behavior when the queue is modified and then shuffle is enabled/disabled?
  - ~~Does it make sense (and is it worth the effort) to calculate the preimage of `shuffleIndices` when modifying the queue with shuffle enabled?~~

### DIFF
--- a/lib/services/audio_service_helper.dart
+++ b/lib/services/audio_service_helper.dart
@@ -64,17 +64,40 @@ class AudioServiceHelper {
     }
   }
 
+  @Deprecated("Use addQueueItems instead")
   Future<void> addQueueItem(BaseItemDto item) async {
+    await addQueueItems([item]);
+  }
+
+  Future<void> addQueueItems(List<BaseItemDto> items) async {
     try {
       // If the queue is empty (like when the app is first launched), run the
       // replace queue function instead so that the song gets played
       if ((_audioHandler.queue.valueOrNull?.length ?? 0) == 0) {
-        await replaceQueueWithItem(itemList: [item]);
+        await replaceQueueWithItem(itemList: items);
         return;
       }
 
-      final itemMediaItem = await _generateMediaItem(item);
-      await _audioHandler.addQueueItem(itemMediaItem);
+      final mediaItems =
+          await Future.wait(items.map((i) => _generateMediaItem(i)));
+      await _audioHandler.addQueueItems(mediaItems);
+    } catch (e) {
+      audioServiceHelperLogger.severe(e);
+      return Future.error(e);
+    }
+  }
+
+  Future<void> insertQueueItemsNext(List<BaseItemDto> items) async {
+    try {
+      // See above comment in addQueueItem
+      if ((_audioHandler.queue.valueOrNull?.length ?? 0) == 0) {
+        await replaceQueueWithItem(itemList: items);
+        return;
+      }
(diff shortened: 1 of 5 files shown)


Chaphasilor:
Regarding shuffle, yeah it's a mess, mostly due to the way the queue is handled in just_audio/audio_service. Given that it's added functionality it's probably fine if there are issues.

Would you mind posting screenshots of how the user interface for this looks? Can't build it at the moment :)

Chaphasilor:
@ray-kast so it's only an option for albums when long-pressing on the album in a list of albums, correct? I think that's fine for the moment :)

Chaphasilor:
Not sure if that PR is still active. You could clone it and test it yourself, but if that implementation isn't significantly better, I think this one should be merged

rom4nik:
Thanks for working on this! I've closed my PR (#142) since I don't have time to revive it, but I hope that at least the discussions there were valuable.

lib/services/audio_service_helper.dart
```diff
@@ -64,17 +64,44 @@ class AudioServiceHelper {
...
   }
 
+  bool hasQueueItems() {
+    return (_audioHandler.queue.valueOrNull?.length ?? 0) != 0;
+  }
+
+  @Deprecated("Use addQueueItems instead")
   Future<void> addQueueItem(BaseItemDto item) async {
+    await addQueueItems([item]);
+  }
```
jmshrv, line 74: Should this be marked deprecated? It would be more ergonomic to allow adding single items, and would remove the need to change everything that adds a single item

lib/services/music_player_background_task.dart
```diff
@@ -198,9 +261,40 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
   }
 
   @override
+  @Deprecated("Use addQueueItems instead")
   Future<void> addQueueItem(MediaItem mediaItem) async {
+    addQueueItems([mediaItem]);
+  }
```
jmshrv, line 267: Same as the note on `AudioServiceHelper` for completeness 🙃

jmshrv (commented):
These changes look good, just the discussion over `addQueueItems` that needs addressing :)

jmshrv:
Just gave this a test, it works nicely. UX will be improved in the redesign, where a more complex implementation is being made (#484).

Merging :)

PR #496, opened 2023-09-19
Transcoded Downloads

### LINKED ISSUE #215: Download transcoded files
I see that this feature was mentioned to be added in the future if possible in 0.3.0 changelog but I still wanted to highlight as an issue since it turned out to be a dealbreaker for me. Feel free to delete this issue if it's already on roadmap :)
My library consists of very fat FLACs where some of the albums take up to 9Gb. While I have enough space on my phone for songs that I might want to have at the moment (not everyone would), sometimes I want to download some more on the road.

jmshrv:
This PR adds the ability to download transcoded files. OPUS is used, as it is both the best lossy audio compression format, and the OGG container doesn't have any weird skip-back-to-the-start bits when writing, which Jellyfin really doesn't handle, as discussed in the issue.

Because iOS doesn't support OGG/OPUS, this feature will initially be Android-only. Ideally, I'd be able to request CAF/OPUS from Jellyfin, but until https://github.com/jellyfin/jellyfin/pull/9192 is released, this will not be possible, as Jellyfin will only consider the container when transcoding. AAC transcoding may be possible when https://github.com/jellyfin/jellyfin/pull/9249 is released, but if I remember correctly I still had weird issues.

Mostly closes #215


jmshrv:
Looks like this will still be a draft, since not specifying a coded will default to `copy` :(

We'll be able to look into this again once https://github.com/jellyfin/jellyfin/pull/9192 lands.

Chaphasilor:
If I'm not mistaken these changes are now live on 11.8.11/11.8.12:

https://github.com/jellyfin/jellyfin/blob/a173d01139357c5913d3293cfb4c6c53019fbed0/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L5616-L5619

Chaphasilor:
Superseded by #568

PR #510, opened 2023-10-12
Jump to Letter when scrolling, closes #125

### LINKED ISSUE #125: [Feature Request] Jump to Letter when scrolling
From a UX standpoint, there should be some way to jump straight to the letter of the artist or album when scrolling. If I want to play an artists whose name starts with a W, I have to scroll to the bottom of 600+ Artists every single time. I feel like a jump-to nav on the right side should solve that problem, kind of like how there is in the Contacts app on the iPhone. 

TinyContactView-Before-and-After

Guillergood:


Guillergood:
Here's the proof of concept of the feature. Actually works very well. The thing is that if the letter is not on the actual page it does not scroll.
https://github.com/jmshrv/finamp/assets/16701917/95290531-f9b8-4c38-aef1-cb58906b747d

Chaphasilor:
Nice! Have you tried using this without a high-precision pointing device like a mouse? xD  
Maybe we could load the needed page on-demand somehow, but there will be issues with anything that's in-between what was already loaded and the new page. Any ideas?

Guillergood:
No, I haven't.
Tomorrow I could test it with my own phone to check the experience.
I was thinking that maybe, we can pull the letters from the elements in the screen.
However, I feel kind of lost in this pagination. If you can guide me on how to pull the elements' names every time it refreshes we could be more accurate with the letters shown.

Chaphasilor:
I honestly have no idea how to do this properly. Maybe there's a way to only fetch a very small amount of information for each track from Jellyfin, and we fetch the rest of the info when the items are scrolled into view?

As for tests, there are none so far. Adding tests would of course be great, but so far no-one could be bothered to write any 😅

Guillergood:
I think we can have leave as it is with this feature, as it does what the users want and seems reasonable.
I have tested with my own phone and it works well.

jmshrv:
Looks good! I'll give it a try at some point, perhaps we can "keep scrolling" in hope of finding a letter that isn't currently loaded.

As for tests, ...yeah. The current codebase is too spaghetti for tests to be written. @Chaphasilor's new QueueService looks like a good candidate for tests though, and new stuff should be modular enough to allow for easy testing.

Chaphasilor:
I'm just not sure if the PR actually solves the issue:
> If I want to play an artists whose name starts with a W, I have to scroll to the bottom of 600+ Artists every single time.

Since we don't load 600+ artists on a single page, pressing W would do... nothing? So it's not really a solution 🤔  
The jellyfin frontend applies some kind of "filter" when pressing a character, so it will *only* show artists with W, with no way to scroll back up. That would at least solve the actual problem, although I don't think it's very elegant.  
@jmshrv do you know of a way to *exclude* certain fields from the API response? We don't really need the `ServerId`, `ChannelId`, `RuntimeTicks`, `Type`, `BackdropImageTags`, etc. for every artist when showing them in a list...

jmshrv:
Yeah I'm not a fan of how Jellyfin does it.

We could exclude fields when loading the list, but then we can't reuse that item elsewhere in the app (when tapping on an album/playlist, that item is directly passed to the detail view instead of us fetching two things).

I think continually scrolling could work for letters that haven't been loaded yet, we could stop if we go past the letter (in cases where the user selects a letter that doesn't exist)

Chaphasilor:
With continually scrolling you mean the app just scrolls down as quickly as it can? I like that idea!  
We could also maybe double the pagination size. We're streaming music after all, and fetching a few hundred kilobytes per page shouldn't be an issue I think.  
Fetching all my ~1800 artists at once is only about 1MB, so a page size of 500 should be fine?

Guillergood:
> I'm just not sure if the PR actually solves the issue:
> 
> > If I want to play an artists whose name starts with a W, I have to scroll to the bottom of 600+ Artists every single time.
> 
> Since we don't load 600+ artists on a single page, pressing W would do... nothing? So it's not really a solution 🤔 The jellyfin frontend applies some kind of "filter" when pressing a letter, so it will _only_ show artists with W, with no way to scroll back up. That would at least solve the actual problem, although I don't think it's very elegant.
> 
> @jmshrv do you know of a way to _exclude_ certain fields from the API response? We don't really need the `ServerId`, `ChannelId`, `RuntimeTicks`, `Type`, `BackdropImageTags`, etc. for every artist when showing them in a list...

It actually does, however if you want me to display the actual letters, I will do.

Chaphasilor:
It does? In the video you posted nothing happens when you click on "M" on the album screen 🤔

Guillergood:
I have dealt with the pagination an it handles the element loaded. As soon I eat I will upload it.

Guillergood:
@Chaphasilor take a look, now all letters work, as they are in the buffer.
[Working letters](https://github.com/jmshrv/finamp/assets/16701917/5d603367-cb9f-4cd5-a1c6-0ae0b56d294b)

Chaphasilor:
Ah okay, so "unavailable" letters are just hidden, got it. I think I would prefer it the other way around though, were all letters are always shown and if you click on that's not in the buffer the app will scroll down multiple times until the needed page is loaded 😁  
This way the letters are also always in the same spot!

Guillergood:
Hello @Chaphasilor and @jmshrv, take a look 😄 

[Scrolling to letter.webm](https://github.com/jmshrv/finamp/assets/16701917/b4e07a38-e57e-459d-8dba-e8ad8b80a547)

Chaphasilor:
That looks good! 😁  
I don't think we need the error message, just scroll to the next available letter (or maybe slightly before it, so that they can see the letter before and the letter after).

If I have the time I'll test it out once I get back home :)

Chaphasilor:
Just checked it out, it does work for the most part. The biggest problem is load performance, with a page size of 100 it takes ages to get to the correct position, but with bigger page size (250, 500) loading becomes slow and choppy. I'll send a video later.

The other issue is that on some tabs the floating action button covers the letters on a small phone, maybe you could give the button a bit more padding or put it on the left side?

Guillergood:
@Chaphasilor I added some padding to the FloatingActionButton. I think we must define the scope of the task instead of throwing feature ideas 😄. Otherwise, this task will last forever.

Chaphasilor:
@Guillergood you are right of course, and I don't want to overwhelm you with requests. In the end, @jmshrv needs to decide if we can merge this :)

Chaphasilor:
Found a small bug when no item for that letter exists, it will scroll to the bottom of the list first and then scroll to the right position shortly after. Probably waiting for an http request to complete?

https://github.com/jmshrv/finamp/assets/18015147/38cdb2dc-9970-42d0-9d0f-b94ac213c3b6

<sup>(yes my playlists are a bit messy)</sup>

Guillergood:
Hello @Chaphasilor , test now. I have refactored it to have less code (and less possible bugs). Also, to manage the timer in a different stage.

Chaphasilor:
Just took another look, it seems fine. Performance is still poor, especially the scrolling to a letter when the full list is already loaded is pretty laggy. And if nothing is loaded yet and I want to scroll to the letter "M" on my album list, it takes roughly 20s to get there. It will work for small libraries, but even medium libraries will not be a great experience I'm afraid.  
But if you want it to be merged, I guess it won't hurt :)

jmshrv:
We can always polish it later. Merged :)

jzalmanowitz:
That looks great.  Kudos!

PR #511, opened 2023-10-13
Add monochrome launcher icon

AhegaHOE:
In this MR the monochrome launcher icon has been added which enables custom colors for Android users that use the Themed Icons feature.

For more info see: https://developer.android.com/develop/ui/views/launch/icon_design_adaptive

Tested on my Device (Google Pixel 6 with Android 14) and it works, the minimum SDK has been bumped to 26 in order to support the monochrome icon.


android/app/build.gradle
```diff
@@ -45,7 +45,7 @@ android {
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
         applicationId "com.unicornsonlsd.finamp"
-        minSdkVersion 21
+        minSdkVersion 26
```
jmshrv, line 48: Is changing the minimum API level required here? I'd rather not drop so many (albeit old) Android versions

jmshrv (commented):
Nice change, just need to make sure about the API level change

android/app/build.gradle
AhegaHOE: I tried building the app with the old version and it threw an error stating that the minimum API version is 26. 🤔

AhegaHOE: > Next, in res/mipmap-anydpi-v26/ic_launcher.xml, create alternative drawable resources in your app for backward compatibility with Android 8.0 (API level 26). Use the <adaptive-icon> element to define the foreground, background, and monochromatic layer drawables for your icons. The <foreground>, <background>, and <monochrome> inner elements support the android:drawable attribute.

This was on the site that I provided in the MR description, seems like anything lower than Android 8.0 (API level 26) is no longer supported with this change.

Chaphasilor:
How about we use `redesign` as the target branch? That one will be a breaking change anyway, and during the beta we can gauge if there is demand for older versions of Android 😁

AhegaHOE:
Sounds good to me 🦾

AhegaHOE:
I tried rebasing onto `redesign` but somehow it get's all weird on my end. So I changed it back to `main` for now as I wasn't able to push the rebased changed to my branch...

 ```
 To https://github.com/AhegaHOE/finamp.git
 ! [rejected]        add-material-design-icon -> add-material-design-icon (non-fast-forward)
error: failed to push some refs to 'https://github.com/AhegaHOE/finamp.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. If you want to integrate the remote changes,
hint: use 'git pull' before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
 ```
 
already tried pulling, fetching and what not but the push is still getting rejected. Sadly I'm not too familiar with GitHub as I'm mainly working with GitLab.

Chaphasilor:
@AhegaHOE if you use VS Code, try checking out your branch and the using the "Merge branch" command and select the redesign branch. If that doesn't work, I guess you should be able to perform a `git reset origin/redesign --hard` on your `add-material-design-icon` branch and then either cherry-pick or manually re-apply your changes, since there are only two changed lines and some files added :)

AhegaHOE:
Merged `redesign` into my `add-material-design-icon` hopefully this is correct now! I somehow completely forgot about merging 😂

Chaphasilor:
Looking at the changed lines, something seems to have gone awry 🙃  
I'll take a look tomorrow...

AhegaHOE:
That seems a lot better now 😂

Chaphasilor:
Yeah, looks good! @jmshrv any idea why the build might be failing? 😅

AhegaHOE:
When building I came across the same issue. Running `flutter pub cache repair` before running `flutter pub get` fixed it. The build fails because flutter has still some invalid caches that need to be defreshed first and only `flutter pub get` won't work.

Chaphasilor:
In that case, maybe try running these commands and the committing the updated `pubspec.lock` file?

AhegaHOE:
Well, the `pubspec.lock` doesn't change so it won't impact the Pipeline. We either have to wait for the caches to be invalidated/updated or temporarily change the definition of the Pipeline.

You can try to reproduce it locally:

```
flutter pub cache clean
git checkout origin main
flutter pub get
git checkout add-material-design-icon
flutter pub get
flutter build apk --debug
```

Chaphasilor (approved):

Chaphasilor:
Also confirmed that this is working as intended on Android 13 as well 😁

AhegaHOE:
> There we go 😉  
> The lockfile was messed up and contained packages that weren't being used :)

What did you end up doing to fix it?
`flutter pub upgrade` by any chance?

Chaphasilor:
@AhegaHOE sorry didn't see your comment. I deleted the global flutter cache (`fluter pub cache clean`) and deleted `pubspec.lock`, then built the app. This generated a new and valid lockfile :)

@jmshrv should I merge this?

jmshrv:
Yeah go ahead, API 26 is 93% of users ([source](https://apilevels.com/)), so it shouldn't be too bad

AhegaHOE:
@Chaphasilor no worries and thank you for the information 😁

Chaphasilor:
Finally merged :) Thanks a ton!

AhegaHOE:
It ain't much but I hope that a lot of people will enjoy it! I'm happy to contribute at least a little bit to this awesome project. ^^

Issue #482, opened 2023-08-09
Shuffle is broken on android, and skips songs of playlist


AbuD3vTs4113:
Since the last release 0.6.15, the feature shuffle on playlist is broken. I listen some songs and some time after that the playlist was already done. So I can't listen to all my music. If anyone has had this problem before, can you confirm that you are having this bug as well?

Please can you fix it

flloschy:
Shuffle is defaulted to only shuffle 250 songs in one session, can be changed in the settings under audio service.

Its the only thing I can imagine what you are referring to.

AbuD3vTs4113:
But even with playlists of 25 songs playback skips songs, the shuffle don't read every songs of my playlist. I have to go back to the previous version to avoid this bug..

AbuD3vTs4113:
But even with playlists of 25 songs playback skips songs, the shuffle don't read every songs of my playlist. I have to go back to the previous version to avoid this bug..

flloschy:
Just tested with a 28 song playlist (on IOS) (app verison 0.6.15) and all 28 songs appeared in the queue

AbuD3vTs4113:
of course the queue display all songs of playlist after shuffle but the player skips some songs

jmshrv:
@AbuD3vTs4113 could you send logs? Make sure you replicate the bug first - logs are not saved from previous sessions

> Just tested with a 28 song playlist (on IOS) (app verison 0.6.15) and all 28 songs appeared in the queue with the default of 250 shuffled songs

That setting only applies to the "shuffle all" button on the songs tab

flloschy:
Oh. My bad

AbuD3vTs4113:
> @AbuD3vTs4113 could you send logs? Make sure you replicate the bug first - logs are not saved from previous sessions
> 
> > Just tested with a 28 song playlist (on IOS) (app verison 0.6.15) and all 28 songs appeared in the queue with the default of 250 shuffled songs
> 
> That setting only applies to the "shuffle all" button on the songs tab

Here the logs:

MusicPlayerBackgroundTask/INFO] 2023-08-12 14:51:56.432851: Starting audio service

null
[DownloadsHelper/INFO] 2023-08-12 14:52:16.858847: Song eWE.eU~XRjInSdM}M|j[R%D%RjRjR%R%RjbDR%Rjofxut7xuofRjn% exists offline, using local file

null
[...]

PR #515, opened 2023-10-18
Fix #514

### LINKED ISSUE #61: Shuffle starts on first song

### LINKED ISSUE #514: Shuffle on Artist does not queue correct
If you shuffle on an artist it plays the first song and then the last in the queue and then stops.

Chaphasilor:
Fixes the "shuffle always starts on first song" issue in the same way as originally done in #442


Chaphasilor:
I added the fix for albums and playlists as well, but it seems like there is a conflict between fixing this and fixing #482 😕  
I personally cannot reproduces the issue from that other issue, but we should try to test for it before merging my PR...

Chaphasilor:
Given that there are still issues with shuffle starting on the first track (see https://github.com/jmshrv/finamp/issues/545), I'll merge this in for now. Unless something severely breaks, this will likely stay in place until the redesign hits stable...

PR #484, opened 2023-08-11
[Redesign] Brand New Queuing System ✨

Chaphasilor:
I can't believe this is finally in a state where I feel like we can think about getting this merged in.  
It was an absolute pain to implement this, and I'm barely happy with the end result, but it should be (much) better than the old queueing system, both functionality and design-wise.

Here's a quick break down:

- New `QueueService` for managing everything queue-related
- Queue is split into four parts: previous tracks, current track, next up, queue
- Support for adding tracks to "next up" (fixes #92)
  - Not affected by shuffle
- Consistent behavior (this was the most work)
- Quality of life improvements (show what playlists/album/etc. is currently playing = source)
- Improved queue design, with more functionality

### Screenshots

Screenshot_20230811-235039

Screenshot_20230812-000620

Screenshot_20230812-000620

### TODOs

- [x] Fix "shuffle always starts on first song" bug
- [x] Remove x and 3 dots icons, make list items dismissable and add long-press handler
- [x] Implement collapsible previous tracks? (like in redesign mockup)
- [x] Support starting playback at a specific index (e.g. when tapping on a track within an album or playlist)
- [x] Handle more playback sources (playlists show up as albums right now)
- [x] Go to parent item when tapping on the source string on the player screen
- [x] Support adding albums/playlists/etc. to queue or next up (shuffled or original order)
- [x] Finish up the design (dynamic colors, background, conditional rendering)
- [x] Finish up playback history view (basic functionality-wise, still needs a proper design through another PR)
- [x] Fix like button not having the correct state
- [x] Replace strings with localized versions
- [x] Add vibro-feedback for improved UX
- [x] Sticky list headers?
- [x] Add clear button for next up list
- [x] Improve Playback Reporting (based on #477)
- [x] Fix rewinding
- [x] Remove excessive debug logging and re-enable chopper logs
- [x] Use linear playback order for Instant Mixes (shuffle breaks song mixes, and the mix should be fully determined by Jellyfin anyway)
- [x] Fix label thickness
- [x] Reword "recently played" to "previous tracks"
- [x] Add "Album/Playlist/etc. added to Next Up" as source, handle it in the player header
- [x] Check if `minBufferDuration` is smaller than `maxBufferDuration`  
   - Otherwise the app needs to continuously fetch new data
- [x] Re-shuffle on shuffle enable (perhaps when turning shuffle *off* so that at the start it isn't shuffled twice)
- [x] Navigate to source when clicking on "Playing from" label (not working with single-track albums?)
[...]

PR #517, opened 2023-10-22
✨ Audio Loudness Normalization (Replay Gain)

Chaphasilor:
(This PR is based on #484 and should only be merged afterwards)

This adds LUFS-based loudness normalization ("replay gain"), based on https://github.com/jellyfin/jellyfin/pull/9222 and https://github.com/jellyfin/jellyfin/pull/10085.  
For this PR to have an effect the server has to be on the "unstable" release channel of Jellyfin. The feature will be released to the stable channel as part of Jellyfin 10.9.  
~~This currently only works on Android, because it uses [`AndroidLoudnessEnhancer`](https://pub.dev/documentation/just_audio/latest/just_audio/AndroidLoudnessEnhancer-class.html) which doesn't have an equivalent on iOS.~~ I'm looking into emulating this for iOS simply by changing the player's volume accordingly.

There's a new settings page for this, which lets the user customize the target LUFS (default is `-14.0`, like on i.e. Spotify), a factor which controls by how much the loudness should be changed towards the target LUFS (the "aggressiveness", default is `1.0`), and the normalization mode.  
For iOS and other non-Android platforms, there's also a setting that lets the user specify a base gain that is used for emulating proper output gain. This value should be negative, so that the volume can be increased to boost the volume of quiet tracks. The default is -5 dB (75% volume), which leaves some headroom but doesn't reduce the volume too much. My test library almost exclusively contained tracks that had a LUFS value greater than the default target of -14 LUFS, so a gain greater than 1 doesn't seem to be too common.

There are three modes:
- "Hybrid" (default): Tracks will be normalized based on each track's LUFS value, but when playing an album the album's LUFS value is used (only for tracks from the same album)
- "Tracks only": Normalization happens only based on the track's LUFS value, the album's LUFS value is ignored
- "Albums only": Normalization only happens when playing albums, based on that album's LUFS value

Screenshot_20231022-173137

### TODOs

- [x] Emulate gain changing on other platforms by modifying the player volume
- [ ] ~~Add feature detection so the setting is only available if the server supports it~~
  - Maybe add some kind of framework for feature detection in another PR?


jmshrv:
The code in this PR looks good, I'd be happy to merge it :)

Chaphasilor:
Ahh good catch, I added some code to handle that. If you could make sure I didn't breaking anything before you merge it that would be great 😅

Titaniumtown:
Wanted to say that it's awesome this got merged, huge improvement to user experience! Thanks for your continued hard work @Chaphasilor!

Chaphasilor:
Thanks @Titaniumtown! Were you able to try it out already? :)

Titaniumtown:
> Thanks @Titaniumtown! Were you able to try it out already? :)

Yes I built it and tried it out, works like a charm.

Issue #108, opened 2021-08-18
Easy way to add currently playing song to playlist


count-null:
To add a song to playlist, you have to find it in the main list (or search for it). 90% of the time I want to add a song, it's because it's currently playing. 

Great work all! This is the best mobile streaming app I've ever used for self hosted libraries.

UnicornsOnLSD:
Added a button on the player screen in 878f2b66b0c6049a88075bf9b76114775cabe55a

Issue #106, opened 2021-08-17
Failed to connect to server - HandshakeException


endervad:
I've set up a self-signed certificate on my Jellyfin server and installed it also as a trusted root CA on my devices. It works fine without security warnings on Chrome (both Windows and Android) and on Android apps, like official Jellyfin and Gelli, except Finamp. When I try to connect to my server in Finamp 0.5.1, I get
`HandshakeException: Handshake error in client (OS Error: CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(handshake.cc:359))`

Certificate extensions:
`X509v3 extensions:
    X509v3 Key Usage: critical
        Digital Signature, Key Encipherment
    X509v3 Extended Key Usage:
        TLS Web Client Authentication, TLS Web Server Authentication
    X509v3 Subject Alternative Name:
        DNS:<redacted>
    X509v3 Basic Constraints: critical
        CA:TRUE
    X509v3 Subject Key Identifier:
        <redacted>`

UnicornsOnLSD:
Try this APK, based off the allow-bad-certs branch: 
[app-arm64-v8a-release.zip](https://github.com/UnicornsOnLSD/finamp/files/7001443/app-arm64-v8a-release.zip)
[app-armeabi-v7a-release.zip](https://github.com/UnicornsOnLSD/finamp/files/7001445/app-armeabi-v7a-release.zip)
[app-x86_64-release.zip](https://github.com/UnicornsOnLSD/finamp/files/7001446/app-x86_64-release.zip)

This just allows all bad certificates, so I won't put this in the main branch. Just wondering, why don't you just use Let's Encrypt to get a proper certificate?

endervad:
This one works, thanks :)
I have a probably rather unique case: I've got a DNS name to my public IP from ISP. It's not part of contract, and I'm not even sure that it was intended, but I don't pay for it so might as well use it as is.

endervad:
Upon further usage of allow-bad-certs version it seems like I was careless to assume, that if I can see the library, everything works. Neither downloads, nor playing songs work.
In case of downloads there's only a single `PlatformException((0) Source error, null, null, null)` in Main Thread > AudioServiceHelper with null stack trace. And in playing there's the exact same handshake exception for every attempt to play any song in Audio Service > MusicPlayerBackgroundTask with null stack trace.

UnicornsOnLSD:
[...]

PR #518, opened 2023-10-22
Add Initial Certificate Override Support

OdinVex:
Back-end work to provide certificate override functionality added. User Interface to manage history and provide context should be added. Clarification when new certificates for a `host:port` combination (if invalid) are added versus never having been approved should be given to the user. Allow them to reject if the remote end has changed (or is behind a load-balancer with multiple certs/ports...)

Things still needed to complete. When users connect **any time** *and there is a certificate error* several things should happen:

1. The back-end should be queried to determine whether any invalid certificate has ever been overridden for the `host:port` combination.

2. If no certificate has been overridden present a dialog to allow the connection temporarily, permanently, or cancel. The dialog should **warn the user that their connection may not be secure as the certificate supplied is invalid\*.**

3. If a certificate has been overridden in the past but this certificate is a different invalid certificate **warn the user that the previous certificate used to access the server has changed.** Prompt the user to allow the connection temporarily, permanently, or cancel.

Provide a means of editing these overrides. The front-end just needs development now. I didn't want to implement any bridge between the front-end and back-end (such as a way to retrieve the actual overrides), I figured that would be best left to someone that can use Dart/Flutter much more effectively.

\* Also note that you'll need to modify the "on connection error dialog" area of code to actually differentiate what kind of error occurred instead of just showing that small notification. Try to determine if it is invalid because the certificate is self-signed, expired, or could not be verified. I'm not sure of the specific error codes, but it looked like this uses code-words (unique and in latin, so ASCII). That will make it easy if you can't do it by error code. \*

Helpful Code Specifics:

```
Files:
lib/main.dart
lib/models/finamp_models.dart
lib/models/finamp_models.g.dart
lib/services/finamp_settings_helper.dart

Classes:
MyHttpOverrides (lib/main.dart)

Functions:
FinampSettingsHelper.addCertificateOverride(host, port, thumbprint)
FinampSettingsHelper.deleteCertificateOverride(host, port, thumbprint)
FinampSettingsHelper.resetCertificateOverrides()
FinampSettingsHelper.hasCertificateOverride(host, port, thumbprint)

Variables:
Map<String, Set<String>> FinampSettings.overriddenCertificates \* Expects "host:port" key, "thumbprint" value.
```

\* Invalid could also simply mean self-signed or not trusted for whatever reason. If it's expired, inform them. If it's self-signed, inform them. If it's an unrecognized authority, inform them.

Edit: Corrected three brain-fart typos, `cert.sha1` to `thumbprint`.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -9,7 +9,7 @@
          In most cases you can leave this as-is, but you if you want to provide
          additional functionality it is fine to subclass or reimplement
          FlutterApplication and put your custom class here. -->
-  <application android:name="${applicationName}" android:label="Finamp" android:icon="@mipmap/ic_launcher" android:usesCleartextTraffic="true" android:requestLegacyExternalStorage="true" android:allowBackup="false" android:fullBackupContent="false">
+  <application android:name="${applicationName}" android:label="Finamp" android:icon="@mipmap/ic_launcher" android:usesCleartextTraffic="true" android:requestLegacyExternalStorage="true" android:allowBackup="false" android:fullBackupContent="false" android:networkSecurityConfig="@xml/network_security_config">
     <activity android:name="com.ryanheise.audioservice.AudioServiceActivity" android:launchMode="singleTop" android:theme="@style/LaunchTheme" android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode" android:hardwareAccelerated="true" android:windowSoftInputMode="adjustResize" android:exported="true">
       <!-- Specifies an Android theme to apply to this Activity as soon as
                  the Android process has started. This theme is visible to the user
--- a/android/app/src/main/res/xml/network_security_config.xml
+++ b/android/app/src/main/res/xml/network_security_config.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <base-config cleartextTrafficPermitted="true">
+        <trust-anchors>
+            <certificates src="system" />
+            <certificates src="user" />
+        </trust-anchors>
+    </base-config>
+</network-security-config>


jmshrv:
Thanks for working on this! This has been a popular request for ages. The code looks good, we just need a frontend. The most secure way of handling this would be checking thumbprints as you said in #106, as you may as well not use HTTPS at all if you're just going to ignore any certificate errors.

If I have time tomorrow, I'll take a look at the TODOs here and finally get this issue fixed.

jmshrv:
> If I have time tomorrow

I did not... Tomorrow hopefully :)

Chaphasilor:
Just some thoughts regarding the integration into the login flow (because that's the part where this feature will be used most of the time):

1. The user starts by entering the server URL. We *could* try to fetch the server information after nothing has been entered for ~750ms, with certificate checking fully disabled (just to see if the URL is correct and pointing to a Jellyfin server)
2. After the user confirms the URL, we again fetch the server and user info, but this time with certificate validation enabled. So if the server uses a self-signed cert, we would prompt the user about it here, presenting the a short description saying what the issue is, showing some info about the cert, and two buttons. This way we also don't need a checkbox or advanced settings.
3. After the user logged into the app, they can manage the known certs in a new section within the settings. The section would contain all certs/thumbprints for the currently used domain only, since we would need to logout anyway to use a different domain. Next to each thumbprint, there should be a delete button, and tapping on a thumbprint should open a modal or bottom sheet showing any information that we have about the cert, with a close and a delete button.

Also, I'm not sure if the current backend code solves the issue of native code using its own certificate validation, e.g. when streaming music with Exoplayer or downloading songs. Hopefully there are ways to re-use the existing logic for these, e.g. by using some just_audio/flutter_downloader constructor argument...

PlexSheep:
Another idea that might cover some usecases while staying true to security would be to let the user add their own CA. That would help distinguish between malicious or bad certificates and real certificates, that are not recognized by the app.

PlexSheep:
@OdinVex I know the library does not allow it (as mentioned in #108). My idea was rather to be able to add the CA in the app, not the system.

But I don't know a lot about android development, maybe it's not doable.

jmshrv:
We may need to provide a network security configuration for Android - https://developer.android.com/privacy-and-security/security-config

Also I may have to move my deadline up to the weekend, I've had no time to work on Finamp this week 😔

jmshrv:
Android started ignoring the user certificate store a while ago for security:tm:. See https://developer.android.com/privacy-and-security/security-config#base-config

jmshrv:
ugh, iOS also needs manual intervention...

Almost as if self-signed certs are a bad idea ;)

jmshrv:
hmm maybe not? I may have messed up my certificate configuration

jmshrv:
Yep, needed to specify the IPs that I was using. Luckily this is only a slight pain to fix now 🙃

jmshrv:
Did some more work on this today (not ready to commit yet), mostly plumbing for showing the certificate error screen

Chaphasilor:
Why did you close the PR? :)

Chaphasilor:
Haha okay, that makes sense! I'll create a new branch here on this repository and re-implement your changes there. Should I commit it under your name/email?

PR #522, opened 2023-11-01
:eyes: Disable scrollbar when not sorting by name

Maxr1998:
Apologies for the messy PR, there was apparently a lot of incorrectly formatted code as well which Android Studio auto-formatted for me.


Chaphasilor:
I can't figure out what the actual change is. Which scrollbar is being disabled and why? 🤔

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -397,59 +397,62 @@ class _MusicScreenTabViewState extends State<MusicScreenTabView>
...
+                              isGrid: true,
+                              gridAddSettingsListener: false,
+                            );
+                          }
+                        },
+                      ),
+                widget.sortBy == SortBy.sortName
+                    ? AlphabetList(
+                        callback: scrollToLetter, sortOrder: lastSortOrder)
+                    : const SizedBox.shrink(),
```
Maxr1998, line 455: This is the actual change (1).

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -480,86 +483,89 @@ class _MusicScreenTabViewState extends State<MusicScreenTabView>
...
+                                    .contentGridViewCrossAxisCountLandscape
+                                : box
+                                    .get("FinampSettings")!
+                                    .contentGridViewCrossAxisCountPortrait,
+                          ),
+                        ),
+                  widget.sortBy == SortBy.sortName
+                      ? AlphabetList(
+                          callback: scrollToLetter, sortOrder: lastSortOrder)
+                      : const SizedBox.shrink(),
```
Maxr1998, line 568: This is the actual change (2).

Chaphasilor:
Okay, now I get it. Makes sense 👍🏻  
If you want me to I could just force-push the branch and split it into two commits before merging it...

Maxr1998:
Did just that myself now :smile:
I actually wanted to do that since I made my last comment, but forgot about it. Thanks for giving me another poke.

Chaphasilor:
Awesome. I think we can ignore the failing build for now, that problem needs to be solved somewhere else...

Maxr1998:
It's probably because I rebased onto the latest master..

jmshrv:
The GitHub Action always uses the latest stable Flutter, which might've changed something tiny that broke the build

Chaphasilor:
I think this can be merged. I'll do that in a few days unless anyone sees a problems with it...

Maxr1998:
That would be great, thank you!

Chaphasilor:
Merged, thanks :)  
Would you be willing to create a small PR that adds a setting for turning off the scrollbar completely? 😁

Maxr1998:
Probably, bit busy though at the moment. Maybe within the next few days.

Issue #237, opened 2022-05-20
[Feature Request] Cross Fade


Ge082:
What I like about PlexAmp is that once a song end by itself, there is a crossfade (Which they call "Sweet Fade) so the transition between the current song and the next song is smooth, it's very effective on Live albums because it eliminates the micro moment where there is silence that is there when the next song loads
Would be cool to have the same feature

B16WLL:
Or (if it's possible) pre-cache the opening few seconds of the next track in the queue, and implement a capless playback option for live/mixed albums.

rom4nik:
Isn't the playback gapless already? I've listened to some live albums and always got smooth transition between tracks without any pauses (unless the artists put them in, that is).

waynepaulward:
It would be great if there was a smooth fade between tracks I suppose it would have to cache tracks when you start to play the album so it can turn it can start by playing the track - is this a feature that will be added …

waynepaulward:
I should have explained this better I mean like a cross fade option so the songs fade together smoothly

jmshrv:
This isn't implemented in just_audio due to platform support (https://github.com/ryanheise/just_audio/issues/815), it may be possible once I get around to the new player backend (#177)

BloodyIron:
I too am interested in this for streaming, but gapless does seem to work in my brief testing with downloaded content.

waynepaulward:
They use this feature on a few open source apps ie play sub and avsub and it works well they are open source and may help

ryanwwest:
@BloodyIron Do you know if gapless audio worked with downloaded content _when not in offline mode_? I'm guessing Finamp is smart enough to always use the downloaded/cached version and so the seamless audio would apply, but unsure.

jmshrv:
Yeah Finamp always defaults to downloaded content when it's available

BobcatNoah:
Would it be possible to implement cross fade by creating two instances of the music player (It seems like it may be possible)? I'd be happy to look into it if it's not too impractical.

Chaphasilor:
@BobcatNoah yes, that was the approach suggested by Ryan Heise over in the just_audio issue. However, I'm not sure how well thing would work, it could severely complicate things regarding pre-caching and queue handling. We would have to split the `ConcatenatingAudioSource` into two sources, one for even and one for odd indices. But this would result in parallel pre-caching, which is less efficient data and memory usage-wise.  
I'm not sure if it's worth the tradeoff...

[...]

PR #523, opened 2023-11-01
Cross fade #237

BobcatNoah:
This PR is work in progress and will implement cross fade #237 between songs. The code is not complete because I want someone to review it before I move forward and there is a major issue I need help with.

I don't have a video example right now, but the issue is that the cross fade is being paused before the seeking is finished. This causes a noticeable gap on slower devices or when the device is connected to something like a car stereo. The expected behavior is that it waits until the seeking is finished before pausing. I suspect this is being caused by a delay between when the seek function finishes (thus returning and letting the code continue) and when the song actually starts playing again.

```
_player.currentIndexStream.listen((event) async {
        if (_player.processingState == ProcessingState.ready) {
          //fadeIsReady = false;
          _player.setVolume(1);
          if (event != null && nextItem == _getQueueItem(_player.currentIndex?? 0)) {
            await Future.delayed(const Duration(milliseconds: 1));
            //final test = Stopwatch();
            //test.start();
            await _player.seek(_crossFadePlayer.position);
            await Future.delayed(Duration(milliseconds: 15));
            _crossFadePlayer.pause();
            //test.stop();
            //_audioServiceBackgroundTaskLogger.info(test.elapsedMilliseconds);
            //finishedSwitchingSong = true;
          }
        }
      });
```

Below is a video of expected behavior. This example isn't perfect though due to the issue.

https://github.com/jmshrv/finamp/assets/84159993/57ed007e-c16e-4a84-a307-03257b7bd9e9

### DIFF
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -111,6 +111,8 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
   /// progress reporting.
   bool _isStopping = false;
 
+  Duration crossFadeLength = const Duration(seconds: 10);
+
   /// Holds the current sleep timer, if any. This is a ValueNotifier so that
   /// widgets like SleepTimerButton can update when the sleep timer is/isn't
   /// null.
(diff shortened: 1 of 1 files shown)


jmshrv:
I believe @Chaphasilor was working on this, they may be able to help :)

Chaphasilor:
I was not, but my suggestion would be to pause the crossfading player a few seconds later. Maybe fade the volume down again until the actual track is playing?  
Otherwise, if the timing still isn't perfect, it might be necessary to create two full-blown player instances and manage them somehow, which seems like a nightmare...

BobcatNoah:
The timing isn't perfect and I can't think of any way to account for the latency added by seeking. Fading the volume down may work, but I'm not sure how well. I have no way knowing when the track is actually playing after the seek and just adding a few seconds would sound bad on devices that seek really fast.
I don't think crossfade is worth creating two full-blown player instances. It definitely would be a nightmare.

Chaphasilor:
Hi @BobcatNoah, sorry that I never managed to take a closer look.  
I was planning to do that once the initial beta release is done. Is there any particular reason why you closed the PR? :)

BobcatNoah:
From my testing, I noticed that different devices had different delays for when you call the play method and when it actually starts playing. This meant I can never perfectly sync up the transition between the songs. On my phone, it might sound fine but on a car stereo, there might be a noticeable delay.

BobcatNoah:
Alternatively, I could try the original method of having two separate players in which Finamp would alternate between the two.

image

Chaphasilor:
Hmm, okay. Maybe just_audio will get cross-fase support at some point, but that might be a while. The two player approach would be the best solution for cross-fading, but I don't think it would play nicely with the queue and the player itself.  
But you're welcome to give it a try, if you feel like it!

BobcatNoah:
I didn't even consider the queue being a problem. I guess I'll have to put this on hold until just_audio add crossfade.

PR #534, opened 2023-11-13
Add sync button to playlist screen

shayaantx:
Downloads playlists on the playlist screen (will add anything missing and remove anything that doesn't exist in playlist anymore that is downloaded)


Chaphasilor:
Thanks for the PR!  
Is this similar to #460?  
Maybe you could add a small description or a short video that clarifies how this works :)

shayaantx:
added summary (lemme know if you want some screenshots or a video)

@Chaphasilor it is similar, however my understanding of https://github.com/jmshrv/finamp/pull/460 is that it will sync every single parent (which i think could be an album or playlist) that exists on the downloads screen.

The difference here is this lives solely on the playlist screen (not albums), this also removes any songs that are no longer part of the playlist anymore (need to test some edge cases with this tho). This also only syncs whatever playlist you are currently viewing (not all the playlists like the other button)

shayaantx:
@Chaphasilor the only clean way I see to reuse that code would to _not_ reuse it and replace that code with this code (since this removes items too)

I can try to refactor it and see what I can come up with it

Chaphasilor:
Okay, then maybe you could split the logic into a separate file and function(s) so that it's easier to replace the existing code later on. Once I get around to testing this out, we can try to get this merged 😁

Chaphasilor:
Not sure why the build it failing. I'll take a look soon!

shayaantx:
@Chaphasilor I was looking into the case where you have a downloaded item shared across playlists/album

so it won't remove the downloaded item as long as the user synced both album and playlist (or playlist and playlist or album and album etc etc)

So the user would have to had to press sync on the album/playlist with the song, then also sync on the other, then the DownloadSong::requiredBy won't be empty.

I also wasn't passing "deleteFor" to DownloadsHelper::deleteDownloads() which was making it always remove it everytime (I fixed this in my recent push).

I refactored it out as requested, still looking into possibly reusing this logic for the SyncDownloads stuff (just been busy - will try to look more this weekend)

Chaphasilor:
That sounds great! My own big PR got merged in recently so I'll be taking a look at the other open PRs soon, including yours. Should I wait until you looked into the reusability?

shayaantx:
> That sounds great! My own big PR got merged in recently so I'll be taking a look at the other open PRs soon, including yours. Should I wait until you looked into the reusability?

Yea I'll repost once once I've figured out the reusability part

shayaantx:
@Chaphasilor so I refactored a few things locally:

1. Replaced download button on album/playlist with just a delete button if the album has been downloaded (so I removed the download button in lieu of the sync button I added)
2. Then replaced the underlying logic sync downloads button on downloads page with my sync logic

However I found one issue, specifically with the delete part of the logic. If you recall before I mentioned I wasn't using "deletedFor" attribute when DownloadsHelper::deleteDownloads()

This particular function always removes whatever the deletedFor value is from the parent caches (_downloadedParentsBox for example). This results in every sync removing the parent item (which basically makes the app look like nothing was downloaded)

I'm investigating the best way to address this, basically seems like I might need to refactor some of the delete logic to be separate so I don't delete parent unless the user asks to delete the entire downloaded album/playlist and in the sync logic just delete items (unless they sync and the album/playlist is empty)

Chaphasilor:
Ah I see, that makes sense. I haven't looked at the code yet, but maybe adding a `removeParentItem` flag would help, or even better detecting if the parent has any downloaded children and not deleting the parent if that is the case.  
But I guess that depends on how the downloading logic handles deleting a parent (including any children), for example if yoi want to get rid of a previously downloaded album.

I'm not familiar with the downloading stuff myself, maybe @jmshrv has some additional input...

shayaantx:
@Chaphasilor yea I did something like that locally, need to test it a bit, then ill repush

- Specifically I renamed the existing deleteDownloads to deleteParentAndChildDownloads since that really attempts to force delete all the downloads (including parent if you give)
- So I split out deleting children and deleting parent, then in my sync delete function I explicitly check if the parent has anymore children left, and delete it if it does not.

Chaphasilor:
Awesome! Looking forward to testing it out 😁

shayaantx:
@Chaphasilor i tested out the changes with my own usage (seems fine), let me know if you find anything weird

Behavior now:
- Now there is a delete and sync button (delete wipes out all offline downloads), sync will add/remove any changes to playlist/album
image
- If you have never synced the playlist/album, the delete button doesn't show, but the sync button will always show
image
- If you remove all items from a playlist (in jellyfin), then sync, it will remove all downloaded items, but leave behind the playlist (which can be manually removed) - the entries for each downloaded item will be gone.
- If an item is shared across offline downloaded album/playlists, that item should not be removed (at least in my testing it was not). The caveat to this is, you must actually sync that offline item for both album/playlists (i.e., if you never click sync on playlistB but did on playlistA, the requiredBy field for the downloaded item has no reference to playlistB - which makes sense I think)

Chaphasilor:
Did some quick testing, so far everything works 😁  
Also merged the main branch to fix some build errors due to conflicting dependencies for me.  
I'll look into merging this once #197 is merged. I'm guessing there will be a few merge conflicts, as both PRs deal with the download button...

Chaphasilor:
Does your implementation add the new downloads to the downloads "queue"? So that it shows up on the "Downloads" screen as `enqueued` or `running`?  
Because I noticed that the "Retry" button that attempts to re-download failed items does not increase the "enqueued" amount at all...

shayaantx:
> Does your implementation add the new downloads to the downloads "queue"? So that it shows up on the "Downloads" screen as `enqueued` or `running`? Because I noticed that the "Retry" button that attempts to re-download failed items does not reliably increase the "enqueued" amount...

I would assume this issue has always existed, cause looking at original code there is nothing explicit for that (that looks to all be within the downloads helper). I've changed the stuff above that (i.e., what to download and what to remove)

I will also mention that button (sync button on downloads screen), doesn't really do anything with UI when you click it, however I see it sync-ing my downloads if I leave the screen and come back.

Chaphasilor:
Yeah, the UI part also needs some work. But that's a different story...

Is this PR ready to merge from your side?

Chaphasilor:
I noticed the download button is only replaced with the sync + delete buttons after leaving and coming back to the album/playlist. Not sure if that's caused by the merge, but we should definitely try to fix that :)

shayaantx:
> I noticed the download button is only replaced with the sync + delete buttons after leaving and coming back to the album/playlist. Not sure if that's caused by the merge, but we should definitely try to fix that :)

Prior to your merge, I clicked download, then sync button appeared + delete button.

After your merge, I do same thing, and see buttons update.

(All I did was make a new playlist and saw appropriate button states)

Chaphasilor:
So it's working like before? 🤔

shayaantx:
> So it's working like before? 🤔

For me it is, not sure if you consistently reproduce the behaviour you mentioned - if there are some specific steps to reproduce let me know and I can take a look

shayaantx:
> :

So I'm not able to reproduce that behavior you showed in your video on my phone or the emulator. When I click download, the songs in the playlist will start to show their individual downloading state icons. Then the button switchs at the end.

I'm not sure why this is happening in your case

I'll upload a video later tonight with what I see

shayaantx:
> Hmm, maybe try it with a longer playlist or on Android?

So I saw this behavior on my phone (which I can take a look into).

I saw another issue, basically if i I have a playlist that has never been downloaded, but it has items that were downloaded from another playlist, it doesn't link those existing items to the new playlist (I'll look into this)

shayaantx:
yea so I logged out more, basically at some point there was some downloaded items for a given playlist, then they were not fully removed, so they are stuck (they don't display in downloads page either).

Then the delete button doesn't remove them, cause that button only removes whatever downloads screen thinks is downloaded (so if they are orphaned they won't be found)

So I think one change we may want to consider, if someone deletes their downloaded album or playlist, we should get all items in playlist and make sure we remove it from everywhere (all the cached download collections). Then at least users can restart in case of corruption or what not.

The other option would be a much larger change, basically stop using these collections to drive whether you downloaded something, instead use like parent id and file paths (i.e., whether the file existss)

Thoughts?

Chaphasilor:
> seems like we should be just refreshing the tab when it its selected/comes into focus.

I think pull-to-refresh should work, and maybe we could patch an album's info in the album list response if we open the album page, but refreshing every time we go back seems unnecessary 🤔

Chaphasilor:
Regarding the download improvements, the download system is due to be rewritten and I think work has already kind of started. But clearing orphans seems like a good solution for now.  

Is there a way to detect these orphans even if other tracks are properly downloaded, or do we need to delete all other downloads first and check what is left?  
Because if we can detect them, we could simply remove all orphans on startup.

shayaantx:
> Regarding the download improvements, the download system is due to be rewritten and I think work has already kind of started. But clearing orphans seems like a good solution for now.
> 
> Is there a way to detect these orphans even if other tracks are properly downloaded, or do we need to delete all other downloads first and check what is left? Because if we can detect them, we could simply remove all orphans on startup.
> 
> Edit: Oh and just keep most of the logging in place please. Maybe adjust the level (fine, finer, finest, etc.), but the logs we'ren't super helpful so far, so any improvement is welcome!

lemme see if i can troubleshoot which downloads collection is missing the data and we can handle that case

shayaantx:
@Chaphasilor ok so I figured out whats going, the requiredBy dependencies is a list, and for some reason it had duplicate ids within it. So I modified the remove() for requiredBy to remove all occurrences. Also when determining what songs to sync, I was logging out no-change/update songs (i.e., songs that don't require any action) - however by not doing this we avoid scenarios where we want download helper to correct requiredBy in case it was orphaned.
```
[DownloadsHelper/INFO] 2023-12-09 17:50:58.658966: Removing f4af7665025022570ccae3c87b12fcea dependency from bb584c4b33344586334a730fd1f0db9d, remaining dependencies [f4af7665025022570ccae3c87b12fcea, f4af7665025022570ccae3c87b12fcea, f4af7665025022570ccae3c87b12fcea]
```

So now the issue where existing downloaded songs that got orphaned is fixed (so now you can sync, delete, in whatever combo and the album/playlist reflects the current state)

I'll look into the button states thing tonight or tomorrow

Chaphasilor:
Perfect! We can fix the button states in another PR and get this one merged if you want? 😁

shayaantx:
> Perfect! We can fix the button states in another PR and get this one merged if you want? 😁

Honestly up to you, I'd probably do one more retest yourself (just to be sure I'm **_not_** insane).

Chaphasilor:
@shayaantx I noticed that if I download an album and then enable offline mode, it doesn't show up in the albums list until I restart the app. Can you reproduce?

Chaphasilor:
Another thing I noticed: if I delete an album but a track from that album is still in a playlist, the track stays downloaded on the device, but the album image does not. So the track is missing the album image, which isn't ideal. Not sure if that is in-scope for this PR though...

shayaantx:
> I noticed that if I download an album and then enable offline mode, it doesn't show up in the albums list until I restart the app. Can you reproduce?

Thats the thing I mentioned earlier about refreshing a bit more aggressively. So if you do following:
1. Sync fresh album
2. Enable offline mode
3. I do see it doesn't show up, but if you tap the arrow button, it will appear.

So when the tab is in focus, its not doing a fresh load (offline enabled or not) - at least this is what I've seen.

>Another thing I noticed: if I delete an album but a track from that album is still in a playlist, the track stays downloaded on the device, but the album image does not. So the track is missing the album image, which isn't ideal. Not sure if that is in-scope for this PR though...

Lemme take a look at this one

Chaphasilor:
Ahh okay I see, yeah that sounds like it could be improved. But I think we can get this merged first, since it's not a regression!

shayaantx:
> Ahh okay I see, yeah that sounds like it could be improved. But I think we can get this merged first, since it's not a regression!

Up to you boss, I can still look into the images removing kerfuffle too, doesn't matter to me.

Chaphasilor:
Yeah the album covers fix would be nice. But no need to look into the refreshing :)

shayaantx:
@Chaphasilor 
> Another thing I noticed: if I delete an album but a track from that album is still in a playlist, the track stays downloaded on the device, but the album image does not. So the track is missing the album image, which isn't ideal. Not sure if that is in-scope for this PR though...

How are you reproducing it? Are you
1. Create 2 playlists with one shared song that has an image
2. Go to playlist A, sync
3. Go to playlist B, sync
4. Delete playlist A
5. Then see missing image on playlist B

If I follow above steps I can't seem to reproduce the behavior you described, however when I debug the code, it seems to delete the image, but the UI doesn't reflect it - dunno, but it def seems like it is executing the remove download image future erroneously

```

  Future<void> addDownloads({
    required List<BaseItemDto> items,
    required BaseItemDto parent,
    required bool useHumanReadableNames,
    required DownloadLocation downloadLocation,

    /// The view that this download is in. Used for sorting in offline mode.
    required String viewId,
  }) async {
....
      for (final item in items) {
        _downloadsLogger.info("Attempting to download ${item.id}");
        if (_downloadedItemsBox.containsKey(item.id)) {
          // If the item already exists, add the parent item to its requiredBy field and skip actually downloading the song.
          // We also add the item to the downloadedChildren of the parent that we're downloading.
          _downloadsLogger.info(
              "Item ${item.id} already exists in downloadedItemsBox, adding requiredBy to DownloadedItem and adding to ${parent.id}'s downloadedChildren");

          // This is technically nullable but we check if it contains the key
          // in order to get to this point.
          DownloadedSong itemFromBox = _downloadedItemsBox.get(item.id)!;

          itemFromBox.requiredBy.add(parent.id);
          addDownloadedSong(itemFromBox);
          _addItemToDownloadedAlbum(parent.id, item);
          continue;
        }
....

``` 

This is the download jellyfin items function snippet, basically if the song already exists, it will basically put the song in all the cache/collections/requiredBy fields then continues to next item. What's missing here is adding requiredby updates to the downloaded image itself (if it exists) for the item we are downloading (which already exists). So this seems like an existing bug cause I didn't change this logic - but I can push a commit to fix 2moro or sometime this weekend.

Chaphasilor:
What I did (iirc) was:  
1. Download playlist
2. Download album of song in playlist
3. Delete album

Could also be that I did 1 and 2 in reverse order...

If you could figure out how to fix it, that would be awesome! 😁

shayaantx:
@Chaphasilor I think I need to rewrite this entire download/remove function to make sense of this, like the way downloadImage gets managed is with requiredBy collection as well (however the ids we pass into this are song ids themselves). This is behavior I'm seeing so far:

1. Sync both playlists
image
2. Delete "te1" playlist
image
3. Switch to offline mode
image

You can see above the playlist loses parent image, but retains song images. I saw some more weird logic (not sure if its from debugging), where after deleting first playlist, the downloaded song was missing - but I can't reproduce that without breakpoints (maybe the futures weren't executing properly cause of my breakpoints)

Now if you modify steps to reproduce:
1. Sync both playlists
2. Delete one song from "te1" - "Inner Assassins"
3. Then come back to downloads page
image

You will get the entire shared song missing (this doesn't happen with the main sync button), but more cause the downloads helper does a hard delete in deleteDownloadChildren (which I kind of workaround in the download sync helper), the lil bit below
```
        if (downloadedSong.requiredBy.isEmpty || deletedFor == null) {
```

I dunno honestly, things get weird deleting, and all these collections used randomly everywhere to maintain persistent state isn't great. There seems like a greater need for a simpler model to store this data imo, but I'm super lazy :)

shayaantx:
Lemme debug it some more and see what I come up with

shayaantx:
@Chaphasilor ok I fixed the 2nd bug I mentioned by taking out the logic out of deleteDownloadChildren() that just deletes the song not including the deleteFor parameter (as that parameter causes the issue in this case and caused me some other problems down the line). The only annoying part is I duplicated that logic in a new method deleteSong(), cause the existing method honestly should be dismantled.

I'm still not able to reproduce the first issue you mentioned, maybe I can add push some logging changes, and you reproduce and post logs?

Chaphasilor:
I'll try to repro it with your changes first, I'll take a look at those tomorrow! Sorry for not responding earlier...

shayaantx:
> I'll try to repro it with your changes first, I'll take a look at those tomorrow! Sorry for not responding earlier...

fyi the recent change isn't for your reported issue (although I think* that happens with this entire PR), its for when someone manually deletes a single shared song

Chaphasilor:
Repro is here: 

https://github.com/jmshrv/finamp/assets/18015147/2e00c13e-a1fa-45f1-9943-75c0be0aadcc

It really is just:   
1. Download playlist
2. Download album of track in playlist
3. Delete album again
4. Track is now missing the cover

shayaantx:
@Chaphasilor oh you meant the cover image of the playlist/album not the individual track right? Cause I can reproduce the cover image disappearing with these steps, but the individual song still retains its image

Chaphasilor:
Actually no, at the end on the songs tab you can see the song is missing the cover. The album is also missing the cover, but since it's deleted it disappears right after refreshing anyway...

shayaantx:
> Actually no, at the end on the songs tab you can see the song is missing the cover. The album is also missing the cover, but since it's deleted it disappears right after refreshing anyway...

FYI I've gone a different direction with this one, basically trying to decouple the download storage (which I'm calling the download store or whatever) from the downloads helper while trying to refactor how all this storage works and hopefully in the end these type of cases become easier (I hope)

It is also possible I give up and just fix this specific use case :), but it really seems like all the Hive boxes shouldn't be all over the downloads helper cause then its always a case of did I put the right object in the right collection in new or existing methods (just makes seeing the downloaded songs and assets relationships much nicer I think).

Lemme know if this is overkill too

image

Chaphasilor:
@shayaantx are you aware of the [planned changes to the download functionality](https://github.com/jmshrv/finamp/blob/main/DOWNLOADS_PLAN.md)? Also discussed here: https://github.com/jmshrv/finamp/issues/213

Best read through it and make sure you're not putting in too much effort for something that will be completely re-built eventually :)

shayaantx:
> @shayaantx are you aware of the [planned changes to the download functionality](https://github.com/jmshrv/finamp/blob/main/DOWNLOADS_PLAN.md)? Also discussed here: #213
> 
> Best read through it and make sure you're not putting in too much effort for something that will be completely re-built eventually :)

Lol agree with all those bullet points, yea maybe I won't go into this I guess - I'll go back to trying to fix this specific edge case

Chaphasilor:
Also, #568 just dropped, take a look at this. That is a bigger project though and will probably take a bit longer to hit stable, so don't worry about your changes becoming obsolete :)

We can get this PR merged as soon as you're happy with it!

shayaantx:
> Also, #568 just dropped, take a look at this. That is a bigger project though and will probably take a bit longer to hit stable, so don't worry about your changes becoming obsolete :)
> 
> We can get this PR merged as soon as you're happy with it!

Yea that PR completely removes the downloads helper, which this PR hinges on

Looking at that PR I see _syncDownload() does what appears to be a much easier to grok way of downloading, no idea how it would affect this PR

So I guess you could merge it in this current state and [Komodo5197](https://github.com/Komodo5197) could merge in these changes into his or we could wait till he is done and merge this?

Chaphasilor:
I'd merge this first. The other PR should be rebased anyway.

So we'll leave out the missing covers for now? It's not a huge deal anyway...

shayaantx:
fyi @Chaphasilor merged https://github.com/jmshrv/finamp/pull/534/commits/dec8e85ac72f4669efbc027e9b4ef1b96c4ac5ca to resolve conflicts

Chaphasilor:
Merged. Merry Christmas 🎄

PR #547, opened 2023-12-07
[Redesign] Recover Now-Playing Queue on Startup

Komodo5197:
Occasionally saves the current queue and position in song while playing.  The queue auto-loads on app startup if enabled, or earlier queues can be restored from a menu.  Added a loading variant of the Now-Playing bar while the queue loads, and tweaked a bunch of theming/transition code for the bar while cleaning it up.  Queues are stored as a collection of raw song ids.


Chaphasilor:
Oh that sounds really awesome! Taking a quick glance at the code, it seems like there are a lot of goodies and fixes in there as well.  
I should be able to give a proper review either tonight or on Saturday.

If you're using just the song ID for persisting queues, that means the queue source for each track will be lost, correct?

Also, how does the logic for handling multiple different queues work? At what point do you separate the queues, each time `QueueService.startPlayback` is called?

What happens if a queue can't be restored (i.e. because Finamp is in offline mode)?

Really looking forward to testing this out! 😁

Chaphasilor:
> The track source is indeed lost. I considered trying to save it but it didn't seem worthwhile.

Maybe it would be possible to at least store the "original" source, so not the source of individual tracks, but just the global source that's shown in the queue list for example. This could possibly be modified to indicate that the queue was restored, but I'm not sure that's needed at all. The restoring should be as transparent to the user as possible.

Regarding the saved queues, maybe 60 is a bit too much. In any case, I noticed that the sorting on the "Restore Queue" screen is from oldest to newest, but it should probably be the other way around, similar to the "Playback History" screen :)

I also encountered an issue where my phone didn't have internet and none of the tabs loaded on startup, and the queue didn't get restored either. While that's fine, it would be nice to show the Now Playing Bar with a button to retry loading the latest queue.  
After I restarting Finamp a few times until the connection was restored, I went to manually restore the queue and noticed an empty queue being persistet. Trying to restore it didn't do much, apart from creating another empty saved queue:



Otherwise this has workes really well so far! I also love that you fixed the accent color calculation so it isn't tied to the player screen anymore :D

jmshrv:
I really appreciate this! iOS really likes to kill Finamp, I really need to know how apps like Spotify keep their players forever :sweat_smile:

Chaphasilor:
I feel like the `getTrackFromId()` method doesn't really belong into the queue service. It definitely shouldn't be a public member of the main class. There are a few other files, like `generateMediaItem()`. I think it would be best to make `getTrackFromId()` private and move it below the other helper methods. What do you think?

Oh and I'm not a huge fan of the syntax with the inline for loop (queue_service.dart:336), right below it you're using a `fold`, so I think a `filter` and `map` would make this more consistent...

I'll do some more testing with the two commits you've added in the coming days!

Chaphasilor:
> iOS really likes to kill Finamp, I really need to know how apps like Spotify keep their players forever 😅

I thought that's what the "Enter Low-Priority State on Pause" setting is for, if I don't disable it Android also always kills the app in the background...  
Or is that setting Android-only?

Chaphasilor:
Just encountered an issue where the accent color for a cover couldn't be calculated, probably because the cover failed to fetch in the provider. At some point the cover was loaded correctly in both the player screen and now playing bar, but the color was still the one from the previous cover:

Komodo5197:
I believe the problem was that PaletteGenerator had a 2 second timeout, and then once it failed would never re-run because the album image never changed.  I upped it to 5 seconds and moved a bunch of provider code around so that I could force a recalculation on track change.  This also detached theme updating from AlbumImage, so redraw loops shouldn't be a concern now.

Chaphasilor:
Thank you for that! That's a very welcome change, and finally a proper solution. Is there anything you're still planning to implement for this PR, or would you consider this ready to merge (after some more testing)?

lib/services/player_screen_theme_provider.dart
```diff
@@ -1,39 +1,69 @@
...
+
+    final PaletteGenerator? palette = await PaletteGenerator.fromImageProvider(
+        image,
+        timeout: const Duration(seconds: 5),
+      ).then((value) => value, onError: (_)=>null);
+    if (palette==null) return null;
+    // Color accent = palette.dominantColor!.color;
+    Color accent = palette.vibrantColor?.color ??
+        palette.dominantColor?.color ??
+        const Color.fromARGB(255, 0, 164, 220);
```
Chaphasilor, line 52: Finamp's default blue accent color (`0, 164, 220`) should be used as a fallback. You can basically get rid of lines 16-28 I think, because that code will only run if the previous attempt at color generation failed, and in that case the fallback should be blue, not gray :)  
It also simplifies the code a bit.

Oh and at the end, please format the files you've changed (using your editor or `dart format path/to/file`

Komodo5197: I did some comparisons of the different default colors.  Do you think the raw blue color is best?

Current color:
Screenshot (48)
Screenshot (52)

Default blue color with accent calculation steps:
Screenshot (47)
Screenshot (51)

 return ColorScheme.fromSeed(seedColor: Color.fromARGB(255, 0, 164, 220),brightness: brightness):
Screenshot (55)
Screenshot (57)

Komodo5197:
The last plan I had for this PR was the switch to batched item lookups, which I got in a day or two ago.  I'm ready to merge as soon as all your concerns/testing are cleared up.

lib/services/player_screen_theme_provider.dart
Chaphasilor: Yeah I think it should be fine. But I was thinking more about a generall fallback if a cover doesn't load/exist. The "restoring queue" widget could stay gray if you prefer that :)

Chaphasilor:
Okay, from my testing I didn't see any issue.  
It's not ideal that we have three different locations loading album covers, sometimes the album cover on the player screen is loaded first and sometimes the accent color is calculated first. But that's only a cosmetic issue, and nothing you need to worry about with this PR. I'll probably merge it some time tomorrow :)

lib/services/player_screen_theme_provider.dart
Chaphasilor: Also, the contrast checking is not really useful if a cover fails to load, since then the background (blurred cover) and accent color don't match anyway. See example here;

Cover for themeprovider failed to load, blurred cover loaded eventually (notice the low contrast):



After restart, when all covers loaded successfully:

Komodo5197: The theme provider now returns grey for the initial queue loading, but blue if palette generation fails.

Komodo5197:
I thought about the multiple image loads and messed with the idea of feeding the album covers from the currentAlbumImageProvider, but the problem is you need image sizes from the UI to make the requests.  I spent a bunch of time trying to figure out sizing things, but now that I'm writing this I've noticed that the audio service seems to be loading full size images for the OS.  So I'll just grab a full size image in currentAlbumImageProvide and then distribute it to everyone to sync everything up better.  I can get that finished up tomorrow.

Chaphasilor:
Okay that sounds nice. Yeah we're getting at least one full-size cover for the current track anyway, and it won't hurt to use the full-size image in places like the now playing bar or queue list. As long as the other items in the list items use a low resolution for bandwidth and performance reasons, everything should be fine.

While you're looking at replacing the image source, maybe you can figure out why the cover on the player screen always reloads when you open the screen 😅

Komodo5197:
Okay, I centralized the current album cover loading and precaching.  I also tweaked the theme provider to hopefully respond to loading errors better instead of just having a 5 second timeout.  Regarding the reloading, hopefully this fixes it due to the now-playing bar using the same image as the player so all the caches and providers are already ready, but I'm not planning to investigate if it doesn't.

Chaphasilor:
You are awesome! 😁

I'll take a look today and get back to you tonight (hopefully) :)

Chaphasilor:
Okay so I didn't perform any additional testing, but the queue restoring works with the gray accent and the album cover is *much better* like this! 😁  
The whole player screen feels so much more polished now, so thanks again for taking the time to fix this <3

I'll just fix up the merge conflict and merge this...

Chaphasilor:
Merged, thank you so much for the effort! Hoping to see another PR from you soon 😁

PR #550, opened 2023-12-08
fix folder structure

Tiefseetauchner:
there was a folder /com/example/finamp/ which did not match up with the namespace in MainActivity.kt

I know I know this is so very minor to make a pull request but you know how the internet people go, they see an issue and then loose sleep over it for the next 25 years


Tiefseetauchner:
Ok proof I'm bad at git, I deleted the file and didn't add it afterwards...

Chaphasilor:
Seems to work just fine, and I found no other references to the old directory structure in the code :)

Issue #551, opened 2023-12-08
[Feature Request] Add Artist "Most played" list


Tiefseetauchner:
Coming from Mucke, I am missing the feature to play the most played songs of an artist in order. Ideally you'd have a list of the most played songs on the artist page and when you press the first, you can go listen through those, or listen to all artist songs with shuffle. A lot like Mucke and Feishin do it

It should be possible (Feishin does it) but I am not sure how it'd be done. I went looking at the code and frankly, coming from Winforms, this is above me. I'll try get into the dart thing more, and I even managed to render a top bar like for albums, but that took two hours, god knows how long it'll take to actually put text in there...

Just wanted to give an heads up and make sure this isn't actively being worked on already :)

Best, Lena

Tiefseetauchner:
You're right, I mistook that, it just gives 10  top songs without ordering on the main Artist screen, Mucke does that but that project has way better tracking I think. What I thought of was going to all songs and having it sorted by plays, which I already do so automatically that I didn't realize that was something I did. But I think using the data Jellyfin stores is a good starting point, even if a little bit unreliable it should be fine overall (except for offline play, but there is/was a PR for that iirc so we could just add that on top to be a little "less wrong"). I'm currently looking at how to get a list of songs displayed and I'd care for the logic after, thank you for the offer, I might come back to that :)

PR #552, opened 2023-12-09
Add artist top songs

Tiefseetauchner:
From issue #551

Please don't be shy to roast the code, first time ever working in flutter, I can't imagine it being any good. I don't know what the usual way to do the application id is, I left it in because I'll probably debug but I suspect we'll remove it just before merging (if this is merged)

There's a TODO I can't resolve, as I don't really understand the technical use of all the variables of `SongsSliverList`

I realised while pondering over the changes that this affects the genre selection too, I think that's actually a pretty neat feature

One thing that's still left is to actually add the queue, I do not know how to do that. We have the sorted - wait is that what the queue variable is for - oh nevermind, i'll figure that out later, I'll leave this as a draft for now


Tiefseetauchner:
Now queues work, and there's no duplicate code, there's a little bit of unprettyness in the SongListTile class as I used the isInPlaylist for artist views now too (to display the correct name somewhere I don't remember, I think in the queue), that should probably be changed

Tiefseetauchner:
Example image

Example Screenshot showing a top list of songs and albums underneath

Chaphasilor:
Regarding the app ID, I think the way @jmshrv and myself always do it is to just overwrite the currently installed version with the debug version 😅 I also use my old phone for initial testing.  
Might be worth looking into how to make this more seamless...

jmshrv:
I'm pretty much always on a dev build lol, I really need a button to redownload everything I last had because I keep corrupting the databases

lib/components/ArtistScreen/artist_item_info.dart
```diff
@@ -0,0 +1,100 @@
...
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      children: [
+        _IconAndText(
+            iconData: Icons.music_note,
+            text: AppLocalizations.of(context)!.songCount(itemSongs)),
+        _IconAndText(
+            iconData: Icons.book,
+            text: AppLocalizations.of(context)!.albumCount(itemAlbums))
```
Chaphasilor, line 35: The album count isn't correct for me for some artists. It should probably use `albumCount` instead of `childCount`, but that's a more expensive operation and probably shouldn't be used for all artists, that's why it's currently not being requested from the server for the artists tab items. You could try requesting it for all artists, or use a `FutureBuilder` and only show the album info after you re-fetched the artists when a user opens the ArtistScreen...

I quickly tried to also request the `AlbumCount` field from the server, but for some reason that didn't work. Maybe you can give it another try? 🤔

lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -147,15 +152,15 @@ class _SongsSliverListState extends State<SongsSliverList> {
...
-
     return SliverList(
       delegate: SliverChildBuilderDelegate(
         (BuildContext context, int index) {
+          // When user selects song from disc other than first, index number is
+          // incorrect and song with the same index on first disc is played instead.
+          // Adding this offset ensures playback starts for nth song on correct disc.
+          final indexOffset =
+              widget.childrenForQueue.indexOf(widget.childrenForList[0]);
+
```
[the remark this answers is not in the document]
Tiefseetauchner: I get this exception

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following RangeError was thrown building SongsSliverList(dirty, state:
_SongsSliverListState#a6360):
RangeError (index): Invalid value: Valid value range is empty: 0
```

I once understood why moving it works but I don't remember, it's some special case because we only have 5 displayed songs instead of all of them

lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -180,20 +185,26 @@ class _SongsSliverListState extends State<SongsSliverList> {
                 widget.onDelete!(item);
               }
             },
-            isInPlaylist: widget.parent.type == "Playlist",
+            // TODO Rename this to something that makes more sense with this addition
+            isInPlaylist: widget.parent.type == "Playlist" ||
+                widget.parent.type == "Artist",
```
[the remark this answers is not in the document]
Tiefseetauchner: It is to display a different queue source, it seems to not work regardless, I'll try to figure it out after work, reverting it for now

lib/components/ArtistScreen/artist_item_info.dart
Tiefseetauchner: I'll address this in the other PR and linked this comment there, this slid into this one on accident

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -0,0 +1,97 @@
...
+                AppLocalizations.of(context)!.albums,
+                style:
+                    const TextStyle(fontSize: 18, fontWeight: FontWeight.bold),
+              )),
+            ],
+            body: MusicScreenTabView(
+                tabContentType: TabContentType.albums,
+                parentItem: widget.artist,
+                isFavourite: false),
+          ));
```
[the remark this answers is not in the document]
Tiefseetauchner: I'll try to get to it, I'm not unhappy to hear that you wanna change that 😆

lib/components/AlbumScreen/album_screen_content.dart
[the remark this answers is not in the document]
Chaphasilor: Try changing it to `final int indexOffset = widget.childrenForList.isNotEmpty ? widget.childrenForQueue.indexOf(widget.childrenForList[0]) : 0` to avoid the exception. If that doesn't work you can move it back into the delegate :)

lib/components/ArtistScreen/artist_item_info.dart
Chaphasilor: I have no idea how that happened lol 👀

lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -180,20 +185,26 @@ class _SongsSliverListState extends State<SongsSliverList> {
...
-                    // "hide song artists if they're the same as album artists" == true
-                    &&
-                    FinampSettingsHelper
-                        .finampSettings.hideSongArtistsIfSameAsAlbumArtists
-                    // song artists == album artists
-                    &&
-                    setEquals(
-                        widget.parent.albumArtists?.map((e) => e.name).toSet(),
-                        item.artists?.toSet())),
+            showArtists: widget.showArtist &&
```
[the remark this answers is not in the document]
Chaphasilor: Yes it's needed, or yes it is `true` anyway? 🤔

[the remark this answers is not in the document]
Tiefseetauchner: Sorry, yes it's needed

I xhanged it to check afterwards whether it's an artist page, and I'll think about how one could improve this formula

lib/components/ArtistScreen/artist_screen_content.dart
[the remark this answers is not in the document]
Tiefseetauchner: I am done, this was a lot lol

I frankly don't think I caught everything, my screen is decisively too small for this, but it works (for me >.<)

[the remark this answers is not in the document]
Tiefseetauchner: (Wanna say I'd appreciate you review these commits)

lib/components/AlbumScreen/album_screen_content.dart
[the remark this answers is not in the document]
Tiefseetauchner: Not needed anymore since everything was moved to its own classes anyways

lib/components/ArtistScreen/albums_sliver_list.dart
```diff
@@ -0,0 +1,72 @@
...
+            item: item,
+            children: widget.childrenForList,
+            index: index,
+            parentId: widget.parent.id,
+            parentName: widget.parent.name,
+            onDelete: () {
+              final item = removeItem();
+              if (widget.onDelete != null) {
+                widget.onDelete!(item);
+              }
```
Chaphasilor, line 61: Why do we need an `onDelete`? It's not being used in the AlbumListTile.  
Don't forget to remove the `removeItem` function as well :)

lib/components/ArtistScreen/album_list_tile.dart
```diff
@@ -0,0 +1,583 @@
...
+  Widget build(BuildContext context) {
+    final screenSize = MediaQuery.of(context).size;
+
+    final listTile = ListTile(
+      leading: AlbumImage(item: widget.item),
+      title: RichText(
+        text: TextSpan(
+          children: [
+            // third condition checks if the item is viewed from its album (instead of e.g. a playlist)
+            // same horrible check as in canGoToAlbum in GestureDetector below
```
Chaphasilor, line 84: Seems like some leftover comments. You can also remove everything that is underlined in yellow or blue, as that is not being used at all

lib/components/ArtistScreen/album_list_tile.dart
```diff
@@ -0,0 +1,583 @@
...
+                      .textTheme
+                      .bodyMedium
+                      ?.color
+                      ?.withOpacity(0.7)),
+            ),
+            TextSpan(
+              text:
+                  " · ${printDuration(Duration(microseconds: (widget.item.runTimeTicks == null ? 0 : widget.item.runTimeTicks! ~/ 10)))}",
+              style: TextStyle(color: Theme.of(context).disabledColor),
+            ),
```
Chaphasilor, line 119: Duration for an album? Where is that coming from 👀

lib/components/ArtistScreen/album_list_tile.dart
```diff
@@ -0,0 +1,583 @@
...
+      onTap: () {
+        if (widget.item.type == "MusicArtist" ||
+            widget.item.type == "MusicGenre") {
+          Navigator.of(context)
+              .pushNamed(ArtistScreen.routeName, arguments: widget.item);
+        } else {
+          Navigator.of(context)
+              .pushNamed(AlbumScreen.routeName, arguments: widget.item);
+        }
+      },
```
Chaphasilor, line 137: This is an **album** list tile, and should only support the case of the parent being an album. Trying to handle all cases in one place it what got us into this mess in the first place :D

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -0,0 +1,87 @@
...
+
+          albums ??= jellyfinApiHelper.getItems(
+            parentItem: widget.parent,
+            filters: "Artist=${widget.parent.name}",
+            sortBy: "ProductionYear",
+            includeItemTypes: "MusicAlbum",
+            isGenres: false,
+          );
+
+          return FutureBuilder(
```
Chaphasilor, line 52: You could use something like `Future.wait()` to combine both futures, that seems a bit better than using two FutureBuilders...

lib/components/ArtistScreen/albums_sliver_list.dart
Chaphasilor, line 1: This could be moved into the `components` directory, as it could be used in other screens as well (or am I missing something?)

lib/components/ArtistScreen/albums_sliver_list.dart
Tiefseetauchner: I misthought it to be an onDispose, thanks for clarifying 😅

Eta 1800 UTC 😃

lib/components/ArtistScreen/album_list_tile.dart
Tiefseetauchner: I have no idea to be frank, I saw it works and I thought that's neat

lib/components/ArtistScreen/album_list_tile.dart
Tiefseetauchner: Woops, pebkac I forgot what I was copying from

lib/components/ArtistScreen/albums_sliver_list.dart
Tiefseetauchner: Gladly moved it, overlooked that and kinda just blindly followed the albumscreen example

lib/components/ArtistScreen/album_list_tile.dart
Chaphasilor: Oh it works? Neat! Seems like Jellyfin is actually properly aggregating children for once 😅

lib/l10n/app_en.arb
```diff
@@ -601,5 +601,7 @@
   "topSongs": "Top Songs",
   "@topSongs": {},
   "playCountInline": " · {playCount} plays",
-  "@playCountInline": {}
+  "@playCountInline": {},
+  "couldNotLoadAlbum": "Couldn't load album",
+  "@couldNotLoadAlbum": {}
```
Chaphasilor, line 606: Maybe use a select here instead, to support different types as well? Like in line 588

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -17,66 +17,60 @@ class ArtistScreenContent extends StatefulWidget {
...
   Widget build(BuildContext context) {
-    songs ??= jellyfinApiHelper.getItems(
-      parentItem: widget.parent,
-      filters: "Artist=${widget.parent.name}",
-      sortBy: "PlayCount",
-      includeItemTypes: "Audio",
-      isGenres: false,
-    );
+    final futures = Future.wait([
+      jellyfinApiHelper.getItems(
```
Chaphasilor, line 25: Maybe add to comments here to say what each request does (request top songs, request albums) (and that it will be index 0/1 in the `futures` variable)

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -17,66 +17,60 @@ class ArtistScreenContent extends StatefulWidget {
...
+    ]);
 
     return FutureBuilder(
-        future: songs,
-        builder: (context, songSnapshot) {
-          var orderedSongs = songSnapshot.data?.reversed.toList() ?? [];
+        future: futures,
+        builder: (context, snapshot) {
+          var songs = snapshot.data?.elementAt(0) ?? [];
+          var albums = snapshot.data?.elementAt(1) ?? [];
```
Chaphasilor, line 45: `elementAt()` throws an exception if the index doesn't exist. You want `elementAtOrNull()` instead :)

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -17,66 +17,60 @@ class ArtistScreenContent extends StatefulWidget {
...
 
-          albums ??= jellyfinApiHelper.getItems(
-            parentItem: widget.parent,
-            filters: "Artist=${widget.parent.name}",
-            sortBy: "ProductionYear",
-            includeItemTypes: "MusicAlbum",
-            isGenres: false,
-          );
+          var orderedSongs = songs.reversed.toList();
```
Chaphasilor, line 47: Try changing the `sortOrder` in the call to `jellyfinApiHelper.getItems()` instead to make it more clear what's doing on

Tiefseetauchner: Oh is thede a way to have jellyfin reverse it?

Tiefseetauchner:
Thanks for the auto builds btw, now I can work without being able to build it xD

Chaphasilor (commented):
Okay, I think that should be the last review.  test it in the coming days and merge it asap :D

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -0,0 +1,78 @@
...
+          return Scrollbar(
+              child: CustomScrollView(slivers: <Widget>[
+            const SliverPadding(padding: EdgeInsets.fromLTRB(0, 15.0, 0, 0)),
+            SliverToBoxAdapter(
+                child: Text(
+              AppLocalizations.of(context)!.topSongs,
+              style: const TextStyle(fontSize: 18, fontWeight: FontWeight.bold),
+            )),
+            SongsSliverList(
+              childrenForList: songs.take(5).toList(),
```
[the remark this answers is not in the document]
Tiefseetauchner: Actually there is kind of, it's used for the queue. We might be able to do it asynchronously though after the widget is loaded, or just accept that only five songs are in the queue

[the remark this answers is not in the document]
Chaphasilor: Ah I see. Yeah the initial page with 5 tracks + fetching the other songs after playback starts would be a good approach I think. I'm also planning to implement this for the regular songs tab to enable in-order playback starting at a specific song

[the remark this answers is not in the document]
Tiefseetauchner: Okay, I don't think I can do that with my current knowledge of dart, I suspect we'd pass a delegate through but I cannot figure out how to await the result in a future in a synchronous method to pass it in as a delegate. I'll set it to limit to 5 for now and see whether I can figure it out, any help would be greatly appreciated as that's a primary usecase for me currently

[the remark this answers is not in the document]
Chaphasilor: I'm planning to support this in the QueueService somehow, where you can specify some initial tracks as well as the initial index, and then to callbacks/futures(?) that fetch songs to be prepended (first track to first of initial tracks, if any) and songs to be appended (last of initial tracks to last tracks, if any) to the queue. I imagining that should work, but I haven't tested it yet

Chaphasilor:
Okay, this looks good to me! We can merge from my side, if that's fine with you? :)

Tiefseetauchner:
Yep

Chaphasilor:
Merged, thanks a ton! 😁  
Onto the next one...

Chaphasilor:
@Tiefseetauchner re: application ID, I actually "looked into it" and there is a way in the gradle config to set a suffix for debug builds, so I'll change that up in the `redesign` branch soon: https://www.perplexity.ai/search/how-can-I-_XOgxvRgQT.gEmngt1jOXw?s=u :D

PR #561, opened 2023-12-18
:eyes: Rework Material 3 theme

Maxr1998:
After some dependency bump the new Material 3 style is enabled by default, which looks kind of odd with the current color schemes.

<details>
<summary>Old design</summary>

</details>

Thus, I generated a nice one with the [M3 Theme Builder](https://m3.material.io/theme-builder#/custom) and imported it into the project.

<details>
<summary>New design</summary>

</details>


Maxr1998:
Please give it a test, had to resolve some other issues as well and hope I caught everything.

lib/main.dart
```diff
@@ -283,9 +283,6 @@ class Finamp extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    const Color accentColor = Color(0xFF00A4DC);
```
Chaphasilor, line 286: I'd like to keep the original accent color here for reference :)

Chaphasilor:
Can we do something about that start padding on the tab bar?  
image  
I added a hacky fix in the `redesign` branch, but maybe you know a nicer solution?

Contrast on the Logs screen is a bit low...  
image

Maybe we can have less difference in color between the play/pause button and the progress slider?  
image

Chaphasilor:
Aside from that everything looks nice, definitely much better than the "default" / whatever crap that was there before 😅  
Thanks for putting in the effort! <3

lib/main.dart
Maxr1998: Added it back in `lib/color_schemes.g.dart` with a comment.

Maxr1998:
Fixed the tab bar issue as well, it's actually pretty easy:

> /// If [TabBarTheme.tabAlignment] is null and [ThemeData.useMaterial3] is true,
> /// then [TabAlignment.startOffset] is used if [isScrollable] is true,
> /// otherwise [TabAlignment.fill] is used.

So, just set `tabAlignment: TabAlignment.start`.

Maxr1998:
Fixed the remaining things that I completely missed last night, as well as some other things.

jmshrv:
Thanks for this! Merged :)

Maxr1998:
Sure, happy to contribute! I might actually further tweak some colors in later, but for now, this should be good.

PR #562, opened 2023-12-18
:sparkles: Add toggle to disable fast scroller

Maxr1998:
As discussed in #522.

### DIFF
--- a/lib/components/LayoutSettingsScreen/FastScrollSelector.dart
+++ b/lib/components/LayoutSettingsScreen/FastScrollSelector.dart
@@ -1,0 +1,24 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+import 'package:hive/hive.dart';
+
+import '../../models/finamp_models.dart';
+import '../../services/finamp_settings_helper.dart';
+
+class FastScrollSelector extends StatelessWidget {
+  const FastScrollSelector({Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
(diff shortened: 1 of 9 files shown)


lib/components/LayoutSettingsScreen/FastScrollSelector.dart
```diff
@@ -0,0 +1,24 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<Box<FinampSettings>>(
+      valueListenable: FinampSettingsHelper.finampSettingsListener,
+      builder: (_, box, __) {
+        return SwitchListTile.adaptive(
+          title: Text(AppLocalizations.of(context)!.showFastScroller),
+          value: FinampSettingsHelper.finampSettings.showFastScroller,
+          onChanged: (value) => FinampSettingsHelper.setShowFastScroller(value),
```
Chaphasilor, line 19: A `subtitle` with a short description of what the "fast scroller" is would be nice here :)

Chaphasilor:
You forgot to use the [arcane arts](https://github.com/jmshrv/finamp/blob/main/CONTRIBUTING.md#the-arcane-arts-code-generation), so the setting wasn't actually being saved across restarts :)  
I've quickly done that for you, everything else looks good!

Thanks a ton!

Chaphasilor:
Forgot about the subtitle, never mind then...  
I should go to bed 🙃

Maxr1998:
Ah yes, the arcane arts.. 🫠
Thanks for handling those for me! About the summary, I didn't think one was necessary, but if it creates any confusion we can still add one.

PR #563, opened 2023-12-20
:eyes: Further tweak color scheme

Maxr1998:
Couldn't help myself and made some further refinements, as teased in #561.

- Use original Jellyfin colors in dark theme (for the light theme, they're too light, unfortunately)
- Base secondary color upon desaturated primary color (reduced chroma)
- Add a proper tertiary color based on the purple Jellyfin accent - right now, it doesn't seem to be used anywhere, though
- Make dark theme backgrounds slightly darker
- Use the secondaryContainer color for the drag-to-queue indicator
- Make progress slider inactive color 50% opaque to be less harsh on the eyes

### DIFF
--- a/lib/components/PlayerScreen/progress_slider.dart
+++ b/lib/components/PlayerScreen/progress_slider.dart
@@ -37,6 +37,8 @@ class _ProgressSliderState extends State<ProgressSlider> {
 
     _sliderThemeData = SliderTheme.of(context).copyWith(
       trackHeight: 4.0,
+      inactiveTrackColor:
+          Theme.of(context).colorScheme.surfaceVariant.withOpacity(0.5),
     );
   }
 
--- a/lib/color_schemes.g.dart
+++ b/lib/color_schemes.g.dart
@@ -1,34 +1,39 @@
 import 'package:flutter/material.dart';
 
-/// The accent color that this color scheme is based on,
-/// using the [M3 Theme Builder](https://m3.material.io/theme-builder).
-const defaultAccentColor = Color(0xFF00A4DC);
+const jellyfinBlueColor = Color(0xFF00A4DC);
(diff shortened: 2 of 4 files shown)


lib/components/PlayerScreen/progress_slider.dart
```diff
@@ -37,6 +37,8 @@ class _ProgressSliderState extends State<ProgressSlider> {
 
     _sliderThemeData = SliderTheme.of(context).copyWith(
       trackHeight: 4.0,
+      inactiveTrackColor:
+          Theme.of(context).colorScheme.surfaceVariant.withAlpha(0x80),
```
Chaphasilor, line 41: I've used `.withOpacity()` everywhere else so far, do you think that's fine or should I be using `.withAlpha()` instead?

Maxr1998: I didn't really think about it when using it 😅
So, it seems like they basically do the same, just that one takes an int from 0-255, the other a float. `withOpacity` actually calls `withAlpha` after multiplying by 255, so it has more overhead, but I think `withOpacity` is easier to read and understand, so I'll update my code to use that.

Chaphasilor: Good to know. On the web in CSS it actually behaves differently, that's why I was curious. But yeah, a float should be easier then, thanks!

Chaphasilor:
<3

lib/components/PlayerScreen/progress_slider.dart
Maxr1998: Right, I read that when I was looking up   the Flutter method for it. The name difference is a tad confusing when it does the same but that's most likely because Dart doesn't support function overloading..

PR #564, opened 2023-12-20
:eyes: Improve logs interface

Maxr1998:
Tweak style and other code improvements.

### DIFF
--- a/lib/components/LogsScreen/log_tile.dart
+++ b/lib/components/LogsScreen/log_tile.dart
@@ -37,47 +37,50 @@ class _LogTileState extends State<LogTile> {
         color: _logColor(widget.logRecord.level, context),
         child: ExpansionTile(
           controller: _controller,
-          leading: _LogIcon(level: widget.logRecord.level),
           key: PageStorageKey(widget.logRecord.time),
+          leading: _LogIcon(level: widget.logRecord.level),
           title: RichText(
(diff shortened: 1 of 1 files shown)


lib/components/LogsScreen/log_tile.dart
```diff
@@ -37,47 +37,50 @@ class _LogTileState extends State<LogTile> {
...
-                TextSpan(
-                  text: "[${widget.logRecord.time}] ",
-                  style: const TextStyle(fontWeight: FontWeight.bold),
-                ),
-                TextSpan(
-                  text: widget.logRecord.loginCensoredMessage,
-                ),
-              ],
+              text:
+                  "[${widget.logRecord.loggerName}]\n${widget.logRecord.time}",
```
Chaphasilor, line 47: can we have the timestamp at the top right to save another line?

Maxr1998: In my experience it usually wraps anyways because the tag & timestamp are too long, so I think it's preferable to have it in two lines by default.

Chaphasilor: Alright. I'll see if I do a bigger overhaul of the logs screen at some point, but until then we can keep it like this :)

Chaphasilor:
tysm!

PR #566, opened 2023-12-20
Tweak download screen action buttons

Maxr1998:
Provide tooltip for the sync playlists button and generally refactor download screen action buttons.


Chaphasilor:
Looks good!  

Some tests would be nice for making sure everything still works, but I can't be bothers to add them yet ^^

PR #569, opened 2023-12-21
Fix adding playlists/albums to queue in offline mode

Chaphasilor:
Up until now albums and playlists were fetched from the server when adding them to the queue, even if offline mode was active.  
This PR changes the behavior to instead use the downloaded items for these actions instead.

This fix will probably also need to be implemented for the redesign.


jmshrv (approved):
Added a null check for `children` just in case - I can't imagine why it'd be null though.

PR #589, opened 2024-02-06
Swipe song to insert it into queue as next item

JAIABRIEL:
I don't see any point in the current swipe-to-append functionality, since my playlists are pretty long and playing it after the entire playlist feels useless. However, this feature is disabled by default.
Since I switched from Spotify, it feels more intuitive to swipe a song to play it next.
Idk your git history style maybe I could squash these commits into a single one.

#### Added a new setting to `Audio Service Settings` to toggle functionality

It's called `swipeInsertQueueNext` and since most users are pobably used to the old swipe queue functionality, it's disabled by default. 
Idk if `Audio Service Settings` is the right place for this option.

#### Swipe to play next

When a song is swiped, it will check if `swipeInsertQueueNext` is enabled and if so, existing functions are called to add insert it in queue.

#### Localization

Added English and German. Other languages are missing yet and I'm not sure if the German translation really matches the other translations.


Chaphasilor:
Hi, thanks for the changes. This behavior will be the default in the upcoming redesign (beta release soon™), but we can probably get this merged into the stable version in the meantime.  
I'm currently on mobile, so I'll take a closer look in a bit :)

JAIABRIEL:
> Hi, thanks for the changes. This behavior will be the default in the upcoming redesign (beta release soon™), but we can probably get this merged into the stable version in the meantime. I'm currently on mobile, so I'll take a closer look in a bit :)

Sounds great, didn't know that there's already someone working on this :) I thought it was a perfect feature to getting familiar with this project and flutter in general.  Are you using something like .editorconfig for this? I had some issued with whitespace/tab using nvim.

Chaphasilor:
Okay, I made some changes, with your new setting it made sense to introduce a whole new settings page for interactions. I also fixed the snackbar message that is shown when inserting next via swipe, seem like you missed that one ^^  
You can take a look at it, and if it still does what you intended it to do, we can get this merged!

Regarding the formatting config, we don't have such a thing since it's built into the dart cli (`dart format`). You can probably get it to work as the default formatter for nvim as well. I'm not a huge fan of the formatting strategy it uses, but at least it's consistent.  
Not all files are properly formatted, so I'd suggest you make your changes and format all the files you modified at the end *in a separate commit*. You could go ahead and try it out with this PR :)

Chaphasilor:
`dart format` accepts optional arguments, the target files/directories you wish to format.

The command should then look like this for multiple files:  
```
dart format path/to/file1.dart path/to/file2.dart path/to/file3.dart
```

The LSP server is probably doing just that in the background, so you should be fine! 😁

JAIABRIEL:
The dart formatting tool made a few changes. I think it can be merged now. And thanks for your help, my next PR will be better :)

Chaphasilor:
Changes look good!

Your PR was absolutely fine! Thanks for your contribution, and I'm looking forward to the next one :D  
Just make sure to take a look at the `redesign` branch and scroll through https://github.com/jmshrv/finamp/issue/220 to make sure you're not re-implementing something that has already been done, that would be a pity :)

PR #590, opened 2024-02-08
Add Airplay button

ihatetoregister:
Here is an attempt to add an AirPlay selector button. 

Tested working on ios device. 

Simulator Screenshot - iPhone SE (3rd generation) - 2024-02-08 at 22 41 11

### DIFF
--- a/lib/screens/player_screen.dart
+++ b/lib/screens/player_screen.dart
@@ -90,25 +90,21 @@ class PlayerScreen extends StatelessWidget {
                             SongName(),
                             ProgressSlider(),
                             PlayerButtons(),
-                            Stack(
-                              alignment: Alignment.center,
+                            Row(
+                              mainAxisAlignment: MainAxisAlignment.spaceEvenly,
                               children: [
-                                Align(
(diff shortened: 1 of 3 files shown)


Chaphasilor:
Hi @ihatetoregister, thanks for the PR! I don't have an iOS device so I can't test it out sadly, but would you mind rebasing this onto `redesign` instead?  
It's not a critical feature and seems like a better fit for the redesign overall. I'd also like to keep the branches from diverging too much.

So clicking the button simply opens the native AirPlay / output selection dialog?

ihatetoregister:
Sure, ill give it a shot. 

Exactly, this enable you to modify output device from the app directly insted of jumping out and go via the music controls. 

I'm not sure how well this works on non-ios devices so if possible, please test to see if I need to add any conditionals to it.

jmshrv:
Made a little change to only show the AirPlay button on iOS, since there isn't much point in it being there on Android (and it'd presumably just crash/complain).

Once this is rebased onto redesign, it can be merged :)

ihatetoregister:
Had some issues rebasing so created a new PR #591

Issue #231, opened 2022-05-11
support for audiobook needed


pinkworld123:
Please support audiobook. thanks

njakes:
I would love to have this as well. The only reason I use Plex is their mobile audiobook support.

Athoskal:
I'll put a $30 bounty on this. It would also be nice if we could pick a specific folder to download the books for offline

jmshrv:
I don't want to take money for developing Finamp, I'll think about working on it once I've completed some of the more foundational changes that I'm working on now (see #213)

MagicRB:
@jmshrv Don't think of it as taking money but merely a thanks from one human to another, one dev to another. You've allowed me to enjoy my music library on the go without burning through my mobile data limits. I'd buy you beer and shake hands bjt well idk where you live and we'll probably never meet (unless youre in amsterdam that is). So the next best thing is donating a few euros and letting you buy the beer yourself :)

Great app btw, you're single handidly making Jellyfin into a better platform 👍

Sp4rky001:
For what it's worth, audiobooks are very important to me and in my experience with Plex and Jellyfin, too much change to the underlying library structure is needed to really support audiobooks.

For me, the perfect solution presented itself in the form of Audiobookshelf.  It was built from the ground up as a book library.  It started with audiobooks and expanded into ebooks and even comics (very early).  It tracks reading stats and I'm surprised at how many hours I put into audiobooks last year.

The only feature keeping my coworker from using it is the lack of carplay integration.  It already has good Android Auto integration.

Chaphasilor:
closing this in favor of #267 since it has additional details and slightly more upvotes

PR #594, opened 2024-02-12
Add playback speed option to song menu

### LINKED ISSUE #412: [Feature Request] Playback speed
Hi and thank you for developing this awesome app!

I would like to request a playback speed change option.

I'm a speed freak, so please give me an option to get a hearth stroke, I need this 250bpm. 
Ideally without changing the pitch of the sound.

lymnyx:
## Description
This PR adds a very basic playback speed option to the new song menu in the `redesign` and thereby fixes #412.

Currently, pressing the newly added button computes the new playback speed with `speed % 3.5 + 0.5`, so in steps of 50%.
~~The speed is remembered while the app is open.~~
The speed is saved to the database and with that remembered even after restarts.

As I am very inexperienced with Flutter and Dart, my approach here might not be ideal. If that is the case, just let me know.

## Screenshots
playback-speed-fork-1
playback-speed-fork-2


Chaphasilor:
Hey, thanks for the PR! Really nice addition!

I have a few comments that I would like to discuss:

- Can we show the speed at a different location? Maybe right below the icon or something? It won't be easy without messing up the alignment, but I don't love that the value is shown in parentheses.
  - Maybe use a `Stack` to position the text right below the icon without messing up the rest of the layout?
  - If we decide that it should stay the way it is right now, then the translation token should be updated to accept a value (the numeric speed) and use that value within the translations, as there might be languages where the speed value would come before the text
- Given that the option is only really useful for audio books, and Finamp is primarily a music player, I'd like to add a setting for hiding this button. For now it could be a boolean switch, and down the road there might be a third option ("automatic") that shows the button when an audio book is being played (or something similar)
  - Any ideas where the setting should be located? Where would you look for something like this?
  - What do you think should the default be? Should the button be shown or not shown in the menu?
- The quick toggle is nice, but we could also add a long-press option that shows a dialog with more fine-grained options. That could include buttons for increasing and decreasing the speed freely (allowing the user to slow down playback speed below 1.0 if they desire), an input field allowing arbitrary speeds (within a range), and a "reset to default" button. If you don't want to implement that, I can do it myself, I just want to make sure that there's no issue with adding such a dialog?

Again, thanks for the PR, I'm sure we can get this merged quickly. I just want to make it even better, while we're at it already :D

lymnyx:
I am not that comfortable with the parentheses situation as well ;D, I will change that.

Regarding the settings, I like the idea of adding an "automatic" setting, it could be based on genre ("Speech", "audiobook", ...) and/or maybe album duration (>5h or so)!? That "automatic" setting could then be the default.

I think the setting could be placed into "Layout & Theme" (that's where I would look), "Audio Service" is an option as well, but to me it sounds too technical and backend for a UI setting.

Adding a dialog for that makes sense of course, the codebase (just like the environment) is just so new to me that I wasn't that comfortable immediately trying to implement that, but I think, I would like to try it now.

Chaphasilor:
I was thinking of using more definitive clues for the automatic detection, like the type of the BaseItemDto (is it just `Audio` for audio books, like for songs, or is there a separate type?) or the type of the parent collection (there we definitely have the needed type, but I'm not sure how hard it is to get to it, e.g. if the item is in a playlist).  
This would enable adding support for AudioBook libraries to the library selector, and I'm not sure right now if there's any difference in how such a library needs to be handled. We could also use the active library as a clue.

lymnyx:
While adding audiobook libraries could be an idea, I don't know if doing automatic audiobook detection based on the library type is ideal. Personally, when I started using Jellyfin and tried the Audiobook support under the `Books` library type (which is, if you understand your proposal correctly, what you mean), it was lacking in different ways and I ended up using a normal `Music` type library instead. Maybe that's just me though...

Also not sure what the situation is with the [Bookshelf plugin](https://github.com/jellyfin/jellyfin-plugin-bookshelf) and how that is then supposed to be handled.

lymnyx:
Just checked by `curl`ing a bit.

The audio in a Books library has in fact a type of `AudioBook`, so if we do want to, we could differentiate those audiobooks and the music from one another.

We could also just do both (checking type and genre) to account for different "setups"?

Chaphasilor:
Yeah that sounds good. You're right, I always confuse audio books with music videos, that's why I thought there is a dedicated collection. I'm surprised that there's an actual AudioBook type, but I guess otherwise the web frontend wouldn't know how to show/play the items.  
Did you try selecting a book library in Finamp and seeing if any sensible data is being fetched? My guess is the data types are too different for it to work, so we'll probably have to go with the genre and length detection for now. Although I'm not sure if there is a sensible length threshold that can handle audiobooks with short chapters (are separate chapters a thing for audio books?) and long songs (e.g. classical music pieces).

I'll go ahead and add a new sub-settings page (like tabs) for customization options and move the new setting there, otherwise the layout settings get pretty messy. I'll probably also reorder the settings a bit to move it towards the top. Do you also want me to take a look at adding the details dialog on long press, or do you want to try it yourself first? :)

lymnyx:
I haven't yet checked the returned data for the book library in comparison to a music library, but I will do that.

I would try to fetch the album's duration and use that as a basis, if that is a thing with the current architecture I would prefer that over track duration.
What do you think is an appropriate minimum duration for an album to be considered an audiobook?

A new settings page is nice!
I would like to try the long press myself, thank you for being so considerate by the way <3.

I am not so sure how to implement your idea with the positioning of the playbackSpeed variable as I don't quite know where to add the `Stack`. As a child of `PlaybackAction`?

lymnyx:
Apart from having to [traverse down](https://jellyfin.org/docs/general/server/media/books) into the actual folders (`Home/Books/Audiobooks/[Artist/Album || Album]`) with the files to get any sensible metadata, at least on first look the `BaseItemDto` of the audio files seems to be fairly similar, much of the typical stuff one would need, is there, I think.

So maybe adding support for the official Jellyfin Audiobook implementation is really something that could be fairly easily done down the line.

Chaphasilor:
Okay. Could you check if the duration is properly set for the audiobook album? If that's there, we can use it for the detection. I'd say it should be at least 2h to the automatic detection, do you think that would that work?

> So maybe adding support for the official Jellyfin Audiobook implementation is really something that could be fairly easily done down the line.

That sounds promising, would you be willing to do that in a follow-up PR? With the redesign we can make some breaking changes, so it would be easy to fit it in. I just don't have the resources (nor the interest) to look into audio books right now, but I'm definitely willing to help out a bit!

lymnyx:
Oh, sorry, I just noticed the folder itself (in contrast to the files) seems to somewhat lack metadata actually, it has no RunTimeTicks. The metadata (like RunTimeTicks) would have to be mainly grabbed from the audio items, I guess.

After this PR, I will still see what I can do, there seems to be an issue for this already (#231).

Chaphasilor:
> I am not so sure how to implement your idea with the positioning of the playbackSpeed variable as I don't quite know where to add the `Stack`. As a child of `PlaybackAction`?

Sorry, seems like I didn't see this message 😅  
Yes, there is a `Column` with an `Icon`, a `SizedBox` (for the spacing) and another `SizedBox` for giving the labels of the PlaybackActions a consistent height. You should be able to wrap the `Icon` in a `Stack`, and then put a `Positioned` widget into the `Stack` that wraps a `Text` containing the current value.  
You could make the value a nullable parameter of the `PlaybackAction` constructor, and not show any text if the value is `null`.  
Does that make it clearer? :)

lymnyx:
Yes, it does. Thank you!

Chaphasilor:
I improve upon it a bit, making the icons all have the same size. Do you think the speed text is too small now? 😅

I'm also planning to leave a review tomorrow or on the weekend, there are a few things that I'd like you to tweak :D

lymnyx:
I thought about keeping the size the same, but I didn't particularly like the reduced vertical spacing and thought even with the changed size it looked quite nice. Matter of taste, I guess ;).

> I'm also planning to leave a review tomorrow or on the weekend, there are a few things that I'd like you to tweak :D

Sure, looking forward to it ;). Just added the album length detection, right now the seemsLikeAudiobook function is started when opening the song menu (and therefore needs a moment when album length checking is required), we might want to already determine whether the button should be displayed when opening the player screen or so.

I was also trying to add the long-press dialog by wrapping the `IconButton` with a `GestureDetector`, setting its onLongPress parameter. That doesn't work for some reason (even though wrapping the Column inside instead works without problems). Without wanting to annoy you again: Any idea why? Or is there a better way of detecting the long-presses?

jmshrv:
As for design, I quite like how Apple Podcasts does it (clicking shows a dropdown)

Apple Podcasts screenshot, showing a dropdown for speed

lymnyx:
@Chaphasilor

I like that, it would fit right into the design of the sleep timer.

> I wouldn't want to leave the function within the menu if it's asynchronous, because that could cause layout shifts after opening the menu.

Yes, that was my concern.

> For now, go with just the GestureDetector.

Yeah, I just don't know how I'm supposed to fix the IconButton stealing the mouse events from the GestureDetector.
I could use an IgnorePointer or something like that, but then I would interfere with the onPressed and the focus styling.

lymnyx:
> As for design, I quite like how Apple Podcasts does it (clicking shows a dropdown)

@jmshrv I like it too, but in my humble opinion a design similar to AntennaPod's might be a bit more fitting. (Considering Finamp's new queue design, etc.)

finamp-pull-594-1
### After clicking on the speed button:
finamp-pull-594-2

Chaphasilor:
Okay, all good points.  
Firstly, if we go with the long-press stuff, just replace the IconButton with a GestureDetector wrapping an Icon :)

But I don't think we'll do the long-press stuff, the drop-down / picker list is a nice idea. The way it's implemented in AntennaPod is actually similar to what I had in mind for the long-press dialog, just with presets instead of increment/decrement buttons.  
But I guess presets are more useful and faster to interact with, and could also be user-customizable. I'm not a fan of the slider for selecting the speed though, a number input should be enough if we have presets already.

I could try to whip up a mockup tomorrow :)

Chaphasilor:
What do you think about this:  

image

I'm not totally sure where to put the reset option, doing it like this will cause a layout shift when the speed is changed. Any ideas?

You can use `SimpleButton` for the icon+text buttons, and use the Artist-/Feature-Chips as a starting point for the preset chips

lymnyx:
You mean the layout shift when adding the preset through "Apply"?
We could add a fixed width and maybe a small animation of the higher speed presets moving to their new position when the preset is created and added to the view.

I feel like there are a lot of views and layers already, and would therefore like the speed selector to appear as a part of the song menu, maybe (when clicked) as a section underneath the PlaybackActions.

Chaphasilor:
Okay so the "Enter speed manually" section doesn't create a new preset, it just sets the speed. Clicking on a preset will apply the speed as well, and it will be reflected in the input field. There's no way to customize the presets at the moment :)  
The small animation might be doable.

Regarding the placement, that's an interesting idea. Not implementing it as a PlaybackAction, but as a separate section. It might work, but in that case I'd probably go with a different layout. (First Image)
Alternatively, we could show a speed indicator on the player screen, and tapping that opens the dialog directly? But it's also still available through the menu, with proper labels. (Second Image)

lymnyx:
Yes, I think having it to the player screen makes sense, I just wasn't sure whether you wanted to have "that many" buttons there ;). We would also have to decide on a design for that dialog then, I'm not the biggest fan of the full screen dialog (sleep timer-like), for example.

In regards to the idea of the speed menu below the playback actions, we could have the presets and all of that there, the visibility toggle maybe being the existing PlaybackAction. That way we don't have yet another full-screen popup.

What are you using for those UI designs, Figma?

Chaphasilor:
The new menus are implemented as a bottom sheet, but I feel like nesting bottom sheets (e.g. sleep timer sheet opened from menu) isn't a great idea.  
Using the PlaybackActions to reveal/collapse a configuration section instead might work, I'm just not sure if I can make it look good.

The designs are made with Figma, yes :)

Chaphasilor:
Just FYI, since this is still mostly a work-in-progress, this won't be released as part of the initial beta released, but instead with the first beta update. But that should hit much sooner than a stable release would, so it's not too far off either :)

lymnyx:
Sure, thanks for letting me know. I already assumed something like that after reading through the beta release notes ;).

I have been playing around a bit and have a prototype for a menu.
There is no in/out animation so far and I don't think it is ideal that the keyboard is covering the input box by default (when starting to type).
We might also not need a reset button, the 1x button might be enough.
In short, _it is_ a prototype :D.

I will just commit it now, so you can see if you like it and what you might prefer to have changed (design, functionality, code).
If you find it to be a catastrophy, that's fine, I can just revert and will still have learned some more Dart ;).

Here's an example screenshot:

Chaphasilor:
That does look nice! A few thoughts:  

- The presets should all have the same width
- The text for the `x1.0` preset should be (semi-)bold, to make it stand out a little bit more
- Yeah, that input field doesn't fit. Take a look at the login flow on how to style it differently
- Regarding the keyboard covering up the input, I see two solutions:
  - Use the scrollController of the menu bottom sheet and fully extend the sheet to the top when the config section is expanded
  - Add a key to the bottom of the expanded section (e.g. to a `SizedBox.shrink()`) and scroll to that key (+ some offset) when the keyboard is shown. There should be a way to detect that.
    - Thinking about it, it might be better to listen to a focus event for the input field and then scroll the input field into view (+offset) instead. You can take a look at the FloatingActionButton in the queue panel for how to scroll to a key :)
    
I really like the design overall, great job! And thanks for your understanding <3

lymnyx:
Done.

Right now I have written it in a way where the bottom sheet scrolls a bit higher when the config section is expanded as I've had some issues with getting the other approach to work, I might change that again later. Not sure if scrolling from `0.6` to `0.8` is robust in terms of devices with less vertical space.

We could also always scroll the active preset (+ offset) into view (initially and on change), I think that would be nice UX-wise.

Other than that, I've noticed that the animations in the app in general, even with the `Curves.easeInOut`, don't feel that smooth, on my device at least. When expanding for example, it feels like it is lagging a bit (setting the duration higher doesn't really change that for me either). Is that a known Flutter thing and maybe a layout issue or why is that?

lymnyx:
Oops, I kind of forgot about `Column` there, sorry.

> Would also be nice if you could remember the previous extent [...]

Yep, I will try to implement that.

Chaphasilor:
That was quick, but I now realize that what I wanted is that the scroll position is kept if the menu was scrolled manually. So just check if the current extent is different from the extent you set it to when opening the section, and if it's different don't change the extent. If it's the same, you may set the position to the original position (which you just implemented).

lymnyx:
Like this?

Chaphasilor:
Sorry, I was asleep ^^  
That works great! I managed to add a small animation too, let me know what you think.

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+    required this.type,
+    required this.values,
+    required this.activeValue,
+    this.onTap,
+    this.mainColour,
+    this.onPressed,
+  }) : super(key: key);
+
+  // for future preset types other than "speed"
+  final String type;
```
Chaphasilor, line 24: Can you make this an enum please? You can put it as a public member of the file, or as a static member of the `PresetChip` class

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+
+    return LayoutBuilder(builder: (context, constraints) {
+      return SingleChildScrollView(
+        controller: _controller,
+        scrollDirection: Axis.horizontal,
+        child: Wrap(
+          spacing: 8.0,
+          runSpacing: 8.0,
+          crossAxisAlignment: WrapCrossAlignment.center,
+          children: List.generate(widget.values.length, (index) {
```
Chaphasilor, line 71: Maybe move the inner function (`(index) {...}`) into a separate function that is also passed the `Type` and that always returns a `PresetChip`? This way we actually customize the presets based on the type, and because we're using an enum  we can't have the possibility of passing an unknown type.

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+  }) : super(key: key);
+
+  // for future preset types other than "speed"
+  final String type;
+
+  final List<double> values;
+  final double activeValue;
+  final Function()? onTap;
+  final Color? mainColour; // used for different background colours
+  final Function()? onPressed;
```
Chaphasilor, line 30: `onPresetSelected`, and pass the value of the selected preset :)

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+                  ? widget.mainColour?.withOpacity(0.4)
+                  : widget.mainColour?.withOpacity(0.1),
+              isTextBold: currentValue == 1.0,
+              width: widget.chipWidth,
+              onTap: () {
+                setState(() {
+                  nowActiveValue = currentValue;
+                });
+                _queueService.setPlaybackSpeed(currentValue);
+                widget.onPressed!();
```
Chaphasilor, line 98: since the callback is nullable (`Function()?`), don't use the `!` operator here. Instead use `widget.onPressed?.call()`, so that in the case of a `null` value, it simply does nothing instead of throwing an exception.

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
+import 'dart:math';
+import 'package:flutter/material.dart';
+import 'package:finamp/services/queue_service.dart';
+import 'package:get_it/get_it.dart';
+
+const _radius = Radius.circular(4);
+const _borderRadius = BorderRadius.all(_radius);
+const _height = 36.0;
+final _defaultBackgroundColour = Colors.white.withOpacity(0.1);
+final _queueService = GetIt.instance<QueueService>();
```
Chaphasilor, line 10: Please always move the `GetIt` calls into a widget. You should move this one into `_PresetChipState`

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+    this.onTap,
+    this.mainColour,
+    this.onPressed,
+  }) : super(key: key);
+
+  // for future preset types other than "speed"
+  final String type;
+
+  final List<double> values;
+  final double activeValue;
```
Chaphasilor, line 27: Does this always use the active value? Isn't it more of a `defaultValue` or `initialValue`?

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+
+  final chipWidth = 55.0;
+
+  @override
+  State<PresetChips> createState() => _PresetChipsState();
+}
+
+class _PresetChipsState extends State<PresetChips> {
+  final _controller = ScrollController();
+  var scrolledAlready = false;
```
Chaphasilor, line 40: `bool`

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+            return PresetChip(
+              value: newValue,
+              backgroundColour: currentValue == nowActiveValue
+                  ? widget.mainColour?.withOpacity(0.4)
+                  : widget.mainColour?.withOpacity(0.1),
+              isTextBold: currentValue == 1.0,
+              width: widget.chipWidth,
+              onTap: () {
+                setState(() {
+                  nowActiveValue = currentValue;
```
Chaphasilor, line 95: updating `nowActiveValue` probably doesn't do anything, since it's immediately re-assigned during the build (line 61). But since `setState()` causes a rebuild, it probably does what you want it to. Try using `widget.activeValue` directly where you need it.

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+    return SizedBox(
+      width: width,
+      height: _height,
+      child: Material(
+        color: backgroundColor,
+        borderRadius: _borderRadius,
+        child: InkWell(
+            onTap: onTap,
+            borderRadius: _borderRadius,
+            child: Center(
```
Chaphasilor, line 138: I think this can be replaced with a simple `TextButton` with the appropriate `ButtonStyle`

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -0,0 +1,153 @@
...
+              child: Text(
+                value,
+                style: TextStyle(
+                    color: color,
+                    overflow: TextOverflow.ellipsis,
```
Chaphasilor, line 143: no need to set the overflow behavior twice :)

lib/components/AlbumScreen/song_menu.dart
```diff
@@ -125,6 +130,11 @@ class _SongMenuState extends State<SongMenu> {
   ColorScheme? _imageTheme;
   ImageProvider? _imageProvider;
 
+  // Makes sure that widget doesn't just disappear after press while menu is visible
+  var speedWidgetWasVisible = false;
+  var showSpeedMenu = false;
```
Chaphasilor, line 135: `bool`

lib/components/AlbumScreen/song_menu.dart
```diff
@@ -170,6 +180,62 @@ class _SongMenuState extends State<SongMenu> {
...
+      for (var i = 0; i < genres.length; i++) {
+        if (["audiobook", "speech"].contains(genres[i].toLowerCase())) {
+          return true;
+        }
+      }
+
+      try {
+        var parent =
+            await _jellyfinApiHelper.getItemById(widget.item.parentId!);
+        if (parent.runTimeTicks! > 72e9.toInt()) {
```
Chaphasilor, line 202: how many minutes is this? add a comment please!

lib/components/AlbumScreen/song_menu.dart
```diff
@@ -170,6 +180,62 @@ class _SongMenuState extends State<SongMenu> {
...
+      } catch (e) {
+        errorSnackbar(e, context);
+      }
+    }
+
+    return false;
+  }
+
+  final inputStep = 0.9;
+  var oldExtent = 0.0;
```
Chaphasilor, line 214: Use the value from line 250 (`initialChildSize`) here. You'll have to define it as a variable.

lib/components/AlbumScreen/song_menu.dart
```diff
@@ -249,6 +317,11 @@ class _SongMenuState extends State<SongMenu> {
                                       ?.playbackOrderShuffledButtonLabel ??
                                   "Shuffling",
                         };
+                        final playbackSpeedTooltip =
+                            AppLocalizations.of(context)
+                                    ?.playbackSpeedButtonLabel(
```
Chaphasilor, line 322: Just rely on `AppLocalizations` being initialized here, and use the `!` operator, no need to provide a fallback text. The existence of the actual token is checked during compile time.

lib/components/AlbumScreen/song_menu.dart
```diff
@@ -266,90 +339,129 @@ class _SongMenuState extends State<SongMenu> {
...
+                        }
+                        return FutureBuilder<bool>(
+                            future:
+                                shouldShowSpeedWidget(playbackBehavior.speed),
+                            builder: (context, snapshot) {
+                              if (snapshot.connectionState ==
+                                      ConnectionState.done &&
+                                  snapshot.data == true) {
+                                speedWidgetWasVisible = true;
+                                sliverArray.insertAll(2, [speedWidget]);
```
Chaphasilor, line 446: I'm not a big fan of this, it makes it hard to read the code and find the relevant section for the widget. Can you try supplying the sliverArray directly to the `SliverCrossAxisGroup`, and using the FutureBuilder there, so that it's in the right position? You can return a `SliverPadding` as long as the speed controls shouldn't be shown.

lib/components/AlbumScreen/song_menu.dart
```diff
@@ -170,6 +180,62 @@ class _SongMenuState extends State<SongMenu> {
     }
   }
 
+  Future<bool> shouldShowSpeedWidget(currentSpeed) async {
```
Chaphasilor, line 183: There is no `SpeedWidget` (it's a button and a menu). Maybe call it `SpeedControl(s)` instead?

lib/components/AlbumScreen/song_menu.dart
```diff
@@ -874,13 +988,28 @@ class PlaybackAction extends StatelessWidget {
...
+                  ),
+                  Positioned(
+                    bottom: -2,
+                    child: Text(
+                      value ?? "",
+                      style: Theme.of(context).textTheme.labelSmall,
+                    ),
+                  ),
+                ],
+              ),
```
Chaphasilor, line 1011: We probably don't need this value text anymore, right?

lib/components/AlbumScreen/speed_menu.dart
```diff
@@ -0,0 +1,152 @@
...
+                        _queueService.setPlaybackSpeed(1.0);
+                        setState(() {
+                          currentSpeed = 1.0;
+                          _textController.text = currentSpeed.toString();
+                        });
+                      },
+                    ),
+                    Expanded(
+                      child: Container(
+                        margin: EdgeInsets.symmetric(horizontal: 20.0),
```
Chaphasilor, line 99: a little too big for my taste, and I prefer to use `paddings` instead. They increase the tapable area of widgets.

lib/components/AlbumScreen/speed_menu.dart
```diff
@@ -0,0 +1,152 @@
...
+                                    100;
+
+                            _textController.text = valueDouble.toString();
+
+                            _queueService.setPlaybackSpeed(valueDouble);
+
+                            setState(() {
+                              currentSpeed = valueDouble;
+                            });
+                          },
```
Chaphasilor, line 129: This doesn't seem to be working for me. Clicking the confirm button on the keyboard does not change the speed, it changes only after I hit apply

lib/components/LayoutSettingsScreen/CustomizationSettingsScreen/playback_speed_control_visibility_dropdown_list_tile.dart
```diff
@@ -0,0 +1,37 @@
...
+                      child: Text(e.toLocalisedString(context)),
+                    ))
+                .toList(),
+            onChanged: (value) {
+              if (value != null) {
+                FinampSettingsHelper.setContentPlaybackSpeedType(value);
+              }
+            },
+          ),
+        );
```
Chaphasilor, line 33: You could add a "More Info" button that explains how the automatic detection works. Check out the ReplayGain settings for an example

lib/l10n/app_en.arb
```diff
@@ -443,6 +443,30 @@
...
+  "speed": "Speed",
+  "@speed": {},
+  "reset": "Reset",
+  "@reset": {},
+  "apply": "Apply",
+  "@apply": {},
+  "on": "On",
+  "@on": {},
+  "off": "Off",
+  "@off": {},
```
Chaphasilor, line 469: They're not used, right? But I guess they don't really hurt

lib/l10n/app_en.arb
```diff
@@ -443,6 +443,30 @@
...
+  "playbackSpeedControlSetting": "Playback Speed Visibility",
+  "@playbackSpeedControlSetting": {},
+  "playbackSpeedControlSettingSubtitle": "Whether the playback speed controls are shown in the player screen menu",
+  "@playbackSpeedControlSettingSubtitle": {},
+  "automatic": "Automatic",
+  "@automatic": {},
+  "shown": "Shown",
+  "@shown": {},
+  "hidden": "Hidden",
+  "@hidden": {},
```
Chaphasilor, line 459: Please add descriptions that explain how and where these tokens are used, and (if necessary) how they should be worded

lib/l10n/app_en.arb
```diff
@@ -778,6 +802,15 @@
   "@playbackOrderShuffledButtonLabel": {
     "description": "Label for the button that toggles the playback order between linear/in-order and shuffle, while the queue is in shuffle mode"
   },
+  "playbackSpeedButtonLabel": "Playing at x{speed} speed",
+  "@playbackSpeedButtonLabel": {
+    "description": "Label for the button that changes playback speed, {speed} is the current playback speed.",
```
Chaphasilor, line 807: The button itself doesn't change playback speed anymore

lib/models/finamp_models.dart
```diff
@@ -1011,3 +1024,46 @@ enum ReplayGainMode {
   @HiveField(2)
   albumOnly,
 }
+
+@HiveType(typeId: 64)
+enum ContentPlaybackSpeedType {
```
Chaphasilor, line 1029: I'd prefer a name like `PlaybackSpeedVisibility`

lib/models/jellyfin_models.dart
Chaphasilor, line 1: Ideally, try to not format files you didn't make any changes to. Just discard the changes when staging/committing

lib/screens/customization_settings_screen.dart
```diff
@@ -0,0 +1,45 @@
...
+          // TODO add button to reset to defaults
+          // IconButton(
+          //   onPressed: () {
+          //     setState(() {
+          //       FinampSettingsHelper.resetTabs();
+          //     });
+          //   },
+          //   icon: const Icon(Icons.refresh),
+          //   tooltip: AppLocalizations.of(context)!.resetTabs,
+          // )
```
Chaphasilor, line 33: Would you be willing to add this? :)

lib/services/finamp_settings_helper.dart
```diff
@@ -182,6 +182,13 @@ class FinampSettingsHelper {
         .put("FinampSettings", finampSettingsTemp);
   }
 
+  static void setContentPlaybackSpeedType(ContentPlaybackSpeedType contentPlaybackSpeedType) {
```
Chaphasilor, line 185: Make sure to also rename this method

lib/services/queue_service.dart
```diff
@@ -728,6 +739,21 @@ class QueueService {
...
+
+    _playbackSpeedStream.add(speed);
+
+    _audioHandler.setSpeed(speed);
+
+    FinampSettingsHelper.setPlaybackSpeed(playbackSpeed);
+
+    _queueServiceLogger.fine(
+        "Playback speed set to ${FinampSettingsHelper.finampSettings.playbackSpeed}");
+  }
```
Chaphasilor, line 753: This can probably be condensed a bit?

Chaphasilor:
Also, the layout shift after fetching the item information is still an issue, correct? I might need to look at that...

lib/l10n/app_en.arb
lymnyx: I thought so as well, but I will remove them. If they are reintroduced, they are reintroduced.

lib/models/jellyfin_models.dart
lymnyx: Yes, noticed that as well. Sorry!

lib/components/AlbumScreen/speed_menu.dart
lymnyx: @Chaphasilor Yes, I implemented this like the input for the sleep timer dialog (same thing there, having to click "Apply"), but I will try to change that.
Speaking of which, we might want to redesign the sleep timer settings? After all, the preset chips are usable for any preset types ;).

lib/components/AlbumScreen/song_menu.dart
lymnyx: I thought you maybe wanted to keep it, but I will remove it then.

lib/components/AlbumScreen/speed_menu.dart
lymnyx: Okay, now pressing the confirm button works too.

lib/components/AlbumScreen/preset_chip.dart
lymnyx: Because of the rebuilds, it is always the active value, I think. The `nowActiveValue` was unnecessary though.

lib/components/AlbumScreen/preset_chip.dart
lymnyx: After trying this, it seems to make things more complicated. I would just keep it like it is right now.

Or is there a better way than doing the `TextButton` like this:
```
TextButton(
  style: ButtonStyle(
    backgroundColor: MaterialStateProperty.resolveWith<Color?>(
        (Set<MaterialState> states) {
      return backgroundColor;
    }),
  ),
  onPressed: onTap,
  child: Text(
    value,
    style: TextStyle(
      color: color,
      overflow: TextOverflow.visible,
      fontWeight: isTextBold! ? FontWeight.w700 : FontWeight.normal,
    ),
    softWrap: false,
  ),
),
```

Even with that, there is still a border radius one has to get rid of in the `ButtonStyle`.

lib/components/AlbumScreen/song_menu.dart
lymnyx: I think the reason for me doing it this way was that simply adding the `FutureBuilder` to the `CrossAxisGroup` already takes up the typical button space (like a child `<div>` with `align-items: stretch` ;)). Maybe there is a way to prevent that, I just haven't found it yet.

lymnyx:
Has #568 somehow broken/removed `item.genres` for the song info specifically? It doesn't matter which media I play, my genre detection doesn't work because `item.genres` and `item.genreItems` are empty even though the objects very much have genres. Might again be an error on my part somehow though.

Could you therefore please play something (that has a genre) on your device/emulator with the current `redesign` codebase, open the song menu and tell me if the "Go to Genre" button is there :D?

Chaphasilor:
I'll take a look at it later, but it seems like you're onto something

lib/components/AlbumScreen/preset_chip.dart
lymnyx: I see that I can definitely optimize it, but we need it to rebuild the `SpeedMenu` and update the text field, right?

lib/screens/customization_settings_screen.dart
lymnyx: Added, for the `customization_settings_screen` at least.

Chaphasilor:
@lymnyx so the culprit seems to be 9b983864b9f1c1a899fc4d7c6c3f6b351b862683. It prevents the default `fields` from being applied when fetching items, so the genres and some other fields are not being returned by the server.  
I just committed a fix, please let me know if it works!

Also, thank you for reporting the issue with the FavoriteButton! It definitely shouldn't throw an error, I'll disable it :)

lymnyx:
Great, everything seems to work again :)!

lib/components/AlbumScreen/preset_chip.dart
Chaphasilor: We can do it like this, yes.  
Alternatively, you could use a `ValueListenableBuilder` and listen to `FinampSettingsHelper.finampSettingsListener`. This way, every time the speed is changed, the speed menu will rebuild.

lib/components/AlbumScreen/song_menu.dart
Chaphasilor: I see. I guess it's fine for now...

lib/components/AlbumScreen/song_menu.dart
Chaphasilor: Yes, I did it for you :)

Chaphasilor:
Sorry for not responding earlier, I'm rather busy atm. I'll try to get the TextButton to work tonight.

> Other than that, I've noticed that the animations in the app in general, even with the Curves.easeInOut, don't feel that smooth, on my device at least. When expanding for example, it feels like it is lagging a bit (setting the duration higher doesn't really change that for me either). Is that a known Flutter thing and maybe a layout issue or why is that?

Have you tested it with a release build yet? `flutter build apk --release` + `flutter install --release` or just `flutter run --release` should do the trick. That should definitely improve the performance, although it might still not be perfect.

Regarding your question about the sleep timer, yes, we should probably migrate to the same interaction style, but the controls should probably be different. Presets for specific durations probably doesn't hurt, but highlighting the currently active preset doesn't really make sense. I'd also like to add an option for "Play last song until the end".  
I think it should be done in a separate PR, once this one is merged.

lymnyx:
Sure, no problem.

Ah, thanks, on release it is definitely a lot better. Interesting.

Chaphasilor:
I updated the layout a bit, let me know what you think. I tried to make the layout a bit more explicit and avoid using default styles where-ever possible, but Flutter makes that really hard. Hopefully it looks fine on your device as well!

lib/components/AlbumScreen/preset_chip.dart
Chaphasilor: I changed that, it's not much shorter but now semantics are correct :)

lymnyx:
It works well.

Great, fixing the layout issue and one more review,
if you don't have any other concerns, we can merge thereafter, I think :).

Chaphasilor:
Just saw another example in the Telegram desktop app:  
image

I'm starting to like the slider more, especially since I caught myself thinking that I can hit the "apply" button to collapse the menu.  Maybe if we can pull off a slider that's fine-grained enough (at least 5% steps should be possible), we could use that instead of a number input. I'm thinking of a fat slider, not a thin one. What do you think?

lymnyx:
(Sorry for the late answer.)

As that was my initial idea for the menu, I have no problem with it ;).

I would then remove the "Apply" button obviously, but keep the "Reset" button, right?

I also assume, we keep the presets and only change the second column of the menu.

---

Other than that, I have apparently found another bug. (By the way, are you okay with me throwing those at you every time, or should I always open an issue?)

When trying to remove an item from a downloaded playlist (in online mode), the item disappears in UI as expected, it does not seem to be actually sending the API request for removing though, not succeeding in removing the track while sending a "positive" `GlobalSnackbar`.

Another small UI thing is that in the different tabs (Albums, Artists, Playlists, Genres, ...), the item's favourite icon and the vertical "shortcut by letter" element on the right have basically no padding or margin inbetween one another. We might want to change that.

Chaphasilor:
Thanks for the feedback. I'm actually not sure how exactly the UI should be structured with the slider, I'd prefer to not have the reset button there, and instead highlight the "x1.0" preset a bit more. The less buttons the better, and the slider needs as much horizontal space as we can give it. Somewhere in the slider it should also show the currently selected speed, maybe using a tooltip as well?  
I might play around with the design tomorrow...

Regarding your bugs and fixes, thanks for those. I would prefer to have them separately though, especially the bug reports. And feel free to open a PR even for tiny changes, then we can get those merged in quickly.  
I'll take a look at the playlist issue, but I'm not sure if I'll change the like indicator for the initial release. The track list tiles will be redesigned anyway...

lymnyx:
Yes, I think it's better situated in the song menu.

lymnyx:
Alright, we might want to, for example, add buttons to easily decrease and increase speed, but here's a simple slider with tooltip.

Chaphasilor:
Okay, thanks for bearing with me. Just took a look at it's heading in a nice direction. I'd actually make that slider much thicker though, and lowlight the "thumb". Depending on what range we choose in the end, we can keep the increment/decrement buttons.  
I think x5 is probably a bit too high, I can barely understand anything at x2.5 already. x0.2 as the lower limit might also be too low.  
How about x0.35 to x3?

lymnyx:
Sorry for the late response. Sure, I will change the design a bit and reduce the speed range.

IIRC the reason I didn't always apply the new speed while dragging was the app lagging quite a bit with that, but I might just apply it less often (after bigger drag value changes for example). Haptic feedback shouldn't be a problem, which `FeedbackType` should I use? `impact`, `selection`, `success`?

Chaphasilor:
Just use what feels best, but `light` or `selection` is probably a good choice.

You could also try to debounce the speed change by only applying after 300-500 ms without a change.

Chaphasilor:
@lymnyx do you still have time to work on this, or should I take over?

lymnyx:
Sorry, I am currently away from home and not able to work on this until approximately next tuesday.

Feel free to add the changes discussed or whatever you deem to be necessary (would be great if you could take a look at pre-fetching the `shouldShowSpeedSelector` data) :).

Chaphasilor:
Dammit, I did the same an hour or so ago but didn't get around to push it yet. Hold on

Chaphasilor:
Okay, let me know what you think of the changes!

lymnyx:
Looks good to me, visually I liked the horizontal margin on the bottom row of the menu though, this way the plus and minus buttons feel a little cramped and out of place as everything above and below it has more padding/margin. Personally I think 2.5x is a bit low as a maximum speed as I sometimes listen to audiobooks on higher speeds, but maybe that is just me.

Chaphasilor:
Okay, we can increase the max value again. The problem with adding margin is that the width of the slider decreases, which makes it harder to select a certain value.

lymnyx:
Yes, I just think a bit of margin wouldn't change this significantly while still being visually more appealing.

lymnyx:
Like this, maybe?

Chaphasilor:
In the end, it depends on the width of the device. But we can probably add a bit of margin again, you're right.  
I can do it later.

So the only thing missing now is the pre-fetching of metadata. I can take a look at it, since something similar will be needed for getting lyrics as well.

Chaphasilor:
> Like this, maybe?

I think that's a bit too much, but I'll have to check on my phone later. Maybe it's still possible to select a certain step, in that case all is well.

lymnyx:
At least on my phone it wasn't a problem.

lymnyx:
An alternative would be to have the slider go to 2.5x, but still provide "3x" and "3.5x" presets, for example. Would you like that better?

Chaphasilor:
> At least on my phone it wasn't a problem.

Which one do you have?

> An alternative would be to have the slider go to 2.5x, but still provide "3x" and "3.5x" presets

So the exact opposite of what we have right now? ^^  
Possible. The slider has no value indicators anyway, so it probably would be fine.  
I think we can try it out and gather some initial feedback with that.

lymnyx:
> Which one do you have?

Google Pixel 6a, so, to be fair, quite big. Not sure about the Apple SEs ;).

> So the exact opposite of what we have right now? ^^

Yep :D, sorry.

> I think we can try it out and gather some initial feedback with that.

Okay!

Chaphasilor:
I have an Asus Zenfone 10, so rather small and tall

Chaphasilor:
Well, selecting the x3.5 results in an exception, a simple clamp should fix that.

lymnyx:
Good point, what a stupid mistake, should have tested for that.

I've also added a small indicator for when the user selects a preset value above the maximum the slider can display.
Let me know what you think.

Chaphasilor:
Yeah that looks good and works well!

Chaphasilor:
@lymnyx just to give you a quick update, I've now implemented a metadata provider that give us info for the upcoming tracks in #693, which I should be able to re-use for deciding if the button should be shown or not. That means I can get this PR finished up once the lyrics PR is merged.  

I noticed that the current check might also return true for playlists (depending on where the `parentId` is coming from, I'll have to check). Do you have any other ideas to make the detection more reliable (even if it would be a bit more involved)?

lymnyx:
Great, thanks for the update!

> I noticed that the current check might also return true for playlists (depending on where the parentId is coming from, I'll have to check).

In case the `parentId` of the playlist items is your concern, that, after a quick test with a playlist of mine, seems to be, like expected, the album id. The menu doesn't show the speed options for me when starting tracks from huge playlists. Not sure if this is what you meant.

> Do you have any other ideas to make the detection more reliable (even if it would be a bit more involved)?

I mean, theoretically we could check the pre-fetched audio for (repeated/prevalent) near-silence and detect typical audiobooks that way, maybe skipping the first few seconds of audio to account for music having to start. Not sure if you would call that reliable or worth it though ;).

As we are already checking for the album length (provided that it works as expected which I've assumed) and the item genres, I am not sure what would be left for us to easily and reliably analyze as we obviously have to differentiate normal music tracks and audiobook tracks that might be, as far as the backend is concerned, categorized in precisely the same way.

Chaphasilor:
yeah that's what I meant, good to know. thanks for testing.

Analyzing the audio itself wouldn't be a great idea I think, and much to complex either way.  
At some point we might have a way to mark a whole library as containing audio books, either in Finamp or in Jellyfin itself, that would work better for this.  
We should also try to support podcasts, so maybe checking the duration of the current track isn't a bad idea either.

I guess if the speed menu is "accidentally" shown, it's not a big deal...

Chaphasilor:
@lymnyx could you give the new changes some testing? Everything should be working now, and if you give the go, I'll include it in the update :D

Chaphasilor (approved):

Chaphasilor:
@lymnyx many thanks for your efforts!

lymnyx:
@Chaphasilor Thank you so much for your not-so-minimal assistence and patience <3, really helped a lot. Great to have this in the beta now!

PR #624, opened 2024-03-08
:bug: Fix sort by DateCreated not working when offline

### LINKED ISSUE #444: Bug: Sorting as "Date Added" changes between Online and Offline Mode
Just as the title says when sorting tabs as "Date Added" the ordering of the items changes when between Online and Offline (this is for the Album and Song tabs). 

- When Online and connected to the server, feature works as intended showing most recently added Albums/Songs first
- When Offline the sorting changes and is as far as I can tell completely random

https://user-images.githubusercontent.com/7260835/236656062-e13ba349-7e08-4372-8258-0bff331654f9.mp4

Maxr1998:
By default, the DateCreated field is missing from the Jellyfin Items response. This causes the DateCreated field to always be null on BaseItemDtos, so it's not available for sorting when offline.

By adding the field name to the defaultFields, this is resolved. However, all items will have to be resynced before sorting works offline.

I also sorted the fields and changed the field name capitalization to match the Jellyfin API docs. The unchanged fields aren't mentioned there, so I'm not sure they're even necessary.

Fixes #444. Targeting the main branch, but should be trivial to port to the redesign branch.

### DIFF
--- a/lib/services/jellyfin_api.dart
+++ b/lib/services/jellyfin_api.dart
@@ -14,7 +14,7 @@ import 'jellyfin_api_helper.dart';
 part 'jellyfin_api.chopper.dart';
 
 const String defaultFields =
-    "parentId,indexNumber,songCount,childCount,providerIds,genres,tags,Etag,albumPrimaryImageTag,parentPrimaryImageItemId";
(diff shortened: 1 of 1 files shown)


Chaphasilor:
Hey, thanks for the PR!  
Changes look good, I'll give it a try in a bit and then merge it.

About the redesign, I think there already was an attempt by @jmshrv to make the sorting more consistent, although I think that was limited to ignoring articles (like "the"). So that should indeed be a simple fix :)

Chaphasilor:
Seems to be working 👍🏻  
tysm!

PR #627, opened 2024-03-10
[Redesign] Improved end of queue behaviour with alternate loop modes

Komodo5197:
Tried to improve behavior when playing or skipping past end of queue with loop mode set to one track or none.  The queue now ends up paused at the start of the first track.

### DIFF
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -284,7 +284,7 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
           : (_player.currentIndex ?? 0) + offset;
       if (queueIndex >= (_player.effectiveIndices?.length ?? 1)) {
         if (_player.loopMode == LoopMode.off) {
-          await _player.pause();
+          await _player.stop();
         }
         queueIndex %= (_player.effectiveIndices?.length ?? 1);
       }
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -233,6 +233,7 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
         doSkip = _player.position.inSeconds < 5;
       }
 
+      // This can only be true if on first track while loop mode is off
       if (!_player.hasPrevious) {
         await _player.seek(Duration.zero);
       } else {
@@ -282,11 +283,17 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
               offset
           : (_player.currentIndex ?? 0) + offset;
       if (queueIndex >= (_player.effectiveIndices?.length ?? 1)) {
-        await _player.pause();
-        queueIndex = 0;
+        if (_player.loopMode == LoopMode.off) {
+          await _player.pause();
+        }
+        queueIndex %= (_player.effectiveIndices?.length ?? 1);
       }
       if (queueIndex < 0) {
-        queueIndex = 0;
+        if (_player.loopMode == LoopMode.off) {
+          queueIndex = 0;
+        } else {
+          queueIndex %= (_player.effectiveIndices?.length ?? 1);
+        }
       }
       await _player.seek(Duration.zero,
           index: _player.shuffleModeEnabled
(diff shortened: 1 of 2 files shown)


Chaphasilor:
Thanks for the PR!

Hmm, I'm a bit torn here regarding the behavior when loop is off. On the one hand, it does make sense that pressing play after the queue ended will resume the queue from the start. But on the other hand, the queue should end on the last item, and not on the first item.  
Ideally we could do both, but I'm not sure how easy that is to do...

I also believe the way you implemented it might interfere with playback reporting. I can't test it right now because I'm testing something else, but maybe you could confirm that after the queue ends, only the end of the last track is scrobbled, but not the start of the first track (at least until play is pressed again)?

Looping the whole queue when loop one is enabled probably comes down to personal preference. I don't use the feature, but we could add is as an option to the audio service settings?

Komodo5197:
The current no-loop behavior is that the player stops, and then everything gets a bit confused by having a queue but no current track.  I played a queue in no loop, it completed and stopped, and then later when I restarted the app it put me at the start of the last song which is definitely not what I want to happen.  Both the now-playing screen and the queue also got a bit janky.

Properly implementing a playback completed state in the now-playing screen seems like more trouble than it's worth, and it would still be a bit weird if it shows the last song but then you press play and suddenly it jumps to the first song.  I think the cleanest and most consistent way to do this would be to fully drop out of now playing when reaching the end, but with the current system that would also drop all access to the queue which is obviously unacceptable.  And trying to disentangle the queue from now-playing seems really difficult for little benefit.

Regarding the playback history, I get a 'reporting track change event' log but do not get a 'Starting playback progress' log.  The first track does seem to show up in the playback history tab but I don't think it gets reported to the server.

Regarding looping one track, I just saw an issue that wanted it to work like this and threw it in.  It makes more sense to me then the current behavior of jumping back to the start of the same song when you skip forwards at the end, which doesn't seem like it would ever be useful.

Chaphasilor:
Okay, I managed to make the playback history also consistent. The stop event for the last track is reported, then the player switches to the first track but doesn't report it to the playback history or Jellyfin yet. Only when pressing play again is the new track (first track) reported to both the playback history and Jellyfin. So that is working great.

But I'm not yet happy with the skipping with loop one enabled. While it does skip from the last track to the first track, there are two issues:  
- Even though the previous track was playing, the new track isn't playing. This is inconsistent with skipping from any other track to the next track, where the new track starts playing immediately
- Skipping back doesn't wrap-around to the end (that is probably an easy fix)

Could you take a look at the first issue?

Chaphasilor:
Okay, finally have some time to look at your 6 (!) PRs 😁  
I changed the action performed when manually skipping past the end when not looping from pause to stop, so that the "has finished playing" event is properly reported to the server. I believe this also makes a bit more sense semantically, and since the player behavior stays largely the same it shouldn't be an issue. If you disagree, I can think about changing this later on, but I'll merge this now.

Komodo5197:
@Chaphasilor Stop does make more sense.  The primary issue with stop is that it causes a re-shuffle.  This is probably not desired, and also hits the 'first track doesn't shuffle' bug in the form of the last track from the old shuffle always being the first track of the new shuffle, so if you just hit play to keep going it always plays the same song twice in a row.

Chaphasilor:
Ahhhh you gotta be kidding me 😅  
It's always *something* with just_audio...

I guess I'll revert to pause then, and change the detection logic in the PlaybackHistoryService.

Komodo5197:
@Chaphasilor Hey, are you planning on reverting this before the next release?  I was messing with this a bit more, and noticed that _player.stop() while shuffled completely confuses the queue until you perform a complete queue replacement.  I think the queue service and just_audio seem to persistently disagree about where the new queue starts/ends?  Also, I also noticed that this only happens when skipping, because you didn't revert the change to the playback complete event.

Chaphasilor:
Thanks for making me aware of that. I spent another two hours trying to consolidate the behavior so that is as similar as possible in all cases, but because the Playback History Service is relying directly on just_audio events, this isn't easy.  
We probably need to introduce our own events on top of the just_audio and audio_service events at some point to properly get the information we need to the placed where it's needed, but for now my workaround is to skip to the end of the last item, so that the player produces a `ProcessingState.completed` event of its own.

I'll push the update with this, but let me know if there are still issues that we can fix with the next updates!

PR #636, opened 2024-03-16
[Desktop Beta] Improved file Locations

Komodo5197:
Moved all database files to ApplicationSupport directory on desktop platforms, which is in appdata/roaming on windows.  Added the ability to set a custom download location as the default.

### DIFF
--- a/lib/components/AddDownloadLocationScreen/custom_download_location_form.dart
+++ b/lib/components/AddDownloadLocationScreen/custom_download_location_form.dart
@@ -131,9 +131,10 @@ class _CustomDownloadLocationFormState
             },
           ),
           const Padding(padding: EdgeInsets.all(8.0)),
-          Text(AppLocalizations.of(context)!.customLocationsBuggy,
-              textAlign: TextAlign.center,
-              style: const TextStyle(color: Colors.red)),
+          if (Platform.isIOS || Platform.isAndroid)
+            Text(AppLocalizations.of(context)!.customLocationsBuggy,
+                textAlign: TextAlign.center,
+                style: const TextStyle(color: Colors.red)),
         ],
       ),
     );
--- a/lib/components/AlbumScreen/album_screen_content.dart
+++ b/lib/components/AlbumScreen/album_screen_content.dart
@@ -61,73 +61,71 @@ class _AlbumScreenContentState extends State<AlbumScreenContent> {
       }
     }
 
-    return Scrollbar(
-      child: CustomScrollView(
-        slivers: [
-          SliverAppBar(
-            title: Text(widget.parent.name ??
-                AppLocalizations.of(context)!.unknownName),
-            // 125 + 64 is the total height of the widget we use as a
-            // FlexibleSpaceBar. We add the toolbar height since the widget
-            // should appear below the appbar.
-            // TODO: This height is affected by platform density.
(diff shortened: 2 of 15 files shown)


Chaphasilor:
I took a look at this while testing #639, and noticed some strange behavior (keep in mind that my install wasn't fully up to date with #639, but it should be up to date with this PR). When I download files to the internal storage, the download works successfully, as does playback, but I cannot find the files anywhere.  
The app reports the path as being in `AppData\Local\com.unicornsonlsd\Finamp`, but such a folder don't exist in `Local`:
image

A similar path exists (`AppData\Local\com.unicornsonlsd\finamp`), but while that one does contain the current logs and folders for `images` and `songs`, those folders are both empty.   
I also checked the custom location I had set up, but the files aren't in there either. 

The logs also say that the images are being fetched from the path shown in the download settings (which doesn't exist). So I'm really not sure what's going on...

Edit: Turns out the files are actually in `AppData\Local\Packages\com.unicornsonlsd.finamp_305mjwt9xw89r\LocalCache\Local\com.unicornsonlsd\Finamp`, which seems to be related to me using the msix-bundle to install the app. Not sure if there's a way to get the actual paths at runtime somehow? And I also don't know why this is only being used for downloads, but not settings...

Uninstalling remove the downloaded files (in the `Packages`) folder, but leaves the settings untouched. That means after re-installing the tracks are still marked as downloaded, but your DownloadsService actually detects that and will automatically download the needed files! However, it does this even in offline mode, which should never happen...

Komodo5197:
It looks like msix app see the real and virtual appdata folders overlayed, so the app still saw and updated the existing logs and settings presumebly from an older non-msix install, but all new songs/images went into the virtual folder.  I don't think there's a great way to deal with this, but people should be using custom locations if they want to mess with the files anyway.  Windows app packages are just kind of weird.  I think this is also why the songs deleted but the settings did not.

Regarding the tracks getting redownloaded, I don't think missing files like this get noticed until you either try to play them or a downloads repair gets run.  Were files actually downloading in offline mode, or just getting updated to not downloaded?  Because both the sync loop and enqueue loop have offline mode checks in them which I would expect to stop things before any downloading can occur.

Chaphasilor:
I'll try a clean install with the msix tomorrow (if I don't forget), hopefully it will just put everything into the "virtual" folder then.
  
As far as I can tell the files were actually being downloaded (files were deleted before, I then went to the album page and the download indicators started appearing, even though the app started in offline mode (through persisted settings). But I will try to repro that, because I'd also be surprised if the offline mode checks would be missing...

Chaphasilor:
Okay, removing all Finamp files and re-installing resulted in all files (settings and downloads) ending up in the virtual folder. The download locations screen still reports the non-virtual folder (which doesn't exist) as the download location, is there any way to get the actual path?

Chaphasilor:
And I did *not* manage to reproduce the downloading in offline mode :)

Komodo5197:
Okay, sounds good.  I certainly don't know of a way to get the real path to the downloads.  I don't see any functionality to do that in the msix flutter package, and I don't have any other ideas from within flutter.  With the weird overlayed folders setup it's probably not possible to do this without writing at least some amount of native code, so I don't think this is getting fixed.

Chaphasilor:
Alright, it's fine then. I can take another look when I redesign the settings...

PR #639, opened 2024-03-17
[Desktop Beta] Add Splitscreen Mode

Komodo5197:
Added splitscreen mode when window is wide enough.  The player/queue will be shown on the right, all other screens show up on the left.  Probably works on tablets too.  Additionally, I've tried to make the song menu modal behave a bit better with a mouse.  I've messed with album image loading with blurhashes, and added a variant of the music screen grid mode that doesn't make a crazy amount of image requests when the window gets resized.  Built on top of https://github.com/jmshrv/finamp/pull/636.

Screenshot 2024-03-16 204542

### DIFF
--- a/lib/components/AddDownloadLocationScreen/custom_download_location_form.dart
+++ b/lib/components/AddDownloadLocationScreen/custom_download_location_form.dart
@@ -131,9 +131,10 @@ class _CustomDownloadLocationFormState
             },
           ),
           const Padding(padding: EdgeInsets.all(8.0)),
-          Text(AppLocalizations.of(context)!.customLocationsBuggy,
-              textAlign: TextAlign.center,
-              style: const TextStyle(color: Colors.red)),
+          if (Platform.isIOS || Platform.isAndroid)
+            Text(AppLocalizations.of(context)!.customLocationsBuggy,
+                textAlign: TextAlign.center,
+                style: const TextStyle(color: Colors.red)),
         ],
       ),
     );
--- a/lib/components/AlbumScreen/album_screen_content.dart
+++ b/lib/components/AlbumScreen/album_screen_content.dart
@@ -61,73 +61,71 @@ class _AlbumScreenContentState extends State<AlbumScreenContent> {
       }
     }
 
-    return Scrollbar(
-      child: CustomScrollView(
-        slivers: [
-          SliverAppBar(
-            title: Text(widget.parent.name ??
-                AppLocalizations.of(context)!.unknownName),
-            // 125 + 64 is the total height of the widget we use as a
-            // FlexibleSpaceBar. We add the toolbar height since the widget
-            // should appear below the appbar.
-            // TODO: This height is affected by platform density.
-            expandedHeight: kToolbarHeight + 125 + 80,
-            collapsedHeight: kToolbarHeight + 125 + 80,
-            pinned: true,
-            flexibleSpace: AlbumScreenContentFlexibleSpaceBar(
-              parentItem: widget.parent,
-              isPlaylist: widget.parent.type == "Playlist",
-              items: widget.queueChildren,
-            ),
-            actions: [
-              if (widget.parent.type == "Playlist" &&
-                  !FinampSettingsHelper.finampSettings.isOffline)
-                PlaylistNameEditButton(playlist: widget.parent),
-              FavoriteButton(item: widget.parent),
-              DownloadButton(
-                  item: DownloadStub.fromItem(
-                      type: DownloadItemType.collection, item: widget.parent),
-                  children: widget.displayChildren.length)
-            ],
+    return CustomScrollView(
+      slivers: [
+        SliverAppBar(
+          title: Text(
+              widget.parent.name ?? AppLocalizations.of(context)!.unknownName),
+          // 125 + 64 is the total height of the widget we use as a
+          // FlexibleSpaceBar. We add the toolbar height since the widget
+          // should appear below the appbar.
+          // TODO: This height is affected by platform density.
+          expandedHeight: kToolbarHeight + 125 + 80,
+          collapsedHeight: kToolbarHeight + 125 + 80,
+          pinned: true,
+          flexibleSpace: AlbumScreenContentFlexibleSpaceBar(
+            parentItem: widget.parent,
+            isPlaylist: widget.parent.type == "Playlist",
+            items: widget.queueChildren,
           ),
-          if (widget.displayChildren.length > 1 &&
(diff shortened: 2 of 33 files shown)


Chaphasilor:
Oh man, you're pushing out these amazing improvements to quickly I don't have a chance to keep up! Totally missed this one.  
I have exams coming up this week and the week after, but I'll see if I can try this out on the side.

Chaphasilor:
Had to try it out right-away :D  

Some issues I noticed:

- When resizing while the player screen is open, it will create two player screens ^^
- The player side could have a maximum width
- The vertical separator looks a bit out-of-place. I think just leaving it out completely would look better, or maybe using a thin, gray separator instead
- Snackbars are shown on both sides
- The grid improvement is a good idea, but could use some further tweaks:
  - Just because the covers don't scale with the width anymore, doesn't mean that the existing settings for number of columns can't be used anymore. It should be applicable to the new layout as well, since that is also using columns
  - I think the covers should still scale to fit the available space, but the automatic fetching based on the size should be disabled (or at least limited to specific thresholds). Do you think that's possible to do nicely?

Komodo5197:
I agree the separator could be better.  Right now I'm using a dependency that allows the split to be dragged around, but the grippable area is exactly the same size as the separator, so making it thinner means it is hard to change.  I'll mess around with this a bit to try and decouple them or something, or maybe pivot more towards auto-sizing the player ratio.

Regarding the new grid layout, I don't feel that the existing fixed column layout makes much sense at all in the context of resizable windows.  I want album covers as small as possible while still being easy to see, and I want them to stay that size across resizing.  If I make the window really small I need to still be able to actually see things, and if I make it really big I want to see a lot of albums, not just a few really big ones.  I don't think resizing the covers is ever really a helpful behavior, and it's not really something the current code does intentionally because it's built for fixed screen sizes, it just naturally fell out of the desire to cleanly align the grid.

In the context of the new grid layout, still using the column count setting seems unintuative.  What would it be, the count of columns at this particular time?   What would happen if you resized the window in settings?  Would you suddenly switch from portrait to landscape counts when passing a square window?  It just doesn't map cleanly.  About the scaling clamping, I've added code that does that when using the old grid view if the view may be resizable, and it it seems to work well enough.

Chaphasilor:
Ah, I didn't even notice that the separator could be dragged. And I agree that it's not necessarily needed, if we have sensible default sizes.

About the grid layout, maybe I missed something? In my tests the covers stayed the same size, and when dragging the window wider the space between the covers simply increased. So the amount of columns did stay the same. Maybe I didn't resize enough to trigger a wrap?  

Either way, we should take a look at how other apps are doing this, and stay consistent with that 👍🏻

Komodo5197:
The layout does reflow as soon as another cover can fit, yeah.  You must have resized a pretty small amount for it not to have happened.  Or maybe something bugged?  Anyway, I checked VLC, Musicbee and the default windows media player and they all have fixed size covers with reflow like this.  iTunes is similar but resizes covers between reflows instead of adding padding.  I think the current behavior seems fine, myself.

Anyway, I found a fork of the Splitview that already had the grip size and grabbable regions separated, so the divider is gone now.  I've also changed to try keeping the player width in pixels consistent across resizing, so it just stays at 400px wide if you never change it, which seems reasonable.

Chaphasilor:
I noticed the now playing bar stretches across the screen while the queue is being restored, and then the app switches to splitscreen mode and hides the now playing bar. Maybe the "restoring queue" placeholder could be moved to the player screen for splitscreen mode, if that isn't too complicated?

Also, the scaling doesn't seem to work properly for me:
image
The panel is way to narrow, both with a display scaling of 150% and 100%. 

Edit: seems like that was due to old settings from a previous installation. Manually dragging the screen wider fixes the width (no scaling anymore) like it should. Dragging the screen smaller than 400px will result in scaling though.

Komodo5197:
Yeah, I re-used the setting so it was trying to hit 0.5px wide until you updated it.  The default on a fresh install should be 400px wide.  The way it works now is that it tries to hit the last saved width in pixels, but will never let the player be >70% or less than 20% of the screen, so it will deviate and start scaling to keep in that range.  It might make more sense to set the minimum to a fixed ~300px or so, because that's where the player starts breaking.  And maybe minimum menu width should be ~400px instead of a percentage too?

Also, the cutoff for showing the split view is at least 800px wide, 500px tall.  Does this feel right to you?

Chaphasilor:
I think the split screen width should definitely be wider than the split-mode player width, so that when split-mode kicks in, there's enough space for the whole UI. Double the split-mode player width (or maybe even a tad more) should be a good starting point.  
I'm not sure if the split-mode needs a minimum height, since the non-split mode won't have better scaling either.

But I have some time tomorrow and will play around with the current values!

Chaphasilor:
The current changes look nice and work well. I think we can merge this, unless there's anything else you want to add...

Komodo5197:
I don't think I have anything else.  Don't forget that the #636 changes are in this, so you should probably make sure you're happy with that first.

Chaphasilor:
Okay, I'll merge this then. Thanks a ton!

Since split screen mode would also be beneficial for tablets, do you think there would be any problem with enabling the setting on mobile platforms, but defaulting it to `false` there?

Komodo5197:
Yeah, having it should be fine.  I was putting at least some thought into tablets while making it, and the biggest concern is that the split dragging is even less discoverable than on desktop, which is pretty minimal.  The code doesn't have any OS gating right now, so the only thing limiting it showing up on mobile phones with this build is the size restriction.  500x800px theoretically means you need a 5 inch by 8 inch screen for this to do anything.

Chaphasilor:
@Komodo5197 I just merged the `redesign` branch into the desktop branch, would you mind taking a look to make sure I didn't mess up? There were a ton of changes, and it was a bit hard to keep track of that the desired merge outcome should be in all cases...

Chaphasilor:
Also, I just wanted to ask if you could try to make separate commits for formatting changes in future PRs? For example by first formatting the files you're working on and then committing your actual changes (might be hard to do), or only formatting *after* committing your changes?  
When I tried to merge the changes into the desktop branch, it was hard to see which were the actual changes due to lots of formatting "noise"...

Komodo5197:
I can maybe try something?  I kept forgetting to run the formatter on some files, so I've had 'Format on save' turned on for awhile.  Most of the really messy diffs for this seemed to be more related to the menu refactoring than the formatting.  The excess formatting changes get mostly suppressed if you turn on ignore whitespace.

PR #643, opened 2024-03-20
[Redesign] Jump to Letter Improvments

Komodo5197:
Jump to letter now reliably hits the target letter, even in grid mode.  It also will always match the screen height without using a scroll bar.  Also, we now defer loading images while jumping to prevent loading large amounts of unneeded images for passed items.

### DIFF
--- a/lib/components/MusicScreen/alphabet_item_list.dart
+++ b/lib/components/MusicScreen/alphabet_item_list.dart
@@ -36,14 +36,13 @@ class _AlphabetListState extends State<AlphabetList> {
 
   @override
   Widget build(BuildContext context) {
-    return Container(
-      alignment: Alignment.centerRight,
-      child: SingleChildScrollView(
-        padding: EdgeInsets.only(
-          right: 2,
-          bottom: MediaQuery.paddingOf(context).bottom,
-        ),
-        child: Column(
+    return Positioned(
+      right: 3 + MediaQuery.paddingOf(context).right,
+      top: 0,
+      bottom: MediaQuery.paddingOf(context).bottom,
+      child: LayoutBuilder(builder: (context, constraints) {
+        var listHeight = constraints.maxHeight;
+        return Column(
             mainAxisAlignment: MainAxisAlignment.center,
             children: List.generate(
               alphabet.length,
@@ -55,13 +54,16 @@ class _AlphabetListState extends State<AlphabetList> {
                 child: Container(
                   padding:
                       const EdgeInsets.symmetric(horizontal: 10, vertical: 2),
-                  child: Text(
-                    alphabet[x].toUpperCase(),
+                  height: listHeight / alphabet.length,
+                  child: FittedBox(
+                    child: Text(
+                      alphabet[x].toUpperCase(),
+                    ),
                   ),
                 ),
               ),
-            )),
-      ),
+            ));
+      }),
     );
   }
 
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -7,6 +7,7 @@ import 'package:flutter/material.dart';
 import 'package:get_it/get_it.dart';
 import 'package:hive/hive.dart';
 import 'package:infinite_scroll_pagination/infinite_scroll_pagination.dart';
+import 'package:scroll_to_index/scroll_to_index.dart';
 
 import '../../models/finamp_models.dart';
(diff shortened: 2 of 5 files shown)


Chaphasilor:
Awesome, I hope I can take a look at this soon!

With the fast scroller adapting to the available height, it might be hard to hit a specific letter on small screen, or is it fine?  
There are many users who requested the ability to swipe over the fast scroller, selecting the letter that is currently under their finger. Since loading takes a bit longer, we could initiate the scrolling once the user lets go of the letter?

I'm imagining something where the left side of the screen will show the selected letter in a really big font size, with a smaller text below saying "Release to scroll to this letter", or something similar. But I'm not sure if the current implementation of the fast scroller can support that properly.

Also, not that the covers aren't loaded anymore, it might be worth attempting to use larger page sizes by default, to reduce the scroll time. Last time I tried it there was to real speed-up, but maybe that's different now!

Chaphasilor (commented):
In `music_screen_tab_view.dart`, line 252, we can probably consolidate this with the offline mode sorting logic that ignores additional articles.

Komodo5197:
I tried out a larger page, and it didn't really make much difference.  Although I don't think covers which were never on screen were loaded before, so I don't think this tweak would have been expected to change anything anyway.

Regarding the letters being small, I don't think that really would have mattered that much, but it's even more irrelevant now because I got the sliding from one letter to another working.

Chaphasilor:
Okay, I made some few changes:
- The selected letter will now be briefly shown even on tap, to make sure that the user can see which letter they tapped
- Limited the height to get clear of the floating action button, but that is rather whack atm. There's probably a way to get the height of the FAB (if any is shown), but for now I just doubled the bottom padding. I also reduced the vertical padding of the letters to make them bigger
- Added haptic feedback when the selection changed from one letter to another one while dragging

Also, in my testing pagination didn't work properly in grid mode. Scrolling was reliable if the relevant page had already been fetched, but if the page required loading, the list was just stuck in a loading animation:

Dragging to already loaded letters/pages in that state worked, but there were some issues loading the cover images. Please have another look, and if you need additional testing I can also do that of course!

But overall it worked really well!

Chaphasilor:
Oh and we should probably add a background to the letters, otherwise they have terrible contrast on top of the grid view...

Komodo5197:
Okay, I'm not sure why you would be stuck loading, because the scroller just scrolls down.  I have fixed up some ways that the scroller could get stuck, which would prevent images from ever loading.  I've added a shadow to the scroller to add some contrast.  I don't really like cutting the scroller off above the floating button, it looks odd and gets crazy cramped on small screens.  I've pushed the button further left instead.  I also noticed that it seems the server does not remove a/an/the when sorting songs, but does for albums?  So I've fixed nameForSorting to match that.  It seems like song sorting is also case sensitive, but I haven't changed anything to match that yet.

Chaphasilor:
Okay, thanks for your changes. I'll take a look once I wake up again...

Chaphasilor:
I switched to using the scaffold color as the container background instead of a shadow, since the shadow didn't really help in my testing. The background is invisible in list mode, but increases contrast in grid mode. I hope you like the look 🙈

The covers are now always properly loaded in grid mode, but scrolling still doesn't work. It might have something to do with the pagination trigger. I noticed that the next page is loaded before I reach the end of the current page in list mode, but when I use the fast scroller, it jumps right at the end of the page. Maybe it's "skipping over" the trigger element / threshold for fetching the next page, and that's why it doesn't load.  
Scrolling back up and then down again also loads the next page just fine, which makes that explanation pretty plausible.  
I'm not sure why you're not seeing this behavior, maybe it's the screen size? I tried with different amounts of columns, but that had not effect.

Chaphasilor:
Regarding the sorting behavior, I remembered there is a `sortName` property on `BaseItemDto`. If we add that as a default field (or when fetching offline metadata), we can use that instead of cooking up or own solution.

Komodo5197:
You were completely right about why the grid loading was broken.  I think I just don't have enough albums in a single library for it to be actively paging.  I had to switch the songs tab over to grid view to reproduce.

Regarding the sortName, I think we've got it about as good as it will get.  Previous testing with songs revealed that they have the tracknumber leading their sortname, so my guess is that when we get song sorted by sortname the server is actually sorting by the regular name as a sort of workaround for this.

Chaphasilor:
Everything seems to be working now! Many thanks! <3

PR #646, opened 2024-03-22
[Redesign] Prevent text ovelap in queue

Komodo5197:
Fades out top of queued tracks list to prevent overlap with text.  Also hides jump to current track button when not scrolled away.

Screenshot 2024-03-22 002328

### DIFF
--- a/lib/components/PlayerScreen/queue_list.dart
+++ b/lib/components/PlayerScreen/queue_list.dart
@@ -145,8 +145,9 @@ class _QueueListState extends State<QueueList> {
     ];
 
     widget.scrollController.addListener(() {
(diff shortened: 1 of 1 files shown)


Chaphasilor:
This looks really good, thanks! It's sad that there really seems to be no proper way to detect if a key is currently visible on the screen, so that we have to settle for the offset logic. I tweaked it a bit to show the button a bit earlier, and hopefully adapt to different screen heights as well. The material transparency you enabled also has the side effect of the app bar allowing the user to click the element beneath it, so I tried to increase the padding of the queue source text slightly.

But overall this is working really well, so I'll merge it now!

PR #648, opened 2024-03-23
[Redesign] Player screen spacing tweaks and landscape mode

Komodo5197:
Various tweaks to the player screen layout to further increase album cover size on small screens and add a dedicated landscape layout.

Very small Phone:
Screenshot 2024-03-22 224917
Screenshot 2024-03-22 224901 

<br/><br/><br/>
Large phone:
Screenshot 2024-03-22 225324
Screenshot 2024-03-22 225305

<br/><br/><br/>
Tablet:
Screenshot 2024-03-22 225201
Screenshot 2024-03-22 225137


Chaphasilor:
Could you do a comparison between the old and new design for the different screen sizes (in regular portrait mode). I think the previous layout has a larger cover on large screens, that really extends close to the edge of the screen (as requested by many users). Would be nice if we could preserve that for the new widget tree.  
For the landscape mode, thanks a ton for the implementation! I'd prefer to have a 50/50 split on all screens, with the cover centered on its side.

Komodo5197:
The cover has padding at 7% of screen size on both sides.  I can shrink that to get arbitrarily close to the edges on large screens.  It looks like 1.3 percent is a pretty close match for the existing design.  I'm not sure why you'd prefer a 50/50 split in landscape, but I've done that too.

Chaphasilor:
The 50/50 split really is just an aesthetic preference, nothing more. I hope you don't mind too much.

I added some more conditions for hiding/resizing various elements depending on orientation and sizing constraints, but there's still some additional work needed. I used the wrong queries in a few places, and using the correct contexts is also really important (e.g. for hiding the feature chips if the total height of everything but the cover is too low).  
I want to make sure the app looks and works well both in regular landscape mode, but also in a tight split-screen mode, like this:


I'll try to add these changes tonight or tomorrow

Komodo5197:
I've cleaned up the queries, so it should adapt a lot better now in both orientations.  It does rely on a whole bunch of precalculated stack heights scattered all over to figure out how to scale, so I'll have to clean that up later.

Chaphasilor:
Nice, thanks. I don't think we can really get around hard-coding these different widths and heights, and in that case it might be best to have them as close to the widgets they're actually affecting as possible, as opposed to keeping track of them in one central place. Let me know if you feel differently, and when you want me to take another look at this :)

lib/components/PlayerScreen/player_buttons.dart
```diff
@@ -42,10 +41,11 @@ class PlayerButtons extends StatelessWidget {
...
               _RoundedIconButton(
-                width: targetHeight > 264 ? 62 : 48,
-                height: targetHeight > 264 ? 62 : 48,
-                borderRadius:
-                    BorderRadius.circular(targetHeight > 264 ? 16 : 12),
+                width: controller.shown(PlayerHideable.bigPlayButton) ? 62 : 48,
+                height:
+                    controller.shown(PlayerHideable.bigPlayButton) ? 62 : 48,
+                borderRadius: BorderRadius.circular(
+                    controller.shown(PlayerHideable.bigPlayButton) ? 16 : 12),
```
Chaphasilor, line 48: At this point, maybe we should just create two instances of `_RoundedIconButton`, one called `SmallPlayButton` and the other `BigPlayButton`, that construct a `_RoundedIconButton` with the appropriate size and border radius?

lib/screens/player_screen.dart
```diff
@@ -173,3 +163,107 @@ class _PlayerScreenContent extends StatelessWidget {
...
+      width += element.width;
+    }
+    return Size(width, height);
+  }
+
+  Size getTarget() {
+    return _target ?? _getSize();
+  }
+
+  bool shown(PlayerHideable element) {
```
Chaphasilor, line 266: I'd prefer `isShown()`, `shouldBeShown()` or `canBeShown()` to prevent confusing it with `show()`

lib/screens/player_screen.dart
```diff
@@ -173,3 +163,107 @@ class _PlayerScreenContent extends StatelessWidget {
...
+      if (_getSize().height < targetHeight) {
+        return;
+      }
+      _visible.remove(element);
+    }
+  }
+
+  Size _getSize() {
+    double height = 162;
+    double width = 144;
```
Chaphasilor, line 254: Where are these values coming from / what do they represent? Please add a comment

lib/screens/player_screen.dart
```diff
@@ -173,3 +163,107 @@ class _PlayerScreenContent extends StatelessWidget {
...
   }
 }
+
+enum PlayerHideable {
+  bigPlayButton(14, 14),
+  queueButton(0, 27),
+  progressSlider(0, 14),
+  twoLineTitle(0, 27),
+  codecInfo(0, 20),
+  loopShuffleButtons(96, 0);
```
Chaphasilor, line 173: `width` and `height` is a bit generic I think, it isn't really obvious that these sizes are added up to determine the total used-up space. It's also completely separated from the *actual* size specified in the player widgets, not sure how best to solve that issue.

lib/screens/player_screen.dart
```diff
@@ -173,3 +163,107 @@ class _PlayerScreenContent extends StatelessWidget {
...
+          _visible.remove(PlayerHideable.bigPlayButton);
+        }
+      }
+    }
+
+    for (var element in verticalHideOrder) {
+      if (_getSize().height < targetHeight) {
+        return;
+      }
+      _visible.remove(element);
```
Chaphasilor, line 218: That might result in many recomputes while rendering, but I'm not sure. Somehow using a local copy and then updating `_visible` might still be a good idea though

lib/screens/player_screen.dart
```diff
@@ -173,3 +163,107 @@ class _PlayerScreenContent extends StatelessWidget {
...
+  ];
+
+  List<PlayerHideable> _visible = [];
+  Size? _target;
+
+  void setSizePortrait(Size size) {
+    _visible = List.from(PlayerHideable.values);
+
+    var maxPadding =
+        FinampSettingsHelper.finampSettings.playerScreenCoverMinimumPadding + 8;
```
Chaphasilor, line 197: Why is this 8 added here? Afaik the padding is expressed as a percentage, so adding 8 would make this 800% ?

lib/screens/player_screen.dart
```diff
@@ -173,3 +163,107 @@ class _PlayerScreenContent extends StatelessWidget {
...
+    var targetHeight =
+        size.height + size.width * ((maxPadding / 100.0) * 2 - 1);
```
Chaphasilor, line 199: What's the logic behind this calculation?

lib/screens/player_screen.dart
```diff
@@ -173,3 +163,107 @@ class _PlayerScreenContent extends StatelessWidget {
...
+    var targetWidth = size.width;
+    _target = Size(targetWidth, targetHeight);
+
+    if (_getSize().width >= targetWidth) {
+      _visible.remove(PlayerHideable.bigPlayButton);
+      if (_getSize().width >= targetWidth) {
+        _visible.add(PlayerHideable.bigPlayButton);
```
Chaphasilor, line 206: The side effect of changing `_visible` to affect the result of `_getSize()` isn't super obvious, and it's really hard to tell what is going on. Is there a more explicit way to configure this, that also scales well in the future?

Chaphasilor (changes requested):
Haven't tested the actual app yet (will do that tomorrow), but the controller overall seems like a decent solution, but still has a few rough edges :)

lib/screens/player_screen.dart
Komodo5197: It's stored as a percentage, not a decimal.  So this adds 8%.  Basically this allows up to 8% extra padding per side to show more player controls if needed.  But if when we hit that point we begin hiding controls instead of shrinking the album further.

lib/screens/player_screen.dart
Komodo5197: It's just screen height - (album cover height when using maximum padding).  This is the max possible control height to avoid excessive album cover shrinkage.

lib/screens/player_screen.dart
Komodo5197: The hidden widgets get calculated all at once before we continue building the widget tree, so there shouldn't be any recomputes caused by internal operations.

lib/screens/player_screen.dart
Komodo5197: changed to shouldShow()

lib/components/PlayerScreen/player_buttons.dart
Komodo5197: Maybe, but it doesn't seem worth the effort of changing now without stronger reasoning.

lib/screens/player_screen.dart
Komodo5197: All the widths/heights in the controller were calculated empirically by putting the controls into increasingly small boxes until they started overflowing.  I've added a comment explaining.

lib/screens/player_screen.dart
Komodo5197: I've refactored a couple things and added a bunch of comments to hopefully make the everything a bit clearer.  Tell me if it actually helps.

lib/components/PlayerScreen/player_buttons.dart
Chaphasilor: Yeah it doesn't have to be part of this PR, or maybe I'll just add it myself ^^

lib/screens/player_screen.dart
Chaphasilor: Right, I totally forgot I used percentages right-away xD  
Should be fine then!

Chaphasilor:
Changes do look good, the comments are nice. I'll probably go ahead and change the wording a bit or add some more info, but nothing major.

Taking a look at the actual app, the scaling is prioritizing the cover a bit too much for my taste:

image

The controls are really squished at the bottom, would be nice if we could instead shrink the cover a bit further and give the controls more "room to breathe". Maybe this can be done by setting a minimum height for the control/info area. This is especially noticeable if a lot of space is available:

image

At least for the larger layout, the feature chips could also still be shown.

The layout for tiny landscape screens should get rid of the padding for the song_name_content, not sure why that scales differently than the progress bar, when both have the same padding in regular landscape mode:
image

But the breakpoints themselves seem to work well, and the correct elements are hidden!

Komodo5197:
Okay, so regarding the "squished" controls - are you looking for padding, or more elements?  I can maybe mess with the tradeoffs a bit in the case of unusually square screens to prioritize controls more, especially as we reach the more important elements.  But in general, I have given basically 0 priority to padding when we could be showing more of the album cover or controls, and I don't see much reason to change that.

About the song_name_content padding -  It looks like the progress bar padding is dynamic with controls width while the song_name_content padding is static.  Is that what you were talking about?

Chaphasilor:
Yes, I'm looking for padding, or rather spacing between the controls. The priority should be:
1. Shrink album cover
2. Shrink space between controls
3. Resize controls
4. Hide controls

The reason, just like in many other cases, is simply that it looks better, and has a higher usability than a large cover image.

Do you think that's doable?

And yes, that was what I was talking about with the song_name_content padding. I noticed that I forgot to add a screenshot, so here it is. Compare it to the regular landscape mode:
image

It's already improved with your latest changes, but it could probably be even wider, so that the menu icon is aligned with the start of the progress slider:
image

Komodo5197:
I can just force the cover a bit smaller so that the controls start padding out, so the question is more about how to determine what to do.  You mention shrink album, then shrink padding, but where do you make that tradeoff?  At this point, I'm thinking there's basically 3 cutoffs - max album shrink before removing vertical padding, max album shrink before shrinking controls and hiding queue button, and max album shrink before hiding feature chips/seek bar.  Right now these are basically 0% 8% 8%.  I'm thinking I'll expose these as settings, because I personally really don't want that first value raised.  Also, do you think any of these should be looking at absolute sizes instead of shrink percentages, or maybe using some other form of cutoff?

Chaphasilor:
I'm not sure how best to approach this. An absolute padding would probably be enough for the controls. Maybe you could add some padding enums that can be shown or hidden based on available size.   
Also, it doesn't necessarily have to be either-or. We could scale the cover down by a bit and *also* reduce the padding at the same time whenever there's not enough space. But that probably doesn't work well with the current code.

In the end, as long as the full-size layout looks like before and there's no overflow happening, the actual scaling doesn't matter too much. If you have a certain preference, you can go with that. After all, you're the one putting all of the effort into this!

Komodo5197:
Okay, I ended up just throwing some padding elements into the hideable enum.  There isn't any padding elements they control, they just shrink the album cover so that the Column has more free space to distribute.  There's a low priority 12px tall one and an additional 8px one that doesn't get hidden until after the queue button.  Feel free to mess with the heights.  I also added a couple settings to hide that padding, hide the queue button, or mess with album sizing priority a bit.  Additionally, I think the song_name_content padding should fully match the slider now.

Chaphasilor:
Okay, let's just use the previous layout, without the added padding. The changes just ended up adding more padding to the cover, and squishing the controls even further in the full-size layout:

image
(Left is old / previous commit, right new/current)

So removing the padding will probably result in the best default layout.

You're right that the slider and menu button are now aligned, but in this case I actually wanted to reduce the padding of the song_name_content, not increase the padding of the slider 😅 This should result in more artist and album info being shown in tiny split-screen mode.  
Apologies if that wasn't clear.

image
(Again left is old, right is new)

With all these new settings it is probably time to add a sub-menu for the player screen layout, that includes the cover padding, cover priority, and queue button for now.  
There will soon follow additional settings, like for using a marquee instead of multi-line text, etc.

Komodo5197:
It looks like I forgot to clamp the max cover height.  The new version you have on the right is the minimum control size before the album cover starts getting squished.  It's supposed to be identical to the old version on taller displays.
Screenshot 2024-03-31 133653

If the controls width is <260, which in your example it is, there is already 0 horizontal padding added.  Note how close the favorite button is to the right side - that's the minimum padding baked into an iconButton.  The left looks like more because of the extra space around the album to hit the 50% cover/controls split.  I could mess with the calculations for that - maybe prioritize the controls more at once the width is small, even before the minimum control width?  Should we only be willing to cut the horizontal album padding, or actually start shrinking the cover at some scale?

Chaphasilor:
Ahh yes, that's much better now!   
And the 50/50 split should then probably also have a minimum width, before we just prioritize the controls. The cover can definitely shrink if needed, but getting rid of the padding should already be enough.  
I'll move the subsettings to the top of the page as part of #594, just as an FYI.

I think we can merge this tomorrow. More users have more screen sizes to test with anyway ^^

Komodo5197:
Okay, we now avoid creating any horizontal padding on the album cover whenever the controls are below their default width of ~250px, but we still limit the max control width to 65% of the screen until we hit the minimum size if we need to shrink the album cover.

PR #650, opened 2024-03-24
Enable drag handle for items in tabs list

Ivanf1:
Hi! First of all, thank you for building this awesome app and making it open source!

This PR contains just a tiny change that makes the drag handle icon visible for the items in the Tabs list.
Since there was no drag handle on the items of the tabs list, I was not aware of the possibility to reorder the tabs, and I wanted to implement it. But taking a look at the code, I saw that the feature was already there. So I just uncommented one line to make the handle visible.
I do not know if that line was left commented out because it caused other issues or for other reasons, but from my (not extended) tests I did not notice any issue.

Hope this can help!

### DIFF
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -321,7 +321,8 @@ class _MusicScreenTabViewState extends State<MusicScreenTabView>
               settings.tabSortOrder[widget.tabContentType],
               settings.onlyShowFullyDownloaded,
               widget.view?.id,
-              settings.isOffline);
+              settings.isOffline,
+              settings.tabOrder,);
           if (refreshHash == null) {
             refreshHash = newRefreshHash;
           } else if (refreshHash != newRefreshHash) {
--- a/lib/components/TabsSettingsScreen/hide_tab_toggle.dart
+++ b/lib/components/TabsSettingsScreen/hide_tab_toggle.dart
@@ -19,7 +19,7 @@ class HideTabToggle extends StatelessWidget {
       builder: (_, box, __) {
         return SwitchListTile.adaptive(
           title: Text(tabContentType.toLocalisedString(context)),
-          // secondary: const Icon(Icons.drag_handle),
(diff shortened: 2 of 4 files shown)


Chaphasilor:
Hey, thanks for the PR!  
Sorry for getting back to you about this only now, I was busy with exams :)

I think there are (or were) some issues with reordering the tabs, where the new order isn't actually being used, maybe that's why it was uncommented...  
I'll take a look at your change tomorrow, and if needed will try to fix any remaining reordering issues!

Chaphasilor:
Okay, I improved the drag handle so that dragging the handle will reorder the list items right-away, instead of still requiring a long press.  
And I also fixed the reordering of tabs, although it still isn't perfect. 

I guess we can merge this then?

Ivanf1:
Hi! I'm still experiencing an issue. An exception occurs when the `Songs` tab changes position (either directly or as a consequence of other tabs moving). The exception occurs [here](https://github.com/jmshrv/finamp/blob/2804f198b1db94690585ac9bd7d5ea8f029c3bbc/lib/models/finamp_models.dart#L724). May it be that the exception is triggered only with the `Songs` tab because it's the only one in the list that has a `DownloadItemType` different from the others?

I also went back to a7ef411 and 1508c0c and the issue is still present.

Chaphasilor:
Yeah I noticed that too. I tried to mitigate it, but I couldn't get the tab content to properly refresh for some reason. The old behavior didn't have the exception for me, but the tab content still had the from item types in the list. But maybe that's also a problem in the new version and it's just dependent on where the songs tab is, like you suggested.  
Either way, it's probably not too big of an issue for now, a simple swipe-to-refresh will get rid of the error.

But if you want to try to fix this, I can also wait before merging the PR :)

Ivanf1:
Hi! Yes, since it does not really break anything I think we can merge this. I will see if I can figure out what exactly is the problem and if I can fix this in the future.

Thank you for everything! :)

Chaphasilor:
Alright, merged! Good luck! :D

Issue #654, opened 2024-03-28
[Redesign][Bug] artist top songs sorting borked


Tiefseetauchner:
When clicking a song in the top songs list of an artist that has the same listening amount as another song the wrong song sometimes plays

Mainly reportkng this so I don't forget it, I'll use a more defined algorithm for sortkng the queue and the list here

Chaphasilor:
Awesome. There we also requests to change the order of albums from alphabetic to release date, maybe we can tackle that in the same PR :)

Tiefseetauchner:
So I understand it correctly, when pressing play on the artist the albums are sorted by alphabet?

The display seems correct, at least on whatever build I'm running

![Screenshot_20240328-114848.png](https://github.com/jmshrv/finamp/assets/48085877/f721106e-0d84-49a3-a3c1-245e139295d4)

Chaphasilor:
The related issue is #609 ^^

Tiefseetauchner:
I commented there, I think it requires more discussion on that thread

PR #655, opened 2024-03-28
#654 - fix inconsistent sorting

Tiefseetauchner:
clamping down on sorting of songs so that the display is equivalent to the queue


Tiefseetauchner:
Added Disabling of the top songs to the settings

lib/models/finamp_models.dart
```diff
@@ -215,6 +217,9 @@ class FinampSettings {
   bool hideSongArtistsIfSameAsAlbumArtists =
       _hideSongArtistsIfSameAsAlbumArtists;
 
+  @HiveField(47, defaultValue: _showArtistsTopSongs)
+  bool showArtistsTopSongs = _showArtistsTopSongs;
+
```
Chaphasilor, line 222: Please move this to the end, so it's easier to figure out what the next ID should be

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -112,11 +115,12 @@ class _ArtistScreenContentState extends State<ArtistScreenContent> {
...
-            sortedsongs.sort(
-              (a, b) =>
-                  b.userData?.playCount.compareTo(a.userData?.playCount ?? 0) ??
-                  0,
-            );
+            sortedsongs.sort((a, b) {
+              int compareCount = (b.userData?.playCount ?? 0)
+                  .compareTo(a.userData?.playCount ?? 0);
+              if (compareCount != 0) return compareCount;
+              return (a.name ?? "").compareTo(b.name ?? "");
```
Chaphasilor, line 122: There should be a `nameForSorting` attribute on the `BaseItemDto` to emulate Jellyfin's sorting behavior (e.g. excluding articles for albums, ignoring case, etc.), please compare that instead!

lib/l10n/app_de.arb
```diff
@@ -301,6 +301,10 @@
   "@showCoverAsPlayerBackgroundSubtitle": {},
   "hideSongArtistsIfSameAsAlbumArtists": "Verstecke Titel-Künstler falls derselbe wie Album-Künstler",
   "@hideSongArtistsIfSameAsAlbumArtists": {},
+  "showArtistsTopSongs": "Meistgehörte Lieder bei Künstler anzeigen",
```
Chaphasilor, line 304: Maybe "Meistgehörte Lieder von Künstlern anzeigen" or "Meistgehörte Lieder auf Künstler-Seite anzeigen"?

lib/l10n/app_de.arb
```diff
@@ -301,6 +301,10 @@
...
+  "@showArtistsTopSongs": {},
+  "showArtistsTopSongsSubtitle": "Ob oder nicht die meistgehörte Lieder bei Künstlern angezeigt werden sollen.",
```
Chaphasilor, line 306: "Ob die meistgehörten Lieder eines Künstlers angezeigt werden sollen, oder nicht."  
or maybe just
"Sollen die meistgehörten Lieder eines Künstlers (auf der Künstler-Seite) angezeigt werden?"

Chaphasilor (changes requested):
Nice change overall, but I haven't looked at the actual app yet, just the code. I'll give it a try tomorrow.  
Just a few things that should be changed, but I could take care of those if you don't have the time...

Chaphasilor:
While the correct song is selected, it seems like the content of the queue is not identical to the top songs list:
image
image

The queue should imo either be identical to the top song list, or should start a song mix, or, if you want to keep all artist songs part of the queue, the selected song should be at first position in the queue. What do you think?

Tiefseetauchner:
I fully agree, I think that was a point of discussion, I'm fully for using the same list

lib/l10n/app_de.arb
Tiefseetauchner: I would prefer the first one without the "oder nicht", I thought maybe that's just a Germany-exclusive colloquialism that I didn't know about, and I followed the other translations there. I'll change all occurrences as I hate the phrasing

Chaphasilor:
I'm still getting inconsistent sorting if multiple tracks have not play count (e.g. all 0 plays). In that case, the selected song is always the last in queue, and the order of the other songs is inconsistent with the top songs.  

For artists where all songs have different play counts, it works well now 👍🏻

Tiefseetauchner:
Yeah I was fighting with futures when I got distracted and tried to setup git extensions for the past hour, if you could have a look at how to solve that I'd greatly appreciate it, currently stuck on how to load all songs asynchronously

Chaphasilor:
Alright. What I did was simply fetch *all* songs sorted by play count (didn't take much longer than just the first 5 on my phone, I think the page size of 100 is now used?), and only use the first (up to) 5 for the song songs section. We could now also have a "See All" button, or something similar.  

Let me know if this causes trouble on your phone with artists that have many songs! The only real bottleneck could be a slow connection or slow JSON parsing...

Chaphasilor:
@Tiefseetauchner if everything works for you I can merge this :)

linux/flutter/generated_plugin_registrant.cc
Tiefseetauchner, line 1: Does this belong in here?

linux/flutter/generated_plugin_registrant.h
Tiefseetauchner, line 1: Does this belong in here?

linux/flutter/generated_plugins.cmake
Tiefseetauchner, line 1: Does this belong in here?

macos/Flutter/ephemeral/Flutter-Generated.xcconfig
Tiefseetauchner, line 1: Does this belong in here?

macos/Flutter/ephemeral/flutter_export_environment.sh
Tiefseetauchner, line 1: Does this belong in here?

Tiefseetauchner (commented):
Seems like build tools from the desktop branch got commited

Tiefseetauchner:
Sorry, I was distracted, seems to work well, even faster than before tho that might be placebo. You commited some desktop files, was that on purpose?

Chaphasilor (approved):

PR #656, opened 2024-03-29
[Redesign] Transcoded downloads tweaks

Komodo5197:
Switch AAC transcodes to use .aac container.  It seems to already work fine with jellyfin 10.8, and I think it works with iOS?  And we know that AAC in .m4a actually doesn't seem to work.  It also seems to scrub properly on android, unlike m4a.  Also, now uses correct file extension on local paths.

### DIFF
--- a/lib/models/finamp_models.dart
+++ b/lib/models/finamp_models.dart
@@ -1560,7 +1560,7 @@ enum DownloadLocationType {
 @HiveType(typeId: 65)
 enum FinampTranscodingCodec {
   @HiveField(0)
-  aac("m4a", true, 1.2),
+  aac("aac", true, 1.2),
   @HiveField(1)
   mp3("mp3", true, 1.0),
   @HiveField(2)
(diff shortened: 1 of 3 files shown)


Chaphasilor:
I'll try it out tomorrow, and maybe @jmshrv or @StayBlue could confirm that this works on iOS?

jmshrv:
This doesn't seem to work on iOS, the `.m4a` container is required to make Apple happy. According to `file`, ffmpeg creates a `MPEG ADTS` when outputting `.aac`.

Komodo5197:
Really?  Well that's unfortunate.  I've confirmed that the m4a aac's don't seem to work on iOS, even with jellyfin 10.9, so I guess iOS is still stuck with un-seekable MP3s.  This still seems to fix AAC on android, so it might as well still go in.

Chaphasilor:
Okay, so to recap:  
- This fixes seeking for downloaded tracks which were transcoded to AAC on Android, for JF 10.9
- This has no effect on iOS, transcoding to AAC doesn't work with .aac nor with .m4a?

Komodo5197:
It actually makes AAC work on Android in jellyfin 10.8 as well, although this doesn't really matter much because of OPUS.  For iOS, I've tried transcoding to AAC against jellyfin 10.9 on my phone with 0.9.3 and it doesn't work, so yeah, I guess this has no impact there.

Chaphasilor:
Okay, merged! If anything comes up we can deal with it then ^^  
Thanks :)

Issue #657, opened 2024-03-29
[Redesign][Feature Request] Toggle for haptic feedback/vibration on button tap


Ivanf1:
Hello!

In the redesign version of the app, an haptic feedback/vibration is produced on some buttons tap. I, personally, would prefer to not have any feedback of this kind when a button is tapped.

I think a toggle could be added in the Interactions section of the Settings page. The default would be to leave enabled haptic feedback/vibration.

If this is something that you think may be worth to add, I can start to work on it.

Chaphasilor:
Hey, yeah that sounds like a very reasonable request! Should be not a lot of effort to add this really, so I can add this in the coming days.

But if you want to contribute this yourself, feel free to do so! The [contribution guidelines](https://github.com/jmshrv/finamp/tree/main/CONTRIBUTING.md) should help you get started, and then it's just copying a setting and renaming a few variables, adding a could of new strings to the translations, and using the settings helper to check if the vibration should be enabled.  
Maybe the vibration stuff could also be wrapped in a class to only have one global place where the settings are checked...

PR #659, opened 2024-03-29
Add toggle to disable vibration #657

Ivanf1:
Hello!
This PR addresses issue #657.

As suggested by @Chaphasilor in [this comment](https://github.com/jmshrv/finamp/issues/657#issuecomment-2027025098) (if I understood correctly), I wrapped the vibration setting check in a class.

I will be waiting for feedback. Hope this can help!

### DIFF
--- a/lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
+++ b/lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
@@ -5,6 +5,7 @@ import 'package:finamp/components/Buttons/cta_medium.dart';
 import 'package:finamp/components/global_snackbar.dart';
 import 'package:finamp/models/finamp_models.dart';
 import 'package:finamp/services/queue_service.dart';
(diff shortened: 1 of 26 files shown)


Chaphasilor:
Thanks for the PR!

The changes look very good overall, and are exactly what I had in mind!  
Would you mind renaming the class to `FeedbackHelper` (so we could theoretically support audio (etc.) feedback on platforms that don't support vibration in the future)?  

And another thing, please call the setting "*enable* vibration" and make it true by default. So that disabling the setting disables vibration. That makes a bit more sense compared to enabling a setting to disable something, just a minor usability improvement ^^

Ivanf1:
Thanks for the feedback!

I implemented the requested changes. Let me know if there is more I can do.

Chaphasilor:
Thanks, that looks perfect! I'll give it a try in a few minutes!

Chaphasilor:
Merged, since everything is working just fine! Thank you!

PR #665, opened 2024-04-02
[Redesign] Blurred background performance improvment (maybe)

Komodo5197:
Attempts to improve performance when using blurred background covers by caching the blurred image to a texture and sharing between widgets.  This seems to be especially helpful if impeller is enabled.  Only tested in emulator, so it's possible that the performance impact on actual hardware is different, but hopefully this is an improvement.

Also, I've updated the .gitignore based on newer flutter versions to properly exclude desktop artifacts and fixed the player_button_more theme transitions.


Chaphasilor:
Thanks! I didn't feel any apparent difference on Android, but that was in debug mode, but I also didn't really have performance issues to begin with. We'll see how release performance is, but it shouldn't be worse :)

And good catch with the button theme!

jmshrv:
Based on feedback on Discord, it looks like this worked :D

PR #666, opened 2024-04-03
[Redesign] Show unknown artist on player screen

Komodo5197:
When a song has no artist/album artist, show and 'Unknown Artist' chip instead of just leaving a hole in the layout.
Screenshot 2024-04-02 193440
Screenshot 2024-04-02 193357


Chaphasilor:
Looks good, thanks!

PR #670, opened 2024-04-04
[Redesign] Fix blurred cover background when switching orientation.

Komodo5197:
Fixes blurred background cache not immediately updating image size when screen size changes to to orientation shifts or other reasons.  I thought I tested this as working, but apparently not.

### DIFF
--- a/lib/screens/blurred_player_screen_background.dart
+++ b/lib/screens/blurred_player_screen_background.dart
@@ -62,20 +62,31 @@ class CachePaint extends SingleChildRenderObjectWidget {
   final String imageKey;
 
   @override
-  RenderObject createRenderObject(BuildContext context) {
-    return RenderCachePaint(MediaQuery.sizeOf(context), imageKey);
+  void updateRenderObject(BuildContext context, RenderCachePaint renderObject) {
+    renderObject.screenSize = MediaQuery.sizeOf(context);
+  }
+
+  @override
+  RenderCachePaint createRenderObject(BuildContext context) {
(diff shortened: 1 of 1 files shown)


Chaphasilor:
Could you add one or two short videos showing the issue? Is it a critical problem?

We justed pushed an update to the stores and are waiting for approval, but if this is critical we can follow up with another one right-away, or even pull the update and replace it with a new version...

Komodo5197:
It’s not critical, it just uses the wrong image size and looks a bit awkward if you switch orientations while already on the player screen.

EFB4BE72-0C9C-482D-8D3E-D4D0F3D547E0

Komodo5197:
Testflight pushed the update an hour or two ago, that's where I caught this.  The caching seems to actually improve the queue performance on my phone by a good bit, so that's nice.

Chaphasilor:
Awesome, seems to be working well! Thanks for the fix ^^

Chaphasilor:
I was wondering if there were any other cases where the hash / cach key should change, but I think I was too tired last night.  
Unlocking my phone today solved the mystery right-away: theme changes (e.g. OS-level dark/light mode) need to update the background to prevent contrast issues.

I'll push a fix for that in a bit :)

Komodo5197:
@Chaphasilor I've thrown a fix for this into my latest PR becuase it was pretty straightforward.

Chaphasilor:
That's good, because I totally forgot. My brain is swiss cheese at this point xD

PR #673, opened 2024-04-05
[Redesign] Fix non-transcoded downloads + tweaks

Komodo5197:
Fixes the .song extension applied to all downloaded original songs, which breaks iOS playback.  Also includes a couple other random tweaks I had in the branch:

- A basic fix to the fast scroller accent issue by just ignoring any items with characters beyond z.
- Queue item padding tweak.
- Apply ''Prioritize cover" player setting to landscape mode as well as portrait.
- Just make airplay button white because it can't change theme when switching songs.

Tell me if you want a cleaner version.

### DIFF
--- a/lib/components/AlbumScreen/song_menu.dart
+++ b/lib/components/AlbumScreen/song_menu.dart
@@ -202,7 +202,7 @@ class _SongMenuState extends State<SongMenu> {
         return DraggableScrollableSheet(
           snap: true,
           initialChildSize: size,
-          minChildSize: size * 0.9,
+          minChildSize: size * 0.75,
           expand: false,
           builder: (context, scrollController) {
             return Stack(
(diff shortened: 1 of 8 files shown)


Chaphasilor:
Could you exclude the fast scroller tweaks? I'm working on a "proper" fix right now :)

And about the AirPlay button, what about putting it into some kind of builder that updates when the theme updates? Wouldn't that work?

Komodo5197:
I can force it to animate by setting a key from the iconColor, I think.  But the performance implications of disposing and rebuilding a native widget for every frame of the theme animation seems questionable.  Maybe I can make it just jump at the transition start along with the album art?

Chaphasilor:
Yeah, it doesn't have to be animated. Maybe you can use a stream from the current theme provider, which would trigger the theme animation.  
Let me know if you want to try that out first, or if I should merge the PR already!

Komodo5197:
I put the icon into an animated switcher, so we only build two icons and just fade between them.  The docs say that opacity stuff works fine on native views, so hopefully this doesn't cause problems.

Chaphasilor:
Thanks a ton. I can't test the iOS stuff either, let's just hope it works...

PR #680, opened 2024-04-06
[Redesign] Queue restore improvments when stopping playback.

Komodo5197:
Save empty queue as latest when stopping playback, but still prevent empty queues from being archived with timestamps.  This prevents the stopped queue from being automatically reloaded if the app is restarted, and also allows the stopped queue to be immediately restored from the restore queue screen if it was closed accidentally.  Also, disables stop gesture in line with existing setting.


Chaphasilor:
Thanks! I noticed the restore after stop too, but was short on time and it slipped my mind. I'll take a look tomorrow.

Chaphasilor:
Working nicely. Thanks!

PR #685, opened 2024-04-07
[Redesign] Added setting to hide artist chip images

Komodo5197:


lib/l10n/app_en.arb
```diff
@@ -1306,5 +1306,6 @@
   "lockDownload": "Keep on Device",
   "@lockDownload": {
     "description": "Text shown on button to require an item which is already downloaded by a parent collection, preventing it from being deleted if the parent is removed."
-  }
+  },
+  "showArtistChipImage": "Show image on artist chip"
```
Chaphasilor, line 1310: Would prefer a less technical description, like "Show artist images with artist names", what do you think?

lib/screens/layout_settings_screen.dart
```diff
@@ -50,3 +54,30 @@ class LayoutSettingsScreen extends StatelessWidget {
...
+                  FinampSettings finampSettingsTemp =
+                      box.get("FinampSettings")!;
+                  finampSettingsTemp.showArtistChipImage = value;
+                  box.put("FinampSettings", finampSettingsTemp);
+                },
+        );
+      },
+    );
+  }
+}
```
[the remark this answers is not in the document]
Komodo5197: Sure, I can move it.  I personally prefer to keep all these settings widgets in the screen file though.  It makes them easier to access, and it's not like they're going to get re-used anywhere.

lib/l10n/app_en.arb
Komodo5197: Sure.  Maybe it should mention that it's specifically the chips and not in the main list, but I can't think of a great way to phrase that.  'with artist names' doesn't feel as specific.

lib/screens/layout_settings_screen.dart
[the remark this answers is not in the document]
Chaphasilor: I don't have a strong preference either way, I just want to keep things consistent.  

If you feel like refactoring all settings to have  them in the same file instead, feel free to submit a PR ^^

lib/l10n/app_en.arb
Chaphasilor: Yeah, I also struggled with that. Maybe adding a subtitle is a good idea then?

Komodo5197: Added subtitle: "This affects small artist image previews, such as on the player screen."

lib/screens/layout_settings_screen.dart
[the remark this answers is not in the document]
Komodo5197: It's already a bit inconsistent because I used this style for all the settings I've already added/moved.  regardless, I've put this into a separate file for now.

Chaphasilor:
Thanks!

PR #687, opened 2024-04-08
[Desktop Beta] Merge cleanup and imageProvider changes

Komodo5197:
The player split view PR had partially-complete set of tweaks trying to speed up and improve image loading with the goal of preventing the song menu from changing theme/background after loaded when opened for tracks not matching the current player theme.  This PR attempts to fully flesh out that code and generally improve image loading.  The song menu should now load without either the background or theme changing.  The blurhash placeholder code has been improved, and is now faster with less blinking.  Both file and network images are more likely to be pulled from the ImageCache now, so placeholders are seen less.


Chaphasilor:
Thanks!  
Is there a specific reason why you introduced the blurhash placeholders only in the desktop branch? Are there performance issues on mobile, or did it just fit more natural there? Is there anything we should watch out for before merging this into the redesign branch?

Could you also add the blurhash placeholder for the artist chips? That would get rid of the layout shifts. But only if it isn't too much trouble.

As for the PR, here's what I noticed:
- The menu is always has a fixed height on desktop. I think we could limit it to the needed height for showing all options instead, like we do on mobile? Or was that a deliberate decision?  
  image
  
- When loading the song menu in the songs list (or somewhere else that's not the player screen), the tiny blurhash is shown in the center of the menu for some reason:
  image  
  It also seems like the default menu background is the one used on the player screen, instead of a generic white/dark one

- The subtitle of the new cache setting says "for albums or some songs". What do you mean by that? For which songs doesn't it work? Or do you mean that only some songs will have their own image instead of using the album's image?  
  Also, if images are cached, images for playlists and artists should probably be cached as well.  
  And I take it the "sync" action would perform a full sync and refresh all album images that were updated on the server, yes?

Chaphasilor:
Another thing I was thinking about: I love that Finamp always uses the up-to-date data from the server when in online mode. Any newly added music or changed images are immediately available, unlike in other apps like Symfonium where you always have to wait for a sync.  
With the downloaded covers, that is not the case anymore, especially with quick syncs enabled (not a new issue, it also applies to regular downloads). Do you think it would be possible to (at least have an option to) use the downloaded images as a placeholder in online mode that is only used until the new image has been loaded?  
I'm not sure if that would cause any issues if the cover actually changes (the blurred background would need to be reloaded in that case). At least the blurhash data should be correct, because afaik we don't use the downloaded metadata in online mode, just the downloaded files.  

Or maybe I'm just not up-to-date with the inner workings of Finamp anymore, and it's already doing all of this? 😅

Komodo5197:
Okay, the menu height and tiny blurhashes should be fixed now.  I'm not sure what you mean about the default background, isn't is just low opactity white/black?  There isn't a particular reason the blurhashes ended up in desktop.  I was messing with the song menutrying to get it to play a bit nicer with a mouse, and trying to reduce blinking there lead me to the blurhashes.

Regarding downloaded images, they should never be out of sync.  Because we always use fresh metadata, the updated blurHash/imageId for the item don't point to the old image anymore, so the image isn't downloaded and we just fetch it from the server like normal.  Then later when syncing we delete the old image and download the new one.

Komodo5197:
Regarding artists, they should have show the blurhash already because they just use AlbumImage.  The problem is that we don't have that blurhash, or even know if the image exists, until we fetch the actual artist BaseItemDto.  So the layout lag is the wait for that request to go through.  I've added a cache for artist BaseItemDtos, so at least this lag happens only once.  You could probably add some sort of precaching logic somewhere if you wanted to make this even better?

Komodo5197:
Regarding the album image cache, any song with a matching blurHash will use the same cover.  Some songs have different images, but also some songs use basically the same image but are just different enough for the blurhashes to not be an exact match.  The fact that song and album images generally match is most of the reason I wanted specifically album covers cached.  I can add playlists/artists into the setting if you think that is useful, or maybe throw in another toggle.  For my personal usecase, I don't really care if those cache.  And yes, syncing works as expected to replace updated images.

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -415,6 +414,11 @@ class _MusicScreenTabViewState extends State<MusicScreenTabView>
   }
 }
 
+class MusicRefreshCallback {
+  void call() => callback?.call();
+  void Function()? callback;
+}
+
```
Chaphasilor, line 421: Is there a particular reason why you used a class instead of just a regular function?

lib/screens/music_screen.dart
```diff
@@ -260,6 +265,13 @@ class _MusicScreenState extends ConsumerState<MusicScreen>
                       )
                     ]
                   : [
+                      if (!Platform.isIOS && !Platform.isAndroid)
```
Chaphasilor, line 268: The platform filter is just there because there isn't enough space, correct?  
If so, is there a way to use this refresh behavior also for the swiping down to refresh? Currently the only way to hard refresh is to toggle offline mode...

lib/components/MusicScreen/music_screen_tab_view.dart
Komodo5197: It’s because the function is basically being passed backwards.  The callback is called by the parent, but needs to be set by the child.  I think I can do the same thing by passing down the map, though.

lib/screens/music_screen.dart
Komodo5197: The swiping down, any settings/filter changes, and this button do the same refresh.  I messed with it to harden the refresh when I unified the online and offline paging during the downloads rework, so the pull refresh should already be just as good as the offline toggle.

Chaphasilor: Okay, maybe I'm just still used to the old behavior and didn't "trust" the refresh enough. Now that you mention it, every time I did use the refresh, it actually did load correctly, so I guess it works fine :)

Chaphasilor:
> Regarding downloaded images, they should never be out of sync

Right, because we use the blurhash as the key, not the ID of the BaseItemDto. Nice.

> I can add playlists/artists into the setting if you think that is useful

Yeah I think so. It should just be "cache images" and it caches (almost) everything, maybe except for the song-specific images like you mentioned

The issues all seem to be fixed now. I thought the menu background was the same blue as the menu for the currently playing track during testing, but I'm not sure. Either way, not the blurhash is correctly used and everything looks nice.  

I did notice that the background updates to the actual image before the cover at the top of the menu does, is there a way to use the same image for both instead of making two requests? And there also isn't a transition right now, so it just "blinks" from blurhash to actual image and updates the accent color as a result. You might wanna use the same transition duration for this as the player uses when switching between tracks (maybe a bit faster if you think that's a good idea), at least for the accent color (I'm not sure if it's a good idea to transition the blurred background as well).

Oh and the way how you integrated smtc? Chef's kiss 😚👌🏻  
Ideally the package should already do that, but until that happens that's a very elegant solution.  
I'm not sure why the `audio_service_platform_interface` package has no version though, is it like a fall-through case where the next-specified version is used (so it's always in sync with `audio_service`)?

Komodo5197:
I'd like to break the image caching apart at least somewhat.  Maybe album/playlist and artist/genre?  Or breaking it up by library would be nice, but I can't think of a way to easily implement the UI.

An earlier build used the wrong widget and could show the blurhash decoding placeholder, which was blue.  You were probably seeing that.

The goal of this is to load the theme/background very quickly, hopefully before a noticeable portion of the menu is visible.  The fact that you saw the background/theme transition was a bug.  I believe I've fixed it, so they should now show up immediately.

The package has no version because I forgot to add one.  I'll fix that.

lib/components/AlbumScreen/song_menu.dart
```diff
@@ -39,8 +40,8 @@ import 'download_dialog.dart';
 Future<void> showModalSongMenu({
   required BuildContext context,
   required BaseItemDto item,
-  ColorScheme? playerScreenTheme,
   bool showPlaybackControls = false,
+  bool usePlayerTheme = false,
```
Chaphasilor, line 44: Can we rename this to what it actually does (not change theme if the current track changes)? Maybe something like `staticTheme`.  
And is there any scenario where we *would* want the menu theme to automatically update if the current track changes?

Chaphasilor:
If you think separate caches would be nicer, we can split it up. If we split, I think a full split by category would be better than arbitrary grouping (there's no relation between artists and genres, they just happen to use the same widgets). So having four different categories for caching is absolutely fine, and I'll handle it somehow in the new design.  
Caching everything by library would be even nicer (because we cache all images for the library, which is more consistent), I'm just not sure if there's significant gains to be had there (e.g. if the libraries are very unbalanced in size). Did you want to split both by category *and* library, so what is the reason you want to split in the first place?  

Regarding UI for the caching by library, you could change the setting to cache all images for the currently active library (or sync/delete the cache). As long as we can pass a library/collection ID to some function to do that, we can easily build a more flexible UI down the road... 

The cover loading seems to be much faster now, and the correct background and cover are used immediately. Amazing.

Depending on what you still want to change, we could get this merged tomorrow!

lib/components/AlbumScreen/song_menu.dart
Komodo5197: There aren't any situations we want the theme to change, because the menu content never does.  The theme should always be static, but sometimes the caller supplies it and sometimes we want to use the item that's playing when the menu gets opened, which is what this flag flips between.

Komodo5197:
I just wanted to split the image cache up a bit to keep the minimum size down.  I've changed it so all album/artist/genre/playlist covers from the currently active library get cached.  The primary point of this was to make the song menu background/theme show immediately, so if that's working for you I think this is complete.

Chaphasilor:
Okay, everything seems to be working, both on desktop and mobile.  
I noticed that even if fully "cached" (downloaded), the blurhashes are (very) briefly shown in online mode, but never in offline mode. I assumed there would be no need to show the blurhashes at all if the images are downloaded, but maybe I'm missing something (like so often)...

I'll merge this now because nothing is broken and it's much better than before, and if there follow-ups we can handle that in a separate PR.

Either way, thanks for another massive improvement to the Finamp experience!

Komodo5197:
Good to hear it works well on mobile too.  Downloaded images are handled identically on online and offline modes, so the difference is a bit strange.  My guess is that there's just more images to load and render in online mode because only some things were downloaded, and flutter takes a few frames to catch up with the larger workload.

PR #692, opened 2024-04-17
Add properly sized icon

jclsn:
The current logo file has a too big canvas to be used as an icon. This icon could be used by package maintainers to integrate it into the desktop environment.

See https://aur.archlinux.org/packages/finamp-git


Chaphasilor:
Hi! Thanks for the PR, but I believe the logo you're looking for is either [this one](https://github.com/jmshrv/finamp/blob/main/assets/icon/icon_foreground.png) or [that one](https://github.com/jmshrv/finamp/blob/main/assets/icon/icon_combined.png).

These should be large enough, and most importantly, have a high resolution. There's also [an SVG](https://github.com/jmshrv/finamp/blob/main/assets/icon/icon_foreground.svg) that you could use to generate a specific resolution, if needed!

jclsn:
Thank you for pointing me to the image files. Unfortunately all of them have too much whitespace around them or are not quadratic. Therefore they are not usable as an icon without cropping. 

[This one](https://github.com/jmshrv/finamp/blob/main/assets/icon/icon_combined.png) is quadratic, but should have at least rounded corners. The SVG would be nice if it were quadratic. Should I crop them and make a new PR?

Maxr1998:
I think it should be possible to use the SVG and fit it into a new square canvas with a few Inkscape commands. If you like, I could pick this up from you, and generate all the sizes that are required for Linux packages.

Chaphasilor:
I would've assumed that the DE would apply some sort of mask on top of the icon to get the rounded corners. That's how it works on Android too. "Hard-coding" rounded corners into the icon doesn't feel right, but it that's the regular way to do this for Arch feel free to do it. You can use the PNG or SVG, just make sure that the resulting icon has a high-enough resolution :)

I could also make quadratic versions for the SVGs if that helps?

Maxr1998:
It *appears* that Linux icons don't actually need a background, comparing with a few other apps I've looked at. I've experimented with it in my fork, you can view the results here: https://github.com/jmshrv/finamp/compare/desktop-beta...Maxr1998:finamp:desktop-icons.

If that works for you @jclsn, I can open a PR for these.

jclsn:
@Chaphasilor Nope,  you would get the whole square on e.g. KDE Plasma. Icons are displayed just as they are.

@Maxr1998 Those icons are perfect. All correct sizes are there. Would be great if you could make a PR! I will close this one now :)

Maxr1998:
> Those icons are perfect. All correct sizes are there. Would be great if you could make a PR! I will close this one now :)

Thank you! And done ✅

PR #693, opened 2024-04-17
✨Lyrics Support

Chaphasilor:
This adds support for lyrics, which the Jellyfin server supports starting with Jellyfin 10.9.  
Since 10.9 is getting close to a release now (planned for end of April), this PR aims to get lyrics support ready for the redesign beta before that release.

The fetching of lyrics, rendering, UI, and things like auto-scrolling are already mostly working. What's still missing is:

### TODO

- [x] Pre-fetch lyrics for previous and upcoming tracks
   - Seems like this is a bit more complicated than pre-fetching album covers, since that used helper classes provided by Flutter made explicitly for image fetching
- [x] Look into using `JellyfinApiHelper.getPlaybackInfo()` in the metadata provider, that might require everything we need for now and is better suited for the task
- [x] Properly highlight lyrics starting at 0:00
- [x] Re-fetch lyrics when toggling offline mode
- [x] Include lyrics with song downloads
- [x] Handle errors when fetching lyrics (e.g. connection timeouts)
- [x] Add a gesture (probably swiping to the left on the controls section) for opening the lyrics screen
- [x] Use a different appbar button for closing the lyrics. The current arrow doesn't fit
- [x] Fix scrolling being locked until the second line is highlighted for synced lyrics
- [x] Use proper localization tokens
- [x] Disable auto-scroll (possibly only temporarily) when the user manually scrolls
- [ ] Add some configuration options, like hiding timestamps or centering lyrics
- [x] Allow tapping on a lyric line to jump to that position in the track
- [x] ~~Allow the user to disable auto-scroll~~
- [x] ~~Use the player's sizing controller to reduce the height of the control section~~
- [ ] Optional: Make the lyrics view part of the player screen, allowing the user to toggle between lyrics view and regular player view. That would have the benefit of allowing users to minimize the lyrics screen immediately without first going back to the player screen, and also re-open to the lyrics view directly


Komodo5197:
Regarding precaching, I think all you need to do is swap out all the ref.read calls on the metadataProvider future in current_track_metadata_provider to calls to ref.watch.  The provider initialized in the prefetch is seeing that no-one is watching and disposing itself before we get to that track.  If you use watch, it should stay alive and the value should be available instantly when the track switch occurs.

Chaphasilor:
Hmm, it seems like the types are not compatible for some reason. It won't let me `watch()` the future, nor the request.  
It appears like ref.watch can't be used with an auto-disposable provider since it requires an `AlwaysAliveProviderListenable`.  
Getting rid of the auto disposing works just fine, but I'm not sure of the implications of that. Seems like it might cause some unnecessary requests (not a huge deal) and also that it won't retry requests after an error when re-opening the player screen.

I've pushed the changes in c1c78ca, maybe you can take a quick look?

Chaphasilor:
I also forgot to add one TODO yesterday, we should probably also download the lyric data along with the tracks and images for offline lyrics. What would be your approach to do this? You're already downloading track metadata somewhere after all (to show the grayed-out tracks), so maybe we can sneak the `LyricDto` in there somewhere?  

Or instead of downloading the regular `BaseItemDto`, we use the new `MetadataProvider` instead (possibly after renaming it and moving it into `finamp_models`), since that one includes both the `BaseItemDto` and `LyricDto` (with the possibility of adding additional data in the future)?  
Let me know your thoughts!

Komodo5197:
We want metadataProvider to autoDispose because otherwise the lyrics will never be moved out of memory.  In general, I don't think having a non-autoDispose family provider is good.  I think the best option is something like this:

```
/// Provider to handle pre-fetching metadata for upcoming tracks
final currentTrackMetadataProvider =
    AutoDisposeFutureProvider<MetadataProvider?>((ref) async {
  final List<FinampQueueItem> precacheItems =
      GetIt.instance<QueueService>().peekQueue(next: 3, previous: 1);
  for (final itemToPrecache in precacheItems) {
    BaseItemDto? base = itemToPrecache.baseItem;
    if (base != null) {
      final request = MetadataRequest(item: base, includeLyrics: true);
      unawaited(ref.watch(metadataProvider(request).future));
    }
  }

  final currentTrack = ref.watch(currentSongProvider).value?.baseItem;
  if (currentTrack != null) {
    final request = MetadataRequest(
      item: currentTrack,
      includeLyrics: true,
    );
    var out = ref.watch(metadataProvider(request)).valueOrNull;
    if (out != null) {
      ref.keepAlive();
    }
    return out;
  }
  return null;
});
```


Regarding downloading lyrics, I'm not completely sure what the best way to go about it is.  We don't want to download for all synced songs, only ones that are actually downloaded, both to reduce stored data and to reduce unneeded server requests.  I don't see any way to batch lyrics or attach them to the BaseItemDto or anything, so that means an individual server request for every downloaded song with lyrics.  The nastiest problem is that none of the lyrics endpoints have any sort of ID or version, so the only way to tell if the lyrics have changed is to re-fetch them for every song.  I guess we re-fetch on full sync, skip on quick-sync, and do the actual downloading inside the sync step.  I'm thinking the best way to actually store this info is probably in a separate Isar database, but using the same isarId as the parent song.

Chaphasilor:
Thanks for the suggestions, they seem to work. I think I was missing replacing the `FutureProvider` type with a `AutoDisposeFutureProvider` to get the `ref.watch` to work.  
Right now the metadata is only being requested while the player or lyrics screens are open, but I'd like for them to be requested as long as there is a queue to make the app more snappy. Compared to fetching the actual tracks it shouldn't cause any overhead anyway. Do you have any experience with that? Should I just add some kind of global listener for the provider? Or is fetching metadata in the background a bad idea for some reason?

We could create an issue for adding the needed info in the server repo. It probably isn't a big change. I'm also not a fan that lyrics are now split up, but it does makes sense to have them as media streams, just like subtitles.  
Maybe we can also get the server to add an endpoint that returns all information about a track at once, but I think it's unlikely, since a media stream usually isn't text-based, and I don't think they'd add an edge case for that. Same for batched lyric / media stream fetching.

What would be the benefit of a separate database for lyrics? More flexibility for handling `BaseItemDto`s and `LyricDto`s independently?

Komodo5197:
Regarding fetching the lyrics in the background, I believe just adding some sort of global watcher is the way to go about it.  Maybe just put it into the now playing bar to match the album image, just try not to accidentally cause unneeded rebuilds on metadata change.  This all somewhat conflicts with the other idea you had earlier of wanting to re-fetch failed requests when closing and re-opening the player screen.

Putting the lyrics in a separate database is just a bit more flexible, and helps separate any lyric related logic from the core DownloadItem syncing/updating logic, which is already complex enough as-is.  It allows managing the lyrics in a way closer to how we manage downloaded files than how we manage the core metadata, which matches better with the behavior we want.

Chaphasilor:
Okay, I'll see what I can do about it. I wasn't really aiming to re-fetch lyrics when re-opening the player screen, I was just trying to figure out the downsides a non-auto-disposed provider would have.  
Errors should be handled explicitly if possible, but I haven't dealt with that yet. Right now errors still result in layout errors and overflow, so it needs additional work anyway ^^

I'll try to figure out how the download system works and how to add the lyrics downloading. I'd be happy about any pointers or guidance you could give me! :)

Komodo5197:
Yeah, without autoDispose the provider will never get disposed, ever, unless one of it's dependencies refreshes.  The cached value will just be stored forever.  This makes sense for things with a single provider, like currentAlbumImageProvider, unless they are very large, and even makes sense for things with a small family like playerScreenThemeProvider, but is pretty bad for something with a large family like metadataProvider that creates a new provider per song.

For the downloads, you want to turn the LyricDto into an Isar collection and register it in main.  Then inside _downloadSong() you fetch the lyrics, if needed, outside the isar write transaction.  Then you actually write it to the new collection inside the same final write transaction in that function, using the same isarId as the parent DownloadItem.  Isar transactions cover all collections at once, so this means that it we can never get desync where a notDownloaded song has lyrics downloaded.  Finally, add a delete on the lyrics collection using the song's isarId to the transaction in deleteDownload()

Chaphasilor:
@Komodo5197 I've taken a shot at implementing lyrics downloading now in bd3ec69. I'm sure there's something I did wrong or more complicated than necessary, especially w.r.t. the `LyricsStub` and `LyricsItem`. Would be great if you could take a look and maybe leave a review or something!

lib/models/finamp_models.dart
```diff
@@ -1721,3 +1721,171 @@ enum TranscodeDownloadsSetting {
   @HiveField(2)
   ask;
 }
+
+
+/// A reference to a downloadable lyric with no state.  Can be freely created
+/// from a BaseItemDto and LyricDto at any time.  LyricsStubs/LyricsItems are considered
+/// equivalent if their types and ids match.
+@JsonSerializable(
```
Komodo5197, line 1729: You are correct that this is far more complicated than needed.  I think the below code should be all you really need.  The caller should already have the isarId to use from the DownloadStub, so just require it in the constructor. You might even be able to directly embed the LyricDto with the embedded tag and avoid the overhead of JSON.  I had problems embedding BaseItemDtos due to the limitations on embedded objects in Isar, but LyricDtos are much simpler and might work.
```
@collection
class DownloadedLyrics {
 DownloadedLyrics._build({
    required this.jsonItem,
    required this.isarId,
  });

  factory DownloadedLyrics.fromItem({
    required int isarId,
    required LyricDto item,
  }) {
    return DownloadedLyrics._build(
        isarId: isarId),
        jsonItem: jsonEncode(item.toJson()),
    );
  }

  /// The integer ID used as a database key by Isar
  final Id isarId;

  /// The LyricDto as a JSON string for storage in isar.
  /// Use [lyricDto] to retrieve.
  final String? jsonItem;

  @ignore
  LyricDto? get lyricDto => _lyricDtoCached ??=
      ((jsonItem == null) ? null : LyricDto.fromJson(jsonDecode(jsonItem!)));
  @ignore
  LyricDto? _lyricDtoCached;
}
```

lib/models/jellyfin_models.dart
```diff
@@ -3778,3 +3781,124 @@ class ClientDiscoveryResponse {
...
+
+}
+
+/// LyricResponse model.
+@JsonSerializable(
+  fieldRename: FieldRename.pascal,
+  explicitToJson: true,
+  anyMap: true,
+)
+@HiveType(typeId: 46)
```
Komodo5197, line 3885: Do we actually need to be able to store these in hive?

lib/services/downloads_service_backend.dart
```diff
@@ -1420,10 +1426,11 @@ class DownloadsSyncService {
     }
     // We try to always fetch the mediaSources when getting album/playlist, but sometimes
     // we download/sync individual songs and need to fetch playback info here.
-    List<MediaSourceInfo>? mediaSources = downloadItem.baseItem!.mediaSources ??
+    BaseItemDto? extendedItem = (downloadItem.baseItem?.mediaSources != null) && (downloadItem.baseItem?.mediaStreams != null) ? downloadItem.baseItem! :
         (await _jellyfinApiData.getItemByIdBatched(item.id,
-                "${_jellyfinApiData.defaultFields},MediaSources,SortName"))
-            ?.mediaSources;
+                "${_jellyfinApiData.defaultFields},MediaSources,MediaStreams,SortName"));
```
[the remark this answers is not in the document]
Komodo5197: Actually, I think a better option might be to rip this section out entirly, and completely rely on those other fetches.  There's a section in _syncDownload that forces a metadata update if sortName or childCount is missing, we should probably just add mediaSources/mediaStreams checks to that.  You still need to make sure that all the existing fetches with MediaSources also have MediaStreams to make that work.

Komodo5197:
You're right about the migration not being great.  As of right now, you need to fully delete and redownload a song to fetch the lyrics.  That goes along with treating the lyrics as a part of the download and not the metadata.  The best solution to this is probably to add another step as step 4 to the downloads repair that just fetches any missing lyrics. Anyone already on the beta can run it manually, and anyone new should be caught up by the run during the migration.  This is assuming that we put the mediaSources/medaiStream metadata update into the sync as I've mentioned in another comment.

lib/services/metadata_provider.dart
```diff
@@ -55,23 +53,52 @@ final AutoDisposeFutureProviderFamily<MetadataProvider?, MetadataRequest>
...
+    if (downloadItem != null) {
+      metadataProviderLogger.fine("Got offline metadata for '${request.item.name}'");
+      itemInfo = downloadItem.baseItem!;
+    }
+  } catch (e) {
+    metadataProviderLogger.warning("Couldn't get the offline metadata for track '${request.item.name}'");
+  }
 
+  // try to fetch from server if not found in downloads
+  itemInfo ??= await jellyfinApiHelper.getItemById(request.item.id);
```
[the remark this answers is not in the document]
Chaphasilor: Because that one doesn't contain the needed fields (mediaStreams). I guess I could put a check before that...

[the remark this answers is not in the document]
Komodo5197: Ah, makes sense.  A check if the provided BaseItemDto is already sufficient would indeed be nice.  It would also be a bit more consistent with other code if we always fetched the metadata from the server in online mode instead of using the downloaded set.

lib/models/jellyfin_models.dart
Chaphasilor: Does it hurt? I'd like to avoid someone trying to store it and it not being stored, and I think there's very little overhead to include it?

Komodo5197: I doesn’t think it has overhead, but I’m not sure why someone would want to store lyrics outside of the download system.  Also, you didn’t register the adapters so they would still need to add code anyway.

lib/services/downloads_service_backend.dart
```diff
@@ -1452,6 +1459,20 @@ class DownloadsSyncService {
...
 
+    // fetch lyrics if track has lyrics
+    LyricDto? lyrics;
+    if (mediaStreams?.any((element) => element.type == "Lyric") ?? false) {
+      _syncLogger.finer("Fetching lyrics for ${item.name}");
+      try {
+        lyrics = await _jellyfinApiData.getLyrics(itemId: item.id);
+        _syncLogger.finer("Fetched lyrics for ${item.name}");
+      } catch (e) {
+        _syncLogger.warning("Failed to fetch lyrics for ${item.name}. Continuing without lyrics.");
```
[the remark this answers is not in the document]
Chaphasilor: I've kept the catching for more detailed logging, but rethrow the error now. That should work, right?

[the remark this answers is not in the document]
Komodo5197: Yeah, a rethrow should be fine.  I'll note that any item which repeatedly fails to sync should get a full stack trace of the error logged already, though, so additional info may not be critical.

Chaphasilor:
@Komodo5197 I've pushed some changes now that hopefully address all of these issues. Could you take a look to check if that's actually the case? It seems like there's still something wrong with the lyrics repair (lyrics were only downloaded for some tracks in my testing), and that code should be refactored, but the rest should be fine, hopefully.

I'd also like to reflect the lyrics fetching as an actual download in the UI, like the song and image downloads. So that is shows up on the active downloads screen and is counted as an enqueued item during sync and repair. Do you think that makes sense / is possible? If so, what would I need to do?

I'm trying to push this update before 10.9 is fully released, which should be this weekend. So I'm a bit in a hurry, even though most things are working already...

lib/services/downloads_service_backend.dart
```diff
@@ -1424,13 +1424,9 @@ class DownloadsSyncService {
...
-    // We try to always fetch the mediaSources when getting album/playlist, but sometimes
-    // we download/sync individual songs and need to fetch playback info here.
-    BaseItemDto? extendedItem = (downloadItem.baseItem?.mediaSources != null) && (downloadItem.baseItem?.mediaStreams != null) ? downloadItem.baseItem! :
-        (await _jellyfinApiData.getItemByIdBatched(item.id,
-                "${_jellyfinApiData.defaultFields},MediaSources,MediaStreams,SortName"));
-    List<MediaSourceInfo>? mediaSources = extendedItem?.mediaSources;
-    List<MediaStream>? mediaStreams = extendedItem?.mediaStreams;
+    // At this point the baseItem should always have the needed attributes
+    List<MediaSourceInfo>? mediaSources = downloadItem.baseItem?.mediaSources;
+    List<MediaStream>? mediaStreams = downloadItem.baseItem?.mediaStreams;
```
Komodo5197, line 1429: You shouldn't need the blocks that copy these items into the BaseItemDto now that these attributes are not being updated.

lib/services/downloads_service.dart
```diff
@@ -681,15 +683,56 @@ class DownloadsService {
...
+        if (_isar.downloadedLyrics.where().isarIdEqualTo(item.isarId).countSync() > 0) {
+          continue;
+        }
+        idsWithLyrics[item.isarId] = null;
+      }
+    }
+    //TODO move this into a separate function in the background service
+    final JellyfinApiHelper _jellyfinApiData = GetIt.instance<JellyfinApiHelper>();
+    for (var id in idsWithLyrics.keys) {
+      var canonItem = _isar.downloadItems.getSync(id);
```
Komodo5197, line 700: You should probably just call this item.  We're not in a transaction.

lib/services/downloads_service.dart
```diff
@@ -681,15 +683,56 @@ class DownloadsService {
     // Step 4 - Make sure there are no unanchored nodes in metadata.
     _downloadsLogger.info("Starting downloads repair step 4");
     downloadCounts[repairStepTrackingName] = 4;
+    final Map<int, LyricDto?> idsWithLyrics = HashMap();
+    var allItems = _isar.downloadItems.where().findAllSync();
```
Komodo5197, line 687: This can be pre-filtered to just songs which are not notDownloaded for a minor efficiency gain.

lib/services/downloads_service.dart
```diff
@@ -681,15 +683,56 @@ class DownloadsService {
...
+      var canonItem = _isar.downloadItems.getSync(id);
+      if (canonItem?.baseItem != null) {
+        LyricDto? lyrics;
+        try {
+          lyrics = await _jellyfinApiData.getLyrics(itemId: canonItem!.baseItem!.id);
+          _downloadsLogger.finer("Fetched lyrics for ${canonItem!.baseItem!.name}");
+          idsWithLyrics[id] = lyrics;
+        } catch (e) {
+          _downloadsLogger.warning("Failed to fetch lyrics for ${canonItem!.baseItem!.name}.");
+          rethrow;
```
Komodo5197, line 709: Errors in the downloads repair should probably be caught, to make sure we get through all the steps.

lib/screens/player_screen.dart
```diff
@@ -199,6 +241,97 @@ class _PlayerScreenContent extends StatelessWidget {
...
+    );
+  }
+
+  Widget _buildBottomActions(BuildContext context, PlayerHideableController controller) {
+    return Consumer(
+      builder: (context, ref, child) {
+
+        final metadata = ref.watch(currentTrackMetadataProvider).value;
+
+        final isLyricsLoading = metadata?.item.mediaStreams?.any((e) => e.type == "Lyric") == null;
```
Komodo5197, line 287: I'm not sure these flag names make sense?  Also, why are manually checking the streams when we have the hasLyrics attribute?

lib/services/metadata_provider.dart
```diff
@@ -0,0 +1,140 @@
...
+        other.includeLyrics == includeLyrics &&
+        other.item.id == item.id;
+  }
+
+  @override
+  int get hashCode => Object.hash(item.id, includeLyrics);
+}
+
+enum MetadataState {
+  loading,
```
Komodo5197, line 35: You shouldn't need this.  Riverpod has this built in.  Watching the provider gives you an AsyncValue, which has an isLoading attribute.  Also, is it actually possible to return a MetadataProvider in a loading state in this setup?

lib/services/current_track_metadata_provider.dart
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+  final currentTrack = ref.watch(currentSongProvider).value?.baseItem;
+  if (currentTrack != null) {
+    final request = MetadataRequest(
+      item: currentTrack,
+      includeLyrics: true,
+    );
+    var out = ref.watch(metadataProvider(request)).valueOrNull;
```
Komodo5197, line 32: I think 
```
return ref.watch(metadataProvider(request).future);
```
is actually better, and will properly reflect the underlying loading state.  And then you can just stick an unconditional ref.keepAlive() into this function so that we don't need a listener in the now playing bar.

lib/services/metadata_provider.dart
```diff
@@ -0,0 +1,140 @@
...
+  //!!! only use offline metadata if the app is in offline mode
+  // Finamp should always use the server metadata when online, if possible
+  if (FinampSettingsHelper.finampSettings.isOffline) {
+    try {
+      final downloadItem = await downloadsService.getSongInfo(id: request.item.id);
+      if (downloadItem != null) {
+        metadataProviderLogger.fine("Got offline metadata for '${request.item.name}'");
+        itemInfo = downloadItem.baseItem!;
+      }
+    } catch (e) {
```
Komodo5197, line 73: getSongInfo doesn't really have any reasonable failure modes.  You probably don't need a catch block here.

lib/services/metadata_provider.dart
```diff
@@ -0,0 +1,140 @@
...
+      metadataProviderLogger.warning("Couldn't get the offline metadata for track '${request.item.name}'");
+    }
+  } else {
+    final requiredAttributes = [request.item.mediaStreams];
+    // try to fetch from server if not found in downloads
+    if (itemInfo == null || requiredAttributes.any((e) => e == null)) {
+      metadataProviderLogger.fine("Fetching metadata for '${request.item.name}' (${request.item.id}) from server due to missing attributes");
+      try {
+        itemInfo = await jellyfinApiHelper.getItemById(request.item.id);
```
Komodo5197, line 82: Does this actually include MediaStreams?

lib/services/metadata_provider.dart
```diff
@@ -0,0 +1,140 @@
...
+      DownloadedLyrics? downloadedLyrics;
+      try {
+        downloadedLyrics = await downloadsService.getLyricsDownload(baseItem: request.item);
+        if (downloadedLyrics != null) {
+          metadataProviderLogger.fine("Got offline lyrics for '${request.item.name}'");
+        }
+        else {
+          metadataProviderLogger.fine("No offline lyrics for '${request.item.name}'");
+        }
+      } catch (e) {
```
Komodo5197, line 115: Are there any reasonable ways for this to throw an error?

Komodo5197:
I think the fixes generally look good.  I've got some comments, but most are relatively minor.  I see you caught the issue with the async write transaction already.

The way we've embedded the lyric download into the syncing process makes displaying them as downloads in that manner a bit awkward.  We also don't queue them up in the same way.  In most circumstances it shouldn't be that noticeable anyway, and should just disappear into a slightly slower sync.

lib/services/downloads_service.dart
Chaphasilor: I initially misunderstood what you meant, but that's probably a good thing, since I'm now using one loop less ^^

lib/services/downloads_service.dart
Chaphasilor: like so?

lib/screens/player_screen.dart
Chaphasilor: Not sure what happened there. Should be better now

Chaphasilor:
Haven't had time to push the latest changes yet, so don't be confused by my comments. I'll do it in 2-3 hours

lib/services/metadata_provider.dart
Chaphasilor: It would fail if the item id is null, but I guess catching or not makes little difference in that case...

lib/services/metadata_provider.dart
Chaphasilor: It seems to return the full `BaseItemDto`, hence there being no filters or fields that can be configured/requested

lib/services/metadata_provider.dart
Chaphasilor: I guess not ^^

lib/services/current_track_metadata_provider.dart
Chaphasilor: thanks for pointing that out, I'm now doing that. I definitely was tired when implementing the loading state enum 😂  

I also changed the passive listen in the now playing bar to a call to closing the `KeepAliveLink` of the current track metadata provider, which should make sure that settings changes are properly applied (in case the metadata can be customized in the future), and it's possible to recover from failures.  
But the main reason for this was properly re-fetching the lyrics when toggling offline mode, which now seems to work fine.

Chaphasilor:
okay, changes are pushed now :)

lib/services/downloads_service.dart
Komodo5197: Yeah, that looks good.

lib/services/current_track_metadata_provider.dart
Komodo5197: I don't think I fully though this through either.  Keepalive isn't going to get you you prefetching.  I checked the docs and the suggested method is wrapping everything in a consumer that watches the ref.  You can give consumers a child outside their build methods to prevent rebuilding everything unnecessarily. Also, as an alternative to the keepalive link, there's a provider version of the finamp settings you can watch to autodispose when it changes without needing the parents to worry about it.  You can also use select to specifically only rebuild when offline changes and not other attributes, because I there's a couple settings like current library that you really don't need to dispose for.

Chaphasilor: Okay, I've pushed another change. I went with a ref.listen instead of the ref.watch outside the build method, not sure if that's equivalent? Watching and selecting the settings seems to work as well.  

However, now the prefetching/caching doesn't seem to work at all. I noticed that the was I was using the provider, it gave me the previous value if available, instead of showing the loading state. I've fixed that now, but the loading state is now briefly shown with every track change. Not sure if that's how it's supposed to we working and my loading state logic just has to be changed, or if the prefetching is actually not working...

Chaphasilor:
> The way we've embedded the lyric download into the syncing process makes displaying them as downloads in that manner a bit awkward. We also don't queue them up in the same way. In most circumstances it shouldn't be that noticeable anyway, and should just disappear into a slightly slower sync.

I think it's fine. Was there any way to check if a repair is still in progress? I noticed that during the repair the "nodes syncing" went to 0 but there were still lyrics being fetched. But I dimly remember asking you to add some way to check for a running repair in your original PR, even if it wasn't used yet...

I also changed the metadata fetching a bit, by using `getPlaybackInfo()` instead, and creating a custom `MediaSourceInfo` when offline. I've also made sure to use a combination of both when a downloaded track is played in online mode, since Finamp will currently use the downloaded track in that case (with different metadata).  
I'm using `downloadsService.getFileSize()` for that, but I think it should be fine since the target is always a single track, so there should never be a recursive calculation that might take longer, correct?

Komodo5197:
Unless something broke, there should be a line like 'downloads repair active' on the downloads screen. Get file size should be fine and shouldn't be recursive, although you may want to assert that you've got a song somewhere.

lib/services/current_track_metadata_provider.dart
Komodo5197: If you're talking about the super quick blinks, that's async lag, becuase the async current metadata provider doesn't run until after the build method that calls it has completed..  You can get around it by making the current metadata provider an AutoDisposeProvider<AsyncValue<MetadataProvider?>>.  Then the provider can syncronously hand over the child value without adding an unneeded async cycle.

Komodo5197: I think the delay is async lag.  You should be able to convert the current metadata provider into an AutodisposeProvider<Asyncvalue<MetadataProvider?>> to work around this.

lib/services/current_track_metadata_provider.dart
```diff
@@ -24,13 +23,14 @@ final currentTrackMetadataProvider =
     }
   }
 
+  unawaited(ref.watch(FinampSettingsHelper.finampSettingsProvider.selectAsync((settings) => settings?.isOffline))); // watch settings to trigger re-fetching metadata when offline mode changes
```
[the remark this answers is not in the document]
Chaphasilor: I seems to be working, but I'll try moving it. I've also had the keepalive stuff in here, so I assumed it would be the right place...

lib/services/metadata_provider.dart
```diff
@@ -53,40 +55,77 @@ final AutoDisposeFutureProviderFamily<MetadataProvider?, MetadataRequest>
...
-  // Finamp should always use the server metadata when online, if possible
-  if (FinampSettingsHelper.finampSettings.isOffline) {
-    final downloadItem = await downloadsService.getSongInfo(id: request.item.id);
+  final downloadStub = await downloadsService.getSongInfo(id: request.item.id);
+  if (downloadStub != null) {
+    final downloadItem = await ref.watch(downloadsService.itemProvider(downloadStub).future);
     if (downloadItem != null) {
       metadataProviderLogger.fine("Got offline metadata for '${request.item.name}'");
-      itemInfo = downloadItem.baseItem!;
+      var profile =
```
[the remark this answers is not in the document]
Chaphasilor: Yeah that makes sense. I'm still getting familiar with working with the download system, thanks your your reviews!

lib/services/current_track_metadata_provider.dart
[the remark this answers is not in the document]
Komodo5197: Does it?  I wouldn't really expect it to - I'd expect invalidating the current provider to just re-fetch the same cached metadataProvider.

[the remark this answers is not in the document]
Chaphasilor: Yeah, not anymore. It was working with the keepalive though

lib/services/metadata_provider.dart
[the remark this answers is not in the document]
Chaphasilor: I'm guessing `DownloadItemState.complete` is the only state that would be considered "fully downloaded"? Or are `needsRedownload` and `needsRedownloadComplete` also valid?

lib/services/current_track_metadata_provider.dart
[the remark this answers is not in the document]
Komodo5197: Ah, that makes sense.  Without somebody watching, the current provider doesn’t rebuild so the metadataProvider autodidposes.

lib/services/metadata_provider.dart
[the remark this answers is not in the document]
Komodo5197: Needsredownloadcomplete also counts.  There’s an iscomplete flag on the state enum you can check.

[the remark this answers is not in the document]
Chaphasilor: Ahh, didn't scroll down far enough. I only saw `isFinal`. Guess I need a bigger monitor xD

lib/services/current_track_metadata_provider.dart
Chaphasilor: worked like a charm! But I have idea why :D

Chaphasilor:
Okay, that should be all of it. Main features should be fully working now:
- Fetching metadata and lyrics and displaying them nicely
- Pre-fetching metadata for upcoming tracks
- Downloading lyrics and more metadata
- Repairing existing downloads to include lyrics and metadata

I think I'll leave this as it is now for the 10.9 update, and then use the initial feedback for gauging which configuration options are wanted by users.  
Would appreciate another round of testing by others!

lib/screens/player_screen.dart
```diff
@@ -174,17 +199,34 @@ class _PlayerScreenContent extends StatelessWidget {
...
+                                .finampSettings.disableGesture) {
+                              queueService.skipByOffset(1);
+                            }
+                          } else if (direction == SwipeDirection.right) {
+                            if (!FinampSettingsHelper
+                                .finampSettings.disableGesture) {
+                              queueService.skipByOffset(-1);
+                            }
+                          }
+                        },
```
Komodo5197, line 217: You should add a tap gesture to toggle playback as well.

lib/services/current_track_metadata_provider.dart
Komodo5197: Basically, the problem is that the current provider was a future provider that ran an async function to build.  Async code blocks return futures, and futures cannot be evaluated synchronously, ever.  This is a problem because the entire build sequence for a frame is synchronous, which means no futures can evaluate until it is fully complete.  This is why future providers return asyncvalue, which either synchronously has the future’s result already, or synchronously doesn’t.  So after the current provider invalidated we had to build the widget twice.  On the first build we trigger the provider build, but it cannot evaluate the returned future and provides a loading async value.  Then during the async gap after the widget tree build completes, the future is evaluated and the provider updates to a data asyncvalue.  This triggers the second widget build with the final data on the next frame.  The new code synchronously returns the chosen metadataProvider’s asyncvalue, which should already be the final data AsyncValue due to the precaching.

lib/screens/player_screen.dart
```diff
@@ -174,17 +199,34 @@ class _PlayerScreenContent extends StatelessWidget {
...
                     mainAxisAlignment: MainAxisAlignment.spaceAround,
                     children: [
-                      SizedBox(
-                          height: min(
-                              constraints.maxHeight -
-                                  controller.getTarget().height,
-                              constraints.maxWidth),
-                          width: constraints.maxWidth,
-                          child: const PlayerScreenAlbumImage()),
+                      SimpleGestureDetector(
```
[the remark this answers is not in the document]
Chaphasilor: I'm planning to refactor this into using a PageView at some point, and I'm not sure if that would properly work from inside the `PlayerScreenAlbumImage` (because the next cover should slide in from off-screen). We also use that widget in some other places (now playing bar, queue panel), where we don't want it to be swipeable.  
So I think we should leave this here, and have code duplication instead?

lib/screens/player_screen.dart
Chaphasilor: As I mentioned in the other thread, I want this to be animated in the future. Do you have an idea how we could animate play/pause so that it's obvious what happend because of the tap?  
If not, I'd rather not add the gesture for now.

But I also remember someone requesting the swipe gesture, I'll see if they also requested play/pause :)

lib/screens/player_screen.dart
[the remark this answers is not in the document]
Komodo5197: The PlayerScreenAlbumImage is only used here, the others just use AlbumImage.  I think any sort of track switching animation would need heavy refactoring of a lot of the components anyway and shouldn't weigh too heavily on decisions like this now, but code duplication is fine.

lib/screens/player_screen.dart
Komodo5197: I feel like playing/pausing is itself pretty obvious and easy to connect without an animation, and you don't have gestures for skipping yet either.  I guess you could just sort of flash a low opacity play/pause Icon over the cover like youtube does?

lib/services/current_track_metadata_provider.dart
Komodo5197: Yeah, that's pretty much it.

lib/screens/player_screen.dart
Chaphasilor: yeah, I was thinking something similar. I guess that's good enough, I'll add the gesture :)

lib/screens/player_screen.dart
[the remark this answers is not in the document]
Chaphasilor: I could've sworn that the widget was used in other places too, but you're right. I've moved the gesture detector there.  
We lose the "tap cover to go to start of song" functionality on the lyrics screen, but that was a minor and hidden feature anyway, and the new behavior is more consistent

lib/screens/player_screen.dart
```diff
@@ -94,10 +100,19 @@ class _PlayerScreenContent extends StatelessWidget {
           }
         }
       },
+      onHorizontalSwipe: (direction) {
+        if (direction == SwipeDirection.left) {
```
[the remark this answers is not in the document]
Chaphasilor: I guess we can. My original reasoning was that this would be a way to get to the lyrics screen even if lyrics can't be loaded, because that screen shows more info about why they couldn't be loaded.  
But a hidden swipe gesture is probably not the right approach for this.

I guess this would mean refactoring the `_PlayerScreenContent` to be a `ConsumerWidget`?

The other option we have would be to not fully disable the button, instead gray it out and change the icon if lyrics can't be loaded, but still letting users tap on it to see the reason why the lyrics aren't available. Disabled elements are "bad" UX-wise anyway...  
What do you think about that?

lib/services/metadata_provider.dart
```diff
@@ -30,7 +32,7 @@ class MetadataRequest {
   }
```
Komodo5197, line 32: You should update the == override to match the hashCode.

lib/screens/player_screen.dart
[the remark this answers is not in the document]
Komodo5197: Greying the button out but still letting people tap through to the lyrics screen seems pretty reasonable, and doing it with the button seems more reasonable than only having the swipe.  It's also harder to accidentally trigger, which is the main reason I want the swipe disabled when we don't have lyrics.

lib/screens/player_screen.dart
```diff
@@ -199,6 +214,97 @@ class _PlayerScreenContent extends StatelessWidget {
...
+                  const SizedBox(
+                    width: 80,
+                    // child: Text("Output")
+                    child: SizedBox.shrink(),
+                  ),
+                  const SizedBox(
+                    width: 80,
+                    child: QueueButton()
+                  ),
+                  SizedBox(
```
Komodo5197, line 288: This widget is taller than the queue button due to the underlying TextButton not having its visualDensity set to -4 like the QueueButton.  This can break the layout and overflow.

lib/services/metadata_provider.dart
Chaphasilor: Oh, didn't see that. Any reason why we don't just use the `hashCode` for the comparison?

lib/screens/player_screen.dart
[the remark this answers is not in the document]
Chaphasilor: So you want both changes? Making the button not fully disabled, and also converting to a ConsumerWidget so disable the swipe gesture when lyrics are missing?

lib/screens/player_screen.dart
Chaphasilor: So I guess I just check the height difference between a regular button and one with visualDensity of -4, and then subtract that from the breakpoint heights?

lib/services/metadata_provider.dart
Komodo5197: Possible hashcode collisions maybe? I don’t know, I’m mostly just following the docs here.

lib/screens/player_screen.dart
Komodo5197: The breakpoint is tuned to the shorter queue button, so you need to increase it to match the lyric button.  Or shrink the lyric button instead.

lib/screens/player_screen.dart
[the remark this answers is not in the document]
Komodo5197: That would be my preferred setup, yeah.

Komodo5197:
While looking at the volume PR, I noticed that there’s another spot in Downloditem.copyWith you should add some mediastream stuff to prevent accidentally removing it when updating metadata.

lib/services/metadata_provider.dart
Chaphasilor: I'm glad at least one of us is 👀

Chaphasilor:
Was the thing I just added in the latest commit what you were referring to? I'm honestly not sure what the point of that assert was there (just to make sure the list of MediaSources is not empty?), but tried to add similar checks for the MediaStreams. Would that work, or do I have to do something else?

Komodo5197:
Yeah, that’s what I was referring to.  Looking at the assert, I’m honestly not entirely sure why it was put in either.  Regardless, it’s probably best to update it to make sure both are null or non-empty, vs the current logic which checks if either is.

lib/screens/player_screen.dart
Chaphasilor: I've shrunken the button now. I don't like these implicit margins anyway. A visual density of -4 corresponds to a symmetric padding of 8px on each side by the way...

lib/screens/player_screen.dart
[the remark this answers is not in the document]
Chaphasilor: Your wish has been granted ✨

Chaphasilor:
Okay, that should be it then. I'll merge this tomorrow morning, unless something else comes up...

Chaphasilor:
Just noticed I forgot to update the the assert. I'll throw that in tomorrow

jmshrv:
This all looks good to me :)

jmshrv (approved):

Chaphasilor:
@barolo it was just released ^^

But the feature will only work with a Jellyfin 10.9 server!

Chaphasilor:
Yeah, I agree that the current "layered" behavior isn't great, especially because it doesn't allow quickly collapsing the player screen and e.g. searching for a track. I can try making it so that it actually switches between the screen instead of layering them, and that it remembers the default.  
Not sure when I'll get to that though, I'll probably wait for additional feedback after 10.9 has officially dropped...

PR #695, opened 2024-04-18
Add app icons for the (Linux) desktop app

Maxr1998:
Supersedes #692.


Chaphasilor:
Thanks!  
Is there a particular reason why you put each resolution in a separate directory?

And I think there's no need to make this linux-specifc, unless that naming convention is somehow needed :)

Maxr1998:
It makes it easier to install the folders to the package later on, since they'll all be moved to `/usr/share/icons/hicolor/<resolution>/apps/finamp.png`.
Which reminds me I forgot the `apps/` subfolder. Guess I'll have to fix this.

The resolution and naming scheme are all linux specific, that's why I named the folder like that.

Btw, I plan to regenerate the iOS and Android icons too, once this has been merged into the `redesign` branch, since the ones converted with ImageMagick aren't very sharp (because ImageMagick first converts to bitmap and then scales, instead of rendering at target resolution).

jclsn:
> Thanks! Is there a particular reason why you put each resolution in a separate directory?
> 
> And I think there's no need to make this linux-specifc, unless that naming convention is somehow needed :)

This is the common Linux icon folder structure, which is used by many desktop environments. This way they can also be easily imported in custom icon themes.

Chaphasilor:
Alright, thanks to both of you! Hopefully everything can be properly packaged now 😁

Maxr1998:
Thanks for the merge! I'll update the packaging this weekend.

Issue #688, opened 2024-04-08
[Bug] Track repeats three times in Playback History


snppls:
On iOS 17.3.1 and Finamp 0.9.5 every song played appears exactly three times in the Playback History rather than once.

![IMG_0367](https://github.com/jmshrv/finamp/assets/156883144/8559c629-8476-488b-af2f-1dedf5bfd4aa)

Chaphasilor:
I guess that answers the question if anyone else uses the playback history :P

Can you get me some logs? On Android it works fine, so I think this is a platform-specific issue.  

![Screenshot_20240409-112358.png](https://github.com/jmshrv/finamp/assets/18015147/f2c8fc3b-6471-4294-8584-e02a6f32c207)

Never-ending story with just_audio and playback tracking 😅

snppls:
Here is a snippet of my recent logs. Attached is the full log - its pretty massive. 
[logs.txt](https://github.com/jmshrv/finamp/files/14917831/logs.txt)

[QueueService/FINEST] 2024-04-09 21:13:17.164431: Saved new periodic queue previous:[56d7ca4a7389729ef28d58727416cf0f, b606ae871125cf62cbbd082398332e06, 0c57f7d3e07d15fa2cb65e0a7ef55334, a4f5c0de71ed87bfa4d090ef3a41318b, 12ba7edd5b807cec2322296714080bdd] current:526275b2974c05ecf0228eca6d631a57 seek:10440 next:[] queue:[85a3f66bbf4d37a03008991f6010400e, 4f962317991ba3d4c133aaf2abceb276, ce3ea4c785737e2070e49fd4726fe155, 6be3ea8c12331e2920f18b11aec03f29, c1b9edde86c31994d11ce11580e63830, ef2b33d824b1378f272ea5167203602c, d7f5575dba5191da7bfc975a8f458999, ce0e9fc4c7070c9374d58ad36430cd11, 645679c6462bb3391efb3e826b86a21d, cbc6c0133d8f60476c6d4e7aacc2bf39, 4cfd90b1ec48dec8c54f16dc7311ef44, 08e290f043f3fb675c4a850bf5a8e493, d6779ec083098309210025541e13211a, 8985a6f25ce4282b287cb06a1443b898, 2476e901c70435e84891d3b142cc9aed]
[Chopper/INFO] 2024-04-09 21:13:17.216018: --> POST BASEURL/Sessions/Playing/Progress
		Content-Type: application/json; charset=utf-8
[...]

PR #696, opened 2024-04-21
Make sure audio handler knows queue from the start

Maxr1998:
If we swipe to add to the queue after launching the app (when no queue exists), the new audio source will temporarily be added to the `_queueAudioSource`, but since the `_audioHandler` doesn't know this queue yet, `_queueFromConcatenatingAudioSource` will still see an empty queue, and thus ignore the new element. Immediately registering the audio source to the audio handler resolves this issue. Furthermore, since the two elements are only written once, they can be final.

### DIFF
--- a/lib/services/playback_history_service.dart
+++ b/lib/services/playback_history_service.dart
@@ -275,7 +275,8 @@ class PlaybackHistoryService {
         !forceNewTrack && (
           currentTrack == _currentTrack?.item ||
           currentTrack.item.id == "" ||
-          currentTrack.id == _currentTrack?.item.id
+          currentTrack.id == _currentTrack?.item.id ||
+          !_audioService.playbackState.value.playing
         )
       ) {
       // current track hasn't changed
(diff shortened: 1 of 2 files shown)


Chaphasilor:
the change itself works fine, what I'm not super happy with is that a track that was added to an empty queue will be listed in the queue twice.  
I managed to fix the track being reported before playback actually starts which makes it more consistent with what is actually reported to the server, but the track should still only show up once in the local history...

Chaphasilor:
The doubled history entry might actually be related to #688, but I'm not certain. I still can't reproduce that one

Chaphasilor:
Seems like it's a different bug. But I'll merge this anyway as the bug is very minor and the server-side seems to be fine.  
If you can figure out how to solve it, please make a follow-up PR!

Maxr1998:
On my end, I'm only seeing the track once after adding it to the queue. Or do you mean the fact that it's also added to the playback history?
Maybe we could filter that out by playback position and state?

Chaphasilor:
I'm just talking about the Playback History. The queue is fine.

If you can figure out a way to do the filtering without breaking playback reporting in other cases that would be great!

I really need to write something up about how playback reporting is supposed to behave in different scenarios...

PR #705, opened 2024-04-29
[Redesign] Skip theme transition animation if app was in background

Komodo5197:
If the app changed theme while in the background, do not animate from the last foreground theme to the current theme when returning to foreground.  Just switch to the latest theme as quickly as possible.


Chaphasilor:
Seems to work, thanks. I spotted two issues that I thought were caused by this PR, but seems like they're unrelated.

I should probably also make the theme of at least the queue panel adapt to the system theme properly, and maybe even the song menu (unless that causes issues with rebuilding a different track)...

PR #706, opened 2024-04-29
Updated volume normalization

Chaphasilor:
Adapts volume normalization / replay gain based on the upcoming changes in https://github.com/jellyfin/jellyfin/pull/11457

Removes outdated settings and shows normalization gain as a feature chip


lib/components/AlbumScreen/song_list_tile.dart
```diff
@@ -250,13 +250,13 @@ class _SongListTileState extends ConsumerState<SongListTile>
                           AppLocalizations.of(context)!.placeholderSource),
               id: widget.parentItem?.id ?? "",
               item: widget.parentItem,
-              // we're playing from an album, so we should use the album's LUFS.
-              // album LUFS sometimes end up being simply `0`, but that's not the actual value
-              contextLufs: (widget.isInPlaylist ||
+              // we're playing from an album, so we should use the album's normalization gain.
+              // album normalization gain sometimes ends up being simply `0`, but that's not the actual value
```
Bond-009, line 254: With the new changes this shouldn't happen anymore ;)

Chaphasilor: nice. and even if it did, it's not an issue, because the gain change would simply be 0, which is the same as not applying a gain change (which was the fallback). It was different when we were still dealing with LUFS, that would've lead to an 18 dB increase in volume 👀

Chaphasilor:
@Komodo5197 will a downloads repair be enough to update the `BaseItemDto` metdata for offline mode when people upgrade to 10.9? Or do we need to add a migration for that?

Komodo5197:
It could be, but does need some tweaks.  I wanted to reduce unneeded database writes, so the metadata only updates if the function BaseItemDto.mostlyEqual returns false.  You can add a comparison of the relevant attributes there and it should work.

Chaphasilor:
@Komodo5197 if you don't have anything else, I'll go ahead an merge this :)

Komodo5197:
The downloads change looks good,  I haven’t payed too much attention to the other code.  I’ll note with this and the lyrics PR we’re up to ~6 feature chips, and it would be nice to be able to disable some in settings to avoid scrolling on small screens.  That’s only slightly related to this though.

Chaphasilor:
Okay. Yeah customizing feature chips is definitely planned. I'd like to move away from the generic list of switches as settings, to a more tailor-made customization that fits whatever is customized, along with a preview.  
So in the case of the feature chips, we'd show a preview with some sample data, and the preview would be the same size as the chips will end up being on the player screen. Maybe related settings (e.g. player screen) could share a single preview.

For the feature chips I'd like a way to show/hide or add/remove chips, and reorder them. All of this should be reflected in the preview, but the preview shouldn't be interactive. All customization should use separate controls.

I'll see if I can get any work done on this in the coming weeks, including an initial mockup.

But for now, a bit of overflow is fine. The complaints might motivate me to work on it sooner rather than later xD

PR #707, opened 2024-04-29
- fix download counter showing wrong number with playlists

Tiefseetauchner:


Chaphasilor:
Seems to be working, thanks!

PR #708, opened 2024-04-30
[Redesign] Show position in queue

Komodo5197:
Show current track position in queue and total queue length.  I'd also like to show the total and/or remaining duration of the queue, but I can't figure out a good place to put this info.

Screenshot 2024-04-29 215614

### DIFF
--- a/lib/components/PlayerScreen/queue_list.dart
+++ b/lib/components/PlayerScreen/queue_list.dart
@@ -1056,13 +1056,13 @@ class QueueSectionHeader extends StatelessWidget {
     return Padding(
       padding: const EdgeInsets.only(left: 16.0, right: 16.0),
       child: Row(
-        crossAxisAlignment: CrossAxisAlignment.center,
+        crossAxisAlignment: CrossAxisAlignment.start,
         mainAxisAlignment: MainAxisAlignment.spaceBetween,
         children: [
           Expanded(
             child: GestureDetector(
                 child: Padding(
-                  padding: const EdgeInsets.symmetric(vertical: 8.0),
+                  padding: const EdgeInsets.only(bottom: 8.0, top: 12.5),
                   child: Column(
                     crossAxisAlignment: CrossAxisAlignment.start,
                     children: [
@@ -1328,9 +1328,9 @@ class RenderQueueTracksMask extends RenderProxySliver {
             Color.fromARGB(0, 255, 255, 255),
             Color.fromARGB(255, 255, 255, 255)
           ], begin: Alignment.topCenter, end: Alignment.bottomCenter)
-              .createShader(const Rect.fromLTWH(0, 125, 0, 10)),
+              .createShader(const Rect.fromLTWH(0, 131, 0, 10)),
           blendMode: BlendMode.modulate,
-          maskRect: const Rect.fromLTWH(0, 0, 99999, 140));
+          maskRect: const Rect.fromLTWH(0, 0, 99999, 145));
 
       context.pushLayer(layer!, super.paint, offset);
     } else {
--- a/lib/components/PlayerScreen/queue_list.dart
+++ b/lib/components/PlayerScreen/queue_list.dart
@@ -42,9 +42,8 @@ class _QueueListStreamState {
 }
 
 class QueueList extends StatefulWidget {
-
   static const routeName = "/queue";
-  
+
   const QueueList({
     Key? key,
     required this.scrollController,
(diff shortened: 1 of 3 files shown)


Chaphasilor:
Good idea in principle, but I don't think this is the right place to put this info. I could see it being shown next to or below the queue source ("playing from") instead.

Depending on what your use case is, we could also just show the length of the previous tracks and queue instead:
```
Previous Songs (3)
<Current Track>
Next Up (1)
Playing from <Source> (171)
```
What do you think? Would that work?

Chaphasilor:
As for showing the remaining duration, that would work well if we place the information below the "Playing from" text. It could say `4/175 (2h 53m remaining)` there, or something similar

Komodo5197:
I've placed all the information below the playing from line.
Screenshot 2024-04-30 112956

Chaphasilor:
Could you simply make the header bigger so that the playing from line stays aligned with the buttons? 
 
And please format the duration as `Xh Ym`. You can add seconds if you want, but this will be the format I'd like to use for the album and playlists screen. If I end up using something else, we could always update this to match :)

Komodo5197:
Putting the icons into the same row as the title doesn't really work due to the padding in the iconButtons.
Screenshot 2024-04-30 144807



I guess I could just increase the padding before the text column a bit if you don't like the alignment?
Screenshot 2024-04-30 145153

Chaphasilor:
I think the second picture looks good. Do you think we should make the extra line optional with a setting?  
Also, please surround the `/` with a space on each side to make it look nicer :)

Komodo5197:
I don't think we need a setting.  It's not like the line takes up a crazy amount of space, and I can't really think of another reason you might not want it.  Also, I don't want to make the ShaderMask height adjustable.

lib/components/PlayerScreen/queue_list.dart
```diff
@@ -1054,34 +1053,70 @@ class QueueSectionHeader extends StatelessWidget {
...
+                                  (snapshot.data!.currentTrack == null ? 0 : 1);
+                              var trackCount = trackPosition +
+                                  snapshot.data!.nextUp.length +
+                                  snapshot.data!.queue.length;
+                              var remaining = const Duration(seconds: 0);
+                              for (var item in snapshot.data!.queue +
+                                  snapshot.data!.nextUp) {
+                                remaining += item.item.duration ??
+                                    const Duration(seconds: 0);
+                              }
```
Chaphasilor, line 1085: Could you please make these four things (current index, total queue size, remaining queue length, remaining duration, maybe also total duration) getters on `FinampQueueInfo`? That would clean up the code, and it might be useful to have in other places too.

lib/l10n/app_en.arb
```diff
@@ -1308,5 +1308,23 @@
     "description": "Text shown on button to require an item which is already downloaded by a parent collection, preventing it from being deleted if the parent is removed."
   },
   "showArtistChipImage": "Show artist images with artist name",
-  "showArtistChipImageSubtitle": "This affects small artist image previews, such as on the player screen."
+  "showArtistChipImageSubtitle": "This affects small artist image previews, such as on the player screen.",
+  "queueTotalSize": "{ trackPosition } / { trackCount }  ({hoursRemaining, select, 0{} other{{hoursRemaining}h }}{minutesRemaining}m remaining)",
```
Chaphasilor, line 1312: I'd prefer not having the index an total size (`{trackPosition } / { trackCount}`) as a translatable string. I don't really see a reason why this would need to be translated (not sure about RTL languages though), and I can already see the translations messing up the formatting by omitting one of the spaces or something similar. Everything inside the parentheses should be sufficient, with a more fitting name (like `remainingDuration`).

Or was there a deeper reason for putting it all together like this?

Chaphasilor (commented):
Noticed two more things that could be improved here, sorry for not spotting them earlier. I'll probably merge this tomorrow then.

lib/l10n/app_en.arb
Komodo5197: I didn't have a specific reason, I'm just not sure how much flexibility translations might need and so just put everything in.  The spacing to make the trackcount looks right doesn't seem to be more complicated than the spacing needed to make some of the fancier select cases look right.  I think a RTL language would reverse the whole thing, which might be the desired behavior?  It's questions like this that made me put as much of the string into the translation as possible, but I can simplify it if you think that's best.

Chaphasilor: Right, I'm not sure either. The selects aren't as prominent, that's why I'm less concerned there.  
What you could do is move them to a separate translation string at least, so that they're independent and can be reused more easily?

Komodo5197: I'm not entirely sure how useful that would be?  I've decided to just implement your original suggestion.

Chaphasilor: We'll that would still support reversing the order for RTL languages.

But it's fine for now, there's no way to translate beta strings yet anyway. We can deal with it if it actually causes issues :)

Chaphasilor:
Thanks for the PR <3

PR #712, opened 2024-05-04
Desktop Support

Chaphasilor:
I'm sick of merging the changes back into the desktop branch, so I'd like to merge them if possible.  
Unless there are any critical issues? If some of you could give this PR some testing to make sure nothing broke, both on desktop **and** mobile, that would be great.

My plan is to merge this in for the next beta update, and then keep it around for one or two updates without an "official" desktop release to make sure the mobile version isn't negatively affected by any of the changes.  
This time period can then also be used to a soft release, where all of the various stores and repositories can be prepared based on the `redesign` branch to make Finamp available everywhere.

CC @jmshrv @Komodo5197 @Maxr1998 @baarkerlounger @AverageHelper

It would also be nice to compile a list of how to build for each platform in the comments, as well as which files need to be updated when there is an update (e.g. bumping version number, etc.)


Chaphasilor:
I'll start with Windows:

### Building

```sh-session
dart run msix:create
```

### Updating

1. Bump `msix_version` in `msix_config` within `pubspec.yaml`. 
   https://github.com/jmshrv/finamp/blob/e2eb8c816e41fa3d8e541d731e23974b6a59d330/pubspec.yaml#L173
2. Create a new build
   
### Caveats:

- It's not possible to update unless the version number was incremented. The last digit of the number can be incremented in case a re-release is needed

### Stores/Repositories:

- Microsoft Store: https://pub.dev/packages/msix#microsoft-store-icon-publishing-to-the-microsoft-store
- winget: https://github.com/microsoft/winget-pkgs
  - winget also supports installing from the Microsoft Store, so this is optional

jmshrv:
For macOS, the process is described in https://docs.flutter.dev/deployment/macos#create-a-build-archive-with-xcode

For distribution, I'll put it on the Mac App Store and include a built `.app` for people who don't want to use the Mac App Store.

Currently setting up the macOS Xcode project, hopefully it's just copying stuff over from iOS

jmshrv:
Just going to post macOS nitpicks as I go here, will mostly be packaging stuff for me to fix

First off, the app icon needs fixing to fit in with macOS

macOS dock showing Finamp's icon being a perfect square instead of rounded

jmshrv:
Just pushed macOS metadata, first impressions are very good, seems to work as well as iOS :)

Maxr1998:
# Linux

The [official deployment guide](https://docs.flutter.dev/deployment/linux) describes how to bundle a package for the snap store. As snaps aren't exactly popular in the general Linux community<sup>[[1](https://www.reddit.com/r/linux/comments/j3ajnf/whats_wrong_with_snaps_why_so_many_people_hate_it/), [2](https://lemmy.world/post/1477518)]</sup>, I'll instead describe the packaging process that [I use for Arch Linux](https://aur.archlinux.org/packages/finamp-git) (which can likely be adapted for other solutions like deb or rpm).

## Building

First, you simply create a release build with the flutter tool:
```bash
flutter --suppress-analytics build linux --release

# Or, if you split dependency download and build,
# like the PKGBUILD (Arch buildscript) does:
flutter --suppress-analytics pub get
flutter --suppress-analytics build linux --release --no-pub
```

## Packaging

To package everything, you can simply add the created files into your package archive. Things you'll want:

- The release bundle (`build/linux/x64/release/bundle`). This should ultimately be copied to `/opt/finamp` during installation. Unfortunately, it's **not** possible to use the recommended system paths for executables (`/usr/bin`) and libs (`/usr/lib` or `/lib`), as the files from the bundle need to stay in the same folder relative to each other.
- A desktop file to launch the app. There's currently [one included in the finamp-git AUR repo](https://aur.archlinux.org/cgit/aur.git/tree/finamp.desktop?h=finamp-git) (which should probably be moved to this repo eventually).
- Desktop/application icons. These are included in this repo (`assets/icon/linux/`) and can be copied to the package to that they're installed to `/usr/share/icons/hicolor/…`.
- The LICENSE, depending on the package format and distribution. On Arch, it's installed to `/usr/share/licenses/finamp`.

## Updating

Currently, we only have a [`finamp-git` AUR package](https://aur.archlinux.org/packages/finamp-git) on Arch, which grabs a version number automatically from the commit hash. No additional step is necessary.

Later on, I plan to also maintain a variant that builds the latest tagged release. As long as there's a git tag with a version number (used as the tag, for example), we're good. For new versions, the version number needs to manually be updated in the `PKGBUILD` by the maintainer (me).

jmshrv:
For Linux it'd be nice to have deb/rpm and Flatpak, that should cover most cases (and it's pretty standard for more niche distros to adapt deb/rpm packages (see most `-bin` AUR packages))

It'd also be nice to have arm64 builds (personally, for Asahi Linux, and a few people have mentioned Linux phones), but it looks like there may be a few blockers there based on discussion in #40.

This could all likely be done in GitHub Actions, saves someone a package to build lol

Also thanks @Maxr1998 for publishing an AUR package, it's super cool to see Finamp there :)

AverageHelper:
macOS seems to run well enough! I like the new two-column layout on large window sizes.

Seems to have trouble playing OPUS, but that's also an issue on iOS IIRC, so that's probs fine for now.

It does seem like users can shrink the window a bit too much tho :P 
Screenshot 2024-05-06 at 09 23 01

AverageHelper:
It looks like the player UI doesn't respond to settings changes until relayout, which gets extremely obvious in two-column view.

jmshrv:
Oh god yeah I made that assumption when initially developing the player screen, that'll be fun to fix

Chaphasilor:
Shouldn't be hard. `_PlayerScreenContent` is a `ConsumerWidget` since the lyrics PR, so we just need to watch the Finanmp settings provider 😁  
I'll add that in a bit, and also take a look at a minimum size for the desktop window.

Chaphasilor:
Okay, player screen now updates whenever settings are changed.  
I've also added the `window_manager` package to set a minimum width and get some additional goodies, and while that works, it seems to cause issues if used on unsupported platforms (like Android or iOS). So we'll need to be careful to only use it on supported platforms. Some kind of wrapper/proxy class might be useful here.

Would be nice if you could test if it also works fine on other platforms

baarkerlounger:
@Chaphasilor do you have a release bundle for me to point at? If so I'll prepare a PR to add Finamp to Flathub. Currently I'm pointing [here](https://github.com/baarkerlounger/finamp/releases/tag/desktop-beta-flatpak) but obviously pointing at this repo would be preferable.

Chaphasilor:
@baarkerlounger right now we manually build all artifacts, but we could try setting up some GitHub Actions for it. I could try setting up a linux build environment for manual builds too.  
I'm guessing there's no way for Flathub to do they building when creating the flatpak?

Komodo5197:
It might make sense to re-tune the cutoff to displays >20% wider than they are tall, or something like that.  I probably wouldn’t put too much thought into it beyond that, because I imagine most people would use split screen, and everyone else can just choose a less stupid window size.

Chaphasilor:
@Komodo5197 how about only applying landscape mode if the height is less than X px? I'd like to only use landscape mode if the height is relatively low.

Chaphasilor:
I also now managed to use GitHub Actions to build artifacts for all platforms (Android, iOS, Windows, Linux, macOS). I believe for everything except Android, only a single architecture (probably x64, or in the case of macOS Apple-Silicon) is supported.

@baarkerlounger are the artifacts found [here](https://github.com/jmshrv/finamp/actions/runs/8988014836) of any use to you? It seems like they are always either up- or downloaded as zip, but they should contain the files you need. If only the download forces the zip, there might also be a way around that, to download the files directly.

Chaphasilor:
Yes, we could then use these release artifacts and attach them to a release, just like we currently do with APKs.  
I guess in this case we'll silently attach them to an upcoming release so you can try if everything works and then make your PR to Flathub, yes?

Another unrelated thing, on Windows I can only have one instance of Finamp running, because there's a lock on the Hive settings file. Sometimes after closing the app this lock is also kept open, seemingly because some background process didn't properly close.  
Is that similar on other platforms as well?

baarkerlounger:
@Chaphasilor that would be perfect, thanks!

Komodo5197:
@Chaphasilor Why?  If the window is 2:1 or something, no matter how big it is, I feel like portrait will look stupid.  Also, the portrait layout algorithm will probably start hiding control elements if the aspect ratio gets too wide, even with fairly large windows.

Chaphasilor:
Yeah I did notice the controls being shown even though there was enough space. But regarding the layout, I'd argue that stretched portrait mode actually looks better than the landscape mode with the huge gaps between the sections on the right side.  
But if there's no consensus, I'd say we wait with any major changes for now (aside from not hiding controls if there's enough space, maybe) until we receive some further user feedback.

Maxr1998:
> I also noticed that on Windows at least (but probably also on Linux), playback of FLAC files with 16 bit resolution and 48 kHz sampling rate doesn't sound great. There's some weird distortion going on, similar to what happens when changing the speed. On Android the track sounds much better.

Sounds like it's being (improperly) resampled. What audio settings do you have on your computer at the moment? Also, on Linux, at least pipewire (a more modern audio stack that's slowly replacing PulseAudio on a few distros) has a default sample rate of 48 kHz already, so that should be fine. I also didn't notice any issues when testing it a while back with 24/96 files.

Chaphasilor:
Yeah seems like it's a problem with my bluetooth headphones on Windows, they're locked to 44.1 kHz

@Komodo5197 I noticed when opening the app / unlocking the screen after the track has changed, a white cover is shown until the actual cover loaded. Seems like at least a blurhash should be shown instead, but the cover should actually already be prefetched...  
I'll try to reproduce this better.
Mind taking a look?

Komodo5197:
None of the currentAlbumImage paths were using blurhashes.  I've rectified that.  Regarding the prefetching, I believe that doesn't fire in the background due to none of the widgets building, so if you advanced more than three tracks this is expected.  If not, I don't know what's up.

Chaphasilor:
Yeah, it could be that I ran into the three tracks limit, which would explain why I wasn't able to reproduce it before.  
I've managed to reproduce it now, by waiting a few minutes and skipping ahead multiple tracks. Will try your changes in a bit!

We could probably move the watching of the metadata provider out of the now playing bar into some kind of widget-less service, unless that's a bad idea?

Komodo5197:
I don't know if watching providers outside of widgets or other providers is actually possible.  I'm also not sure if avoiding a slight load on reopen is worth continually fetching album covers in the background when they aren't being used.

Chaphasilor:
I'll see how it looks with the blurhash, but the blank screen definitely looks a bit jarring and unfinished...

Chaphasilor:
@jmshrv is the macOS version ready, or are there any fixes still needed? 

@Maxr1998 is transcoding working on Linux?

Tiefseetauchner:
When unfocusing, lyrics stop scrolling. This is probably good on mobile but I feel like it's not intended behavior on desktop
Thanks for the desktop version tho, by far the best jellyfin music client, even in it's early stages

Tiefseetauchner:
Also just noticed, it always loops the queue even if looping is disabled (linux), but that's not at all blocking I think, probably some MPV thing

jmshrv:
macOS needs a few little fixes, will hopefully do today

jmshrv:
Fixed a few more issues - the icon is now rounded correctly and the Hardened Runtime is set up, so we can now distribute release builds for macOS (i.e., local distribution, as well as App Store).

Only issue I've noticed now is that the app name isn't correct in the toolbar for some reason.

Screenshot 2024-05-13 at 13 32 13

Downloading also works, although I haven't tested custom paths

Chaphasilor:
> Only issue I've noticed now is that the app name isn't correct in the toolbar for some reason.

I guess it's using the name from pubspec.yaml?

Tiefseetauchner:
[Screencast_20240513_183509.webm](https://github.com/jmshrv/finamp/assets/48085877/8d1b3194-497a-4125-824e-95d4ed823d27)

Sorry for being such an evil beta tester 😈

Maxr1998:
> is transcoding working on Linux?

Yup, seems to work both for playback and downloads.

Maxr1998:
When I try to close the app, it seems to freeze for me.
This is what's (repeatedly) logged to the console:
```
Attempt:11 waiting for isolate vm-isolate to check in
Attempt:11 waiting for isolate InitializerIsolate._mainloop to check in
```

Moreover, ListenBrainz reporting still doesn't seem to work. It does report the playback stop event, but the position is set to 0, so it doesn't record the listen.

Chaphasilor:
@Maxr1998 was the freezing still happening with the latest commits? And if so, did it also happen before 4d40973?  
I'd rather not mess with the playback reporting at this point in fear of breaking Android and iOS, and it's probably an issue with https://github.com/Pato05/just_audio_media_kit anyway

jmshrv:
Flutter 3.22 has broken CI for some reason, I can build fine (at least for Linux)

Chaphasilor:
I'm also fighting with this. I could build fine on 3.19, upgrading to 3.22 broke it, downgrading back to 3.19 didn't fix it 🙃  
Still trying to figure out what is happening...

Chaphasilor:
@jmshrv could you check if the package name change actually fixed the casing in the macOS title bar? Seems like that was causing some build issues, and I've updated the package references for now. But if it doesn't help, I'd prefer to revert that.

The Windows build is still acting up. I've managed to get it to build on my machine with 3.22 now, but had a situation where `flutter run` would work (with a debug build), but `flutter build windows --debug` wouldn't (nor would release builds). A `flutter clean` and OS restart fixed that issue, but without modifying any tracked files, so I have no idea what's going on...

My `flutter doctor` looks like this:
```log
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.22.0, on Microsoft Windows [Version 10.0.19045.4291], locale en-US)
[✓] Windows Version (Installed version of Windows is version 10 or higher)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Chrome - develop for the web
[✓] Visual Studio - develop Windows apps (Visual Studio Build Tools 2019 16.11.1)
[✓] Android Studio (version 2023.1)
[✓] Android Studio (version 2022.1)
[✓] VS Code (version 1.89.1)
[✓] Connected device (3 available)
[✓] Network resources
```

Chaphasilor:
Okay I give up with the CI for now. We'll have to build it manually anyway, as long as that works it's fine. Maybe it'll work in a few days, maybe someone else has an idea...

Tiefseetauchner:
Evil beta tester here breaking your latest builds

[Screencast_20240514_183850.webm](https://github.com/jmshrv/finamp/assets/48085877/bfea06bf-e1f7-4cfa-b048-cbf2b49e852e)

Two main things here: 
- When opening and closing the sidebar, the whole app flickers
- When clicking the player bar at the bottom after it loaded the previous queue but you didn't change song yet, it core dumps.

Linux, built from latest, nothing in the logs as far as I can tell, but for completion sake
[finamp-logs.txt](https://github.com/jmshrv/finamp/files/15311578/finamp-logs.txt)

Maxr1998:
> was the freezing still happening with the latest commits? And if so, did it also happen before [4d40973](https://github.com/jmshrv/finamp/commit/4d40973f589e2ce714620e40e05d8fafc8cfd12c)?

I was on 5f3a6e6a776eb36dcc4d132d63ac3a6e18daff18, so after that change. Let me test the previous one.

> I'd rather not mess with the playback reporting at this point in fear of breaking Android and iOS, and it's probably an issue with https://github.com/Pato05/just_audio_media_kit anyway

Yeah, understandable. Might need to do some debugging there.

Chaphasilor:
Sweet, @Komodo5197 fixed the Windows build issues 😁

Now about the remaining bugs:
- [ ] Windows: Transcoding doesn't work
- [ ] Linux: Segfault when opening the player screen after restore
- [ ] Linux: Flickering when toggling menu drawer
- [ ] MacOS: Lower-case app name in title bar
- [ ] Linux (and probably Windows): Playback reporting is a bit broken

Maxr1998:
4216aa0773b46fe0dd15bc0091aad1de23c283c4 seems to have broken the build on Linux for me, since the Flutter tool distributed in AUR is still 3.19.
I think it broke for you earlier because [the caret syntax](https://dart.dev/tools/pub/dependencies#caret-syntax) behaves differently for versions below 1.0. Thus, `^0.18.1` didn't allow version `0.19.0`. We could, however, change it to something more explicit like `>=0.18.1 <0.20.0`, then it should work across more Flutter versions.

jmshrv:
That sounds good, as far as I know `intl` 0.19 hasn't changed anything for us, and is just an awkward SDK thing.

The Linux version works really nicely (especially with , the only thing that is vaguely annoying is that Flutter uses massive GTK window bars by default, which looks really weird on KDE, but I'm not sure if that's easily configurable

Finamp with a big window bar

jmshrv:
> MacOS: Lower-case app name in title bar

Will check this in the morning (aka ~2pm) :)

Chaphasilor:
@Komodo5197 pointed out there's some issue with extracting accent colors on Linux. I can see that in the screenshot @jmshrv sent (did you also disable blurred backgrounds there too, or is that a bug with the blue background tint?), but it seems to be working fine for @Tiefseetauchner.  
I believe James is using Asahi Linux, and Lena is using Arch, both with KDE? Is there maybe some external dependency/lib needed to get this to work properly? Or did we break it at some point?

jmshrv:
Can confirm that the macOS window title is correct now :)

Chaphasilor:
@Maxr1998 do you have the issues with ListenBrainz also when using the desktop branch on mobile? Or is it really related to media-kit/mpv?

Maxr1998:
> do you have the issues with ListenBrainz also when using the desktop branch on mobile? Or is it really related to media-kit/mpv?
Just tested on mobile and it works perfectly fine with this branch. So, it's indeed mpv/media-kit/just-audio.

Chaphasilor:
Okay, thanks everyone for helping with this! You can now start pointing to the `redesign` branch if you build from source. With the next release, I'll attach the Linux app bundle as a release asset. If you need anything else, let me know!

Chaphasilor:
@baarkerlounger the release with the needed bundle is now available. Does that work for you? :)

baarkerlounger:
@Chaphasilor yes I think so. Added a PR [here](https://github.com/flathub/flathub/pull/5297) to add a flatpak build to Flathub. Could you comment on it when you get a chance?

PR #714, opened 2024-05-05
[Desktop Beta] Cleanup scrollbars

Komodo5197:
Due to an error when scrolling on desktop, I previously removed the Scrollbar widgets from many lists.  However, it seems that while scrollbars are still shown on desktop, they are not on mobile.  This corrects that and shows scrollbars on mobile while also cleaning up the remaining Scrollbar widgets with no explicit ScrollController.  Also, I noticed some blinking on the lyrics screen slide transition, so I fixed that.

### DIFF
--- a/lib/components/AlbumScreen/preset_chip.dart
+++ b/lib/components/AlbumScreen/preset_chip.dart
@@ -1,6 +1,8 @@
 import 'dart:math';
-import 'package:flutter/material.dart';
+import 'dart:ui';
+
 import 'package:finamp/services/queue_service.dart';
+import 'package:flutter/material.dart';
 import 'package:get_it/get_it.dart';
 
 const _radius = Radius.circular(4);
@@ -14,7 +16,7 @@ enum PresetTypes {
 
 class PresetChips extends StatefulWidget {
   const PresetChips({
-    Key? key,
+    super.key,
     required this.type,
     required this.values,
     required this.activeValue,
@@ -23,7 +25,7 @@ class PresetChips extends StatefulWidget {
     this.onPresetSelected,
     this.chipWidth = 64.0,
     this.chipHeight = 44.0,
-  }) : super(key: key);
+  });
 
   final PresetTypes type;
 
@@ -41,11 +43,10 @@ class PresetChips extends StatefulWidget {
 
 class _PresetChipsState extends State<PresetChips> {
   final _queueService = GetIt.instance<QueueService>();
-  final _controller = ScrollController();
-  bool scrolledAlready = false;
+  ScrollController? _controller;
 
   void scrollToActivePreset(double currentValue, double maxWidth) {
-    if (!_controller.hasClients) return;
+    if (_controller != null && !_controller!.hasClients) return;
     var offset = widget.chipWidth * widget.values.indexOf(currentValue) +
         widget.chipWidth / 2 -
         maxWidth / 2 -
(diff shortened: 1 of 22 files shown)


Chaphasilor:
Changes all look good. Thanks for the lyrics transition fix!

I noticed (as someone already noted before) that trying to favorite a track from the player screen throws an assertion error:
image

Initially I tried to repro this on Windows, but couldn't (and still can't). But it's happening on Android for me. Would you mind taking a quick look?

Komodo5197:
The issue should be fixed.  I forgot to update the resync commands with the new FinampCollection format.  I'm not sure why it doesn't reproduce on windows for you, it does for me. Are you not testing a debug build?

Chaphasilor:
Ahh yes, the `msix:create` command produces a release build which I'm aleays installing to test. That makes sense, and thanks for the fix!

If we already have these asserts, it would be nice to at some point add some logging for release builds as well, maybe by creating our own custom assert. Just a thought...

 I'll merge it when I wake up again...

Chaphasilor:
I also just noticed the song menu isn't taking up the whole width on my phone, while the queue panel is. I guess I messed something up during the merge 😅

Komodo5197:
I intentionally narrowed the song menu on desktop because the drag to close gesture isn't particularly intuitive.  This allows you to easily click outside the menu to close it.  I thought I had it at full width on mobile, but I guess not, so I've fixed that.  I also noticed that the playback speed widget seems to have been completely broken by the merge, so I fixed that too.

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -54,24 +55,21 @@ class _PresetChipsState extends State<PresetChips> {
...
 
-    _controller.animateTo(
-      offset,
-      duration: Duration(milliseconds: 350),
-      curve: Curves.easeOutCubic,
-    );
+    if (_controller == null) {
+      _controller = ScrollController(initialScrollOffset: offset);
+    } else {
+      _controller!.animateTo(
```
Chaphasilor, line 61: why is this an if/else and not just a `_controller =?? ScrollController(initialScrollOffset: offset);`  
Don't we wanna scroll if there's no scroll controller yet?

lib/components/AlbumScreen/preset_chip.dart
```diff
@@ -54,24 +55,21 @@ class _PresetChipsState extends State<PresetChips> {
...
+        offset,
+        duration: const Duration(milliseconds: 500),
```
Chaphasilor, line 63: Any particular reason why you changed the duration? Feels a bit slow now with the cubic curve.

Chaphasilor (commented):
thanks for the fix. just have two small questions, and I think we're not in a hurry.

lib/components/AlbumScreen/preset_chip.dart
Komodo5197: No, if _controller is null we're doing the first build of this widget so we don't want to animate the scroll, we want the offset already applied before the widget shows.

lib/components/AlbumScreen/preset_chip.dart
Komodo5197: I slowed this down to make things a bit easier to see during testing and just forgot to revert it.

Chaphasilor:
Awesome, thanks for the changes!

PR #719, opened 2024-05-09
[Desktop Beta] More player screen layout tweaks

Komodo5197:
Tweaks to player screen layout code to improve appearance with large windows.  Alters portrait vs landscape mode criteria on desktop and caps maximum control area size.

### DIFF
--- a/lib/components/AlbumScreen/preset_chip.dart
+++ b/lib/components/AlbumScreen/preset_chip.dart
@@ -60,7 +60,7 @@ class _PresetChipsState extends State<PresetChips> {
     } else {
       _controller!.animateTo(
         offset,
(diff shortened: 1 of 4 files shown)


Chaphasilor:
Nice, everything working nicely!  
Seeing your drag-scrolling change, I noticed there are issues on desktop with the lyrics screen when *not* drag-scrolling. When scrolling with a touchpad or mouse wheel, auto scrolling will not be disabled and the FAB won't be shown, even though it should, naturally.  
With the whole auto-scrolling stuff I didn't manage to make this work on mobile aside from listening for the touch drag, but maybe you have a better idea that also works for desktop?

PR #724, opened 2024-05-11
Renamed "Song" to "Track" in app_en.arb

lukaslindnermusic:
For now, I only renamed the user-facing strings.

I noticed that some variables and classes use "song" but some use "track" already, is this intentional?  Maybe I will rename the variables and classes as well in the future when I got familiar enough with the language and codebase.

If I missed one of the user-facing strings or accidentally renamed a variable already, please let me know 😊

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -149,7 +149,7 @@
   },
   "unknownName": "Unknown Name",
   "@unknownName": {},
-  "songs": "Songs",
+  "songs": "Tracks",
   "@songs": {},
   "albums": "Albums",
   "@albums": {},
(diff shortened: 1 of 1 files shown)


Chaphasilor:
That looks good, thanks for the PR!  
Everything seems to be converted correctly, I didn't spot any missing conversion nor a changed variable name.

And the mixed naming scheme for classes and widgets wasn't intentional, some were based on preference, others for historic reasons. With the newer ones I tried to use track instead of song, but wasn't totally consistent either.

The next step for now would be also renaming the variable names for the translations, including the placeholders in the actual translation string (this file), and also the string identifiers used in the code :)

PR #729, opened 2024-05-14
Add desktop entry m4 template

Maxr1998:
Instead of making packagers ship this themselves, this should be a predefined template in the upstream repo.

### DIFF
--- a/assets/finamp.desktop.m4
+++ b/assets/finamp.desktop.m4
@@ -1,3 +1,4 @@
+dnl Desktop file entry for Finamp on Linux. Use m4 to generate the final file.
 [Desktop Entry]
 Type=Application
 Name=Finamp
@@ -6,4 +7,4 @@ Icon=finamp
 Exec=__INSTALL_PATH__/finamp
 Terminal=false
 Categories=AudioVideo;Audio;Player;Music;
-Comment=A Jellyfin music client for mobile and desktop
+Comment=An open source Jellyfin music player
--- a/assets/finamp.desktop.m4
+++ b/assets/finamp.desktop.m4
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Name=Finamp
+GenericName=Music Player
+Icon=finamp
+Exec=__INSTALL_PATH__/finamp
+Terminal=false
+Categories=AudioVideo;Audio;Player;Music;
+Comment=A Jellyfin music client for mobile and desktop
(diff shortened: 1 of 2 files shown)


assets/finamp.desktop.m4
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Name=Finamp
+GenericName=Music Player
+Icon=finamp
+Exec=__INSTALL_PATH__/finamp
+Terminal=false
+Categories=AudioVideo;Audio;Player;Music;
+Comment=A Jellyfin music client for mobile and desktop
```
Chaphasilor, line 9: please use the description from pubspec.yaml:

> An open source Jellyfin music player

assets/finamp.desktop.m4
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Name=Finamp
+GenericName=Music Player
+Icon=finamp
```
Chaphasilor, line 5: does that works already, or do maintainers need to adapt that to their package?

assets/finamp.desktop.m4
```diff
@@ -0,0 +1,9 @@
...
+Exec=__INSTALL_PATH__/finamp
```
Chaphasilor, line 6: is this a variable that needs to be changed? if so, a comment (if possible) would be nice.

Maxr1998:
(this will be used by the AUR package, which currently ships this file itself)

Chaphasilor (commented):
Thanks!

Maybe you could also throw in a section into CONTRIBUTING.md to mention that this template exists?

assets/finamp.desktop.m4
Maxr1998: As long as the icons are also installed according to the [XDG icon theme spec][spec], yes. 

[spec]: https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

assets/finamp.desktop.m4
Maxr1998: It does, it's meant to be templated with m4. I'll add a note to CONTRIBUTING.md, that should be ok.

Maxr1998: Note added, I think this should be enough for people packaging the app for their distribution.

Chaphasilor:
perfect, thanks!

PR #730, opened 2024-05-15
[Desktop Beta] Windows media-kit fixes

Komodo5197:
Attempts to dispose player instance when window is closed on windows/linux.  This seems to fix the issue where a you couldn't reopen finamp due to the Hive file locks still being held.

Fixes transcoding on windows by switching to a newer version of MPV via a fork of media_kit_libs_windows_audio.  This just grabs the latest release from https://github.com/zhongfly/mpv-winbuild, so the .dll is way bigger due to having all the video stuff.  Hopefully media-kit releases a new update with a later MPV one of these days.

Also I found an application name field in the Macos files, so I capitalized it.  Hopefully that does something positive.


Chaphasilor:
Nice, I'll give that a try later! You mentioned in a comment that the download link will expire, so that would entail creating a new link in your fork and then upgrading the package rederence until media-kit catches up?

Chaphasilor:
I've also been trying out the library image caching some more, and noticed that the cached images don't seem to be used in the queue list or the songs tab. I assume this is because only album images are fetched, but no track images?

Given that is most cases, tracks do use an album cover, it might be possible to search by blurhash even for tracks, in hopes of finding a matching album cover. But maybe there's a reason why this isn't happening/working?

Komodo5197:
Currently, I have it set to always grab the latest release.  This means it shouldn’t expire, but every time you do a clean build for windows you might end up with a newer MPV version.  I did this because I didn't want to keep updating the fork, and also want to lock Finamp to a specific commit rather than a branch.  I can go back to the fixed links that expire for more consistency, or maybe just rehost one of the releases and use that.

The downloaded/cached images should be searched by blurhash already, so at least some songs should be using the cache.  However, in my testing I noticed that a lot of the time the song blurhash was not actually 100% identical to the album blurhash, even if they looked the same.  This is probably down to some tracks resolution differences or something.

Chaphasilor:
Okay, I don't feel strong either way about how to handle the links. Do it however you prefer, I just need to know how/when to update.

Yeah I noticed some songs working and others not, but assumed that was due to having some of them actually downloaded. I guess blurhash simply isn't ideal for that usecase.

Chaphasilor:
Okay, I can confirm transcoded playback works now. Gapless isn't working in that case sadly, but it's definitely better than before.  
Anything else you want to get in, or is this ready to merge? :)

Komodo5197:
I did some testing with Linux, so I've thrown those changes in here too.  There appears to be something up with the native engine's image handling there, because PaletteGenerator just isn't providing any theme colors and I'm pretty sure that the CachedPaint widget for the blurred backgrounds was causing segfaults.  I've hopefully fixed the segfaults, but I don't have any ideas for getting the theme colors working.  The blinking app when closing the sidebar didn't occur for me.

I think everything's pretty much how I want it.  I'm leaving the link to always point to the latest MPV so we shouldn't have to do any updating, and hopefully there isn't a dud build uploaded at some point.  This should be good to merge.

Chaphasilor:
Alright, I'll merge this then. Then we can get some more feedback by others. I could also try installing the AUR version on my Steam Deck to do some additional testing, that's the only Linux device with a desktop environment that I have at the moment 😅

PR #734, opened 2024-05-17
[Redesign] Download orphan tracks with library

Komodo5197:
Download orphan tracks which are not contained in any album when downloading library.  Actual functionality not yet tested as I don't have any orphan tracks.

Also forces full metadata refresh during download repair, adds some comments, cleans up unneeded childCount references, fixes showing deleting instead of syncing on subitems on the downloads screen, update playlist image on sync, and properly remove syncFailed state.

### DIFF
--- a/lib/components/DownloadsScreen/item_file_size.dart
+++ b/lib/components/DownloadsScreen/item_file_size.dart
@@ -22,10 +22,11 @@ class ItemFileSize extends ConsumerWidget {
         ref.watch(isarDownloader.itemProvider(stub).future).then((item) {
       switch (item?.state) {
         case DownloadItemState.notDownloaded:
-          if (isarDownloader.getStatus(item!, null).isRequired) {
-            return Future.value(syncingText);
-          } else {
+          if (isarDownloader.getStatus(item!, null) ==
+              DownloadItemStatus.notNeeded) {
             return Future.value(deletingText);
+          } else {
+            return Future.value(syncingText);
           }
         case DownloadItemState.syncFailed:
           return Future.value(syncingText);
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -1,11 +1,10 @@
 import 'dart:async';
 import 'dart:math';
 
-import 'package:collection/collection.dart';
 import 'package:Finamp/services/finamp_user_helper.dart';
+import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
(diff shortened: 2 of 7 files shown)


Chaphasilor:
Whew. That's a lot of improvements!  
How much testing would you want me to do?

As for orphaned downloads, I just dumped a few tracks into the root of my library directory and rescanned, in case you want to give it a try yourself.

Komodo5197:
The things to verify are just generally verifying that downloading/syncing/deleting are still working normally, verifying that the orphans are downloading now, and checking if updating an image actually works right.  I was previously under the impression that changing an image would replace the imageId, but that isn't the case at all, so some verification that they now properly change after a sync would be good.  Playlists will update on a quick sync, all other items may take a full sync/downloads repair, depending on the exact dependencies.

Chaphasilor:
Alright, I'll give it a test now!

Chaphasilor:
- Downloading and syncing orphans works perfectly, no duplicates were created either
- Playlist properly resync tracks after a restart in online mode
- Playlists images are properly updated during a regular sync. It actually also worked for artists without forcing a sync or repair, just restarting (with prefer quick syncs enabled). For track image changes, a repair was needed, but that did work fine

I noticed that after deleting an album from the device in offline mode, the albums tab kept loading until I toggled the "fully downloaded" filter. Not sure if that's new or also was the case before. I had marked some tracks of that album to always stay on device, and deleted from the album screen, if it matters

Komodo5197:
The reason the artist image refreshed was because we were requiring mediaSources/mediaStreams on all downloaded items, but those seem to always be null on collections.  I've changed it to only require those on songs.

The music screen locked up because we called pagingController.refresh() on offline deletes instead of the correct _refresh() which updates the deduplication state.  This was an existing issue.

The cached covers, and any other downloaded image, will be properly ignored due to the blurHash change of the item.  If blurHashes are broken on the server they will still be used, but that doesn't seem avoidable and is not a new issue.

Chaphasilor:
Okay, perfect! So the issue with the artists and other types was that we always tried to re-fetch the BaseItemDto because the existing ones were missing the media sources/streams, and we want to avoid that? Because refreshing the image wouldn't be so bad after all (but it's fine that it doesn't do that anymore now) ^^

Thanks for the PR!

Komodo5197:
Yeah, refreshing the image would be nice, but the amount of extra requests can really pile up.  And we can't do it for albums due to skipping them completely when quicksyncing, and I believe quicksyncing would also skip any info-only artist due to skipping the tracks that point to them, so trying to refresh things like artists every time would be very inconsistent and doesn't seem like a great idea.

PR #735, opened 2024-05-19
[Redesign] Offline favorites improvments

Komodo5197:
Uses an info download of the favorites finampCollection to provide more accurate and up-to-date favorite information when online.  Switches favorites provider to use this when appropriate, and re-enables favorite filter on music screen.

Also, adds an option to download info for all playlists, allowing partially downloaded playlists to be shown when offline.


lib/models/finamp_models.dart
```diff
@@ -1209,20 +1221,24 @@ enum DownloadItemStatus {
...
+  musicVideo("MusicVideo", false, []);
+
+  const BaseItemDtoType(this.idString, this.expectChanges, this.childTypes);
 
   final String? idString;
   final bool expectChanges;
+  final List<BaseItemDtoType>? childTypes;
+
+  bool get expectChangesInChildren =>
+      childTypes?.any((x) => x.expectChanges) ?? true;
```
Chaphasilor, line 1241: Right now everything that has children of type album also has songs, which makes sense. I'm wondering if this could be done implicitly, where we check if the child type also has children that might change.  
This is just an idea to make supporting downloads of collections in the future easier, which can contain basically and child type. I am however not sure if this would make sense wrt the code, because I don't know if there's any benefit in recursively modelling the contained children and their types, or if just using the type `folder` (which always expects changes) is enough already.

lib/models/jellyfin_models.dart
```diff
@@ -2268,7 +2270,13 @@ class BaseItemDto with RunTimeTickDuration {
...
   factory BaseItemDto.fromJson(Map<String, dynamic> json) =>
       _$BaseItemDtoFromJson(json);
-  Map<String, dynamic> toJson() => _$BaseItemDtoToJson(this);
+  Map<String, dynamic> toJson({bool setOffline = true}) {
+    var json = _$BaseItemDtoToJson(this);
+    if (setOffline) {
+      json["FinampOffline"] = true;
+    }
+    return json;
+  }
```
Chaphasilor, line 2279: Is lacing the `BaseItemDto` with Finamp state information a good idea? Sure we have custom getters, but introducing a custom property which is also included in the JSON (and therefore potentially sent to the server at some point) seems drastic.

lib/services/favorite_provider.dart
```diff
@@ -5,43 +5,78 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
...
 import '../components/global_snackbar.dart';
 import '../models/jellyfin_models.dart';
+import 'downloads_service.dart';
 import 'feedback_helper.dart';
 import 'finamp_settings_helper.dart';
 import 'jellyfin_api_helper.dart';
 
 part 'favorite_provider.g.dart';
 
 /// All DefaultValues should be considered equal
```
Chaphasilor, line 15: this comment is outdated

lib/services/favorite_provider.dart
```diff
@@ -5,43 +5,78 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
...
 }
 
 @riverpod
 class IsFavorite extends _$IsFavorite {
+  bool _changed = false;
+  Future<void>? _initializing;
+
   @override
   // Because DefaultValue is always equal and we never invalidate, this should only
   // be called once per itemId and all other DefaultValues should be ignored
```
Chaphasilor, line 37: also outdated I believe

lib/models/jellyfin_models.dart
Chaphasilor: As far as I can tell, the property is used to mark items where we prefer the favorite data from the server? Maybe tracking this in a separate list or map would be better, especially since lookup times don't appear to be critical here as it happens in the background anyway?

lib/screens/downloads_settings_screen.dart
```diff
@@ -37,52 +37,51 @@ class DownloadsSettingsScreen extends StatelessWidget {
           ),
           if (Platform.isIOS || Platform.isAndroid) const RequireWifiSwitch(),
           const ShowPlaylistSongsSwitch(),
+          const SyncFavoritesSwitch(),
+          ListTile(
+            // TODO real UI for this
+            title: const Text("Show all playlists offline"),
+            subtitle: const Text(
+                "Sync metadata for all playlists to show partially downloaded playlists offline"),
```
Chaphasilor, line 45: Is there a particular reason you didn't move this to the localizations yet, or did you just miss it?

lib/screens/downloads_settings_screen.dart
```diff
@@ -121,6 +120,35 @@ class RequireWifiSwitch extends StatelessWidget {
...
+          subtitle:
+              Text(AppLocalizations.of(context)!.trackOfflineFavoritesSubtitle),
+          value: syncFavorites ?? false,
+          onChanged: syncFavorites == null
+              ? null
+              : (value) {
+                  FinampSettings finampSettingsTemp =
+                      box.get("FinampSettings")!;
+                  finampSettingsTemp.trackOfflineFavorites = value;
+                  box.put("FinampSettings", finampSettingsTemp);
```
Chaphasilor, line 144: I noticed you not using the `FinampSettingsHelper` recently. Is that on purpose? If so, I'm interested in your reasons.  
To my understanding the helper just makes it easier to change settings from anywhere in the app, and I don't think there's a downside to that, but I might be wrong of course.

Chaphasilor:
Thanks for taking the time to add this!  
I'll take a look at it today and try to do some testing involving other clients as well to see how it behaves.

Since we now have the ability to cache playlist metadata, I think we could use that (if available) to pre-select the `ToggleableListTile`s in the playlist actions menu? The metadata should usually be pretty up-to-date, and if someone modified a playlist on another device, sees the old status in Finamp, and then updates the state again, it shouldn't have any effect anyway (since duplicates in playlists are disabled by default, and removing a track "twice" also doesn't hurt).

Also, I noticed your solution for getting a `BuildContext` within a regular class using the `GlobalSnackbar`, that's really clever. I assume I could also use this for the queue service and the Android Auto stuff to provide localized strings to the OS itself?

lib/screens/downloads_settings_screen.dart
```diff
@@ -37,52 +37,51 @@ class DownloadsSettingsScreen extends StatelessWidget {
...
+          ListTile(
+            // TODO real UI for this
+            title: const Text("Show all playlists offline"),
+            subtitle: const Text(
+                "Sync metadata for all playlists to show partially downloaded playlists offline"),
+            trailing: DownloadButton(
+                infoOnly: true,
+                item: DownloadStub.fromFinampCollection(
+                    FinampCollection(type: FinampCollectionType.allPlaylists))),
+          ),
```
Chaphasilor, line 50: This prompts for transcoding options (and probably also download location), which doesn't make sense

Chaphasilor:
There is no collection item for the playlist metadata:

Chaphasilor:
Just tried liking a track in Symfonium and then seeing how Finamp handles it.

If Finamp is in *online* mode and started (or resumed after more than 4h, I assume), a sync is automatically started and favorites are synced properly.

If Finamp is in *offline* mode and started, sync is briefly attempted, but then marked as completed. Switching to online mode will not restart a sync in this case. This should probably be changed, either by always syncing when switching from offline to online mode, or "queueing up" the sync after startup so that it is performed when offlinr mode is turned off for the first time after a restart.

Playing a downloaded track in online mode will show the correct favorite status, and that will be *mostly* preserved when switching to offline mode. The track will be shown as liked, but the favorite-only filter will not find it. Changing to a different track and refreshing the current view will then remove the favorite status again.

Manually triggering a sync will always update the favorite metadata, and will also immediately download the favorites if all favorites are downloaded.

Here are some logs:

([finamp-logs_favorite-sync-issues.txt](https://github.com/jmshrv/finamp/files/15370113/finamp-logs_favorite-sync-issues.txt)), I wasn't sure if all of this was contained in the other logs.

[finamp-logs_favorite-sync-issues-2.txt](https://github.com/jmshrv/finamp/files/15370115/finamp-logs_favorite-sync-issues-2.txt)


<details>

<summary>
And some screenshots
</summary>























</details>

lib/models/finamp_models.dart
Komodo5197: The only place this list is used right now is the expectChangesInChildren bool.  I almost just hardwired in all the values for that instead of this list, and I might still change to that method, because calculating them like this sort of shows the logic, but sort of doesn't.  It's currently false for every type of collection except folder, which we don't even use, so it's a bit speculative.

lib/models/jellyfin_models.dart
Komodo5197: Yeah, it's a bit hacky, but I couldn't figure out any other reasonable way to do this.  Note that it isn't id associated - that is, we might have two BaseItemDtos with identical IDs for the same item, but one should be trusted and one should not be.  This makes any sort of separate lookup difficult.  The attribute needs to be included in the JSON to survive being serialized by DownloadItems and FinampQueueItems.  As of right now we never send a BaseItemDto to the server, and I don't really see why we ever would, only ids or other specific attributes, so I'm hoping that never comes up as an issue.

lib/screens/downloads_settings_screen.dart
Komodo5197: It just seems like a needless abstraction.  Why would we ever want to change this setting from outside this widget?  This is basically the same logic as why I prefer to not break the settings switch widgets out into separate files.  The breakout is especially pointless because most of the settings end up directly using the settings box anyway, so we don't even get to abstract that aspect away to a higher level interface with the submethod, we just spread our box manipulation into multiple methods for no reason.

lib/screens/downloads_settings_screen.dart
Komodo5197: Actually, we need a download location for the images this generates.  I'll make it skip transcoding prompts though.

lib/screens/downloads_settings_screen.dart
Komodo5197: None of the DownloadButton based settings have been moved to localizations yet.  I should probably do that.

Komodo5197:
I can add info items to the downloads screen for these two collections, but it'll be a bit weird.  The favorites collection will keep getting re-generated to match the setting, even if you delete it.  The all playlists collection will always show a zero file size because we only follow require links, so all the info linked images don't count.

Using the offline data to approximate playlist inclusions for the playlist menu is a good idea, I'll try to add it.  You should be able to use the same method I used to get a BuildContext elsewhere, although it should probably only be a last resort, and you want to watch out for accidentally using it during early initialization and getting null.

Regarding the behavior in the face of an external like, that all seems like mostly correct behavior.  The exception is not queuing a sync when started offline, which I've fixed.  Falling back to the offline known status when switching to offline mode is expected, with a slight holdover while favoriteProvider created online still exists.  This would behave cleaner if the offline favorites were up-to-date, which the queued sync from offline to online mode should help with.

Chaphasilor:
No need to show the favorites download collection, that's a toggle in settings. But the playlist stuff should be shown, even if the size is zero. It shouldn't be zero though, since the playlist covers are downloaded/required by it?

I'll be sure to not use a null check when using the global context. I can always fall back to the current solution with untranslated strings, but in most scenarios it should work.

Yeah, I think if a sync is started when leavj
ing offline mode, everything should work out just fine.  
Maybe invalidating the favorite provider when toggling offline mode (same as we do with the metadata provider) is a good idea though, that way it's more obvious to the user that there is stale data. The favorite status simply disappearing at some seemingly random point could be confusing, same as the favorite filter not showing the track, even if the track is shown as a favorite in the player screen / mini player.

Thanks for the fixes and playlist improvements. Really looking forward to testing that tomorrow!

lib/models/jellyfin_models.dart
Chaphasilor: I'm not sure if editing metadata would require sending BaseItemDtos. If not, it should be fine.

But what if we add the needed metadata to the DownloadItem and MediaItem directly? Or is that too complicated/convoluted?

Komodo5197:
I managed to get info linked items included in file sizes in a way that hopefully works - tell me if all the sizes seem to be still correct for you.  I ended up turning the playlist metadata into a fully separate FinampCollection type because trying to make most of the stuff work on info linked items was getting too messy, so now the only info linked thing is the hidden favorites metadata collection.  There is an issue where none of the syncing/deleting/updating file size on change logic works because the item never changes state.  This is because the playlists it points to are never considered downloaded, because we do not, in fact, download them.  I'm not sure this is fixable, but I've forced it to be marked complete so at least you can see the current file size instead of 'syncing'.

lib/screens/downloads_settings_screen.dart
Chaphasilor: It's still/again asking for transcoding settings with the latest commits ^^

lib/models/jellyfin_models.dart
Komodo5197: It gets messy and convoluted fast, because we pass BaseItemDtos around all over the place.  Tracking their origin externally would be pretty hard.  If we somehow end up sending BaseItemDtos to the server, we can just make sure to strip this field out first.

lib/screens/downloads_settings_screen.dart
Chaphasilor: Hmm. I agree that for most settings it doesn't make a whole lot of sense. For some (like playback speed or potentially offline mode) it would make more sense, but I guess in that case we can just add a method to the helper.  
Would some abstraction that removes the need to always duplicate the box handling make sense? So that it's a generic `updateSetting(String settingKey, dynamic value)` for everything?

lib/models/jellyfin_models.dart
Chaphasilor: Okay, let's leave it like that for now.  
FYI there are also plans to migrate Finamp to using https://github.com/TentacleOrg/Tentacle (which is maintained by Kara from JellyBook, and meant as a collaboration between [JellyBook](https://github.com/JellyBookOrg/JellyBook), [Jellyflix](https://github.com/jellyflix-app/jellyflix) and Finamp) instead of our current manually-maintained API client. For that we will probably need to find a compromise between adding additional features (like custom getters and setters) to tentacle, and moving some of our custom stuff (like this BaseItemDto modification) to another abstraction layer in Finamp.

Chaphasilor:
The new playlist metadata on the playlist actions menu works well, and newly-created playlist (e.g. added by a Jellyfin plugin) still show up with the dashed icon like they should. I debated using the metadata to speed up loading of playlists, but in the case of new, unsynced playlists this would either cause a layout shift or mess up the order of playlists, so I think leaving it like this is best for now.

The queued sync after leaving offline mode also seems to work.

I guess we can get this merged then?

Unrelated, but another thing I noticed:  
When library images are cached, a quick sync seems to also sync all nodes for which there are cached images. I guess this is done to detect changed images, and there's no way to sync less nodes?

lib/screens/downloads_settings_screen.dart
Komodo5197: Yeah, more abstractions in the settings would be nice to reduce box handling and/or widget duplication, but I haven't been able to think of a workable way to implement it.

Komodo5197:
Yeah, trying to steal downloaded metadata in more ways than this seems messy.  I think this should be good to merge.

I believe detecting changed images doesn't actually involve syncing the nodes in this case.  They just sync because finampCollections don't have the logic to not add unchanged children to the sync queue like normal collections do with the expectChangesInChildren logic.  The nodes still generally skip themselves when processed because unless an image changed they are already images in state complete, they just have to run through the sync queue to do it.  It doesn't matter too much, pretty much only albums have the volume for this optimization to matter.

Chaphasilor:
That sounds good. If any of the new stuff causes issues, we won't know it before releasing it to users anyway ^^

PR #740, opened 2024-05-23
[Redesign] Fix transcoding prompt

Komodo5197:
Fix transcoding prompt logic.


Chaphasilor:
Oh, woops. Do I understand it correctly that currently the prompt will always be shown for most collections, even if set to "always"?

Chaphasilor:
Either way, thanks for the fix! :D

PR #741, opened 2024-05-23
Restore monochrome icon on Android

Maxr1998:
Removed while regenerating icons in 37897816ca0484d7c47388275bbdf35b73b6d469.


Chaphasilor:
I had a hunch that the regeneration broke it. Thanks for fixing it! <3

Will this happen again the next time icons are generated, or is creating that file manual anyway? 🤔

Maxr1998:
Not sure, since I don't know how they're generated in the first place 😅

PR #747, opened 2024-05-24
[Redesign] Fix syncing legacy FinampCollections.

Komodo5197:
Fixes an issue where a null check error is thrown while syncing older finampCollections that do not have a FinampCollection object attached.  I haven't actually tested it with a legacy item yet.

### DIFF
--- a/lib/services/metadata_provider.dart
+++ b/lib/services/metadata_provider.dart
@@ -96,11 +96,15 @@ final AutoDisposeFutureProviderFamily<MetadataProvider?, MetadataRequest>
           ? profile?.stereoBitrate
           : downloadItem.baseItem!.mediaSources?.first.bitrate;
 
-      // We cannot create a fully accurate MediaStream for a transcoded item, so
-      // just return an empty list.
+      // We cannot create accurate MediaStreams for a transcoded item,so
+      // just return the lyrics stream, as those are not affected and will not
+      // be shown if the mediaStream is not present
       List<MediaStream> mediaStream =
           profile?.codec != FinampTranscodingCodec.original
-              ? []
+              ? downloadItem.baseItem!.mediaStreams
+                      ?.where((x) => x.type == "Lyrics")
+                      .toList() ??
+                  []
               : downloadItem.baseItem!.mediaStreams ?? [];
 
       localPlaybackInfo = MediaSourceInfo(
@@ -151,7 +155,12 @@ final AutoDisposeFutureProviderFamily<MetadataProvider?, MetadataRequest>
     if (localPlaybackInfo != null && playbackInfo != null) {
       playbackInfo.protocol = localPlaybackInfo.protocol;
       playbackInfo.bitrate = localPlaybackInfo.bitrate;
-      playbackInfo.mediaStreams = localPlaybackInfo.mediaStreams;
+      // Use lyrics mediastream from online item, but take all other streams
+      // from downloaded item
+      playbackInfo.mediaStreams =
+          playbackInfo.mediaStreams.where((x) => x.type == "Lyric").toList();
+      playbackInfo.mediaStreams.addAll(
+          localPlaybackInfo.mediaStreams.where((x) => x.type != "Lyric"));
       playbackInfo.container = localPlaybackInfo.container;
       playbackInfo.size = localPlaybackInfo.size;
     }
--- a/lib/services/metadata_provider.dart
+++ b/lib/services/metadata_provider.dart
@@ -40,6 +40,9 @@ class MetadataRequest {
       item.id, queueItem?.id, includeLyrics, checkIfSpeedControlNeeded);
 }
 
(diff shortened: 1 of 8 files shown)


Chaphasilor:
I've had this error pop up a couple times with my main installation, but I also can't test if that fixes it because it's signed by James.

Do you want me to go back to an older version and test the upgrade, or are you confident enough that we can just merge this?

Komodo5197:
Some testing would be nice, but I’m pretty confident in this fix, so you can just move ahead if you want.  It’s basically just pulling the fix I actually tested in the desktop branch out from getfinampcollectionchildren to the Finampcollection property so that the new code from the offline favorites is also using it.

Chaphasilor:
Okay, I'll see what I can do. Is going back to 0.9.6 and downloading a few favorites before upgrading enough, or does it require more specific steps?

Komodo5197:
Downloading any of the finampcollections, such as all favorites, and then switching to 9.7 should break syncing for that item.

lib/services/metadata_provider.dart
```diff
@@ -84,23 +84,20 @@ final AutoDisposeFutureProviderFamily<MetadataProvider?, MetadataRequest>
...
-          // keep the original bitrate from the media stream instead
-          if (profile?.codec != FinampTranscodingCodec.original && profile?.stereoBitrate != null) {
-            mediaStream.bitRate = profile?.stereoBitrate;
-          }
-        }
-        return mediaStream;
-      }).toList() ?? [];
-          
+      // We could explicitly get a mediaSource of type Default, but just grabbing
+      // the first seems to generally work?
```
Chaphasilor, line 88: Not sure if that's a good idea. I was thinking about supporting stuff like music videos, where the bitrate should show the audio bitrate, and not the video bitrate (but the video would most likely be the first stream).  
Preferring type `Audio`, then `Default`, and then falling back to the first stream would be the best approach imo

Komodo5197: Those aren't both possible values.  MediaSources can have types "Default" "Grouping" "Placeholder", while Mediastreams can have types "Audio" "Video" "Subtitle" "EmbeddedImage".

Chaphasilor:
So the core change of the PR seems to work. Downloaded all three finampCollections in 0.9.6, upgraded to 0.9.7, got a bunch of errors, switched to this PR, errors gone, sync completed, repair also working.  
Didn't attempt downloading the playlist info collection on 0.9.7, but it probably shouldn't affect the migration to this PR / 0.9.8...

Komodo5197:
Nice.  Yeah, I don't think the playlist info collection should have any bearing on this.  Also, I noticed that issue where lyrics don't work for downloaded items when online - I'm thinking this is because the merging in metadata_provider was stripping the lyrics mediaStreams.  If so, this PR should fix that because I messed with how the info gets passed around/merged a bit.

Chaphasilor:
Yeah I had the same theory about the media stream merging. I'll take a closer look tomorrow!

Chaphasilor:
Re: lyrics bug, I was simply missing a negation here:
https://github.com/jmshrv/finamp/blob/cec5b3d9256062ff51503fd3c9971a120077f0b9/lib/services/metadata_provider.dart#L152
The idea was to add all non-audio/default streams from the online data, and then add the audio streams from the offline data.

But it seems with your change to returning an empty list for transcoded downloads, you break lyrics for those tracks completely, since the lyrics are only stored as a mediaStream, and not separately like the audio or image :P  
The comments you added are very helpful though!

lib/services/metadata_provider.dart
Chaphasilor: Interesting. In that case maybe grab the first source, but the `Audio` stream?

Komodo5197:
Okay, that makes sense.  Although I've made all the streams but lyrics default to the offline ones default to the offline ones because they are probably part of the file?  Not like it matters when we don't use them.  Anyway, I can't actually check because my server is still on 10.8, but hopefully lyrics for transcoded downloads work now.

lib/services/metadata_provider.dart
Komodo5197: Yeah, that's basically how I've got it set up now.

Chaphasilor:
whoops, didn't see you already fixed that bug. sorry about the merge conflict!

Chaphasilor:
Lyrics for transcoded downloads in offline mode are still broken, because the stream type is called `Lyric` and not `Lyrics`. We should probably use an enum for that anyway 😅

Komodo5197:
Whoops.  Yeah, an actual Enum would be nice.  I guess that's another benefit if/when we migrate the api models to be autogenerated.

I actually forgot I needed to check the sync favorites offline settings as well as the filter status, so it's a good thing that you did put a separate fix in.

Chaphasilor:
I just copied the filter from the one in music_screen_tab_view :P

Lyrics are working now, and the favorite index is also correct, thanks!

PR #760, opened 2024-05-30
Bundle metainfo.xml in linux builds

baarkerlounger:
@Chaphasilor sorry looks like I missed a bit


Chaphasilor:
Assets are updated :)

PR #763, opened 2024-05-31
Appstream validation for metainfo.xml

baarkerlounger:
Fixes Appstream validation of the metainfo file. This should resolve the flathub build failure.

### DIFF
--- a/assets/com.unicornsonlsd.Finamp.metainfo.xml
+++ b/assets/com.unicornsonlsd.Finamp.metainfo.xml
@@ -4,17 +4,20 @@
   <name>Finamp</name>
   <summary>An open source Jellyfin music player</summary>
   <metadata_license>CC0-1.0</metadata_license>
-  <project_license>MPL</project_license>
+  <project_license>MPL-2.0</project_license>
   <recommends>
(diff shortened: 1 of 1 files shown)


baarkerlounger:
@Chaphasilor this should fix the flathub build error. Could you update the release assets again please?

Chaphasilor:
Sure, will do!

Chaphasilor:
done

Chaphasilor:
@baarkerlounger I've updated the ID to be lowercase here, but can't edit the flathub PR. Could you please handle the requested changes there? Keep in mind that the file names in the bundle are different now!

baarkerlounger:
@Chaphasilor will do. The name in the desktop file should probably also be updated but not sure if that'll break the arch build

Maxr1998:
Not sure what's referred to here. Could you please elaborate?

Chaphasilor:
Hmm, I'm not sure either. I assuned there was some kind of ID in the m4 desktop file, but that doesn't seem to be the case?

Maxr1998:
Nope, there's only a display name and an identifier for the icon file.

PR #774, opened 2024-06-11
[Redesign] Fix queue item playlist removal cache

Komodo5197:
Use the correct ID so that queue items removed from a playlist actually stop showing up as in it.

### DIFF
--- a/lib/components/PlayerScreen/queue_source_helper.dart
+++ b/lib/components/PlayerScreen/queue_source_helper.dart
@@ -147,5 +147,5 @@ bool queueItemInPlaylist(FinampQueueItem? queueItem) {
           .contains(queueItem.source.type) &&
       baseItem?.playlistItemId != null &&
       !playlistRemovalsCache
-          .contains(queueItem.source.id + (baseItem?.id ?? ""));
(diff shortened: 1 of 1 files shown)


Chaphasilor:
Thanks. I feel like that was needed to support items that were (re-added) to a playlists while already in the queue, since then the playlist item ID would be incorrect. But I'm not totally sure anymore.  
Does it not work at all without this change? The only thing affected is the "remove from playlist" option in the menu I think?

If it doesn't work without the fix we can definitely merge this. I guess there is no way to make it work for newly-added tracks anyway without manipulating the baseItemDto...

Any idea why the build is failing? Did the nee Flutter version break things again? o.O

Komodo5197:
Re-adding items was never really supported due to issues with having an outdated playlistItemId.  That's why I ended up switching back to the playlistItemId instead of the id, but I forgot to switch the check when switching the insertion.

The removal cache affects showing remove from playlist as well as showing the item as present in the playlist actions menu.  Removed items would still show up highlighted if you re-opened the menu, which is what I noticed to catch this.

I didn't cancel anything, I think github was just acting up.  It was behaving a bit oddly when I created this PR.

Chaphasilor:
Okay, sounds good! Sorry for not getting this in before the update...

PR #783, opened 2024-06-17
Fix release order for appstream validation

baarkerlounger:
Needed to fix https://github.com/flathub/com.unicornsonlsd.finamp/pull/2


Chaphasilor:
They really are picky 😅

While you're at it, you could also update the release date for 0.9.8, I had to take a guess on that one...

baarkerlounger:
I don't make the rules :rofl: 

I think the date should probably point to the Finamp release date so have updated it to that.

Chaphasilor:
Thanks! :)

baarkerlounger:
@Chaphasilor mind updating the release assets?

Chaphasilor:
done :)

PR #792, opened 2024-06-23
- fix settings not showing

Tiefseetauchner:


Chaphasilor:
Nice, thanks for tracking that down! <3

I guess at some point we should move that setting into a subsection, but maybe we'll wait until there are other similar settings...

PR #793, opened 2024-06-23
Fix chinese font incorrectly rendered on Windows Desktop

lifegpc:
Flutter 3 have font renderer issues on Windows. (For details, see https://github.com/flutter/flutter/issues/103811 )
The simplest way to resolve this issue is add zh_CN to supportedLocales.
Before apply patch:
屏幕截图 2024-06-23 203518
After apply patch:
屏幕截图 2024-06-23 204531
Some characters still incorrectly rendered because speficied font weight is not available on Chinese default font and fallback to English font. (See https://github.com/flutter/flutter/issues/105014 )


Chaphasilor:
Thanks for the fix! That seems like a simple patch and I don't think it would break anything else, so I'm happy to merge it!

If font weights are causing bigger issues (which seems to be the case if the wrong characters are used), you could also propose a fix for that. If there's a specific font weight that does work well we could discuss changing the default to that one instead :)

PR #802, opened 2024-07-01
Add an option to keep screen awake in the Lyrics View

hermesespinola:
This adds support to prevent the phone screen from sleeping when you are reading lyrics (it shouldn't happen if the song doesn't have lyrics). I also ran dart fix on the touched files.

I didn't find any test files so didn't add any, if there actually are let me know and I'll add them. I also tried testing this on a device with no success. I kept running into errors related to the Gradle config and/or android dependencies so I eventually gave up trying to do this. If someone is willing to test this that'd be appreciated.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -1570,12 +1570,12 @@
   "@showLyricsTimestampsSubtitle": {
     "description": "Subtitle for the setting that controls if timestamps are shown for lyrics"
   },
-  "keepScreenAwakeInLyricsScreenTitile": "Keep screen awake",
-  "@keepScreenAwakeInLyricsScreenTitile": {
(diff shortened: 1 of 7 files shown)


Chaphasilor:
Hey, sorry for not replying earlier. Taking a look at it now!

Chaphasilor:
The changes look good! I'm wondering if it might be better to use `WakelockPlus.enabled` for checking if a lock is active instead of manually tracking it? Shouldn't add any complexity, but it might be a hassle since it returns a future. Up to you I'd say :)

Otherwise, everything seems to be working like it should, on Android at least. If you're happy with the code I can merge this! I wouldn't mind if you tackled the merge conflicts yourself, but I can also do it myself ^^

PR #803, opened 2024-07-02
Fix URI passed to ContentProvider

puff:
Fixes an issue where artwork doesn't display in the media notification or android auto/automotive if the server uses a port number.

The issue was the port number wasn't removed from the URI when passing to the ContentProvider, so it wasn't able to find it.
image
Additionally, the ContentProvider would encode the URI's authority (breaking the URL) unless we set the encoded authority directly.


jmshrv:
Thanks for this, merging :)

PR #847, opened 2024-08-22
Fix stray "null" in download transcode dialog

Maxr1998:
Fixes *"Download original - x.xx GB null"* transcode option when downloading playlists.

Originally reported by Tarulia in the Finamp Discord.


jmshrv:
Thanks for this, surprised that "null" got through lol

Issue #500, opened 2023-09-24
[Feature Request] Finamp-specific volume control


Venryx:
The songs in my library are apparently quite loud; to match the volume level I want, I need to put my system volume at ~5%.

This is a problem, because it means that then I cannot hear other apps with audio -- such as Google Maps for navigation, which I want to be able to hear alongside my music.

The solution would seem to be to add a volume control within Finamp's player specifically. Then I can reduce its volume enough that the system-level Media volume gives a good result for Finamp and Google Maps at the same time.

****

By the way, I notice there's a feature request for Replay Gain support: https://github.com/jmshrv/finamp/issues/468

That's good of course, but a volume control would still be needed to control the relative volume between Finamp and Google Maps. (or other use-cases, eg. android builds where it only provides ~10 system-level "volume ticks")

Chaphasilor:
Would you want to adjust this volume frequently or could it be somewhere in the settings? 🤔

Venryx:
For my particular use-case, having it in the settings would be "workable" -- although it does seem like something useful enough to be worth adding a little button for it. (eg. click the button, and it shows a vertical slider above it for volume)

Chaphasilor:
I'm thinking about adding this as part of adding volume normalization (#468). Basically, when enabling volume normalization, you could set a "target volume" which all tracks should have (on average)

Venryx:
> I'm thinking about adding this as part of adding volume normalization (#468). Basically, when enabling volume normalization, you could set a "target volume" which all tracks should have (on average)

That would be very nice indeed! Especially in contexts where ui interaction is difficult (eg. while driving), since it would (at least mostly) remove the need to change volumes for tracks.

Venryx:
> Sweet. Also, another question: You mentioned that you can't hear Google Maps announcements while Finamp is playing. Are both apps running on the same device? Finamp _should_ "duck" the volume if navigation apps are playing audio, if that's not working we could probably fix this in an even better way.

Well the problem is that my music tracks are just too loud, so that I have to set the overall system volume at almost the lowest setting; so even though Finamp "ducks" the volume when Google Maps speaks, the system volume is so low that it still makes the Google Maps voice uncomfortably quiet.

I know I could fix this by pre-processing the tracks to be quieter, but it would be really nice to not have to do that. (which is why your suggestion above about a volume normalization feature sounds great to me)

Chaphasilor:
[...]

PR #850, opened 2024-08-25
Add "Play On" receiver compatibility

pinsarda:
This is a very early shot at getting the playon feature from jellyfin to work with finamp.
For now this is a extremely rough draft to start working on the feature and getting rid of an annoying bug. Can only start songs without having media controls and need to enter the url manually.


Chaphasilor:
Thanks for getting this started, really looking forward to it!

So I managed to get this working a bit more reliably and generically some time ago, just pushed those changes :)  
The playback controls are now all shown, and Finamp registers support for all commands I could find. The important ones are already implemented too.  
I didn't work on the actual "play on" functionality though, so that's unchanged, meaning it works for the most part, but sometimes the wrong item is picked (e.g. selecting a specific track from an album will play the first track from the album instead). Ideally, this should work just like it works in Finamp, and make use of the same settings. So, if a single track is selected, depending on the Finamp settings it should either play just that track or start an instant mix from that track.

There are still issues with the websocket, it seems to get disconnected (or at least stops receiving messages) after a short while or certain commands. Haven't looked into that just yet.  
I also noticed when testing this out again that when logging into the server the websocket connection will fail until the app is restarted, because the connection is established before the URL and credentials are know. The initialization should therefore be delayed until setup is complete, or restarted after successfully logging into a server.

pinsarda:
The current implementation works very well on my phone, but can probably be a bit slow as I request each song individually instead of the whole album at once. I don't know if there's an easy way to do it better currently...

Chaphasilor:
Well just take a look at how the content of an album is fetched for the album screen :)  
It's just a matter of setting the parentID filter, I think.

pinsarda:
Yes I thought about that but this would introduce a bug when starting a playlist. The websocket just gives us ids of songs to play. It doesn't ask finamp to play an album or a playlist. This makes it hard to do it differently.

Chaphasilor:
Is there no way at all to differentiate from "where" the track was selected? I guess in that case all we can do is play just that one track (or start the instant mix, depending on settings), and nothing else.  
Is that also the case when you try to play an entire album or playlist? Does it only send us the first track, without any additional information?

pinsarda:
I don't think so, we just receive a list of ids, and optionally the index of the track selected if there was any. The current implementation does what you're proposing. If a specific track is selected, we only play this one or start instant mix. And if there's no precision on the starting index it queues the whole album/playlist. After testing it seems like the best possible behaviour to me.

pinsarda:
As discussed in https://github.com/jmshrv/finamp/issues/500 an internal volume controller will be implemented eventually, we'll rely on that instead of system wide volume.

Chaphasilor:
@Maxr1998 how did you approach this issue on the Android (TV) app? 🤔

Maxr1998:
Can't speak for Android TV, not even sure play on is supported there. On the Android app, we just wrap the web app so the handling is identical to that.

Komodo5197:
This seems like the server is giving us a list of song IDs to form the queue, which means the most similar piece of code is loadSavedQueue() in lib/services/queue_service.dart, so you can probably look at that for inspiration.  Most notably, you can make batch requests for up to 200 item IDs at once, which is significantly faster.  Also, I'm pretty sure the behavior of loading the single song or an instant mix is questionable.  We should be loading the whole id list from the server into the queue and just setting the initialIndex to what the server tells us to.

Chaphasilor:
@Komodo5197 the playing of a single track should only happen when the user selects a single track in the other client, and the API only gives us one ID. If we get multiple IDs, we should play all of them of course.

Thanks for the suggestion about the queue restore, that would make sense.  

However, my main concern was how to identify where these tracks are coming from, so we can show a proper queue source. If the user requests to play an album, I'd like Finamp to show which album is playing as the queue source name. Same for playlists, artists, etc. But it seems that info is simply not given to us, so we'll have to default to some generic ("Queued by Remote Device"?) name instead.

Komodo5197:
We are getting the full queue.  I click a track partway through an album on the web UI and finamp recieves {"MessageType":"Play","MessageId":"560c155a593a497e8bd047610f2e8e4a","Data":{"ItemIds":["46430779c8072941ed8f7e2ddafaebe4","f8ee032bb75db53c1c317c572e3e8814","fda9dba1e21b261bdb050efe5bd95bbb","4b0a4b5c756079eb8322860215fbe5fb","e012de5b9fb821a2258b6e91c9225459","ea8e6422f6266c453469e104e3e4b950","cbb922f4f858571b79dafa8da3578270","d891ee5acb57e58ca338df5518c8ce37","80ed7b24b4878c28a75e6234ff066bd6","ca4e6f99c7ad2bae487aaba0585f869c","d586e2361bab084b04872a4128507db8","67f19176d6582739abfd0678442b45de","c34b8e312fd99b7194d927c64a9700a9","c89368641479249da0b333c09ba1aa40","ba8ade9d002d8c40bae852531bcebba5","8ac7aa4480fa7fcaecf1d03a816431ed","22a0587b3018e44644924535542ace41","ca63bf02ba1a0bee484e2fb6e036d3e0"],"PlayCommand":"PlayNow","ControllingUserId":"de6a683bafb44247881bb62bf125f7e8","StartIndex":8}}.

There are also additional play commands 'PlayNext', corresponding to addToNextUp(), and 'PlayLast', corresponding to addToQueue().  Implementing those allows remote queue building that seems to match up with how the web client works.

Regarding queue source, it does indeed seem that we don't have this information in any form, and will need to use a generic header.

Also, I happen to still be running a 10.8 server, and to get this to work I had modify the ClientCapabilities to explicitly set supportsSync and supportsContentUploading to false, and remove the supported command "SetPlaybackOrder".

Chaphasilor:
Okay, thanks for taking a look. Haven't had a look at the code recently, but there are still instances where a single track is selected, e.g. when playing from a search result. In that case, the Instant Mix behavior should still apply, I think.

Interesting note about the 10.8 server. I also still have one of those lying around, I'll give it a try myself. Maybe the commands had a different name back then? I'd like to keep the playback order functionality if possible, but I'd also like to keep supporting 10.8.  
Maybe it's finally time to query the server info (version, name, etc.) on startup, and have some version-specific features. I guess feature detection is a hopeless endeavour with Jellyfin anyway...

Chaphasilor:
I gave the current version a try yesterday. It was working well at first, and I could see the queue, playback info, and control playback mostly fine. But later on I didn't manage to get the queue or even the playback info to show up anymore, even after restarting the app.  
I also encountered an issue where it seems like commands where being queued for some reason. I tried playing/pausing and sending a few other commands (including a message) through the web client to Finamp, which didn't have an effect. But then many minutes later I tried to pause playback (in Finamp directly), but it was resumed immediately. That happened a few more times, and then finally it actually did pause and stayed paused, and then the message appeared. Maybe there's a command waiting for a Future that never resumes, or something similar?

pinsarda:
For the first issue, it's weird I didn't experience it. I got the same bug regarding command queuing though, and got to the same conclusions as you without nailing it yet. I'll get back to it tomorrow and on thursday. I'll also adjust the code to have the behaviour proposed by @Komodo5197

pinsarda:
@Chaphasilor Couldn't spend as much time as I wanted on this, but as we guessed, using unwaited instead of await when executing the commands (expecially Pause, Unpause and PlayPause) completely fixes the issues I had with unresponsiveness after sending too many commands, and also makes everything smoother. Is it a bad practice to do so ? 
It definitely needs some more testing to be sure but now it seems to "just work" on my device. 
There's one remaining bug when seeking I have to fix. Jellyfin takes a bit of time to update the progress bar on the web client, because finamp doesn't report playback state directly.

Chaphasilor:
Hmm. It's good that we seem to have found the issue, but the best idea is probably to figure out why the play/pause futures don't resolve like they should.

Aside from that, will the websocket reconnect to the server after changing i.e. from WiFi to mobile data?

pinsarda:
Ok, I'll check further, I didn't see anything suspicious last time I checked the futures.
With my current setup (VPNs and stuff) I unfortunately can't check easily if the websockets reconnect when changing network, but I highly doubt it. Disconnecting and reconnecting wi-fi closes the connection. Changing this, as well as fixing the bug when connecting for the first time, is on my to do list

Chaphasilor:
Just wanted to note that the websocket connection should get disconnected when switching to offline mode, before I forget about it.

pinsarda:
There's an issue when using playlist, the songs are completely unordered. This is an upstream bug from jellyfin : https://github.com/jellyfin/jellyfin/issues/8885
I didn't look into the details just yet

Chaphasilor:
That's because the server sends us the raw track list instead of just the playlist ID, right? If there is an ID / some way to figure out which playlist is being requested, be could simply disregard the track list and fetch the playlist ourself?

pinsarda:
Absolutely, but the playlist id is not in the message received by the websocket unfortunately. Plus you have no way to even know if it's a regular album or a playlist being played (apart from checking if all the songs belong to the same album after fetching data). As you guessed, we only get a list of ids of songs to play and a starting index, that's it.
I have dirty workarounds in mind but I would rather have the bug fixed properly jellyfin side. And/or have the playlist/album id supplied directly in the websocket data, which would also be nice to populate the "playing from" label.

pinsarda:
In the last commit, I am not too sure about my "updateState" function in favoriteProvider, I couldn't find an other way to do it directly using riverpod's ref. It works but may be unnecessary to create this function.

lib/components/LoginScreen/login_flow.dart
```diff
@@ -120,6 +120,13 @@ class _LoginFlowState extends State<LoginFlow> {
                 connectionState: connectionState,
                 onAuthenticated: () {
                   Navigator.of(context).popAndPushNamed(ViewSelector.routeName);
+                  final jellyfinApiHelper = GetIt.instance<JellyfinApiHelper>();
+                  jellyfinApiHelper.updateCapabilities(ClientCapabilities(
+                    supportsMediaControl: true,
+                    supportsPersistentIdentifier: true,
+                    playableMediaTypes: ["Audio"],
+                    supportedCommands: ["MoveUp", "MoveDown", "MoveLeft", "MoveRight", "PageUp", "PageDown", "PreviousLetter", "NextLetter", "ToggleOsd", "ToggleContextMenu", "Select", "Back", "TakeScreenshot", "SendKey", "SendString", "GoHome", "GoToSettings", "VolumeUp", "VolumeDown", "Mute", "Unmute", "ToggleMute", "SetVolume", "SetAudioStreamIndex", "SetSubtitleStreamIndex", "ToggleFullscreen", "DisplayContent", "GoToSearch", "DisplayMessage", "SetRepeatMode", "ChannelUp", "ChannelDown", "Guide", "ToggleStats", "PlayMediaSource", "PlayTrailers", "SetShuffleQueue", "PlayState", "PlayNext", "ToggleOsdMenu", "Play", "SetMaxStreamingBitrate", "SetPlaybackOrder"],
```
Chaphasilor, line 128: I think we should have a getter/function in `playon_handler.dart` that generates these `ClientCapabilities`. So that the file that handles the commands also decides which commands are currently supported.  

We can also use the proper type for "Audio", `BaseItemDtoType.song`. I'm also wondering if some other types should be supported, like `MusicAlbum`, `MusicArtist`, `MusicGenre`, `Playlist`, and `AudioBook`. *Maybe* if those are supported the server will actually send us some information about the playlist or album being played, instead of just a list of tracks?

lib/services/playon_handler.dart
```diff
@@ -0,0 +1,245 @@
...
+  
+  Future<void> initialize() async {
+
+    // Turn on/off when offline mode is toggled
+    var settingsListener = FinampSettingsHelper.finampSettingsListener;
+    settingsListener.addListener(() async {
+      if (FinampSettingsHelper.finampSettings.isOffline) {
+        await closeListener();
+      } else {
+        await startListener();
```
Chaphasilor, line 44: we should probably check somewhere if we're already connected, otherwise we'll reconnect every time a setting is changed! And since volume is a setting, that might be causing issues...

Chaphasilor: Volume being a regular setting might be an issue in general. Any `Consumer` widget that uses `finampSettingsProvider` would rebuild on volume changes, not ideal...

pinsarda: Just patched the first problem. And I agreee about the last comment. What could we do instead ?

Komodo5197: Avoiding excess rebuilds is the purpose of all the select() statments and the subproviders.  So a lot of widgets should be fine, even with a setting that changes fairly often.  There's still a couple of widgets using the settings listner, which would update all the time, and but we should be generally getting rid of that.  We're already tracking things like loop mode and playback speed, so I wouldn't be too concerned adding volume.

lib/l10n/app_en.arb
```diff
@@ -942,6 +942,10 @@
   "@yourLikes": {
     "description": "Title for the queue source when the user is playing their liked tracks"
   },
+  "playon": "Remote client",
```
Chaphasilor, line 945: This isn't the translation for the play on feature, so let's call it `remoteClient` or even better `queueSourceRemoteClient` (or `queueSourcePlayOn` if you prefer that).  
And also, please use Title Case for this, that looks nicer in the UI :)  

Also, I just had an idea. On the player screen there's a "title" or type and a "subtitle" or item. Here's what I'm talkin about:
image
I think it might be good to only use the title, so that it says "Playing From Remote Client", with nothing (or an empty string) below it. What do you think?

Chaphasilor: (Unless that's already what's happening, haven't built the commit yet 😅)

Chaphasilor: Oh and you still need to adjust the translation for `playingFromType` and add a new case there, and you should add a new case (with simply a `break`) in `queue_source_helper.dart` :)

pinsarda: image
The last PR does this, is this what you had in mind ?

Chaphasilor: Even though I don't speak French, yes, I think that's what I meant :)

lib/components/PlayerScreen/queue_source_helper.dart
```diff
@@ -64,6 +64,7 @@ void navigateToSource(BuildContext context, QueueItemSource source) {
     case QueueItemSourceType.trackMix:
     case QueueItemSourceType.filteredList:
     case QueueItemSourceType.downloads:
+    case QueueItemSourceType.remoteClient:
```
Chaphasilor, line 67: I don't think we'll ever implement a source target for this. So let's move it further up and add a `break`, like for `nextUp` or `unknown`. Right now you put it into the fall-through case which creates a "Not yet implemented" snackbar...

lib/l10n/app_en.arb
```diff
@@ -964,7 +960,7 @@
   "@savedQueue": {
     "description": "Title for the queue source for tracks were previously in the queue before an app restart and have been reloaded"
   },
-  "playingFromType": "Playing From {source, select, album{Album} playlist{Playlist} trackMix{Track Mix} artistMix{Artist Mix} albumMix{Album Mix} genreMix{Genre Mix} favorites{Favorites} allTracks{All Tracks} filteredList{Tracks} genre{Genre} artist{Artist} track{Track} nextUpAlbum{Album in Next Up} nextUpPlaylist{Playlist in Next Up} nextUpArtist{Artist in Next Up} other{}}",
+  "playingFromType": "Playing From {source, select, album{Album} playlist{Playlist} trackMix{Track Mix} artistMix{Artist Mix} albumMix{Album Mix} genreMix{Genre Mix} favorites{Favorites} allTracks{All Tracks} filteredList{Tracks} genre{Genre} artist{Artist} track{Track} nextUpAlbum{Album in Next Up} nextUpPlaylist{Playlist in Next Up} nextUpArtist{Artist in Next Up} other{} remoteClient{a Remote Client}}",
```
Chaphasilor, line 963: Please try to always have the `other` case at the end, because it has special meaning and is essentially the `else` part

lib/services/playon_handler.dart
```diff
@@ -122,8 +122,13 @@ class PlayonHandler {
...
     }).listen((count) {
-      startListener();
-      _playOnHandlerLogger.info("Attempted to restart listener");
+      if (count < 24) { // We try to connect for two minutes before giving up
+        _playOnHandlerLogger.warning("Attempt $count to restart playon listener");
+        startListener();
+      } else {
+        _playOnHandlerLogger.warning("Stopped attempting to connect playon");
+        _reconnectionSubscription?.cancel();
+      }
```
Chaphasilor, line 128: This is fine for now, but once #1139 lands we can probably try to reconnect once the connection status changes.  
It would be a bit weird if we never tried again until a restart.  
Alternatively, after 24 counts you could increase the duration we wait, from 5 seconds to 30 or even 60 seconds, instead of completely giving up.

lib/services/playon_handler.dart
```diff
@@ -190,7 +191,22 @@ class PlayonHandler {
...
     if (request['MessageType'] != 'ForceKeepAlive' &&
         request['MessageType'] != 'KeepAlive') {
+
+      // Because the Jellyfin server doesn't notify remote client connection/disconnection,
+      // we mark the remote controlling as stale after 5 minutes without input as a workaround.
+      // This is particularly useful to stop agressively reporting playback when it's not needed
+      await _isControlledSubscription?.cancel();
       isControlled = true;
+      _isControlledSubscription =
+        Stream.periodic(const Duration(minutes: 5), (count) {
```
Chaphasilor, line 201: Maybe make this a setting? The default can stay the same, but it would be more flexible and could help avoid battery drain if users can reduce it to 1 minute or so...

Chaphasilor (commented):
Just a few smaller comments. I think we can probably merge this today or tomorrow!

lib/services/playon_handler.dart
pinsarda: Good idea, should I put it in a brand new settings category or an existing one ? I don't feel like there is a perfectly relevant one currently

Chaphasilor: I good see a "Playback Reporting & Play On" category making sense? Then we could move the "Report queue to server" and "Playback Session Update Frequency" in there as well...

Chaphasilor:
Changes look good! I'll probably go over the exact wording an capitalization tonight.  
Now, what is left to do before this can be merged?

I can think of:
- [ ] Make sure output panel doesn't throw errors on non-Android platforms'
- [ ] Report queue to server by default
  - Maybe we even want to add a setting for disabling the Play On websocket connection? Are there any downsides to it (probably battery and data usage)? Then we could automatically send the queue if Play On is enabled, and otherwise use the state of the dedicated setting for sending the queue
- [ ] (Potentially) Integrate changes from #1100 to use the new internal volume
- [ ] Test on desktop

Improvements for Future PRs:
- [ ] Link volume slider in output panel with system volume by default (on mobile), but allow detaching it for controlling the internal volume.
- [ ] And of course, implement controlling *other* clients

Also, for the release notes I'd like a list of notes/requirements for using this:
- Jellyfin bugs to consider (only admin can control, playlists in wrong order)
- Ideal settings to make sure it works reliably (e.g. disable all battery saving
- Supported Features, but more importantly *unsupported features*
  - Switching tracks within the queue is still unreliable
  - Loop/shufflle can't be toggled yet
  
Please add any points you can think of :)

pinsarda:
On linux desktop it works like a charm. 
I'll add the toggle to disable the feature completely in the settings today when I have 2 minutes. 
For the ideal settings to use, I think it needs testing. I should beta test a release build.

I don't have much to add, you already listed a lot of things, but :
- We discussed android auto disabling the feature, do we still want that ? If so I looked into it briefly, but can't get a boolean or settings listener to catch this event.
- A common issue that could arise is using a misconfigured reverse proxy that blocks websockets (nginx does it by default afaik)
- For jellyfin web bugs, sometimes (rarely) the player seems to work unreliably, but a simple socket reconnect from the remote client does the trick in my experience

Chaphasilor:
Okay, the reverse proxy stuff is a good point!

Chaphasilor:
@Komodo5197 since you recently removed the `postLaunchHook`, where should I best initialize the Play On service now? The main challenge is obtaining a `WidgetRef` so that we can update the favorite state...

Komodo5197:
We could grab the providerScope from one of the global snackbar keys and use that.  That would allow us to set this up from main with anothrer user login hook.  Alternatively, it looks like we could manually create a global ProviderContainer that we register with GetIt.  Then we can replace the top level ProviderScope with an UncontrolledProviderscope using that container, and any service code could use the global container to read or listen to providers.  Also, you may want to have a check in the favorite provider update where if the old and new values differ, call ref.keepAlive(); and _changed = true;.

Chaphasilor:
I'll merge this now, since everything still seems to work with the latest changes and I'd like to finish up the release.  
If anything isn't ideal yet we can hopefully fix it afterwards.  
Huge thanks @pinsarda for pushing this through and working on this for so long! <3

Komodo5197:
I was talking about setting _changed inside the new updateState function.  As it is now if the remote updates the favorite and then we jump tracks forwards and backwards I believe the provider will have been rebuilt and the change has been forgotten.  Meanwhile, if we're in the initializer callback the previous value is null or meaningless so we shouldn't be trying to determine if anything changed.  Also, I think we should have some sort of error handling for the _handleMessage function as a whole?  Also, from the code it looks like the service calls closeListner() for any settings change when playOn is disabled?  Also, iis it intentional that sortBy: "IndexNumber" is inconsistent between the different types of plays?  Also, I noticed a valueListenable is used for the new settings toggle.  It theoretically causes less uneeded rebuilds iuf you use the riverpod style.  Also, I noticed the use of a lambda for the setter - you don't need that, you can just directly supply the setter function if there's no additional code.

PR #867, opened 2024-09-06
Screen does not stay on #781

Sp4rky001:
Added Always On option in settings and changed default option to While Showing Lyrics


Chaphasilor:
Okay, seems like a rebase is really needed ^^

Oh and one thing, I'd appreciate if you could use the PR title to describe the actual change, not reference an issue :)  
You can link the issue in the description body of course, and that would be helpful. Thanks :)

Sp4rky001:
hah, ok.  Hopefully, I won't mess up the branch with a rebase.  Git documentation warns against it but i never really understood rebase very much, ha ha

PR #869, opened 2024-09-07
Added Always On option for Keep Screen On feature

Sp4rky001:
Adding to the Keep Screen On feature, the Always On option was added.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -1638,10 +1638,11 @@
   "@keepScreenOn": {
     "description": "Option to keep the screen on while using the app"
   },
-  "keepScreenOnSubtitle": "Keep the screen on while playing music or only while showing lyrics",
+  "keepScreenOnSubtitle": "Keep the screen on while using Finamp, only while playing music or only while showing lyrics",
   "keepScreenOnDisabled": "Disabled",
-  "whilePlaying": "While Playing Music",
-  "whileLyrics": "While Showing Lyrics",
+  "keepScreenOnAlwaysOn": "Always On",
+  "keepScreenOnWhilePlaying": "While Playing Music",
+  "keepScreenOnWhileLyrics": "While Showing Lyrics",
   "keepScreenOnWhilePluggedIn": "Keep Screen On only while plugged in",
   "keepScreenOnWhilePluggedInSubtitle": "Ignore the Keep Screen On setting if device is unplugged"
 }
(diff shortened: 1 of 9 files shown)


Sp4rky001:
oh, my goodness.  I learned a lot about rebasing.

Chaphasilor:
Ahh, I think a regular rebase won't cut it either in this case, sorry!  
I squashed your commit into a single one when merging, and now when you try to rebase, git unable to match your unsquashed commits to the squashed one. That's why it applies *all* of your commits on top of the squash commit :((

Doing the following should work:
1. Checkout `redesign` and make sure it's pulled
2. Create a new branch from it (instead of these two steps you could also run `git reset --hard origin/redesign` on the branch of this PR, but that would make it harder to keep your commit)
3. Copy the whole commit ID (hash) for the commit adding the Always On functionality
4. On the new branch (which is up-to-date with `redesign`, use `git cherry-pick <commit-hash>` to bring the commit over into the new branch. You might have to resolve conflicts.

If you made more than one new commit, repeat steps 3+4 for all of them.

Then it should be a clean branch with just your new commit(s), which you can then use for this PR (e.g. by checking out the current PR branch, and running `git reset --hard <new-branch-name-with-cherry-picked-commit>`.

Sorry about the Git gymnastics. I should've waited before merging the original PR :)

lib/services/keep_screen_on_helper.dart
```diff
@@ -46,14 +46,17 @@ class KeepScreenOnHelper {
     });
   }
 
-  void setKeepScreenOn() {
+  void setKeepScreenOn() async {
```
Chaphasilor, line 49: Any reason why you made this async when you're not `await`ing anything? :)

lib/models/finamp_models.dart
```diff
@@ -2193,8 +2193,10 @@ enum KeepScreenOnOption {
   @HiveField(0)
   disabled,
   @HiveField(1)
-  whilePlaying,
+  alwaysOn,
   @HiveField(2)
+  whilePlaying,
+  @HiveField(3)
```
Chaphasilor, line 2199: It's fine to do this now since the feature hasn't been released yet, but for existing features you shouldn't modify or re-assign these IDs, since that would unexpectedly change some settings for users.

lib/l10n/app_en.arb
```diff
@@ -1638,10 +1638,11 @@
   "@keepScreenOn": {
     "description": "Option to keep the screen on while using the app"
   },
-  "keepScreenOnSubtitle": "Keep the screen on while playing music or only while showing lyrics",
+  "keepScreenOnSubtitle": "Keep the screen on while using the app, while playing music or only while showing lyrics",
```
Chaphasilor, line 1641: You could shorten this to `When to keep the screen on` or something similar. The options are given in the select anyway ^^

PR #871, opened 2024-09-07
Add Always On option for Keep Screen On functionality

Sp4rky001:
This pull request appends to the Keep Screen On functionality by adding an "Always On" option.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -1638,10 +1638,11 @@
   "@keepScreenOn": {
     "description": "Option to keep the screen on while using the app"
   },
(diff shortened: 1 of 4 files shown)


Sp4rky001:
I'm not an expert on Git so I did this the manual way.  I recreated a branch from the current redesign branch and manually edited the files using beyond compare for the model and language files.

That async in my method was a test to see if I can await Wakelock.enabled to accurately gauge if Wakelock is indeed enabled, but that code didn't pan out.  I believe in the last commit, that was already taken out but I did double check for this commit.

Chaphasilor:
Only got around to looking at this just now. But everything checks out, so thanks again! :D

PR #873, opened 2024-09-10
Fix player album image shadow for non-square covers.

Komodo5197:
Attempts to fix the oversized backdrop shadow seen on the player screen with non-square album covers.  Also allows non-square albums to better expand into padding regions if needed to increase their size without affecting controls layout.  Also, I made the keep_screen_on_helper less chatty while tracks are playing.

### DIFF
--- a/lib/components/PlayerScreen/player_screen_album_image.dart
+++ b/lib/components/PlayerScreen/player_screen_album_image.dart
@@ -81,42 +81,39 @@ class PlayerScreenAlbumImage extends ConsumerWidget {
                 }
               }
             },
-            child: AspectRatio(
-              aspectRatio: 1.0,
-              //aspectRatio: 0.5,
-              child: Align(
-                alignment: Alignment.center,
-                child: LayoutBuilder(builder: (context, constraints) {
-                  //print(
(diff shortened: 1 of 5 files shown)


Chaphasilor:
Thanks for the PR! I can't comment on the aspect ratio stuff since I only have square covers, but maybe @Maxr1998 could chime in here?

As for the screen on helper, everything seems to be working as expected, and the change does make sense. I'm afraid it did coincide with merging the always-on addition, so there's a merge conflict and your branch doesn't include the always-on feature yet. Would you mind fixing that?

Chaphasilor:
I just noticed this re-introduces the bug where blurhashes are very tiny before the actual covers are loaded :)  
Here's a screenshot showing the blurhashes getting replaced by the actual covers:
image

Komodo5197:
Okay, blurhashes should fill the square now.  I also noticed that low resolution images weren't filling the square either, so I've bumped those to render up to 4X their native resolution to usually avoid that.

Maxr1998:
Sorry for taking a bit to reply, checked it just now and it's working great. Thanks!

Chaphasilor:
Awesome, thanks for the fix!

PR #877, opened 2024-09-13
[Redesign] Access sidebar via gesture

Komodo5197:
Allows accessing the sidebar by swiping right when on the leftmost tab of the music screen.

### DIFF
--- a/lib/screens/music_screen.dart
+++ b/lib/screens/music_screen.dart
@@ -317,24 +317,122 @@ class _MusicScreenState extends ConsumerState<MusicScreen>
                       : 8.0),
               child: getFloatingActionButton(sortedTabs.toList()),
             ),
-            body: TabBarView(
-              controller: _tabController,
-              physics: FinampSettingsHelper.finampSettings.disableGesture
-                  ? const NeverScrollableScrollPhysics()
-                  : const AlwaysScrollableScrollPhysics(),
-              dragStartBehavior: DragStartBehavior.down,
-              children: sortedTabs
-                  .map((tabType) => MusicScreenTabView(
-                        tabContentType: tabType,
-                        searchTerm: searchQuery,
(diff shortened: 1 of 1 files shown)


Chaphasilor:
The gesture is definitely better than swiping from the edge of the screen, with all of the native gesture navigation now ^^

But in the current planned design for the home screen, the menu is moved to a popup (or maybe a bottom sheet to stay consistent) that is opened by tapping a button in the top right. Here's what I'm referring to:

Home Screen Concept Mockup

The cog (might be replaced with the user's profile picture) would open the menu, and it should be structured similar to Immich's:
image


So I'm not sure how long this change will survive / make sense 😅

Komodo5197:
Yeah, I'm aware this might not last that long.  But I've found myself instinctively making this gesture an unreasonable amount of times, so I decided I might as well get it in until then.

Chaphasilor:
Alright, I guess it won't hurt, and it works just fine in my testing ^^

jmshrv:
Decided to disable this on iOS since it felt weird with the "bounce" at the end of lists, which Android doesn't have (7da719aa55dc020441b09b0f770fa87635789f14).

jmshrv:
oh also that change will apply for macOS/Windows/Linux, I actually just enabled it for Android

Chaphasilor:
@Komodo5197 aren't you on iOS? I guess you had your reasons for wanting the gesture? 😅

Komodo5197:
Actually, my existing phone finally broke a month or two ago, and I ended up with an Android to replace it.  It's way larger, because apparently nobody manufactures reasonably sized phones anymore, which is presumably why I kept trying to get the sidebar open with a gesture instead of just using the button.

Chaphasilor:
Ah okay. Yeah small phones are very hard to get ahold of. But I'm glad you were able to keep using Finamp! 😁

PR #878, opened 2024-09-13
[Redesign] Switch Android client ID source.

Komodo5197:
It seems like the id in the androidDeviceInfo.id is actually a build ID.  I've switched to an ID provided by the app_set_id plugin, which I believe should be unique.

### DIFF
--- a/lib/services/jellyfin_api.dart
+++ b/lib/services/jellyfin_api.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:io' show HttpClient, Platform;
 
+import 'package:app_set_id/app_set_id.dart';
 import 'package:chopper/chopper.dart';
 import 'package:device_info_plus/device_info_plus.dart';
 import 'package:finamp/services/http_aggregate_logging_interceptor.dart';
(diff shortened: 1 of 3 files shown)


Chaphasilor:
Hmm, thanks for attempting to fix this. I'll give it a try later, but my guess is this won't solve the issue. The ID seems to rely on Play Store metadata (which is bad for F-Droid releases), and will be the same for all apps published by the same dev, on Android at least.  
On iOS at least the value seems to be unique per-device, but that would still prevent installing different versions on one device.

I believe a random but stable ID should work. We could generate it and store it to the DB on first start.

Chaphasilor:
Okay, just tested it and it has the same issue, at least when using two versions of the app on the same device. Should we just use a random string, like `finamp-xyzabc`?

Chaphasilor:
I just noticed I didn't check out the correct branch. Let me take another look 😅

Komodo5197:
Generating and storing a random ID would be my next thought too.  I'll switch to that if this take doesn't work out.

Chaphasilor:
Okay, it seems like it actually worked. Same device, different package ID. It could be that a different device but the same package ID won't work, but I can't test that since my other phone died. So I'll merge it for now, and we'll see if anyone complains.  

I think the IDs/names should stay the same in Jellyfin too, since the access token is already generated. So there shouldn't be any issues with users getting logged out.  
Thanks for the fix!

PR #881, opened 2024-09-13
Use vector drawable for adaptive foreground and recreate legacy icons

Maxr1998:
I noticed that the current icons were still PNGs, so I used the icon tool in Android Studio to create some nice new ones. I also replaced the legacy icons, which were a square with sharp edges, with a neat circle icon.


jmshrv:
Thanks! I didn't know Android supported vector icons :)

Issue #536, opened 2023-11-23
Weird Album Sorting on Artist Page


TinDefacto:
Hi there!

Some recent update -- I'm not sure which one -- sent the sorting of albums on an artist's page all out of wack. Now the order of an artist's releases may be (as in one case for me) 2021, 2016, 2019, 2013, 2020, 2013, 2011, 2014, 2019, 2016, 2022. It's not even alphabetical, so I'm really confused! I really appreciated the previous chronological order, and I don't see this issue being raised elsewhere, so I figured I'd point it out. 

Thanks again for all the work y'all do. :)

Chaphasilor:
That might've been https://github.com/jmshrv/finamp/pull/448 or #493

Could you check the full metadata in Jellyfin to see what the PremiereDate of these albums is, compared to the production year?

TinDefacto:
Oh, interesting -- thanks for the response! I hadn't seen #448, and that
looks most likely to be the culprit, but when I look at the metadata on
Jellyfin, literally only one of the albums has a release date, and that's
the 2022 one (which appears last). So I'm not sure what's going on?

TinDefacto:
And doing a quick survey of other artists, it seems like it's correctly
sorting albums that have release dates filled in -- but for those that only
have a year, I can't find rhyme or reason to the order it decides on.

Chaphasilor:
If there is no release date provided, the order is most likely just random, as you are observing. We should fall back to sorting by production year and then name I think

TinDefacto:
Yes, I think that would make the most sense! Preferably by taking any blank
release dates to be YEAR-01-01, so it gets sorted in with the dates ones,
if that makes sense.

Maxr1998:
I remember doing two PRs for this issue, since some changes were undone in between. Maybe that's just an artifact of that, and not actually an issue? @TinDefacto could you verify whether it's still an issue on the latest version?

TinDefacto:
Hi! Okay, thanks so much for your work on this -- though I think it still
needs a bit more work!

Some artists on my server (like those that have multiple releases per year)
have full release dates for each album -- these are now displayed in
perfect chronological order, which is immaculate.

However! For other artists I either have a mix of exact release dates or
blank ones (i.e. just the release year) or only years for each release, no
exact release date. These it seems to just be putting in alphabetical order
by album name, which is very much not what I want.

What I feel might be a good solution is to, for albums with only years and
no release date, sort it by release date by filling in the missing ones
[...]

PR #901, opened 2024-09-26
Sort albums by PremiereDate on artist screen

Maxr1998:
This ensures albums are correctly ordered by release date, if there are more than one within the same year.
I am not sure if the `ProductionYear` sort is still needed here, but it doesn't hurt, I suppose. If the `PremiereDate` is empty, that would mean it's at least sorted by `ProductionYear` like before.

### DIFF
--- a/lib/components/ArtistScreen/artist_screen_content.dart
+++ b/lib/components/ArtistScreen/artist_screen_content.dart
@@ -91,11 +91,11 @@ class _ArtistScreenContentState extends State<ArtistScreenContent> {
           )
         else
           Future.value(null),
-        // Get Albums sorted by Production Year and Premiere Date
+        // Get Albums sorted by Premiere Date
         jellyfinApiHelper.getItems(
           parentItem: widget.parent,
           filters: "Artist=${widget.parent.name}",
-          sortBy: "ProductionYear,PremiereDate",
+          sortBy: "PremiereDate",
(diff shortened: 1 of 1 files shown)


Maxr1998:
For what it's worth, the offline sort algorithm already only uses the`PremiereDate`.

Chaphasilor:
Thanks, but I'm having a deja-vu 😅  
https://github.com/jmshrv/finamp/pull/448

Or did that never end up in `redesign`?  
Also, there's a related issue: #536  
Is that getting addressed here? 🤔

Maxr1998:
Good memory! The previous commit did end up in `redesign`, but as the artist screen got a separate interface, my change was essentially dropped (as `music_screen_tab_view` doesn't handle artists anymore).

The linked issue is a bit of a library issue. If the albums don't have years applied for whatever reason, sorting by either `ProductionYear` or `PremiereDate` will obviously fail. I agree that adding the `SortName` a sort parameter would fix this.
My comment in my previous PR reminded me of something, though. It appears that using `ProductionYear,PremiereDate` as a sorting is pretty useless, since `PremiereDate` already falls back to `ProductionYear` in the server source.

Chaphasilor:
Okay, that seems good! Right, at some point artist were just a list of albums...

The changes look good to me. I'll go ahead and merge them :)

PR #903, opened 2024-09-26
fix: login screens side padding

edeuss:
Fixes the cramped feeling on the page change by moving the padding to the SingleChildScrollView's.

### DIFF
--- a/lib/components/LoginScreen/login_authentication_page.dart
+++ b/lib/components/LoginScreen/login_authentication_page.dart
@@ -54,6 +54,7 @@ class _LoginAuthenticationPageState extends State<LoginAuthenticationPage> {
   Widget build(BuildContext context) {
     return Scaffold(
       body: SingleChildScrollView(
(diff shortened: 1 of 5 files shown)


Chaphasilor:
Thanks for the PR! I just took a look but honestly can't tell a difference on my phone (Asus Zenfone 10). Even took screenshots for side-by-side comparisons.  
Would you mind also sharing screenshots and letting me know which device you're using? :D

edeuss:
@Chaphasilor it's only a visual difference on the transition between pages. The embedded navigator uses the whole width instead of just part of the screen.

I can add a comparison video if needed.

Chaphasilor:
Ahh, it's the transition! I see it now :D  
Yeah that's much better. My original plan was to make the Finamp icon a hero widget that morphs and shrinks to the top after the first page, but that didn't work out for some reason. But the swapped padding looks much better for sure!  
Thanks, merging this ^^

PR #904, opened 2024-09-26
feat: replace blurry png logo with svg

edeuss:
Upped the resolution to the same as the Jellyfin icon in the same folder. 

Might be a good idea to switch to use the SVG instead of an PNG.


Chaphasilor:
Thanks for the PR! We already have an SVG of the logo, should we just switch to using that right now? ^^

edeuss:
@Chaphasilor Switched to use an SVG, Keeping the image commit in the PR for now (looks like the image is used in other places too. I can switch the other places to use the SVG in another PR). Or I can make it apart of this one, thoughts?

edeuss:
Looking in the code it looks like there is only two other locations the image is used, the settings page and the drawer.

edeuss:
@Chaphasilor Added the last to pages to the PR, I have never had any real issues with SVG's on any of Flutters supported platforms.

Chaphasilor:
Got it. Is there any particular reason for creating a new SVG? Was the existing one in a wrong location or had too much padding?

Chaphasilor:
Ah yeah that tracks. I have no idea why it was done that way, but I'm open to restructuring things if that makes more sense! Applies to the entire project, if there's something odd that we could do better, we should do that :)

edeuss:
@Chaphasilor We could do that in a later pr. But I think this PR is good.

Chaphasilor:
Sounds good. Let's do it like that. Thanks for the PR!

PR #930, opened 2024-10-22
Improve locale handling

Maxr1998:
Some minor tweaks to date formatting and locale handling. It would be preferable to not squash these, but to do a rebase merge instead.

### DIFF
--- a/lib/components/LanguageSelectionScreen/language_list.dart
+++ b/lib/components/LanguageSelectionScreen/language_list.dart
@@ -67,19 +67,20 @@ class LanguageListTile extends StatelessWidget {
     this.locale,
   });
 
-  final Locale? locale;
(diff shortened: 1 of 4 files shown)


Maxr1998:
> Thanks!
> 
> Am I correct to assume you discovered this because the date/time strings were formatted using a different locale than your display language?  

Yup. I have selected "system" in the settings and my phone is set to German, but I got AM/PM before.

> We also have a "system" option in the app's language settings, can't that be used here instead of add in a new package? 

The package is already there, and we do use it already. It's just that it's not properly initialized as of right now.

> And you want a rebase because they are not all related to locale handling?

Exactly, one is just a cleanup commit, so I'd prefer to keep them separate.

Chaphasilor:
Awesome, thanks for the clarification!

PR #931, opened 2024-10-22
Fix pub get failure

Maxr1998:
Instead of loading isar_generator from a (now deleted) fork, we can also use a dependency override: isar/isar#1505


Chaphasilor:
Ah, a deleted fork! I was wondering what's going on there, but didn't investigate yet, so thanks for that.

PR #932, opened 2024-10-23
Add fast-forward/rewind buttons

vijayabhaskar78:
The recent update to MusicPlayerBackgroundTask adds configurable skip forward and backward controls with smart visibility features. The implementation includes skip forward (default 30 seconds) and backward (default 10 seconds) controls that can be customized through the new SkipControlSettings class.

The visibility of these controls is managed through a smart system that can automatically show or hide them based on content type. In automatic mode, the controls appear for content longer than 10 minutes or content tagged as podcasts, audiobooks, or speech. You can also configure the visibility to be always shown or never shown through the SkipButtonVisibility enum.

The skip controls are fully integrated with the existing media controls system and can optionally appear in the notification area. The implementation includes Android-specific icons for the skip buttons, which need to be added to the project's drawable resources (ic_skip_back_10 and ic_skip_forward_30).

To use the skip controls, you can either interact with the UI buttons that appear based on the visibility settings, or programmatically call skipForward() and skipBackward(). The skip durations, visibility behavior, and notification presence can all be customized through the SkipControlSettings class when initializing the player.

### DIFF
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -17,6 +17,26 @@ import 'finamp_settings_helper.dart';
 import 'finamp_user_helper.dart';
 import 'jellyfin_api_helper.dart';
 
+enum SkipButtonVisibility {
+  always,
+  automatic,
+  never,
+}
+
+class SkipControlSettings {
+  final Duration forwardSkipDuration;
+  final Duration backwardSkipDuration;
+  final SkipButtonVisibility visibility;
+  final bool showInNotification;
+
+  const SkipControlSettings({
+    this.forwardSkipDuration = const Duration(seconds: 30),
+    this.backwardSkipDuration = const Duration(seconds: 10),
+    this.visibility = SkipButtonVisibility.automatic,
+    this.showInNotification = true,
+  });
+}
+
 // Largely copied from just_audio's DefaultShuffleOrder, but with a mildly
 // stupid hack to insert() to make Play Next work
 class FinampShuffleOrder extends ShuffleOrder {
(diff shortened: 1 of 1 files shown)


Chaphasilor:
Hmm, that's a bit different from what you proposed in your recent comment.

Most importantly, please target the `redesign` branch with your PR. A lot of the functionality you're relying on is only available in that branch, and you should have an easier time adding your changes there because a lot of the groundwork has already been done.  
Aside from that, please consider the following:

- The settings and default values should be put into `finamp_models.dart` along with the other settings.
- The logic for showing the skip buttons should be a simple modification to the existing logic in `metadata_provider.dart`, where some variables are renamed to reflect that not just speed controls are affected by it, and then that `MetadataProvider` should be used to decide if the playback controls should be shown or not on the player screen. Right now you're not showing any new buttons on the player screen, just in the media notification
- You are specifying icons/drawables for the media notification, but these don't exist yet
- You are not using the existing [`fastForward`](https://pub.dev/documentation/audio_service/latest/audio_service/AudioHandler/fastForward.html) and [`rewind`](https://pub.dev/documentation/audio_service/latest/audio_service/AudioHandler/rewind.html) methods from `AudioService` (which is implemented by `MusicPlayerBackgroundTask`. You can also set the default fast-forward and [rewind interval](https://pub.dev/documentation/audio_service/latest/audio_service/AudioServiceConfig/rewindInterval.html) as part of the AudioServiceConfig specified when `MusicPlayerBackgroundTask` is being instantiated.
- You can configure if the fast-forward and rewind buttons are shown within the `_transformEvent()` method in `MusicPlayerBackgroundTask`. There you can access the `MetadataProvider` to determine if the controls should be shown or not. This should be very similar to what is being done to get the favorite status.
- Your PR doesn't compile since you're specifying the media controls incorrectly. Please make sure that your code compiles.

Chaphasilor:
hey @vijayabhaskar78, are you planning to implement the requested changes? :)

PR #936, opened 2024-10-29
Use same logic for playback stop events and offline listen log

Maxr1998:
The boolean expression we currently use for server stop events already works very well, so the same should be used while offline (in the `onTrackChanged` callback). I also added a logging call to the `OfflineListenLogHelper` for good measure.


Chaphasilor:
Thanks, will take a look tomorrow!  

One thing I thought about today, will the currently playing track be properly logged when going offline mid-playback? Probably yes, since it only logs when it ends (since your recent PR), right?

Maxr1998:
Yup, it should, since the check is indeed at the end. I *think* I even tested this once.

Chaphasilor:
Awesome, thanks!

PR #944, opened 2024-11-05
[Redesign] download Migration Fix

Komodo5197:
Skip adding viewIds to migrated playlists as are shouldn't have them.


Chaphasilor:
Thanks! I'll merge this right-away and do the testing in #829.
It could be that this didn't cause issues during migration with a 10.8 server (or are you still using that?). But this is definitely the better solution!

PR #946, opened 2024-11-07
Only show artist top songs with at least one play

Maxr1998:
Fixes the minor issue where the artist top songs section would also show songs with "0 plays".


Chaphasilor:
Makes sense, thanks! I'll make sure that empty lists don't cause any issues.

Maxr1998:
Good point, I didn't think of that. Thank you!

PR #954, opened 2024-11-14
[Redesign] Add SafeArea to playing bar restore widget

Komodo5197:
This prevents the restoring version of the playing bar from clipping under system UI elements.


Chaphasilor:
Ah, I briefly noticed that too but forgot about it again. Thanks for the fix!  

Only just saw the PR, I get too many notifications from GitHub...

Issue #758, opened 2024-05-29
delete button? 


picarica:
Hello i just wonder is there delet bytton? Like delete from server? I would like to delete some songs and do i need to open jellyfinbapp to delete it? And yes my account is administrator.
And yes i mean delete file from server not when files are downloaded for offline playback

Chaphasilor:
Hey, this is currently not possible. We could probably add this, but hide it behind a setting. I want to be very careful around deleting users' media, hope you understand.

I'll look into adding this together with deleting playlists.

picarica:
> Hey, this is currently not possible. We could probably add this, but hide it behind a setting. I want to be very careful around deleting users' media, hope you understand.
> 
> I'll look into adding this together with deleting playlists.

understantable and reasonable, just wish i could delete redundant media without using jellyfin app, or maybe some trash system would suffice.

Chaphasilor:
Yeah I have a playlist of songs to delete myself, so I know the problem! I'll see what I can do :)

NikMAX2077:
+1. A very much-needed feature.

Chaphasilor:
This will be added in the next beta update!

PR #992, opened 2024-12-29
Added Reset Buttons for every Settings Screen

flloschy:
Fyi: Some already existing update functions I used dont update the UI, dont know how to fix that

Also idk why `pubspec.lock` got updated, i guess thats normal?

Please double check the defaults ive put there, I believe they are the real defaults since I copied the settings I had which I didnt change much if at all

### DIFF
--- a/lib/components/AlbumScreen/download_button.dart
+++ b/lib/components/AlbumScreen/download_button.dart
@@ -94,7 +94,7 @@ class DownloadButton extends ConsumerWidget {
             confirmButtonText:
                 AppLocalizations.of(context)!.deleteDownloadsConfirmButtonText,
             abortButtonText:
-                AppLocalizations.of(context)!.deleteDownloadsAbortButtonText,
+                AppLocalizations.of(context)!.genericCancel,
             onConfirmed: () async {
               try {
                 await downloadsService.deleteDownload(stub: item);
--- a/lib/components/AlbumScreen/track_list_tile.dart
+++ b/lib/components/AlbumScreen/track_list_tile.dart
@@ -142,7 +142,7 @@ class TrackListTile extends StatelessWidget {
             viewFilter: finampUserHelper.currentUser?.currentView?.id,
             nullableViewFilters: settings.showDownloadsWithUnknownLibrary,
             onlyFavorites:
-                settings.onlyShowFavourite && settings.trackOfflineFavorites,
+                settings.onlyShowFavourites && settings.trackOfflineFavorites,
           );
 
           var items =
--- a/lib/components/DownloadsScreen/downloaded_items_list.dart
+++ b/lib/components/DownloadsScreen/downloaded_items_list.dart
@@ -61,7 +61,7 @@ class _DownloadedItemsListState extends State<DownloadedItemsList> {
                         confirmButtonText: AppLocalizations.of(context)!
                             .deleteDownloadsConfirmButtonText,
                         abortButtonText: AppLocalizations.of(context)!
-                            .deleteDownloadsAbortButtonText,
+                            .genericCancel,
                         onConfirmed: () async {
                           await downloadsService.deleteDownload(stub: album);
                           if (mounted) {
--- a/lib/components/InteractionSettingsScreen/keep_screen_on_while_charging_selector.dart
+++ b/lib/components/InteractionSettingsScreen/keep_screen_on_while_charging_selector.dart
@@ -19,7 +19,7 @@ class KeepScreenOnWhilePluggedInSelector extends StatelessWidget {
               AppLocalizations.of(context)!.keepScreenOnWhilePluggedInSubtitle),
           value: FinampSettingsHelper.finampSettings.keepScreenOnWhilePluggedIn,
           onChanged: (value) {
-            FinampSettingsHelper.setKeepScreenOnWhileCharging(value);
+            FinampSettingsHelper.setKeepScreenOnWhilePluggedIn(value);
           },
         );
       },
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -106,7 +106,7 @@ class _MusicScreenTabViewState extends State<MusicScreenTabView>
                 : "SortName"),
         sortOrder: sortOrder,
         searchTerm: widget.searchTerm?.trim(),
-        filters: settings.onlyShowFavourite ? "IsFavorite" : null,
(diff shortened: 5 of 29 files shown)


Chaphasilor:
Thanks for the PR!  

Changes look good, here are a few comments:
- The `pubspec.lock` file contains the current versions of all used packages (dependencies). In the main package file, `pubspec.yaml`, we list each needed package along with a version selector, usually something like `^1.2.3`, which means it can be updated to 1.x.y if a new version becomes available. This way packages can be automatically updated without explicitly changing the version. `pubspec.lock` then keeps track of which version was actually used. So yes, it's usually okay if it changes.
- The settings that don't automatically update are stateful child widgets, so their state (e.g. current text in a `TextField`) is preserved even when the parent widget is rebuilt. You can force a rebuild of the children by explicitly giving them a key, and changing that key in the setState() method. [Here's a longer explanation](https://www.perplexity.ai/search/flutter-reload-stateful-child-0ce7MRASTq6MT_8KeZhNDQ)
- You can run `dart format .` in your terminal to automatically apply the correct formatting to your code. Your editor (probably VS Code) also usually has an option to "format on save", which makes sure the code is correctly formatted right-away.  
If you run `dart format .`, make sure to do it at the very start or end, in a separate commit, so that it is easy to see what you actually changed aside from the formatting.
- Right now we're just using same values as the current default and manually setting them to reset everything. If someone changes the default, the reset would still use the old value. I think we should refactor things so that we directly use the default values from `finamp_models.dart` to stay consistent. Right now those are private (leading underscore, so only available in that file), which we would need to change. If you want to give it a try, go ahead! Otherwise I can also try it. I suggest we create a new class that contains all these values as static const properties.

Chaphasilor:
Oh and I just noticed you already added a button for resetting everything! That would've been my next step. Awesome!

Chaphasilor:
Just tested it out, everything seems to work. I have one more suggestion:  
A simple confirmation dialog would be nice, something like "Do you want to reset these settings back to defaults?" With "Reset" and "Cancel" options. To do this, the reset button could be made into a new widget (`SettingsResetButton`) that is passed a function, then shows the dialog on tap, and calls the function if the dialog is confirmed.  
That would also get rid of a ton of duplicate code :)  
There already is a `ConfirmationPromptDialog` that you can use for this!

flloschy:
Instead of a new Class I made a function in the settings helper, kinda against convention but i felt like its cleaner than having a new class.
At least as long as this button is only used for this specific use case.

Because the actions is now in a generic function I didn't implement the UI reloading, cant think of a nice way of doing this now :/ 

And of course app localization needs to be added to the 4 new strings that came with the function, dont know how though

flloschy:
Okay I tried this to force rebuild the UI but that didnt do anything 

```dart
class _AudioServiceSettingsScreenState
    extends State<AudioServiceSettingsScreen> {
  Key _childKey = UniqueKey();
  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(AppLocalizations.of(context)!.audioService),
        actions: [
          FinampSettingsHelper.makeSettingsResetButtonWithDialog(
              context, () {
                _childKey = UniqueKey();
                FinampSettingsHelper.resetAudioServiceSettings();
              })
        ],
      ),
      body: ListView(
        children: [
          if (Platform.isAndroid) StopForegroundSelector(key: _childKey),
          SongShuffleItemCountEditor(key: _childKey),
          BufferDurationListTile(key: _childKey),
          LoadQueueOnStartupSelector(key: _childKey),
          PeriodicPlaybackSessionUpdateFrequencyEditor(key: _childKey),
          ReportQueueToServerToggle(key: _childKey),
        ],
      ),
    );
  }
}

```

Chaphasilor:
Thanks! I mean it's not like we need to save on components, so maybe a regular widget is better :)  

For the state reloading, you can simply pass an anonymous function to `makeSettingsResetButtonWithDialog()` (or the new widget), like so: 

```dart
FinampSettingsHelper.makeSettingsResetButtonWithDialog(
  context,
  () {
      setState(() {
        FinampSettingsHelper.resetAlbumSettings();
        // optionally assign a new key for child widgets here
      });
    },
  )
```
That should work just fine.  
**Edit: I just saw that you did just that in the latest commit ^^**

As for localizations, you go into `app_en.arb` and add the missing tokens either at the end or next to some similar ones. Always add the token itself and some metadata (prefixed with an `@`) that includes a description of how/where the token will be used.  
Then you can either run the "Generate Localizations" command (in VS Code), or use the regular code generation command (from the contributing guidelines):  `dart run build_runner build --delete-conflicting-outputs`. This will register the new tokens so that you can use them in the code. After changing or adding a token, you need to manually hot-reload, since changing the `.arb` file won't trigger a reload.

flloschy:
I believe thats everything done you asked for :D

Chaphasilor:
Firstly, sorry for taking so long to get to this!

Yep, that looks pretty good! You added some new settings (like the download worker count), which is nice! But you didn't use these default settings in the Hive declaration (see Hive field 39), so the hard-coded values are still there. I'll go ahead and fix that :)

I'll also change the wording for the confirmation dialog slightly. I'm a fan of explicitly stating what is confirmed, so "Reset all settings" instead of "Yes I'm sure". This should force users to be more deliberate in their selection, and not just search for a default confirmation button to quickly dismiss the dialog. Not a big deal though.

One more thing: whenever you have Hive or Isar stuff (or any file that has a corresponding `.g.dart` file), you'll have to run code execution (with the `dart run build_runner build --delete-conflicting-outputs` command). Otherwise the generated files aren't updated, which will cause issues at some point.

Chaphasilor:
I added the missing settings and default values, renamed a few settings for consistency, and used them in a couple more place. There are probably a few more places throughout the code that still use hard-coded values, but we can try to replace them whenever we encounter them.

Aside from that everything looks good now, so thanks a ton for the PR!

flloschy:
No problem, Im happy to help out! 
Looking forward to contribute more with my new learned knowledge :D

Chaphasilor:
Anything you're interested in particularly?

flloschy:
Not really but if you got anything thats a bit more advanced that this im down for it

Chaphasilor:
Okay, how about https://github.com/jmshrv/finamp/issues/758 ?  
That should make you a bit more familiar with the API side. I'd suggest to add another entry at the bottom of the track menu for now, and optionally hide it behind a setting (not sure about that one). Either way, there should be a confirmation once again.

Your best friend will be https://api.jellyfin.org (sadly that site is very performance hungry).

flloschy:
Alright, I will see what I can put together!

PR #1001, opened 2025-01-05
[Redesign] Additional fixes to avoid bottom system UI

Komodo5197:
Fixes missing bottom padding to allow scrolling all elements past the bottom system UI when now playing bar is inactive and on all screens using CustomScrollViews.

### DIFF
--- a/lib/components/AlbumScreen/album_screen_content.dart
+++ b/lib/components/AlbumScreen/album_screen_content.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:flutter_sticky_header/flutter_sticky_header.dart';
@@ -7,6 +6,7 @@ import '../../components/favourite_button.dart';
 import '../../models/finamp_models.dart';
 import '../../models/jellyfin_models.dart';
 import '../../services/finamp_settings_helper.dart';
+import '../padded_custom_scrollview.dart';
 import 'album_screen_content_flexible_space_bar.dart';
 import 'download_button.dart';
 import 'playlist_name_edit_button.dart';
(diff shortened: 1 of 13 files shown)


Chaphasilor:
Nice, thanks for that. IMO we can even add a good chunk of additional padding, so that the last list element scrolls to about half the screen height. That very clearly signals that the end of the list has been reached. I've been meaning to add this for a while, but never got around to it. Adding a static (and configurable) bottom padding should be easy, looking at the code.

Also, I noticed that the now playing bar still sometimes shifts up when loading. I remember you adding a missing SafeArea a while back, so I'm not sure what the issue is there...

Komodo5197:
I've added an argument for specifying additional bottom padding, but upped the padding on anything.  Half a screen seems a bit much, I'd lean towards 1 or two items of overscroll max myself, maybe even less.  Although the scroll physics do a decent job of communicating the end of lists already which is presumably why the system apps don't feel the need for overscroll - the only place it feels off is the album lists where it always feels like there could be a track hidden behind the now playing bar even though there isn't.  Feel free to tune it however you like, but I don't feel the need to mess with it much myself.  Regarding the now playing bar, I don't see a shift in this build.  Is it possible you're thinking of the actual beta build, which still doesn't have that change released?

Chaphasilor:
Okay, went with a padding of 32, slightly less than the default list tile, but gives a nice visual indication that no other tile is coming after the last one. Also didn't look weird on the artist screen, where larger paddings did. I had to exclude the drawer, the padding doesn't do any good there. Everything else seemed to behave as it should.  
Thanks again for the PR!

Chaphasilor:
@Komodo5197 before I forget, I just noticed some padding issues on the playlist menu. I'm not sure if they're related to this, would you mind taking a look?

Komodo5197:
This seems to be an unrelated issue with the cachepaint stuff.  I've created a PR with a fix.

PR #1002, opened 2025-01-07
Add fallback cover

Chaphasilor:
This is an old experimental branch, but since missing covers still result in a blank player screen with gray accent color, I thought it probably won't hurt to add this in.  
The placeholder can be improved later on, once we know that it actually works.


Chaphasilor:
@Komodo5197 would appreciate if you could take a look at if what I did to `albumImageProvider` makes sense. For some reason in the old version I didn't have issues with returning a Future, but that's probably just something that got lost im the last rebase.  
Just using the AsyncValue seems to work from a type checking perspective, but I'm not sure if this breaks reactivity in some way.

lib/services/queue_service.dart
```diff
@@ -1103,13 +1104,19 @@ class QueueService {
       }
     }
 
+    final imageUri = artUri ?? (await getFallbackImageFile()).uri;
+
+    _queueServiceLogger.fine(
+      "imageUri: $imageUri\ndownloadedImage: ${downloadedImage?.file != null}\njellyfinUri: ${_jellyfinApiHelper.getImageUrl(item: item)}\nfallbackUri: ${(await getFallbackImageFile()).uri}\nitem.imageId: ${item.imageId}\nitem.imageTags: ${item.imageTags}",
```
Komodo5197: Do we really need this much information logged?  Also isn't the fallback uri consistent across every call?

lib/services/album_image_provider.dart
```diff
@@ -38,8 +43,8 @@ class AlbumImageRequest {
 
 final Map<String?, AlbumImageRequest> albumRequestsCache = {};
 
-final AutoDisposeProviderFamily<ImageProvider?, AlbumImageRequest> albumImageProvider = Provider.autoDispose
-    .family<ImageProvider?, AlbumImageRequest>((ref, request) {
+final AutoDisposeFutureProviderFamily<ImageProvider?, AlbumImageRequest> albumImageProvider = FutureProvider.autoDispose
```
Komodo5197: I would strongly prefer to avoid returning a future here.  It will result in a lot of widget rebuilds we don't really need.  We should just set up this file and store its path in main, so we can then just return that synchronously.

lib/services/album_image_provider.dart
```diff
@@ -58,17 +63,17 @@ final AutoDisposeProviderFamily<ImageProvider?, AlbumImageRequest> albumImagePro
       });
 
       if (request.item.imageId == null) {
-        return null;
+        return FileImage(await getFallbackImageFile());
```
Komodo5197: You probably need the scale: .25 to make this look correct on some larger uses of the album cover.  Also, you might want to use resizeImage as well, depending on how big this fallback is.  If its too large, you can fill up the image cache with a bunch of copies of a large image displayed at a small size.

lib/services/queue_service.dart
Chaphasilor: might's broken that log during a rebase, was probably more useful before. I wanted to log this verbosely to catch and inconsistencies where no cover was shown, but I'm not sure if this is still needed.  
Do you want me to remove it? :)

lib/services/album_image_provider.dart
Chaphasilor: That's a good idea

lib/services/queue_service.dart
Chaphasilor: I guess this might've just been a debug log, I only now noticed the manual line breaks. I've removed it now :)

lib/services/album_image_provider.dart
Komodo5197: The three returns are the same, you should also resize image on the other 2.

Chaphasilor: oh okay

Chaphasilor: done :)

PR #1004, opened 2025-01-09
Delete from server button

flloschy:
A delete button to remove tracks form the servers filesystem/library 
Request from https://github.com/jmshrv/finamp/issues/758


Chaphasilor:
> No description provided.

🧐

flloschy:
Im bad at writing these things :d

Chaphasilor:
@flloschy take a look at `metadata_provider.dart`, and how it's used in the menu:

https://github.com/flloschy/finamp/blob/f5da742a56fe9c9dadd5413239271dce5c21b041/lib/components/AlbumScreen/song_menu.dart#L224-L230

This is the best spot to implement the deletability checks.

As for the download button itself, you've re-implemented the local part of it, which you can find here:
https://github.com/flloschy/finamp/blob/f5da742a56fe9c9dadd5413239271dce5c21b041/lib/components/AlbumScreen/song_menu.dart#L426-L486

The delete button only shows up once you've actually downloaded a track, naturally. So try doing that first. A track can only be deleted if it was downloaded directly, and not as part of an album. If it *is* part of a downloaded album/artist/playlist, the menu will only show the "lock" button, which will separately mark the track as required for downloading. This means that even if the album is deleted, the track will stay downloaded until explicitly deleted.  
It's currently not possible to exclude certain tracks/items from existing downloads (like downloading an entire album except for one track), so deleting a local track is only possible if it was downloaded directly.  
So the confirmation dialog only needs to contain the alternative button if the track can actually be deleted from the device. You can take a look how that check is done.

Deleting a track from the download system then works like this:

```dart
var item = DownloadStub.fromItem(
    type: DownloadItemType.song, item: widget.item);
unawaited(downloadsService.deleteDownload(stub: item));
```

Chaphasilor:
Also, you could try to (conditionally) extend the existing confirmation dialog with additional buttons instead of creating a new widget. That would also help staying consistent with the background colors and such ^^

flloschy:
Thats kinda what I did in [lib/components/confirm_prompt_dialog_with_actions.dart](https://github.com/jmshrv/finamp/pull/1004/files#diff-fd953b6f903f769026f1530f2e78b206745b8c11af02e0396d904733b3e3ff87)
but its a sketchy workaround because a ListView broke the rendering so i made the buttons wide and now they dont order them selfs vertically if the window is wide enough so I decided to do the second menu option instead

flloschy:
Ohhhhhhhhh there is already a delete from device button which only shows up when a track is separately downloaded... So that makes things easier easier for me xd

flloschy:
I updated the already existing delete button to adapt to the 2 deletion types: server and client

If only one of the two is possible its reflected in the button icon and label, and I added the confirm dialog which was missing previously 

If its both the name is more generic with just "Delete", when pressed the user will be ask if they want to delete from Server, Client or both, when clicked it will ask one more time to verify the deletion

Chaphasilor:
Let's keep the server delete separate please. It shouldn't be as common as deleting a download. The prompt was meant as a way to say "hey, are you sure you don't just want to remove the download?" instead of "here are your options, choose whichever you like".

I even thought about a separate setting that controls if deleting from the server is possible at all, which defaults to false.

So just leave the existing button like it was, and add a separate buttom for deleting from the server :)

F-4Dev:
I added a delete-button to the context menu of the album and playlist. @flloschy  and I thought it was a good idea to put this in the same PR.
The page doesn't get refreshed after deletion at the moment. @Chaphasilor  I might need your help for that.

Chaphasilor:
TIL you can collaborate on a PR xD
@F-4Dev maybe it would be better to only have the delete option when actually looking at the album/playlist, instead of in the list of all albums. To refresh, take a look at the search functionality, I think that does something similar (i.e. when closing the search bar).  
I'm also not sure if deleting playlists should be included here, because that work's a bit different, right? Playlist improvements (reordering, renaming, etc.) are also planned for the future, so we could do that in one go.

F-4Dev:
@Chaphasilor I think having the delete option there is pretty useful. If you want to delete multiple Album/Playlists  you don't have tap on them first and then delete them. Maybe you could also create an batch/multi select feature in the future. Personally I think deleting should be possible, both in the list of all albums and at the album screen.
From what I have seen (and it's not that much "^^) I also think that the code is pretty similar. Deleting a playlist and deleting an album uses the same api `_jellyfinApiHelper.deleteItem(widget.album.id);`

Also Playlists renaming and reordering etc would be awesome :)

lib/components/MusicScreen/album_item.dart
```diff
@@ -107,21 +109,52 @@ class _AlbumItemState extends ConsumerState<AlbumItem> {
...
+    if (isOffline) return;
+    // check whether widget is album or playlist
+    if (!(widget.isPlaylist || widget.album.type == "MusicAlbum")) return;
+    // make sure to only update once, else it will be stuck in a loop
+    if (deletableGotUpdated) return;
+    // prevent second execution
+    deletableGotUpdated = true;
+    // no need to bother if setting is disabled
+    Box<FinampSettings> box = Hive.box<FinampSettings>("FinampSettings");
+    if (!(box.get("FinampSettings")?.allowDeleteFromServer ?? false)) return;
```
Chaphasilor, line 129: you might wanna move that check further up

lib/components/MusicScreen/album_item.dart
```diff
@@ -274,6 +307,16 @@ class _AlbumItemState extends ConsumerState<AlbumItem> {
...
               ),
             ),
+
+          if (canDeleteFromServer)
+            PopupMenuItem<_AlbumListTileMenuItems>(
+              value: _AlbumListTileMenuItems.deleteFromServer,
+              enabled: canDeleteFromServer,
+              child: ListTile(
+                leading: const Icon(Icons.delete_forever),
+                title: Text("Delete from Server"),
```
Chaphasilor, line 317: please add a proper translation token for this

F-4Dev: https://github.com/jmshrv/finamp/pull/1004/commits/902dc53cabd784601b641dcef750364d9143276a
I added a all translation tokens for this PR ^^

lib/components/MusicScreen/album_item.dart
F-4Dev: Done [😎 🚀 🔥](https://github.com/jmshrv/finamp/pull/1004/commits/902dc53cabd784601b641dcef750364d9143276a#diff-87cdd9f8734592c56f5eb69a2a70c3f4e6eedc76ad08e1688e436715a53b9167R119-R120)

lib/components/MusicScreen/album_item.dart
```diff
@@ -115,7 +115,9 @@ class _AlbumItemState extends ConsumerState<AlbumItem> {
   bool deletableGotUpdated = false;
   bool canDeleteFromServer = false;
 
-  void checkAllowDeleteFromServer() {
+  void checkAllowDeleteFromServer() {    
+    // no need to bother if setting is disabled
+    if (!(box.get("FinampSettings")?.allowDeleteFromServer ?? false)) return;
```
Chaphasilor, line 120: I think you forgot the line that gets the `box` variable

flloschy: Its now outside the function on line 112

Chaphasilor: Ah so it was duplicate before. That's why there was no change highlighted ^^

Chaphasilor:
@flloschy @F-4Dev if this is ready for review and merging, please mark it as such! :)

flloschy:
@F-4Dev has a problem with his part where the delete button only appears the second time the context menu opens after changing the settings. The confirm dialog is missing, reloading of album/playlist view after deleting and a way to delete albums/playlists in their respective "full screen" view.

I believe my additions are done as far as i can tell, though id like to remove some duplicate code after everything seams to be ready.

I could also help out a bit more but im scared of merge issues when we both work on something in parallel xd

Chaphasilor:
@flloschy if you have time to spare you could also open a new PR and work on a separate issue there! 😁

flloschy:
Anything particular in mind?

Chaphasilor:
Nothing particular, pick your poison.  
There are enough bugs to fix, but it would also help if a few more screens would get a fresh coat of paint. Settings could be a good place to start, especially since you've already worked on those before. You can find some recent mockups in the `#design` channel on discord...

flloschy:
Yesterday i was a bit tired so today it is!

Added said confirm prompt that was missing for playlists and albums

Cleaned the code a bit by putting the duplicate confirm dialogs into `download_service.dart` which might be the wrong place but I thought it would fit since the local delete function is located there. With this I compacted the translations to be adaptive instead of having one prompt for each situation, thats why some translations are now removed.

Updating the album/playlist-list still needs to be done though

lib/services/downloads_service.dart
```diff
@@ -505,15 +505,24 @@ class DownloadsService {
...
   }
 
+   String _convertItemTypeToTranslation(String type) {
+    switch (type) {
+      case "MusicAlbum": return "album";
+      case "Playlist": return "playlist";
+      default: return type;
+    }
+   }
+
```
Chaphasilor: I think there might already be something similar in either `finamp_models` or `jellyfin_models`. Probably in some `toString()` method

lib/services/downloads_service.dart
```diff
@@ -525,9 +534,10 @@ class DownloadsService {
   }
 
   /// type needs to be a key of the itemDeletedSnackbar replacement
-  /// /// As of writing this code the following options exist: album, playlist, artist, genre, song, library
+  /// As of writing this code the following options exist: album, playlist, artist, genre, song, library
```
Chaphasilor, line 540: this could be converted to an enum (or maybe there already is a type for that!)

lib/services/downloads_service.dart
```diff
@@ -561,6 +571,29 @@ class DownloadsService {
             centerText: true));
   }
 
+  Future<bool> canDeleteFromServer({required String itemId, alreadyChecked = false}) async {
```
Chaphasilor: I feel like this function could also feel at home in the `jellyfin_api_helper`, after all that's an abstraction of the actual `jellyfin_api`, and it doesn't interact with the actual download system.

Chaphasilor:
Thanks!

Let's keep widgets separate from logic please :) If you feel like it doesn't belong anywhere, feel free to create a new file! But anything in the `services` directory shouldn't contain widgets...  

I'm not a fan of the wording in the confirmation dialogs, especially the whole "Attention" stuff even when just deleting from the device. But I can reword it later on.  
I also noticed that the alternative option for the delete from server button isn't there, maybe I wasn't super clear there. Just add a third button only to the delete from server dialog, that says "delete only from device instead" and does the same thing as the regular delete button (but without confirmation).
After deleting (either case), you should call `Navigator.of(context).pop()` to dismiss the menu. Same when deleting an album (or playlist) from the album screen

The actual deletion seems to work, nice job.

Do you need help with refreshing the lists, or do you want to figure things out yourself?

lib/services/downloads_service.dart
flloschy: jup, but I changed my code to not need this anymore at all (will push later)

lib/services/downloads_service.dart
flloschy: I also changed this to not be a required parameter anymore, dont really know why this should have been a parameter in the first place (will push later)

lib/services/downloads_service.dart
flloschy: Fair, moved it to the api helper (will push later)

flloschy:
I already made this kind of dialog [here](https://github.com/jmshrv/finamp/pull/1004/files#diff-fd953b6f903f769026f1530f2e78b206745b8c11af02e0396d904733b3e3ff87) but I really dont like how this works or looks, just a hack-y solution by me. Soo I figured why not put it in the menu with all the other stuff, I think its just a bit more user friendly. Additionally its only visible when the setting is active, so most users wont see this anyway and if they see it they asked for it.

In other news, the `Navigator.of(context).pop()` breaks finamp xd 
It results in a blackscreen. I guess because the prompt already closes itself

Chaphasilor:
Okay, let's leave the dialogs the way they are for now.
The pop() not working is unexpected, take a look at how the other options are doing it then? If the menu closes afterwards that's all we want. You should only call this in the 
track/song menu and on the album screen. The album list long-press menu isn't a new context, so that won't work.

Oh an it seems like you got a typo in the new file name, it should either be `prompts` or `prompt` :D

For refreshing, try to call the refresh function that is called when dragging down on the album list :)

flloschy:
The Automatic navigation out of the album/playlist view doesnt work, Probably because the context is the wrong context? At least thats what im assuming

Chaphasilor:
Whoops, didn't mean to mark it as ready yet, but might as well. Now we get build checks too.

Wrong context could be. I'm a bit hazy on the details myself. Although it should be possible to get the parent context somehow. I assume you're doing the same thing the back button on that screen (called FinampAppbarButton I belive) is doing?

Maybe you could use a callback for it that you pass down from the parent context?

flloschy:
I always get this error when tried to pop the context, maybe because the function gets called after an await? Because the back navigation works on local delete 

I both parsed the context of the parent and used a callback from the parent (and parents parent and finally from main.dart)

```
══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY
╞═════════════════════════════════════════════════════════
The following assertion was thrown during a task callback:
Looking up a deactivated widget's ancestor is unsafe.
At this point the state of the widget's element tree is no longer stable.
To safely refer to a widget's ancestor in its dispose() method, save a reference to the
ancestor by
calling dependOnInheritedWidgetOfExactType() in the widget's didChangeDependencies() method.

When the exception was thrown, this was the stack:
#0      Element._debugCheckStateIsActiveForAncestorLookup.<anonymous closure>
(package:flutter/src/widgets/framework.dart:4963:9)
#1      Element._debugCheckStateIsActiveForAncestorLookup
(package:flutter/src/widgets/framework.dart:4977:6)
#2      Element.dependOnInheritedWidgetOfExactType
(package:flutter/src/widgets/framework.dart:4997:12)
#3      Localizations.of (package:flutter/src/widgets/localizations.dart:549:17)
#4      AppLocalizations.of (package:flutter_gen/gen_l10n/app_localizations.dart:100:26)
#5      askBeforeDeleteDownloadFromServer.<anonymous closure>.<anonymous closure>.<anonymous
closure>.<anonymous closure> (package:finamp/components/delete_prompts.dart:58:43)
#6      GlobalSnackbar._message (package:finamp/components/global_snackbar.dart:74:23)
#7      GlobalSnackbar.message.<anonymous closure>
(package:finamp/components/global_snackbar.dart:71:22)
#8      GlobalSnackbar._enqueue.<anonymous closure>
(package:finamp/components/global_snackbar.dart:34:56)
#9      _TaskEntry.run.<anonymous closure> (package:flutter/src/scheduler/binding.dart:101:32)
#10     Timeline.timeSync (dart:developer/timeline.dart:188:22)
#11     _TaskEntry.run (package:flutter/src/scheduler/binding.dart:100:16)
#12     SchedulerBinding.handleEventLoopCallback
(package:flutter/src/scheduler/binding.dart:542:15)
#13     SchedulerBinding._runTasks (package:flutter/src/scheduler/binding.dart:520:9)
#17     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:194:12)
(elided 3 frames from class _Timer and dart:async-patch)


This exception was thrown in the context of a scheduler callback. When the scheduler callback
was _registered_ (as opposed to when the exception was thrown), this was the stack:
#0      new _TaskEntry.<anonymous closure> (package:flutter/src/scheduler/binding.dart:86:31)
#1      new _TaskEntry (package:flutter/src/scheduler/binding.dart:88:6)
#2      SchedulerBinding.scheduleTask (package:flutter/src/scheduler/binding.dart:486:33)
#3      GlobalSnackbar._enqueue (package:finamp/components/global_snackbar.dart:34:33)
#4      GlobalSnackbar.message (package:finamp/components/global_snackbar.dart:71:7)
#5      askBeforeDeleteDownloadFromServer.<anonymous closure>.<anonymous closure>.<anonymous
closure> (package:finamp/components/delete_prompts.dart:57:30)
<asynchronous suspension>
(elided 3 frames from dart:async)
═══════════════════════════════════════════════════════════════════════════════════════════════
═════
```

Chaphasilor:
Yes, that seems to be the reason. Isn't the local delete also using await?

flloschy:
There are both Async and sync delete methods iirc

flloschy:
Never mind, there isnt a Sync delete method for local deletes.

I mean the deletion could be put in an unawait() while navigating back one step. But I dont like the idea of being moved away from something although it fails... Even putting the redirect inside the .then() function doesnt prevent the issue

Komodo5197:
You should probably be using the context given to you by GlobalSnackbar.message in the localization callback.

Chaphasilor:
Redirect should only happen if the album/item isn't accessible afterwards, so when deleting from server *and* local device, or when only deleting from the local device but with offline mode enabled.

Komodo5197:
I'm specifically talking about applocalizations.of(context).  That's still using the widgets context in your latest patch.  The pop should probably be fine where it was?  But anything in the snack bar callback function should be using the callback context because the original widget may already have been torn down.

flloschy:
What da hec 
adding the `context.mounted` check works. I swear i tried this before and it didnt work xd

lib/components/delete_prompts.dart
```diff
@@ -0,0 +1,94 @@
...
+    BuildContext context, DownloadStub stub, {bool popIt = false}) async {
+  DownloadItemStatus status =
+      GetIt.instance<DownloadsService>().getStatus(stub, null);
+  String type = stub.baseItemType.name;
+
+  final jellyfinApiHelper = GetIt.instance<JellyfinApiHelper>();
+  final downloadsService = GetIt.instance<DownloadsService>();
+
+  String deleteType = status.isRequired
+      ? "canDelete"
```
Chaphasilor, line 51: lets add enums for these :)

lib/components/delete_prompts.dart
```diff
@@ -0,0 +1,94 @@
...
+              }
+            } catch (err) {
+              GlobalSnackbar.error(err);
+            }
+          },
+          onAborted: () {},
+          centerText: true));
+}
+
+Future<void> askBeforeDeleteDownloadFromServer(
```
Chaphasilor, line 41: can we rename this? we don't delete a download from the server. we delete a file/item from the server, and potentially also delete a download from the device :)

lib/components/delete_prompts.dart
flloschy: done in a future commit

Chaphasilor:
Since we're getting close to finalizing this, please fix the merge conflicts when pushing the other commits :)

flloschy:
Okay so, final big thing to do (i believe) is the reload when deleting from the album full screen view (works with the context menu inside the list itself).
I tried a few different ways but i cant get it to work...

Chaphasilor:
Reloading the list of albums, right?

What about refreshing the album when deleting a single track, does that work?

flloschy:
Yes reloading works when deleting a single item inside the list, not the full screen view

Chaphasilor:
okay, it took a while to finish this one up, but everything looks good now! this will be included in the next beta release.  
if you could give this (the `redesign` branch) another test before that, that would be nice, just to make sure I resolved the merge conflicts correctly :)

flloschy:
Umm 
image

Chaphasilor:
You'll need to clear storage first, since the Hive IDs werw shuffled around since the last version you had installed :)

flloschy:
Ohh yeah makes sense! 
Do you happen to know the storage location on linux? ^^

Chaphasilor:
The equivalent of `%appdata%`, probably somewhere in the home directory? But no idea really

flloschy:
Appears to work correctly
Well.. I cant delete an Album but I figure thats because I enabled read only? The error suggests its an server problem so probably just me
Playlist deletion works fine so everything should still be neato burrito

PR #1009, opened 2025-01-10
Fix settings drawer padding.

Komodo5197:
The drawer padding needs to be entirely handled by internal safearea due to use of SliverFillRemaining


Chaphasilor:
Alright, thanks! Didn't see your PR earlier, sorry about that!

PR #1013, opened 2025-01-11
add metadata helper classes for device, app, and server information.

aman-boop:


aman-boop:
also, as you asked me, now the finamp_log_helper.dart exports the logs as zip file including the metadata.json

PR #1016, opened 2025-01-12
Refactor logging system to use Hive for log storage and metadata managment.

aman-boop:
the Setup_loggin.dart now uses the metadata_helper.dart to fetch info

### DIFF
--- a/lib/services/finamp_logs_helper.dart
+++ b/lib/services/finamp_logs_helper.dart
@@ -1,48 +1,40 @@
 import 'dart:io';
-import 'dart:math';
-
+import 'package:archive/archive_io.dart';
 import 'package:clipboard/clipboard.dart';
 import 'package:finamp/services/censored_log.dart';
-import 'package:flutter/cupertino.dart';
-import 'package:flutter/foundation.dart';
+import 'package:finamp/services/metadata_helper.dart';
 import 'package:logging/logging.dart';
-import 'package:path/path.dart' as path_helper;
 import 'package:path_provider/path_provider.dart';
 import 'package:share_plus/share_plus.dart';
+import 'package:path/path.dart' as path_helper;
+import 'package:hive/hive.dart';
 
 class FinampLogsHelper {
   final List<LogRecord> logs = [];
-  IOSink? _logFileWriter;
+  late final Box logBox;
+
+  FinampLogsHelper() {
+    _initializeLogBox();
+  }
+
+  Future<void> _initializeLogBox() async {
+    logBox = await Hive.openBox('logs');
+  }
 
   Future<void> openLog() async {
-    WidgetsFlutterBinding.ensureInitialized();
-    final basePath = (Platform.isAndroid || Platform.isIOS)
-        ? await getApplicationDocumentsDirectory()
-        : await getApplicationSupportDirectory();
-    final logFile = File(path_helper.join(basePath.path, "finamp-logs.txt"));
-    if (logFile.existsSync() && logFile.lengthSync() >= 1024 * 1024 * 10) {
-      logFile
-          .renameSync(path_helper.join(basePath.path, "finamp-logs-old.txt"));
-    }
-    _logFileWriter = logFile.openWrite(mode: FileMode.writeOnlyAppend);
+    await _initializeLogBox();
   }
 
   void addLog(LogRecord log) {
     logs.add(log);
-    if (_logFileWriter != null) {
-      // This fails if we log an event before setting up userHelper
-      var message = log.censoredMessage;
-      if (log.stackTrace == null) {
-        // Truncate long messages from chopper, but leave long stack traces
-        message = message.substring(0, min(1024 * 5, message.length));
-      }
-      _logFileWriter!.writeln(message);
-    }
 
     // We don't want to keep logs forever due to memory constraints.
-    if (logs.length > (kDebugMode ? 10000 : 1000)) {
+    if (logs.length > 1000) {
       logs.removeAt(0);
     }
+
+    // Write log to Hive box
+    logBox.add(log.censoredMessage);
   }
 
   /// Sanitises all logs and returns a massive string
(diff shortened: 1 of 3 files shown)


aman-boop:
@Chaphasilor I don't know, I changed the target branch

lib/services/finamp_logs_helper.dart
```diff
@@ -62,32 +54,33 @@ class FinampLogsHelper {
...
-            mode: FileMode.writeOnly);
-      }
-      if (newLogs.existsSync()) {
-        await tempFile.writeAsBytes(await newLogs.readAsBytes(),
-            mode: FileMode.writeOnlyAppend);
-      }
-    } else {
-      await tempFile.writeAsString(getSanitisedLogs());
-    }
+    final zipFile = File(path_helper.join(tempDir.path, "finamp-logs.zip"));
```
Chaphasilor, line 57: can you add the timestamp to the filename here like I suggested on Discord a while ago?

lib/services/finamp_logs_helper.dart
```diff
@@ -62,32 +54,33 @@ class FinampLogsHelper {
...
+    encoder.create(zipFile.path);
+
+    // Add log file to the zip
+    final logFile = File(path_helper.join(tempDir.path, "finamp-logs.txt"));
+    await logFile.writeAsString(getSanitisedLogs());
+    encoder.addFile(logFile);
+
+    // Add metadata.json to the zip
+    final metadata = getIt<MetaData>();
+    await metadata.init();
```
Chaphasilor, line 70: we should probably call the `init()` function on startup, somewhere in `main.dart`

lib/services/finamp_logs_helper.dart
```diff
@@ -62,32 +54,33 @@ class FinampLogsHelper {
...
+    // Close the zip encoder
+    encoder.close();
 
+    // Share the zip file
+    final xFile = XFile(zipFile.path, mimeType: "application/zip");
     await Share.shareXFiles([xFile]);
 
-    await tempFile.delete();
+    // Clean up temporary files
+    await metadataFile.delete();
```
Chaphasilor, line 83: also clean up the logs text file?

lib/setup_logging.dart
```diff
@@ -3,14 +3,19 @@ import 'dart:io';
 import 'package:flutter/foundation.dart';
 import 'package:get_it/get_it.dart';
 import 'package:logging/logging.dart';
-import 'package:package_info_plus/package_info_plus.dart';
-import 'package:device_info_plus/device_info_plus.dart';
+import 'package:hive_flutter/hive_flutter.dart';
+//import 'package:package_info_plus/package_info_plus.dart';
+//import 'package:device_info_plus/device_info_plus.dart';
```
Chaphasilor, line 8: if they're not needed any longer, remove them completely please

lib/setup_logging.dart
```diff
@@ -31,35 +36,14 @@ Future<void> setupLogging() async {
   final startupLogger = Logger("Startup");
   startupLogger.info("App starting, logging initialized.");
 
-  final packageInfo = await PackageInfo.fromPlatform();
-  final deviceInfo = DeviceInfoPlugin();
+  final metadata = MetaData();
+  await metadata.init();
```
Chaphasilor, line 40: if `MetaData` a singledton or not? in `finamp_logs_helper.dart` you used `GetIt` to get the instance...

Chaphasilor (commented):
looks mostly good, but haven't tried it out yet. commented some minor changes, but this is definitely headed in the right direction! 😁

aman-boop:
I can't test it out myself

lib/services/finamp_logs_helper.dart
aman-boop: Main.dart is scary for me

aman-boop:
@Chaphasilor upstream and my forked version of finamp_logs_helper looks a lot different

lib/main.dart
```diff
@@ -76,11 +76,13 @@ import 'services/music_player_background_task.dart';
 import 'services/theme_mode_helper.dart';
 import 'services/ui_overlay_setter_observer.dart';
 import 'setup_logging.dart';
+import 'services/metadata_helper.dart';
 
 void main() async {
   // If the app has failed, this is set to true. If true, we don't attempt to run the main app since the error app has started.
   bool hasFailed = false;
   try {
+    await setupLocator();
```
Chaphasilor, line 85: please either namespace this (`import 'services/metadata_helper.dart' as metadataHelper;` and then `await metadataHelper.setupLocator()`), or move the setup function to main.dart (like all the other setup functions).

lib/services/metadata_helper.dart
```diff
@@ -0,0 +1,156 @@
...
+        };
+    }
+  }
+}
+
+Future setupLocator() async {
+  getIt.registerLazySingleton(() => DeviceInfo());
+  getIt.registerLazySingleton(() => AppInfo());
+  getIt.registerLazySingleton(() => ServerInfo());
+  getIt.registerLazySingleton(() => MetaData());
```
Chaphasilor, line 155: I don't think it's necessary to register four new global singletons. The `MetaData` class should be enough, and you could add any methods that need to be accessed into it. Or provide getters for the other classes through `MetaData`.

Also, maybe rename it to `MetaDataHelper`?

lib/setup_logging.dart
```diff
@@ -31,35 +34,14 @@ Future<void> setupLogging() async {
...
-    deviceInfoString =
-        "Windows ${windowsInfo.displayVersion} on ${windowsInfo.deviceId}";
-  } else {
-    final webInfo = await deviceInfo.webBrowserInfo;
-    deviceInfoString =
-        "Web browser ${webInfo.userAgent} on ${webInfo.platform}";
-  }
-  startupLogger.info("Running on $deviceInfoString.");
+  startupLogger.info(
+      "This is ${packageInfo.version} version ${packageInfo.version}+${packageInfo.originallyInstalledVersion} (Signature '${packageInfo.installationDate}'), installed via ${packageInfo.updateHistory}.");
```
Chaphasilor, line 44: Please don't change the information that is logged. The missing information (like `appName`, `buildNumber`, etc. should be added to the `AppInfo` class)

lib/services/metadata_helper.dart
```diff
@@ -0,0 +1,156 @@
...
+    } else if (Platform.isLinux) {
+      LinuxDeviceInfo linuxInfo = await deviceInfo.linuxInfo;
+      osVersion = linuxInfo.version ?? 'Unknown';
+      deviceModel = linuxInfo.id;
+      deviceType = "PC"; // You can add more logic to determine the device type
+    } else if (Platform.isWindows) {
+      WindowsDeviceInfo windowsInfo = await deviceInfo.windowsInfo;
+      osVersion = windowsInfo.displayVersion;
+      deviceModel = windowsInfo.deviceId;
+      deviceType = "PC"; // You can add more logic to determine the device type
```
Chaphasilor, line 38: how about you just add `platform` instead of device type, and set them accordingly (Android, Windows, Linux, etc.)?

lib/services/metadata_helper.dart
```diff
@@ -0,0 +1,156 @@
...
+      case MetaDataType.ServerInfo:
+        return serverInfo.toJson();
+      case MetaDataType.DeviceInfo:
+        return deviceInfo.toJson();
+      default:
+        return {
+          "appInfo": appInfo.toJson(),
+          "serverInfo": serverInfo.toJson(),
+          "deviceInfo": deviceInfo.toJson(),
+        };
```
Chaphasilor, line 146: This doesn't generate JSON. The resulting string has no quotes at all. Please use the `@JsonSerializable` annotation and `toJSON` method like it's done in `finamp_models.dart`. You'll have to run code generation after adding the annotation. You can do that by running `dart run build_runner build --delete-conflicting-outputs`.

Chaphasilor (changes requested):
Nice progress! Exporting the zip file works, but there are some additional changes needed :)

PR #1022, opened 2025-01-15
[Redesign] Cache paint sizing fix

Komodo5197:
Fix not drawing full screen sized image when caching blurred background images.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -62,6 +62,7 @@ android {
         }
         debug {
             applicationIdSuffix '.debug'
+            minifyEnabled true // avoid dexing issues
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Chaphasilor:
Wow that was fast! Trying it out now.

Chaphasilor:
Interestingly the debug build I had installed (which didn't contain your fix yet) didn't show the issue, so maybe it's only an issue in release build?  
I definitely don't see the issue in the debug build, and I did try a `flutter run --release` (which uninstalled by regular Finamp version, yay, off to download everything again) but that also didn't have the issue (with your fix). So I'll just merge this in for now, as it doesn't seem to break things...  
Thanks!

Issue #1025, opened 2025-01-18
[Redesign] Random-Album-Order in Offline Mode always keeps the same album in first position


lukaslindnermusic:
The title says it 😊 

I tend sorting the Albums in random order to randomly get music suggested/displayed.
In Online Mode, this works fine, but while I am in Offline Mode, the first album always stays the same ("Planet Of The Apes").
Refreshs, closing and re-opening the app, switching tabs... nothing helps.

Chaphasilor:
This will be fixed with the next beta update :)

PR #1037, opened 2025-01-31
[Redesign] Fix sort order 'Random' while offline

Komodo5197:
Addresses #1025

### DIFF
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -1,18 +1,16 @@
 import 'dart:async';
 import 'dart:math';
 
-import 'package:collection/collection.dart';
 import 'package:finamp/components/Buttons/cta_medium.dart';
-import 'package:finamp/components/Buttons/simple_button.dart';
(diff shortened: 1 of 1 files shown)


Chaphasilor:
Thanks for the PR! Can't try it out right now but I have to reason to doubt that it works.

Maybe I'm just tired, but why didn't it work before? Can't be a limitation of `.sort()`, since the other methods don't show this behavior. Some implementation detail of `Random`?

Either way, a comment in the old location pointing to where the case is handled and why it shouldn't be handled in the switch might be good to avoid future issues!

Komodo5197:
The first item thing was because the nextInt upper bound is exclusive, so we could only produce -1,0 instead of -1,0,1.  But the whole idea of random comparisons to shuffle is bad anyways because testing still seemed to show a heavy bias towards items being close to the same positions every time, even after I fixed that.  You probably need a lot of comparisons to all go a single direction to have items really move in the list from their unsorted positions.

Chaphasilor:
Okay, yes that makes sense. Thanks for adding the comment!

PR #1044, opened 2025-02-10
[Redesign] Downloads Size Warning + other fixes

Komodo5197:
Added warning text to downloads dialog for downloads with more than 150 songs.  Download dialog will be shown if this triggers, even if no download settings are needed.  This necessitates a 1 request delay before the dialog shows, which for me can take quite a while if the artist page is still loading because my server is very slow, but which should hopefully be fairly quick for most people.

Additional fixes:
- Reworded custom download location warning to mention the issue with the "Music" directory.
- Padded minutes to two digits for any time over an hour in queue time remaining display.
- Altered app name of android debug builds to make them less confusing when a release build is installed.

### DIFF
--- a/lib/components/AlbumScreen/download_dialog.dart
+++ b/lib/components/AlbumScreen/download_dialog.dart
@@ -13,8 +13,6 @@ import '../../services/finamp_user_helper.dart';
 import '../../services/jellyfin_api_helper.dart';
 import '../global_snackbar.dart';
 
-const songWarningCutoff = 150;
-
 class DownloadDialog extends StatefulWidget {
   const DownloadDialog._build({
     super.key,
@@ -76,6 +74,7 @@ class DownloadDialog extends StatefulWidget {
     // where this can be determined in one query.
     JellyfinApiHelper jellyfinApiHelper = GetIt.instance<JellyfinApiHelper>();
     List<BaseItemDto>? children;
+    print((await jellyfinApiHelper.getItemById(item.id)).toJson());
     if ((item.baseItemType == BaseItemDtoType.album ||
             item.baseItemType == BaseItemDtoType.playlist) &&
         (needTranscode || songCount == null)) {
@@ -99,7 +98,8 @@ class DownloadDialog extends StatefulWidget {
 
     if (!needTranscode &&
         downloadLocation != null &&
-        (songCount ?? 0) < songWarningCutoff) {
+        (songCount ?? 0) <
+            FinampSettingsHelper.finampSettings.downloadSizeWarningCutoff) {
       final downloadsService = GetIt.instance<DownloadsService>();
       var profile = FinampSettingsHelper
                   .finampSettings.shouldTranscodeDownloads ==
@@ -225,7 +225,8 @@ class _DownloadDialogState extends State<DownloadDialog> {
                         .dontTranscode(originalDescription)),
                   )
                 ]),
-          if ((widget.songCount ?? 0) >= songWarningCutoff)
+          if ((widget.songCount ?? 0) >=
+              FinampSettingsHelper.finampSettings.downloadSizeWarningCutoff)
             Padding(
                 padding: const EdgeInsets.only(left: 8.0, right: 8.0, top: 8.0),
                 child: Text(AppLocalizations.of(context)!
--- a/lib/components/AlbumScreen/download_dialog.dart
+++ b/lib/components/AlbumScreen/download_dialog.dart
@@ -13,6 +13,8 @@ import '../../services/finamp_user_helper.dart';
 import '../../services/jellyfin_api_helper.dart';
 import '../global_snackbar.dart';
 
+const songWarningCutoff = 250;
+
 class DownloadDialog extends StatefulWidget {
   const DownloadDialog._build({
     super.key,
@@ -21,6 +23,7 @@ class DownloadDialog extends StatefulWidget {
     required this.downloadLocationId,
     required this.needsTranscode,
     required this.children,
+    required this.songCount,
   });
 
   final DownloadStub item;
@@ -28,6 +31,7 @@ class DownloadDialog extends StatefulWidget {
   final String? downloadLocationId;
   final bool needsTranscode;
   final List<BaseItemDto>? children;
+  final int? songCount;
 
   @override
   State<DownloadDialog> createState() => _DownloadDialogState();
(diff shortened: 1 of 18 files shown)


Chaphasilor:
The wonders of gradle :D  

Thanks for the PR and the additional fixes. I'll give it a try now.  
How about pro-actively making the needed request in the background, or somehow scraping the needed info from the regular requests that are made to load artist's albums and such?

lib/components/AlbumScreen/download_dialog.dart
```diff
@@ -65,7 +69,32 @@ class DownloadDialog extends StatefulWidget {
...
+          includeItemTypes: BaseItemDtoType.song.idString,
+          fields:
+              "${jellyfinApiHelper.defaultFields},MediaSources,MediaStreams");
+      songCount = children?.length;
+    } else if (item.baseItemType == BaseItemDtoType.artist ||
+        item.baseItemType == BaseItemDtoType.genre) {
+      children = await jellyfinApiHelper.getItems(
+        parentItem: item.baseItem!,
+        includeItemTypes: BaseItemDtoType.album.idString,
+      );
```
Chaphasilor, line 90: Just had an idea and confirmed with my server. The artist `BaseItemDto` also includes a `ChildCount`. I believe this includes both albums and individual tracks, but if so it should be possible to subtract the `AlbumCount` from the `ChildCount` to get the track count.  
That should save us a request, since we have the artist info when rendering the artist screen, and possibly even when redering the artist list.

Chaphasilor:
Hmm, track count doesn't match up between artist page and dialog 🤔

Komodo5197:
I believe that may be the artist vs album artist thing I was talking about.  205 songs by the artist, but 779 songs in albums with them as an album artist.  Dos that seem reasonable in this case?

lib/components/AlbumScreen/download_dialog.dart
```diff
@@ -13,6 +13,8 @@ import '../../services/finamp_user_helper.dart';
 import '../../services/jellyfin_api_helper.dart';
 import '../global_snackbar.dart';
 
+const songWarningCutoff = 150;
```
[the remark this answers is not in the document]
Komodo5197: We already have plenty of constants around.  If you actually think it might make sense to have this configurable, I can just add the setting now.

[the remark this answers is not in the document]
Chaphasilor: Yes, that would be good.

lib/components/AlbumScreen/download_dialog.dart
Chaphasilor: You might have to include the field manually. I didn't try in Finamp, but checked the browser requests when opening an artist in jellyfin-web.

Chaphasilor:
Yes that could be. Jellyfin sometimes creates nonsensical albums with a bunch of random track in them, and tons of album artists.  

Which are we currently downloading?

Komodo5197:
We currently download the album artist.

lib/components/AlbumScreen/download_dialog.dart
Komodo5197: It seems like that only gets properly calculated if the item is requested individually?  Anyway, this seems to be the artist songcount instead of the album artist songCount anyways, so it isn't helpful.

Chaphasilor: Hmm, I see. It might be that it's only returned for individual artists, yes. But it might be a faster request than fetching all albums, not sure.

And wouldn't it still be useful as a threshold? You say we're downloading album artist, but that includes all tracks by that artist, doesn't it? It should, imo. When I download an artist, I'd expect every track and album by them to be downloaded.

Chaphasilor:
Hmm, seems like I wasn't really clear in what I meant. I didn't want to expose the setting yet, just have it be part of the `FinampSettings`. But I guess exposing it doesn't really hurt. So thanks for that, and sorry for not being clearer.  

I was also thinking about reorganizing settings soon, with each settings category having an "advanced" section where settings like these could find a home. Sometimes they're useful for debugging, but regular users shouldn't need to mess with them

Chaphasilor:
Oh and if I'm not mistaken the dialog opens quicker now on the artist screen, so using the fetched album list seems to work 👍🏻

lib/components/AlbumScreen/download_dialog.dart
Komodo5197: I mean that the data only seems to be filled in if you use getItem, singular, instead of getItems, like used on the music screen.  I'd imagine all the artist songs would usually be included, but that's only a lower bound so it's not really that helpful.  Fetching the album list exactly matches the downloader behavior, so it should be exactly accurate.  I don't know exactly how all the different requests work out and how jellyfin handles album artists and stuff, so I can't say for certain what happens.  Do our requests get get albums with a song with them as an artist, or jusat albums with them in the album artists field, and can any of this stuff be overridden by the user in unpredictable ways?  I pretty much only browse by album or playlist, so I don't have any hands on experience with how jellyfin handles artists/genres, or how finamp's online mode does either.

Komodo5197:
What I basically meant with my comment was that I didn't see any point putting a setting in that we 'might' want to expose, and then having to deal with all the hassle of codegen and maybe migration code if we ever want to change it.  Unless it's exposed to the user for change, it was fine as a constant local to the relevant code.

lib/components/AlbumScreen/download_dialog.dart
Chaphasilor: Honestly, me neither. I'll have to look into it when adding explicit support for it down the road. But i did notice some inconsistencies in offline mode, where some artist albums weren't shown on the artist screen, but I could see and play tracks from those missing albums, and use the player screen to navigate to the album screen. But that's a different issue.

So I guess we can keep it the way it is right now. The inconsistent track counts are weird, but they just reflect the weirdness of Finamp, and seem to be accurate, right?

Komodo5197:
I've fixed the display bug where the codec was being displayed on 'transcoded' collections with no actual audio files.  I tried to fix some issues I noticed where an image download with an outdated download location wouldn't update or show as failed.

Chaphasilor:
Is there any unresolved discussion for this, or can we merge it?

Komodo5197:
I believe everything has been resolved.

Chaphasilor:
Alright, thanks!

PR #1045, opened 2025-02-10
Add icon for F-Droid

leoheitmannruiz:
This is the icon shown at https://play.google.com/store/apps/details?id=com.unicornsonlsd.finamp, compressed by zopflipng.


Chaphasilor:
Thanks for the PR! This has been bugging me for a while now.  
Is Zopfli required for F-Droid to display it properly, or was it optional?

Chaphasilor:
Yeah I'm sure they do, to save on resources. But we can't do anything about that ^^  
Same as we can't do anything about them not properly supporting beta channels...

PR #1053, opened 2025-02-17
[Redesign] Adding Albums to Playlists and Private Playlists

Komodo5197:
Allows albums, artists, and the like to be added to playlists.  Jellyfin responds to this by adding all songs the item contains to the playlist without us needing to manually list them out.  Also, an option has been added allowing the user to select public or private visibility when creating a playlist.


Titaniumtown:
Thank you for the 2nd part w/ the Private Playlists! It's a pain to constantly set the playlists to private in jellyfin. <3

Chaphasilor:
Just got around to testing this, works well! I also like the warning about only being able to remove one-by-one. I'm amazed by how quickly this works too, I guess sometimes the server really helps instead of getting in the way ^^

Thanks for working on this!

PR #1059, opened 2025-02-20
Refactor "song" to "track" in the entire app.

lukaslindnermusic:
Exception: /lib/services/jellyfin_api.dart

I hope everything is working properly.
I tested the app briefly, and everything seemed to function as before, but I might have overlooked something.

I wasn’t sure what to do with the other language `.arb` files. Since the Polish file was causing my builds to fail, I deleted all language files except `app_en.arb` on my local branch.

How do the translations work? Do I have to edit the other language files (more specifically the keys) manually as well?

If the syntax errors in the Polish file are due to an issue with my local encoding or something similar, I’d be happy to get some help resolving it (I am new to Dart, Flutter and also Visual Studio Code and working with git repos... only did web programming before). 😊


lib/models/jellyfin_models.dart
```diff
@@ -2115,9 +2115,9 @@ class BaseItemDto with RunTimeTickDuration {
   @HiveField(130)
   int? albumCount;
 
-  /// Gets or sets the song count.
+  /// Gets or sets the track count.
   @HiveField(131)
-  int? songCount;
+  int? trackCount;
```
Chaphasilor, line 2120: we can't change this one. the variable has to be the same as in the Jellyfin API (except for capitalization), so that the automatic serialization and deserialization works.  
Line 1586 is affected as well.

The other changes should be fine, as they really are only internal.

Chaphasilor:
Thanks! I've been looking forward to this :D  

Regarding the translations, if you delete the files and we merge the PR, all translations would be deleted. So you should redo the deletion. The easiest way to do that is probably to switch to the redesign branch (where I've fixed the syntax error causing the build issues), and then copy all the translation file into a temporary folder, before switching back into your branch. There you can add the files back where they belong, and re-do the renaming afterwards.  
If you're unsure how to do that, I can also handle it for you :)

lukaslindnermusic:
[4ef762c](https://github.com/jmshrv/finamp/pull/1059/commits/4ef762c43b14732142f12354d28b5998cb4534e5) also includes the rollback of trackCount -> songCount.

Also, I noticed, that the german translations sometimes are really weird and fixed a lot of translations, as this is my main language. (I hope it is okay that I did this directly within this pull request?)

Chaphasilor:
Thanks for refactoring this! I'm sure we'll end up with a few mentions of "songs" again after a while, but those should be easily fixable 😁

Issue #1060, opened 2025-02-20
[Redesign] UI List Glitch in iOS Album, Genre and Artist Screens


lukaslindnermusic:
On iOS on the Album, Genre and Artist Screens, when scrolling down the lists, the top bar with the back button and the title moves up all the way to the end of the screen, clashing with the clock and battery iOS status bar.

![Image](https://github.com/user-attachments/assets/b0de4178-ef57-4697-a20e-adcd30849c39)

Already working on a fix 😊

PR #1061, opened 2025-02-20
iOS UI Fixes in Album/Artist/Genre Screen Lists.

lukaslindnermusic:
Scrolling down on a list in the album, artist or genre screen caused the title bar to clash with the iOS status bar.
Mentioned in https://github.com/jmshrv/finamp/issues/1060.

Added `SafeArea()` around the list bodies, now it looks like this:
Bildschirmfoto 2025-02-20 um 23 47 11


Chaphasilor:
Awesome, good catch and thanks for the fix!

Issue #948, opened 2024-11-08
[Redesign Beta] Player screen has the wrong transition direction


NutchapolSal:
The design seems to imply that the player screen should come in from the bottom, which tracks with the down arrow close button and the swipe down gesture. However currently the player does a transition from the right which implies a "swipe right to go back" but that conflicts with previous track gesture. A slide-in from bottom would make more sense.

maybe related to #619?
also does this have something to do with the side-by-side design on tablet?

Finamp version: 0.9.11
iOS version: 18.0.1

jmshrv:
The "lore" to this is that when I implemented it years ago as an Android user, the default animation happened to line up, and I didn't really test on iOS at that point. A proper swipe animation would be nice though :) (especially since I think Flutter/Android has changed the default animation?)

Chaphasilor:
@NutchapolSal this should be fixed since 0.9.15 :)

PR #1075, opened 2025-03-02
feat(now_playing_bar): slide up transition animation

edeuss:


Chaphasilor:
@edeuss I've made some tweaks to the animation and added a swipe up gesture. Please take a look and let me know what you think. I haven't tested the swipe gesture in a release build yet, but it should be smoother there. If you can figure out a way to use the gesture detector instead, without disabling the dismissible, that would be even better.

edeuss:
@Chaphasilor Not sure the fade out on close is better. Maybe a fade and size together?

I was thinking this a fast fix (since the moving from right to left kinda bugged me a lot lol), then was going to look into improving it to function more smoothly like the Spotify one.

Chaphasilor:
Sure, okay. But then I'd keep it as the fade for now, since the slide down looks a bit weird with the now playing bar.   
Does that work for you?

edeuss:
@Chaphasilor ok sure 👍

Chaphasilor:
@edeuss thanks! And just to be sure, this aims to fix https://github.com/jmshrv/finamp/issues/948, yes?

edeuss:
@Chaphasilor yes correct (didn't check if there was an issue open for it)

Issue #1079, opened 2025-03-04
[Bug] Duplicate buttons in album


Zumux:
In the top right corner of albums there is 2 download buttons. If you press to download you will briefly see 2 trash icons and 2 sync icons. They will then turn to just 2 trash icons.

Finamp Version: 0.9.14
Device: Google Pixel 6a
OS: GrapheneOS 14 (2025021100)

Chaphasilor:
Ah, thanks for reporting that! Looks like a botched merge, should be an easy fix!

edeuss:
@Chaphasilor Did a commit that fixes it for you.

theinfguy:
Spotted this as well, but good to see it’s already been dealt with!

Chaphasilor:
Fixed in [beta update 0.9.16](https://github.com/jmshrv/finamp/releases/tag/0.9.16-beta)!

PR #1081, opened 2025-03-04
fix(album_screen_content): double download button

edeuss:
Fixes: https://github.com/jmshrv/finamp/issues/1079


Komodo5197:
Looking at this, it seems there are logic errors in the code here introduced here by the delete from server, which does not cover incidental downloads.  Also, this logic should probably be inside the DownloadButton anyway.  I'll try to get a PR cleaning this up in sometime.

Issue #1070, opened 2025-03-01
[Feature Request] Sort Tracks by Number of Plays


dot-ammar:
Any plans to add a “Sort by Number of Plays” option for tracks? This is a feature available in the Jellyfin web client that I use often, and it would be great to have it in Finamp as well.

Chaphasilor:
Added in [beta update 0.9.16](https://github.com/jmshrv/finamp/releases/tag/0.9.16-beta)!  
Or was it 0.9.15?

PR #1087, opened 2025-03-05
sort by play count

Decimate1405:
- This time it's proper :) #1070

### DIFF
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -574,6 +574,12 @@ List<BaseItemDto> sortItems(
           } else {
             return a.premiereDate!.compareTo(b.premiereDate!);
           }
+        case SortBy.playCount:
+          if (a.userData?.playCount == null || b.userData?.playCount == null) {
+            return 0;
+          } else {
+            return a.userData!.playCount.compareTo(b.userData!.playCount);
+          }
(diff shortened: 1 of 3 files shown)


Chaphasilor:
I somehow missed this, but looking at it now :)

I think while Jellyfin offers a play count for things like albums and playlists, it never really gets updated, especially not when listening via Finamp. So this probably only really works for sorting tracks...

Chaphasilor:
Can you check if any of your albums or playlists have a play count other than 0? Otherwise I can merge this in but add a note ot the release modes. Or we could only show the sort option for tracks.

Decimate1405:
It only works for tracks, I didn't really see an option that shows playlist/album counts, if this is something that's already implemented I can dig into it more but atm it'd only for tracks (verified this)

Decimate1405:
Also, looking at the feature request they did mention "tracks" so I'm guess that's the scope of sorting by plays for now as jellyfin doesn't update the album/playlist plays

Chaphasilor:
Okay, I've adjusted the `SortByMenuButton` to only show the `playCount` option for the tracks tab, that should avoid confusion :) 
 
Playlist/album play counts are currently not shown, but theoretically the Jellyfin API returns play count data in the `UserData` section of their `BaseItemDto`s. But it seems like Jellyfin never actually updates the play count, to it's effectively useless for now.

Thanks for the PR!

PR #1092, opened 2025-03-05
feat: Music Video support

edeuss:
Allow playing media that is marked as music videos.


Chaphasilor:
Neat! I think there is a branch somewhere that allows selecting music video libraries on the library selection screen after login. Might be worth merging in!

PR #1095, opened 2025-03-07
Rename DownloadItemType.track back to song

Maxr1998:
The rename caused the entire library to be re-downloaded when repairing downloads, as Finamp uses the download type for the hash when computing the isarId.

### DIFF
--- a/lib/models/finamp_models.dart
+++ b/lib/models/finamp_models.dart
@@ -1312,7 +1312,11 @@ class DownloadItem extends DownloadStub {
 }
 
 /// The primary type of a DownloadItem.
+///
 /// Enumerated by Isar, do not modify order or delete existing entries.
+/// The enum name is used by `DownloadStub.getHash` to calculate the isarId for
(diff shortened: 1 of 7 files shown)


lib/models/finamp_models.dart
```diff
@@ -1315,7 +1315,7 @@ class DownloadItem extends DownloadStub {
 /// Enumerated by Isar, do not modify order or delete existing entries.
 enum DownloadItemType {
   collection(true, false),
-  track(true, true),
+  song(true, true),
```
Chaphasilor, line 1318: can you please add a warning here, not to rename this? preferably with the `//!!! <text>` syntax

Chaphasilor (approved):
Looks fine. I'm not sure if we could add some sort of alias to the `DownloadItemType` to have a consistent API outside of the download system, but that might just cause more confusion.

Maxr1998:
I fear that would create more confusion, yes. For later, we could look into replacing the enum name used for the hashed ID with a hardcoded string constant in the enum.

Issue #799, opened 2024-06-28
Fade audio on pause/resume


Ambient-Impact:
As someone who enjoys loud music, I find it way more pleasant to have it fade out quickly (like 200 ms or so?) and it feels less jarring or like I've hit a brick wall when I pause a track. I've used Poweramp on Android for years, and this is the one feature I miss switching over to Jellyfin and Finamp. Awesome work with the redesign by the way.

Chaphasilor:
Hey, thanks!  
That should be possible to do, by simply quickly changing the volume, as suggested here: https://stackoverflow.com/questions/69508649/how-can-i-fade-in-and-fade-out-audio-in-just-audio

Not a priority for me, but if you or anyone else wants to tackle this, I'd be happy to accept a PR! It should be relatively easy to implement :)

Ambient-Impact:
I'd love to if I had any experience working with any of this, but I'm a web developer and never worked on a native app; I know the limits of my knowledge far too well. 🤓

Chaphasilor:
Dart is very similar to JavaScript, and you're not really touching native code. It would be similar to working with the Web Audio API in this case. You'd just need some OOP knowledge to understand how to overwrite methods of parent classes.

But no pressure! Just, feel free to ask if you'd to contribute :)

Ambient-Impact:
True, Dart is pretty similar, and I do a ton of OOP stuff so that part I have a really good knowledge of. I'll think about it, though right now I'm short on time for these sorts of things because I have to gather bill repellent. 😭

Chaphasilor:
@Ambient-Impact this will be added with the next beta update in a week or so!

Chaphasilor:
Added in [beta update 0.9.16](https://github.com/jmshrv/finamp/releases/tag/0.9.16-beta)!

Ambient-Impact:
@Chaphasilor Just installed the update and tried it out. I love it. Thanks!

PR #1100, opened 2025-03-09
Add audio fade-in and fade-out with corresponding settings

fmuehlis:
Hello, 
I tried implementing the feature requested in #799. I'm quite new to Flutter and Open Source so if there is anything I should improve just let me know :)
The huge diff in `finamp_models.dart` seems to be an issue with reformatting (?).
All in all this PR adds:

1. Fade-in setting, for fading in audio on resume. Fade-in is disabled by default by using a value of 0.
2. Fade-out setting, for fading out audio on pause. Fade-out is disabled by default by using a value of 0.

I could not figure out why, but on initial playback the fade-in does not seem to work, only on resume and pause.
I did not want to add too many new settings, thus the step size for fade-in and fade-out are hardcoded to 50ms for now.

### DIFF
--- a/lib/services/finamp_settings_helper.dart
+++ b/lib/services/finamp_settings_helper.dart
@@ -222,6 +222,8 @@ class FinampSettingsHelper {
     FinampSetters.setPeriodicPlaybackSessionUpdateFrequencySeconds(DefaultSettings
         .periodicPlaybackSessionUpdateFrequencySeconds); // DOES NOT update UI
     FinampSetters.setReportQueueToServer(DefaultSettings.reportQueueToServer);
+    FinampSetters.setAudioFadeInDuration(DefaultSettings.audioFadeInDuration);
+    FinampSetters.setAudioFadeOutDuration(DefaultSettings.audioFadeOutDuration);
   }
 
   static void resetNormalizationSettings() {
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -398,26 +398,22 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
           await _player.setVolume(min(
               _player.volume + state.volumeFadeInStepSize,
               state.recoverVolume));
+          fadeState.add(state.copyWith(fadeVolume: _player.volume));
           if (_player.volume >= state.recoverVolume) {
-            fadeState.add(state.copyWith(
-                fadeDirection: FadeDirection.none, fadeVolume: _player.volume));
-          } else {
-            fadeState.add(state.copyWith(fadeVolume: _player.volume));
+            fadeState.add(state.copyWith(fadeDirection: FadeDirection.none));
           }
           break;
         case FadeDirection.fadeOut:
           await _player.setVolume(
               max(_player.volume - state.volumeFadeOutStepSize, 0.0));
+          fadeState.add(state.copyWith(fadeVolume: _player.volume));
           if (_player.volume <= 0.0) {
-            fadeState.add(state.copyWith(
-                fadeDirection: FadeDirection.none, fadeVolume: _player.volume));
+            fadeState.add(state.copyWith(fadeDirection: FadeDirection.none));
 
             fut = _player.pause();
 
             // restore volume
             await _player.setVolume(fadeState.value.recoverVolume);
-          } else {
-            fadeState.add(state.copyWith(fadeVolume: _player.volume));
           }
           break;
         default:
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -19,6 +19,7 @@ import 'package:get_it/get_it.dart';
 import 'package:just_audio/just_audio.dart';
 import 'package:just_audio_media_kit/just_audio_media_kit.dart';
 import 'package:logging/logging.dart';
+import 'package:rxdart/rxdart.dart';
 
 import 'android_auto_helper.dart';
 import 'finamp_settings_helper.dart';
@@ -62,6 +63,9 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
 
   double iosBaseVolumeGainFactor = 1.0;
   Duration minBufferDuration = Duration(seconds: 90);
+
+  final _audioFadeStepDuration = Duration(milliseconds: 50);
+  BehaviorSubject<bool> fading = BehaviorSubject.seeded(false);
 
   final outputSwitcherChannel =
       MethodChannel('com.unicornsonlsd.finamp/output_switcher');
@@ -281,8 +285,13 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
   Future<void> dispose() => _player.dispose();
 
   @override
-  Future<void> play() {
-    return _player.play();
+  Future<void> play() async {
+    if (FinampSettingsHelper.finampSettings.audioFadeInDuration >
+        Duration.zero) {
+      return fadeInAnPlay();
+    } else {
+      return _player.play();
+    }
   }
 
   @override
@@ -291,7 +300,72 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
   }
 
   @override
-  Future<void> pause() => _player.pause();
+  Future<void> pause() async {
+    if (FinampSettingsHelper.finampSettings.audioFadeOutDuration >
+        Duration.zero) {
+      return fadeOutAndPause();
+    } else {
+      return _player.pause();
+    }
+  }
+
+  int getFadeSteps(Duration fadeDuration) {
+    return (fadeDuration.inMilliseconds / _audioFadeStepDuration.inMilliseconds)
+        .toInt();
+  }
+
+  double getVolumeFadeInStepSize(double currentVolume) {
+    final duration = FinampSettingsHelper.finampSettings.audioFadeInDuration;
+    final steps = getFadeSteps(duration);
+    return currentVolume / steps;
+  }
+
+  double getVolumeFadeOutStepSize(double currentVolume) {
+    final duration = FinampSettingsHelper.finampSettings.audioFadeOutDuration;
+    final steps = getFadeSteps(duration);
+    return currentVolume / steps;
+  }
+
+  Future<void> fadeOutAndPause() async {
+    final currentVolume = _player.volume;
+
+    final volumeStep = getVolumeFadeOutStepSize(currentVolume);
+
+    fading.add(true);
+    var volume = currentVolume;
+    await Stream.periodic(_audioFadeStepDuration)
+        .takeWhile((_) => volume > 0.0)
+        .forEach((_) async {
+      volume = max(volume - volumeStep, 0.0);
+      await _player.setVolume(volume);
+    });
+
+    final pauseFut = _player.pause();
+    await _player.setVolume(currentVolume);
+    fading.add(false);
+    return pauseFut;
+  }
+
+  Future<void> fadeInAnPlay() async {
+    final currentVolume = _player.volume;
+
+    final volumeStep = getVolumeFadeInStepSize(currentVolume);
+
+    fading.add(true);
+    await _player.setVolume(0.0);
+    final playFut = _player.play();
+
+    var volume = _player.volume;
+    await Stream.periodic(_audioFadeStepDuration)
+        .takeWhile((_) => volume < currentVolume)
+        .forEach((_) async {
+      volume = min(volume + volumeStep, currentVolume);
+      await _player.setVolume(volume);
+    });
+    fading.add(false);
+
+    return playFut;
+  }
 
   Future<void> togglePlayback() {
     if (_player.playing) {
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -2,7 +2,6 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 import 'dart:math';
-import 'dart:ui';
 
 import 'package:audio_service/audio_service.dart';
 import 'package:finamp/components/global_snackbar.dart';
@@ -12,6 +11,7 @@ import 'package:finamp/services/favorite_provider.dart';
 import 'package:finamp/services/jellyfin_api_helper.dart';
 import 'package:finamp/services/queue_service.dart';
 import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:finamp/l10n/app_localizations.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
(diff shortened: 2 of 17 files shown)


lib/services/finamp_settings_helper.dart
```diff
@@ -554,6 +554,20 @@ class FinampSettingsHelper {
...
+        .put("FinampSettings", finampSettingsTemp);
+  }
+
+  static void setAudioFadeInDuration(Duration audioFadeOutDuration) {
+    FinampSettings finampSettingsTemp = finampSettings;
+    finampSettingsTemp.audioFadeInDuration = audioFadeOutDuration;
+    Hive.box<FinampSettings>("FinampSettings")
+        .put("FinampSettings", finampSettingsTemp);
+  }
+
```
Chaphasilor, line 570: Please also register these new settings in the reset method (in `resetAudioServiceSettings()`)

lib/services/music_player_background_task.dart
```diff
@@ -281,8 +283,12 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
   Future<void> dispose() => _player.dispose();
 
   @override
-  Future<void> play() {
-    return _player.play();
+  Future<void> play() async {
+    if (FinampSettingsHelper.finampSettings.audioFadeOutDuration > Duration()) {
```
Chaphasilor, line 287: You should use `Duration.zero` instead of an empty instance (same in the pause handler)

Chaphasilor: Also, you're comparing with `audioFadeOutDuration`, but it should be `audioFadeInDuration`

Chaphasilor (commented):
Thanks for the PR!  
That should be a nice little addition.

Two things worth mentioning: I believe your current implementation will remain in the "playing" state until the fade out is complete, so the play/pause button on the player screen wouldn't update until after that's done. That might make the app feel less responsive.  
Haven't tried it out yet, but if that is indeed the case, maybe try to update the state of the button somehow before playback is paused?

Regarding the fade in not working the first time, maybe the fade is completed before playback has started? Have you tested this with a very long fade in duration?  
It could also be that the `play()` call is skipped the first time for some reason, or that `play()` and `pause()` don't behave identical when it comes to awaiting the futures.

Chaphasilor:
Oh and you can just move the contents of the two new files directly into `audio_service_settings.dart`, that reduces some boiler plate and keeps things better organized :)

Chaphasilor:
Just tried it out and comfirmed the issue with the play/pause button really appears. Maybe an intermediate state would be a good solution, like a loading spinner or small animation?

I also encountered a bug, when you first pause, and the try to play again while the volume is still fading out, the fade in will use a reduced volume as the original volume. This can lead to the volume becoming 0 permanently.  
I've actually started working on a proper internal volume mechanism as part of #850, maybe take a look at that? I could also transfer your changes over to that PR, or you target it as the "base branch".  
That should get rid of this bug, since the internal volume is not modified, only the player volume itself.

fmuehlis:
Hi, thank you for the review, I fixed some of the lower hanging fruits. I will have to look at the volume and UI problems later.
> target it as the "base branch"

Do you maybe have a hint on how to do that?

Chaphasilor:
Thanks for the fixes!

I'm honestly not sure if targeting the other PR is possible in this case.  
But try changing the target branch of your PR to point at https://github.com/pinsarda/finamp-playon/tree/playon

fmuehlis:
I tried fixing the UI issues, although I am not sure whether modifying the MediaStateStream by adding the audioFading attribute is the best way to achieve this. If you have a better way of fixing this in mind, just let me know and I will implement it.
I also tried targeting the branch you mentioned, but that was not possible. I think it should be possible to fix the fading issues without doing that though, I will try to fix that in the coming days too.

fmuehlis:
I added the changes you suggested, which I think works quite nice! With this implementation the fade-in and fade-out should also not cause issues with volume anymore. There is now a single Stream.periodic for fading in and out where the current FadeState determines whether the volume will be faded in or out from the current state.
It might be worth adding some helper methods to FadeState to update the internal state for reduced boilerplate:
```dart
var state = fadeState.value;
state.fadeDirection = FadeDirection.none;
fadeState.add(state);
```
to something like this:
```dart
fadeState.add(state.updateDirection(FadeDirection.none))
```

Chaphasilor:
Thanks! Regarding the boilerplate, it don't mind it too much, but if you want you could implement the `.copyWith()` method for the `MediaState` class, so that the code then looks like this:

```dart
fadeState.add(state.copyWith(fadeState: FadeDirection.none))
```
You might also need to refactor to using named arguments instead of positional arguments, but that shouldn't take long (just adding curly braces around the parameters).

I was thinking about maybe having a threshold for the fadeout, after which a spinner will be shown during the transition. I've been testing with ~3.5s fade out, and at first it always seems like nothing is happening, because the volume doesn't change noticeably. So maybe keep the current behavior for fade out durations <=2s, and show the spinner afterwards? For the fade in it should be fine, since the difference between nothing playing and something playing is more apparent.

What do you think about that?

I also noticed one bug: Playing and then pausing works and properly fades in an then out again, but pausing and playing doesn't work. During the fade out, the play button appears to be disabled. Could you please adjust that?

fmuehlis:
Alright, I implemented `copyWith`, this would not require named arguments for the normal constructor. Would you still like to have them anyway?

So do you mean that the spinner should always show when pausing if the setting is set to a value higher than 2 seconds? Or that the spinner should be shown after 2 seconds of fading?

I will try to reproduce the bug with the pause then play tomorrow, I could not figure out why this would happen just from the code right now.

Another thing I noticed is that pause is sometimes also used in a context where fade out is not really useful like at the end of the queue or when the sleep timer is up. I would solve this with an overwrite right, so that fading can be disabled when needed. Or would you say the fade out should always happen, even at the end of the queue? I think that should also be possible, although it will probably also be much more complex than an overwrite parameter. For play this should be already working, but the fadeout would have to start automatically before the end of the queue/sleep timer.

Chaphasilor:
Thanks for the changes!

> Would you still like to have them anyway?

Yes, I think it's both more comfortable to use and clearer when reading the code :)

Spinner should be shown the entire time if the configured fade out duration is > 2s. Unless you think that's not a good idea?

Override is fine, but I think in the two cases you mentioned the fade is actually fine. Especially for the slerp timer.  
For the end of the queue, the fade can happen after the track stops playing, no need to actually fade out here :)

Chaphasilor:
That was actually also something I was considering, but my idea was more of a "bar" or background representing the current volume sliding up or down behind the icon. But your approach sounds better, where it's just the circular progress incidator around the icon, if I understand it correctly?  
Haven't tried it out yet.

Having a progress indicator *and* the icon is probably the best we can do 👍🏻

If you feel adventurous you could try aligning the progess bar with the shape of the play/pause button, which is a rounded rectangle and not a circle. That might look sleek? I'm thinking an animated border or something.  
But that's really out there, and I'm happy to merge it without it...

If you want refactor the implementation some more feel free to do that; but I won't be able to actually give the current changes a try before tonight.

fmuehlis:
>If you feel adventurous you could try aligning the progess bar with the shape of the play/pause button, which is a rounded rectangle and not a circle. That might look sleek? I'm thinking an animated border or something.
But that's really out there, and I'm happy to merge it without it...

That sounds like a nice idea! I played around a bit with this and the best way I could make this work is by using [this package](https://pub.dev/packages/progress_border), which implements a custom BoxBorder. It is a single file package, but I'm not sure whether adding a dependency for a feature like this is a good idea. It really does look nice though:
https://github.com/user-attachments/assets/4d981abc-37c6-407b-b806-1b6e8b39d560

The complexity of the package is already really low, so implementing it myself would probably end up close to the same code.

Chaphasilor:
Adding a package really isn't an issue! Especially if it's just a UI feature. If it breaks, we can either switch to a different package or copy and fix the code ourself. And if it doesn't work at all any more, we drop it completely and everything still works, it just won't look as pretty.  

But I agree that the video you posted looks awesome! I've been hoping to get more animations into Finamp eventually, so that's a good start!  
One thing: if there's an option to change the easing curve, I'd prefer an ease-in-out curve instead of the linear animation. So some slight acceleration and deceleration at the start and end. You're probably piping in the volume value directly at the moment, but maybe you could implement a simple mapping function that applies the easing. There should be helper functions that are used during page transitions, they might be called tweens. Let me know if you need help with that!

fmuehlis:
Oh true, I have looked into this a bit today. I have never worked with animations like this before, so I'm not quite sure whether the animation can be reused by just playing it forward or reverse for fade in and fade out. I could not get that to work. However the changes I committed should improve the smoothness of the animation already.
There seems to be a bug where the progress border is shown when first entering the currently playing song view. I don't know why that happens, I will look into that further when I have the time to.

Chaphasilor:
Neat, thanks!  

Seems like another bug sneaked in, currently the animation only starts *after* the fade is already completed. Also seems like the animation plays once for the player screen and once for the now playing bar.

Only one additional suggestion, you could make the name of the new widget/file even more descriptive, like `audio_fade_progress_visualizer_container.dart` or something similar. Don't be afraid of longer file/widget names if they better indicate what's going on :)  

Maybe we need to find a better color for showing the progress on the now playing bar, depending on the cover/accent combo this might be hard to see. But I can also play around with that when the time comes ^^

fmuehlis:
>Seems like another bug sneaked in, currently the animation only starts after the fade is already completed.
Ah true, sorry! It should be fixed now. It seems that the animation durations are still behaving a bit weird, do you maybe have clue what it could be?

>Also seems like the animation plays once for the player screen and once for the now playing bar.
What do you mean by that? The animation has to be run per widget right? Or is there a way to share it?

Chaphasilor:
I'll take a look later.  
Here's a video of the other bug, look at the now playing bar after dismissing the player screen, the animation repeats there. But maybe your latest commit changed the behavior, I haven't built that one yet.

Chaphasilor:
Just built the latest commit, I think the animation behaves correctly (aside from the video I sent), but the fadeIn/fadeOut state might not be totally correct. When quickly switching between playing and pausing, sometimes taps seem to have no effect, with the fade continuing in the same direction as before. That might be it?  
It's not a super severe issue though, so don't hang up on it :)

fmuehlis:
>When quickly switching between playing and pausing, sometimes taps seem to have no effect

Hmm I could not reproduce this, do you maybe have a short clip of it happening?

>aside from the video I sent

That does not seem to be such an easy fix, but I think I figured out why the animation plays twice:
The `NowPlayingBar` is not disposed when opening the player screen, meaning the fade state is still updated inside of the widget. The animation however is only rendered once the widget becomes visible again.

I was able to fix it using the [visibility_detector](https://pub.dev/packages/visibility_detector) package. This way the animation state is only updated in the listener when the widget is actually visible. There sadly does not seem to be a flutter native solution to visibility detection.

Another thing I noticed was that the fade animation was playing during app startup, hence I implemented the fade override to disable fading in the `QueueService` initialization. I also disabled the fading for the other cases we discussed earlier, because why not if the override is implemented anyway (save a CPU cycle or two yay!).

Chaphasilor:
@fmuehlis regarding the visibility, instead of not updating the animation state if it's not visible, how about instead making sure the animation still plays? I found the [`Offstage` widget](https://api.flutter.dev/flutter/widgets/Offstage-class.html) today, which seems like a simple wrapper that ensures the now playing bar is kept quasi-rendered at all times, which should fix this issue.  
It should also help with accent color transitions after closing the player screen.  
Maybe try putting that into the NowPlayingBar class and see if it helps?  
That will sacrifice the two CPU cycles you just gained though :P

Chaphasilor:
https://github.com/user-attachments/assets/02cb65b3-3470-4e6e-a544-d0866454f0c3

Here's a video of the issue. Especially double-tapping seems to be problematic.

fmuehlis:
> @fmuehlis I found the [`Offstage` widget](https://api.flutter.dev/flutter/widgets/Offstage-class.html) today, which seems like a simple wrapper that ensures the now playing bar is kept quasi-rendered at all times, which should fix this issue.

Hmm that's a nice idea. I just tried that but it does not seem to work the way I expected. Whether the widget should be offstage has to be decided manually and passed to the `Offstage` widget as `offstage: bool`. I tried `true` in which case the entire child is not rendered at all, and `false` in which case the old issue arises again. So in this case the visibility would have to be managed in the widget including the `AudioFadeProgressVisualizerContainer`, which would make the usage of `Offstage` unnecessary as I could also just pass the manually managed visibility state down, so that the animation state is not updated. I think this would require all parent widgets to be `StatefulWidget`s though (which is not the case for `NowPlayingBar` and `PlayerButtons`).

>Here's a video of the issue. Especially double-tapping seems to be problematic.

Hmm yeah, that's sort of the issue I experienced too. I think it should be fixed now. It still looks weird sometimes with small fade times though (but it does the correct thing I think). Maybe we should add a toggle in the settings on whether to show the fade animation at all or disable it for really low fade times?

Chaphasilor:
Ah I see. Okay, seems like that `Offstage` widget is a lot less useful than I thought then...  

I'm against a toggle (it's not a significant change, and UI customization should be kept to a minimum), but if you want to disable it for shorter fades, that's fine. Maybe 500 or 750 ms could be a good threshold?

The animation is definitely much more consistent now. I've managed to break it slightly with some intensive spamming and interrupting, but nothing that would need additional fixing. The two animations on now playing bar and player screen also seem to be correctly synced now, so I think this is ready to merge from my side!  
Let me know if you want to add the setting or not :)

Also took another look at the code and everything looks good. We're not even really adding a new dependency since it was already a transitive dependency before ^^

fmuehlis:
Awesome! I think I would rather keep the animation on at all times for consistency. Otherwise it might be confusing on why this animation suddenly starts to play. The connection between the animation and the fade setting might be easier to make then and I think the animation looks just fine under "normal" conditions.

I rebased the feature branch to the redesign branch on my machine to resolve the remaining conflicts, is that ok or should I rather merge? One problem I have is that I somehow cannot build the redesign branch because `device_info_plus` has conflicting dependencies. I was able to build it with `device_info_plus ^11.3.0` and everything works just fine. Kept the `^11.3.3` version in the rebase and only changed it locally, but you might have to confirm that it works.

Chaphasilor:
Sure, if you put in the effort of rebasing, that's absolutely fine! It's actually a little better commit-history-wise, but also requires more effort for larger PRs, so I don't usually do it.  
Just push your changes and I'll take a look! If everything works, I'll merge it :)

Chaphasilor:
Great, thanks! Is it okay if I still squash the commits? I don't think we need all the fixes separately.  
I also noticed that the sleep timer still doesn't use the fade out settings, but that should be an easy fix.  
Everything else is still working.  
I can fix the sleep timer and then squash-merge, unless you want it done differently?

fmuehlis:
>Saw that you did the sleep timer behavior on purpose. How important is that to you?

Actually not important at all! I thought it wouldn't do anything for the sleep timer anyway, but now that I'm thinking about it it does make a difference.
>Great, thanks! Is it okay if I still squash the commits? I don't think we need all the fixes separately.

Sure, go ahead :)
I saw some Repos having guidelines on this, so if this is something you have to do regularly it might be worth adding to the contribution guidelines? Possibly just as an optional step. Then for my next contribution I can take some more work off of you :)
Thank you for your patience!

Chaphasilor:
Awesome! Getting this PR merged took way longer than I initially expected, but I'm very happy with how it turned out. Thanks for adding this!  
If you want, you can also fill out this form to receive some free stickers after the hackathon ends: https://app.formbricks.com/s/cm8tajvx13912s001l9it719v

Oh and don't worry about the squashing, it takes 1-3 clicks to squash-merge a PR on GitHub :) And I really decide this on a case-by-case basis, to keep the commit history as helpful as possible.

PR #1104, opened 2025-03-12
[Redesign] Improve Queue Restore Speed

Komodo5197:
Improves the delay between app startup and queue restore completion by moving the call slightly earlier in the boot sequence and skipping an uneeded wait for the saved queues to be committed to disk.  This allows reasonably sized queues to complete their server queries before the music screen queries start instead of after.


Chaphasilor:
Awesome, I'll try this out later.  
Maybe that helps with performance on tje unstable server version; that has been super slow to load since they started the EFCore migration.  
That was expected though, but is still annoying ^^

Chaphasilor:
"Later" is a subjective term I guess...  
Tested it out and it seems to work at least just as well as before. We'll see how the performance gain for daily usage is.  
Thanks again!

PR #1118, opened 2025-03-22
Redesign downloads screen to group items by type and sort them

Maxr1998:
Instead of showing all downloads in an unsorted list, the downloads screen now has separate sections for artists, albums, playlists, genres and (individually downloaded) tracks.
On changes to the downloads database, the screen is automatically updated. I also tweaked some paddings.

<details>
<summary>Screenshot</summary>

image
</details>

*Please squash merge when reviewed.*


lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -1,78 +1,148 @@
...
-
-  @override
-  State<DownloadedItemsList> createState() => _DownloadedItemsListState();
+enum DownloadsScreenItemType {
+  artists,
+  albums,
+  playlists,
+  genres,
+  tracks,
 }
```
Maxr1998, line 19: I might still replace this enum by simply passing a lambda that returns the query to the widget instead, thoughts about that?

lib/screens/downloads_screen.dart
```diff
@@ -36,8 +36,16 @@ class DownloadsScreen extends StatelessWidget {
...
+          DownloadedItemsTitle(title: AppLocalizations.of(context)!.artists),
+          const DownloadedItemsList(type: DownloadsScreenItemType.artists),
+          DownloadedItemsTitle(title: AppLocalizations.of(context)!.albums),
+          const DownloadedItemsList(type: DownloadsScreenItemType.albums),
+          DownloadedItemsTitle(title: AppLocalizations.of(context)!.playlists),
+          const DownloadedItemsList(type: DownloadsScreenItemType.playlists),
+          DownloadedItemsTitle(title: AppLocalizations.of(context)!.genres),
+          const DownloadedItemsList(type: DownloadsScreenItemType.genres),
+          DownloadedItemsTitle(title: AppLocalizations.of(context)!.tracks),
+          const DownloadedItemsList(type: DownloadsScreenItemType.tracks),
```
Maxr1998, line 48: The repetitions are a bit awkward, but we need to have all categories as direct sliver children of the scroll view.

lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -1,104 +1,182 @@
...
+          leading: AlbumImage(item: stub.baseItem),
+          subtitle: ItemFileSize(stub: stub),
+          trailing: FutureBuilder(
+            future: ref
+                .watch(_downloadsService.statusProvider((stub, null)).future)
+                .then((item) => item.isRequired),
+            builder: (context, snapshot) {
+              if (snapshot.hasData && snapshot.data!) {
+                return const Icon(Icons.lock);
+              }
```
Maxr1998, line 177: Should possibly be changed to a button that allows deletion.

lib/components/DownloadsScreen/downloaded_items_list.dart
Komodo5197: I'd probably write the downloadsService methods as something like
```
List<DownloadStub> getUserDownloaded(BaseItemDtoType? typeFilter) =>getVisibleChildren(_anchor,typefilter);
List<DownloadStub> getVisibleChildren(DownloadStub stub, BaseItemDtoType? typeFilter){
    var whereFilter=switch(typefilter){
        null=>_isar.downloadItems.where().typeNotEqualTo(DownloadItemType.image),
        BaseItemDtoType.track => _isar.downloadItems.where().typeEqualTo(DownloadItemType.track),
        _ => _isar.downloadItems.where().typeNotEqualTo(DownloadItemType.collection)
  };
    var anchorFilter= _whereFilter.filter().requiredBy((q) => q.isarIdEqualTo(stub.isarId));
    return switch(typeFilter){
         BaseItemDtoType.track =>anchorFilter .sortByBaseIndexNumber().findAllSync(),
         _ =>anchorFilter .sortByName().findAllSync(),
     };
  }
```
and then you can just use BaseItemDtoType directly.

lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -1,78 +1,148 @@
...
+      case DownloadsScreenItemType.genres:
+        _query = _downloadsService
+            .queryUserDownloadedCollection(BaseItemDtoType.genre);
+        break;
+      case DownloadsScreenItemType.tracks:
+        _query = _downloadsService.queryUserDownloadedTracks();
+        break;
+    }
+
+    _itemsProvider = StreamProvider.autoDispose((ref) {
```
Komodo5197, line 80: I believe dynamically creating providers is considered bad.  I think the correct code would have a family provider in the global scope.  Something like

> final userDownloadedProvider = Provider.autoDispose .family<BaseItemDtoType, List<DownloadStub>>((ref, type) {
> return GetIt.instance<DownloadsService>().queryUserDownloadedCollection(type).watch(fireImmediately: true);
> }

lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -1,78 +1,148 @@
...
+                          icon: const Icon(Icons.sync),
+                          onPressed: () {
+                            _downloadsService.resync(album, null);
+                          },
+                        ),
+                      IconButton(
+                          icon: const Icon(Icons.delete),
+                          onPressed: () => askBeforeDeleteDownloadFromDevice(
+                                  context, album, refresh: () {
+                                setState(() {});
```
Komodo5197, line 112: Is this needed?  I think the itemprovider should handle the refresh.

Maxr1998: I simply reused the former code here, so it might not be needed anymore?

lib/screens/downloads_screen.dart
Chaphasilor: That's fine for now. The redesign of the downloads screen is still hazy, not sure if I want to go for tabs, screens, or filter chips for the different categories. So this is still subject to change :)

lib/components/DownloadsScreen/downloaded_items_list.dart
Chaphasilor: Yeah it should work through the providers and Consumer widget. Everything download-related is reactive by default, I think.

lib/screens/downloads_screen.dart
Maxr1998: I was actually thinking about filter chips earlier. But for the first iteration, I wanted to keep things simple.

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: Removed that refresh handler now, should still work.

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: True, that reduces the required methods a bit. Applied the change similarly to what you suggested.

lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -1,104 +1,168 @@
...
+  final String title;
+
+  @override
+  Widget build(BuildContext context) {
+    return SliverToBoxAdapter(
+      child: Padding(
+        padding: const EdgeInsets.only(left: 16, top: 8, right: 16),
+        child: Text(
+          title,
+          style: Theme.of(context).textTheme.headlineSmall,
```
Chaphasilor, line 25: That still looks a bit large. Maybe try adding emphasis though font weight instead of size? That often looks sleeker :)

lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -1,104 +1,168 @@
...
+
+    return ref
+        .watch(_downloadsService.downloadedItemsProvider(widget.type))
+        .when(
+          data: (items) => items.isNotEmpty
+              ? SliverList(
+                  delegate: SliverChildBuilderDelegate(
+                    (context, index) {
+                      DownloadStub stub = items.elementAt(index);
+                      return ExpansionTile(
```
Chaphasilor, line 72: Can you try to indent the child items when the tile is expanded? Currently the nested items look identical, except for the missing delete icon

lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -1,104 +1,168 @@
...
+                                  askBeforeDeleteDownloadFromDevice(
+                                      context, stub),
+                            ),
+                          ],
+                        ),
+                        subtitle: ItemFileSize(stub: stub),
+                        tilePadding: const EdgeInsets.symmetric(horizontal: 16),
+                        children: [
+                          if (stub.baseItemType.hasChildren)
+                            DownloadedChildrenList(parent: stub)
```
Chaphasilor, line 101: Probably out-of-scope for this PR, but it's a bit strange that we only allow expanding one level deep. Expanding should always work until getting to the actual tracks, so for an artist or genere one should be able to expand the contained albums to see the tracks...

lib/screens/downloads_screen.dart
```diff
@@ -36,8 +38,16 @@ class DownloadsScreen extends StatelessWidget {
...
+          DownloadedItemsTitle(title: localizations.artists),
+          const DownloadedItemsList(type: BaseItemDtoType.artist),
+          DownloadedItemsTitle(title: localizations.albums),
+          const DownloadedItemsList(type: BaseItemDtoType.album),
+          DownloadedItemsTitle(title: localizations.playlists),
+          const DownloadedItemsList(type: BaseItemDtoType.playlist),
+          DownloadedItemsTitle(title: localizations.genres),
+          const DownloadedItemsList(type: BaseItemDtoType.genre),
+          DownloadedItemsTitle(title: localizations.tracks),
+          const DownloadedItemsList(type: BaseItemDtoType.track),
```
Chaphasilor, line 50: Weird idea, but how about we order these categories i the same order as the tabs on the music screen?

Chaphasilor (commented):
Thanks for the PR!  

Just commented a few design suggestions, but nothing major I hope.  
Definitely looks more organized than before :D

lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -1,104 +1,168 @@
...
+                        ],
+                      );
+                    },
+                    childCount: items.length,
+                  ),
+                )
+              : SliverToBoxAdapter(
+                  child: Padding(
+                    padding: EdgeInsets.symmetric(horizontal: 16, vertical: 8),
+                    child: Text(AppLocalizations.of(context)!.noItem),
```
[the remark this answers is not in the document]
Maxr1998: I was too lazy to add a new string, but maybe I should do that.

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: Fair, I'll do some reading on text styling then.

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: I can do that, but I personally prefer having the same indentation. Maybe we can otherwise separate them, e.g. by background color?

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: Yes, probably out of scope :sweat_smile:
I do already wonder whether they are loaded lazily as expected, or not..

lib/screens/downloads_screen.dart
Maxr1998: The reason I put artists first is that they have that extra sync button and are usually much less, so easier to get past. I'd also argue that the use-case is different. On the music screen, you want to get quickly to something to listen to, so typically albums and contained songs. On the downloads screen, my sorting idea was to start with the least amount of elements/"outermost" entity. Technically, playlists could be moved to the start.

lib/components/DownloadsScreen/downloaded_items_list.dart
Chaphasilor: Whatever you prefer! But BG color might be harder to do: changing the list items could bd messy, chaning the list BG requires passing (indentation) to look okay. A vertical left line/border with just a bit of indentation could work too?

lib/screens/downloads_screen.dart
Chaphasilor: You are making a lot of assumptions about users there ^^

I actually mostly download artists and playlists, and very rarely albums. And my first tab are tracks ^^

The current order is fine, but another order probably isn't worse either. Right now it simply is arbitrary, and sorting based on tabs would at least be less arbitrary.

But it's really not a huge deal :)

Chaphasilor:
Just noticed, you're missing a category ^^  
Special downloads, like favorites, all playlists, or latest albums are not shown at  all anymore. You could put them at the top or bottom?

Maxr1998:
Are those of type finampCollection?

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: I like the line border idea. Let me try that then.

lib/screens/downloads_screen.dart
Maxr1998: Oh, so you meant sorting based on the user preferences, not the default value of the tabs. That sounds reasonable enough. And true, those are definitely assumptions, that's why the music screen can be reordered, thankfully 😁 
I'd say we should keep it as is, and I'll look into filtering with chips after this PR is merged.

lib/services/downloads_service.dart
```diff
@@ -1266,20 +1266,18 @@ class DownloadsService {
...
+        .where()
+        .typeEqualTo(category.type)
+        .filter()
+        .requiredBy((q) => q.isarIdEqualTo(_anchor.isarId));
 
-    final afterWhere = switch (filterType) {
-      BaseItemDtoType.track => where.typeEqualTo(DownloadItemType.track),
-      _ => where.typeEqualTo(DownloadItemType.collection),
-    };
+    final afterFilter = category.baseItemType != null
```
Komodo5197, line 1276: There's a .optional() filter that makes this sort of thing a bit cleaner.

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: Not sure I vibe with that:

<summary>Screenshot</summary>

image
</details>

lib/components/DownloadsScreen/downloaded_items_list.dart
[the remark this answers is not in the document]
Maxr1998: Done :white_check_mark:

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: Done :white_check_mark:

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: A different color might work, some slightly lightened background color. But finding the right one there is hard.

lib/services/downloads_service.dart
```diff
@@ -1257,18 +1264,20 @@ class DownloadsService {
...
-  /// Get all user-downloaded items.  Used to show items on downloads screen.
-  List<DownloadStub> getUserDownloaded() => getVisibleChildren(_anchor);
-
-  /// Get all non-image children of an item.  Used to show item children on
-  /// downloads screen.
-  List<DownloadStub> getVisibleChildren(DownloadStub stub) {
-    return _isar.downloadItems
+  /// Query all downloads by the given BaseItemDtoType.
+  /// Used to show various items grouped by type on the downloads screen.
+  Query<DownloadStub> queryUserDownloaded(DownloadsScreenCategory category) {
```
Komodo5197, line 1269: This doesn't seem generally useful, especially returning a query instead of more concrete results.  I'd just put all the logic directly inside the provider.

lib/services/downloads_service.dart
```diff
@@ -152,6 +152,13 @@ class DownloadsService {
     return isar.downloadItems.watchObject(stub.isarId, fireImmediately: true);
   });
 
+  /// Provider for downloaded items of a specific type.
+  /// Used to show various items grouped by type on the downloads screen.
+  late final downloadedItemsProvider = StreamProvider.family
```
Komodo5197, line 157: You should probably specify user downloaded, not downloaded in general, in the provider name and/or comment.

lib/services/downloads_service.dart
Maxr1998: The query is necessary I think to be able to watch for changes. But sure, I can inline the function into the provider too.

lib/components/DownloadsScreen/downloaded_items_list.dart
Chaphasilor: And you really don't like slightly indenting the items? :P

lib/screens/downloads_screen.dart
Chaphasilor: Yes, I meant applying the user-configured order :)  
Alright, that sounds good. You might also wanna take a look at the #design channel on Discord for inspiration. Or what Spotify is doing with its Songs/Podcasts filter

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: I tried it and honestly wasn't a fan. But if there's no alternative we can also do that I guess.

Chaphasilor:
The headings are definitely better now, though I would reduce the padding between the heading and the items below it.  
You could also decrease the spacing between items just a bit.

The special download itens also aren't expandable any more, although some types like the 5 latest albums or favorites should be :)

Everything else seems fine!

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: The background doesn't look too bad actually:
image
Using `surfaceVariant` since the other surface colors seem to be undefined in our theme, but fixing this is probably out of scope for this. For now, this color would work.

And here's the indentation version:
image

Maxr1998:
The items spacing is defined by the list tile itself and follows the material guidelines, so that should imo stay as-is. I reduced the title spacing though and fixed not being able to expand finampCollections. Also fixed the points @Komodo5197 brought up. Please resolve the discussions once you're happy with my changes, helps me keep track of what needs to be done :grin:

lib/services/downloads_service.dart
Maxr1998: Done :white_check_mark:

lib/services/downloads_service.dart
Maxr1998: Done :white_check_mark:

lib/services/downloads_service.dart
Maxr1998: Done :white_check_mark:

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: Fixed :white_check_mark:

lib/components/DownloadsScreen/downloaded_items_list.dart
Maxr1998: Done :white_check_mark:

Chaphasilor:
Okay, leave the spacing then. But honestly, screw material guidelines; the track list tiles also have reduced spacing and that works fine. The redesign also isn't using material design any more.  
It's not really an accessibility issue either, but becomes a usability issue if people have to scroll for ages ^^  
Not critizising you here!
/rant

From my side this is fine to merge then, I'll update the review discussions later.

Maxr1998:
Alright. Just tell me your decision on the indentation vs. background color topic, I'll still push that change then. I also just noticed that the ExpansionTile doesn't even follow MD3 itself, they ignored that their divider also adds padding, even when invisible. Ugh. I wonder whether we can fix that, or if we should just make our own widget.

lib/components/DownloadsScreen/downloaded_items_list.dart
Chaphasilor: Fine, use the background for now, and when we get to redesigning the downloads screen we'll think about this again ^^  
Temporary victory for you :P

Chaphasilor:
Sure, feel free to apply any fixes you can think of! And we'll go with custom widgets if needed, but sticking to existing widgets and wrapping them as much as possible helps with semantics/accessibility :)  

I think you can go ahead and merge this after you pushed the remaining changes. Thanks!

Maxr1998:
And done! Should be good now. I'll merge in ~20 minutes then if nothing else comes up.

I also realized that the height/size of the items is the same as the albums on the album screen and artists on the artists screen. Only the album track lists are 2dp shorter.

PR #1123, opened 2025-03-26
[Redesign] Fix missing internal downloads location

Komodo5197:
Due to the errors with setting the migration flags, some users are coming from existing installs but skipping the downloads migration, resulting in the new internal downloads storage location never being created.  This will recreate it for those cases, fixing the issue where users don't have any available download locations.  I considered running the full migration, but I'm pretty sure that would just break things if done after the new download system has potentially already had other changes committed.

### DIFF
--- a/lib/services/downloads_service.dart
+++ b/lib/services/downloads_service.dart
@@ -1050,8 +1050,11 @@ class DownloadsService {
       FinampSettingsHelper.addDownloadLocation(downloadLocation);
     }
     await Future.wait([
+      // ignore: deprecated_member_use_from_same_package
       Hive.openBox<DownloadedParent>("DownloadedParents"),
+      // ignore: deprecated_member_use_from_same_package
       Hive.openBox<DownloadedTrack>("DownloadedItems"),
+      // ignore: deprecated_member_use_from_same_package
       Hive.openBox<DownloadedImage>("DownloadedImages"),
     ]);
     _migrateImages();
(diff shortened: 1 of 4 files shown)


Komodo5197:
Also, I fixed the library selector because it seems like I broke it in the previous PR.

lib/services/downloads_service.dart
```diff
@@ -1041,12 +1041,12 @@ class DownloadsService {
...
         .where((element) =>
-            element.baseDirectory == DownloadLocationType.internalSupport)
+            element.baseDirectory ==
+            DownloadLocationType.platformDefaultDirectory)
         .isEmpty) {
       final downloadLocation = await DownloadLocation.create(
-        name: "Internal Storage",
-        baseDirectory: DownloadLocationType.internalSupport,
-      );
+          name: "Internal Storage",
```
Chaphasilor, line 1048: Maybe a global constant would be good for this? there are already two-three affected lines in this PR, and it wouldn't be good to accidentally put a typo in one of them.

Chaphasilor:
Thanks! Code looks fine from a quick glance, and the view picker also didn't give me any issues :)

Just a thought - maybe we should mark the new `Internal Storage` location as the default location during the migration, because a) that's the most reliable and feature-rich location, and b) if people have other custom locations and want to use them, they should be able to figure out how to re-enable the prompt and maybe see the warning about images in custom locations (we might want to put another warning in the download dialog for those locations at some point).

I'm not sure about this myself yet, since it might really be unexpected and detrimental to the experience on phones with little internal storage but an SD card slot. Although those are getting rarer of course... 

We can merge whenever you're ready :)

lib/services/downloads_service.dart
Komodo5197: I've moved the name to a constant.  What I'd really like to do is have the name actually localized, but it needs to be accessed in main() so I don't believe that's possible.

Komodo5197:
Marking it as default seems questionable to me.  It's not like custom locations got less workable after all.  If someone already had one, they probably really wanted it and we should keep prompting.  And most people probably don't have any other locations so they don't need a default marked.  Anyway, this was a pretty simple change, feel free to merge.

lib/services/downloads_service.dart
Chaphasilor: Yeah that might be problematic. We could override the name in the settings screen, and the actual directory isn't user-facing anyway.

Chaphasilor:
Okay, let's merge then. Thanks again!

PR #1126, opened 2025-03-28
[Hackathon] Haptic Feedback for the now playing bar

flloschy:
Im starting early because apparently one cant have two forks? Very weird

Added Haptic Feedback when skipping in both directions and closing the playbar

### DIFF
--- a/lib/components/now_playing_bar.dart
+++ b/lib/components/now_playing_bar.dart
@@ -193,6 +193,7 @@ class NowPlayingBar extends ConsumerWidget {
               if (direction == DismissDirection.down) {
                 final queueService = GetIt.instance<QueueService>();
                 await queueService.stopPlayback();
+                FeedbackHelper.feedback(FeedbackType.success);
               } else {
                 await openPlayerScreen();
               }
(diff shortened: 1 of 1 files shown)


Chaphasilor:
You can't have two forks, but you can have multiple branches on your fork! :D  
You should create a new branch for each PR, so that you can keep them open while working on new stuff. Try to avoid using the `redesign` branch on your fork as the source for the PR :)

Anyway, awesome to see the first contribution coming in! 😁  
Looking at the code I'm wondering if we might want to have the feedback first, and then the `await`s. This way the feedback is instant and not delayed, and lets the user know that they just did something, even if loading the next track or stopping the queue takes a bit longer :)  
Maybe give it a try and report back what feels better?

flloschy:
Im afraid of branches because of merge issues but I guess those gonna happen no matter what in this. Gonna do branches for the two other things i planed then :)

I would like to test but I really dont want to deal with installing android and what not to get the apk to build, im sure this will be fiiiine, not much can go wrong anyway :D

Chaphasilor:
Alright, I'll give it a whirl then ^^

Chaphasilor:
Yep, seems to work and feels nice :)  
The feedback is still a bit delayed, but that's due to us using a Dismissible instead of a PageView for the swiping. But that's another issue that can be fixed as part of the hackathon ^^  
Thanks for the PR!

Chaphasilor:
Oh, also - I've squashed your two commits into one, since keeping them separate had no advantage. But that means that the commits in the upstream redesign branch (in this repo) are different from the comments in your redesign branch.  
So you should go into your repo, into your `redesign` branch (**!**) and then do a `git reset upstream/redesign --hard` to sync up, so you don't run into merge conflicts :)

flloschy:
Alright, thanks!

flloschy:
Ah thanks! Done

Issue #158, opened 2021-12-18
offline mode / downloads: remember storage location for downloads


foss-:
iOS 15.2
Finamp 0.5.1

**Currently**

When tapping on download icon `Add Downloads` dialog pops up asking for location to store files. Once the location has been selected and user taps `Add` files are downloaded. This dance has to be done for every new download.

**Expected**

Remember storage location for next download. Chances are very high user is going to use the same storage location, so remembering it will make the download dance easier.

btzs:
On some phones there is only one location available anyway. So I would suggest not to ask at all in that case.

jmshrv:
This has already been implemented, although when downloading an artist it still asks for a location for some reason.

Chaphasilor:
There are still some improvements needed. The new download dialog doesn't use default values, so you definitely need to change at least one drop down, depending on your settings. Using a default value (ideally the last-used one) would be nice

matty-r:
@Chaphasilor Commenting to try and tackle this during the hackathon!

Chaphasilor:
Awesome! There currently is a way to mark a download location as default, by going to settings > downloads > download locations and then tapping the star icon next to it.  

However, this issue is about cases where a user has multiple download locations, but no default location configured. In that case, a dialog is shown when trying to download something, with a dropdown asking where to save it.  
Currently, no location is pre-selected, but ideally either the first location in the list ("Internal Storage") or, if available, the last-used location should be pre-selected. For this, an internal setting could be created (in `finamp_models.dart`), that keeps track of the last-used location whenever a download is added. The setting doesn't need to be user-changeable.

Chaphasilor:
Added in [beta update 0.9.16](https://github.com/jmshrv/finamp/releases/tag/0.9.16-beta)!

foss-:
Finamp 0.9.16
Android 13.1

No longer seeing additional requests when downloading music for offline usage. Thanks so much @matty-r.

PR #1129, opened 2025-03-29
Remember last used download location

matty-r:
Pull Request for #158 

This should bring about the desired request for the issue. If a default location is set then that will be used, however if the list is shown then the last used download location will be automatically selected - otherwise just the first item in the list will be selected (i.e Internal Storage). If the selection dialog isn't presented, i.e no transcode required, and a default had been set, then that will be the last used download location.


Chaphasilor:
@matty-r thanks for the PR!  
I've tested it out and found a small bug that you couldn't find, it's related to the migration from the stable version to the beta. The stable version uses a deprecated default location which can't be selected for downloads in the beta any more, but it's still the first location in the map. So trying to set it as the pre-selected option resulted in an error. 

I've changed that to use the default internal location of the beta instead as a fallback :)  

Everything else seemed to work perfectly for me!  
I've renamed the new setting to reflect that it's just an ID and not an actual `DownloadLocation` object, but that's just a minor detail...

I'll merge this once the automatic build are done!

Chaphasilor:
Awesome, glad to have that in!  
If you want some stickers after the hackathon and haven't done so already, you can fill out this form: https://app.formbricks.com/s/cm8tajvx13912s001l9it719v

lib/components/AlbumScreen/download_dialog.dart
```diff
@@ -173,28 +178,42 @@ class _DownloadDialogState extends State<DownloadDialog> {
...
+
     return AlertDialog(
       title: Text(AppLocalizations.of(context)!.addDownloads),
       content: Column(
         mainAxisSize: MainAxisSize.min,
         crossAxisAlignment: CrossAxisAlignment.start,
         children: [
-          if (widget.downloadLocationId == null)
+          // Only show if there are multiple download locations
+          if (FinampSettingsHelper
```
Komodo5197, line 198: This needs to filter out the internal documents directory before determining if there are more than 1 valid location, like the items list for the dropdown does.

lib/components/AlbumScreen/download_dialog.dart
```diff
@@ -173,28 +178,42 @@ class _DownloadDialogState extends State<DownloadDialog> {
       }
     }
 
+    DownloadLocation? getFirstSelectedLocation() {
+      // If we haven't selected anything, first try and grab the lastUsedDownloadLocation, then try the default, otherwise just select the first available option
+      selectedDownloadLocation ??= settings
```
Komodo5197, line 183: The default download location should have priority over the last selected location.  Note the the default is not an app default, but a location explicitly selected to be the default by the user.

lib/components/AlbumScreen/download_dialog.dart
```diff
@@ -126,12 +129,14 @@ class DownloadDialog extends StatefulWidget {
...
 }
 
-class _DownloadDialogState extends State<DownloadDialog> {
+class _DownloadDialogState extends ConsumerState<DownloadDialog> {
   DownloadLocation? selectedDownloadLocation;
   bool? transcode;
 
   @override
   Widget build(BuildContext context) {
+    final settings = ref.watch(finampSettingsProvider).value;
```
Komodo5197, line 138: I don't think we actually want to rebuild this widget on settings changes.  We should probably just use FinampSettingsHelper.finampSettings instead of watching the provider.

Chaphasilor:
@Komodo5197 addressed your review in 354ca2be2f9b7fe05878510b5a9046b9ea657ae0 :)  
thanks!

PR #1130, opened 2025-03-29
Remove blank space when 'show album before lyrics' is disabled

F-4Dev:
When the 'Show album before lyrics' setting in the Layout & Theme lyrics view tab is ~~enabled~~ disabled, the album cover is hidden, resulting in a blank space being left behind. Now the lyrics start at the top of the screen. 

Before:
image

After:
image

I also removed the album cover when no lyrics are available becuase I think it makes a bit more scenes:
image

### DIFF
--- a/lib/screens/lyrics_screen.dart
+++ b/lib/screens/lyrics_screen.dart
@@ -235,7 +235,9 @@ class _LyricsViewState extends ConsumerState<LyricsView>
                   constraints: BoxConstraints(
                     maxHeight: constraints.maxHeight - 180,
                   ),
-                  child: const PlayerScreenAlbumImage()),
+                  child: (finampSettings?.showLyricsScreenAlbumPrelude ?? true)
+                      ? const PlayerScreenAlbumImage()
+                      : SizedBox()),
               const SizedBox(height: 24),
               Icon(
                 icon,
@@ -354,20 +356,20 @@ class _LyricsViewState extends ConsumerState<LyricsView>
                         if (index == 0)
                           AutoScrollTag(
                             key: const ValueKey(-1),
-                            controller: autoScrollController,
-                            index: -1,
-                            child: SizedBox(
-                              height: constraints.maxHeight * 0.65,
-                              child: Center(
-                                  child: SizedBox(
-                                      height: constraints.maxHeight * 0.55,
-                                      child: (finampSettings
-                                                  ?.showLyricsScreenAlbumPrelude ??
-                                              true)
-                                          ? const PlayerScreenAlbumImage()
-                                          : null)),
-                            ),
-                          ),
+                              controller: autoScrollController,
+                              index: -1,
+                              child: (finampSettings
+                                          ?.showLyricsScreenAlbumPrelude ??
+                                      true)
+                                  ? SizedBox(
+                                      height: constraints.maxHeight * 0.65,
+                                      child: Center(
+                                          child: SizedBox(
+                                        height: constraints.maxHeight * 0.55,
+                                        child: const PlayerScreenAlbumImage(),
+                                      )),
+                                    )
+                                  : Text("\n")),                           
                         AutoScrollTag(
                           key: ValueKey(index),
                           controller: autoScrollController,
--- a/lib/screens/lyrics_screen.dart
+++ b/lib/screens/lyrics_screen.dart
@@ -369,7 +369,9 @@ class _LyricsViewState extends ConsumerState<LyricsView>
                                         child: const PlayerScreenAlbumImage(),
                                       )),
                                     )
-                                  : Text("\n")),                           
+                                  : SizedBox(
+                                    height: MediaQuery.of(context).size.height * 0.05,
+                                  )),                           
                         AutoScrollTag(
                           key: ValueKey(index),
                           controller: autoScrollController,
--- a/lib/screens/lyrics_screen.dart
+++ b/lib/screens/lyrics_screen.dart
@@ -370,7 +370,7 @@ class _LyricsViewState extends ConsumerState<LyricsView>
                                       )),
                                     )
                                   : SizedBox(
-                                    height: MediaQuery.of(context).size.height * 0.05,
+                                    height: constraints.maxHeight * 0.2,
                                   )),                           
                         AutoScrollTag(
                           key: ValueKey(index),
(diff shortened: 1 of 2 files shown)


Chaphasilor:
Awesome, thanks! I think I originally kept that empty space so the first lyrics line starts in the center of the screen, but I agree that it looks a bit off...

I'm assuming you're only hiding the album cover on when there are no lyrics if the "Show album before lyrics" setting is disabled?

lib/screens/lyrics_screen.dart
```diff
@@ -354,20 +356,20 @@ class _LyricsViewState extends ConsumerState<LyricsView>
...
+                                      true)
+                                  ? SizedBox(
+                                      height: constraints.maxHeight * 0.65,
+                                      child: Center(
+                                          child: SizedBox(
+                                        height: constraints.maxHeight * 0.55,
+                                        child: const PlayerScreenAlbumImage(),
+                                      )),
+                                    )
+                                  : Text("\n")),
```
Chaphasilor, line 372: Have you tried using a SizedBox.shrink() instead?

Chaphasilor:
Change looks good! Just the one comment, and then we're ready to merge :D  
I also noticed the title of the setting wasn't super clear, so I improved that too!

Maxr1998:
> Awesome, thanks! I think I originally kept that empty space so the first lyrics line starts in the center of the screen, but I agree that it looks a bit off...

Could this be set to a screen height percentage instead? I agree it's not that nice when it's so low, but moving the lyrics completely to the top doesn't look great either.

Chaphasilor:
Fine by me. What do you think @F-4Dev ?

lib/screens/lyrics_screen.dart
F-4Dev: I tried that but then the lyrics partially overlapped by the top bar. And I can't set the min height or something for a SizedBox.shrink right?
image
I could use a normal SizedBox and set the height but I thought we could make i depended on the font size but i am not 100% sure if this is a good idea.

F-4Dev:
@Chaphasilor @Maxr1998 
image


What about that? Now its depended on the screen height ^^
[f45c361](https://github.com/jmshrv/finamp/pull/1130/commits/f45c361ca902686dd8131d187817721e75779147)

lib/screens/lyrics_screen.dart
```diff
@@ -354,20 +356,22 @@ class _LyricsViewState extends ConsumerState<LyricsView>
...
+                                  ? SizedBox(
+                                      height: constraints.maxHeight * 0.65,
+                                      child: Center(
+                                          child: SizedBox(
+                                        height: constraints.maxHeight * 0.55,
+                                        child: const PlayerScreenAlbumImage(),
+                                      )),
+                                    )
+                                  : SizedBox(
+                                    height: MediaQuery.of(context).size.height * 0.05,
```
Maxr1998, line 373: I wonder if this could also just use `constraints.maxHeight`. Maybe a bit more appropriate than using the screen height, since the rest of the views use that too. I'd also increase the blank space a little, obviously not as much as before (0.65/0.55), but maybe something like 0.25 or 0.30?

F-4Dev: what about `constraints.maxHeight * 0.2`? Then its the same as the SizedBox under the lyrics.

image

It's stil feels a bit much for me personally, maybe `0.15`?

Maxr1998: I'd personally go higher, but fair. What would 0.15 look like?

F-4Dev: image
This is 0.15, what do you think?

Chaphasilor: I'd target either the top (0.05), middle (0.5) or upper third (0.33) for a sensible design that doesn't feel arbitrary. But it's not a huge issues, so you guys agree on something ^^  

Just please don't make it customizable xD

Maxr1998: Yeah, I definitely wouldn't want that either. I'm fine with 0.15 too, somehow prefer it over 0.2. I'd personally lean towards 0.33 too, could you make one last screenshot for us with that? 😅

F-4Dev: Is it okay for you @Maxr1998  if we  go with 0.05? 
The other options just doesn't seem right to me "^^

F-4Dev: image

This seems like not removing the blank space to me xD

Maxr1998: Thank you! Yeah, 0.33 is not that great. Comparing all screenshots on my phone in fullscreen, I like 0.2 and 0.05 best. Now it's time for @Chaphasilor to make a final decision 😁

Chaphasilor: I have to admit that 0.2 *feels* the way I expected 0.33 to feel. It makes it clear that this is the top, and there are no other lyrics below. If @F-4Dev doesn't hate it, I'd go for that.  
But @F-4Dev made the PR, so you go ahead and decide. Either option is good :)

Maxr1998: > I have to admit that 0.2 *feels* the way I expected 0.33 to feel. It makes it clear that this is the top, and there are no other lyrics below. If @F-4Dev doesn't hate it, I'd go for that.

I think that's because the spacing is in addition to the app bar. Thus, it probably ends up around the 0.33 breakpoint when looking at the whole screen. So your suggestion was definitely correct.
 
> But @F-4Dev made the PR, so you go ahead and decide. Either option is good :)

Fine for me too!

F-4Dev: Okay I think 0.2 it is ^^
I admit that its more clear that this is the top of the lyrics, while not taking up to much space.

Maxr1998: Alright! Thanks a lot for the iterations and generally your contribution!

Maxr1998 (approved):

Maxr1998:
Thanks again for your contribution!

I've squashed your commits into one, since keeping them separate had no advantage. But that means that the commits in the upstream `redesign` branch (in this repo) are different from the comments in your `redesign` branch.
So you should go into your repo, into your `redesign` branch (!) and then do a `git reset upstream/redesign --hard` to sync up, so you don't run into merge conflicts. And for the future, I recommend using feature branches, so after getting your `redesign` branch up-to-date, checkout a new branch with `git checkout -b feature-name`.

Make sure to register for stickers, too!
https://app.formbricks.com/s/cm8tajvx13912s001l9it719v

PR #1131, opened 2025-03-29
Allow specifying keystore for debug builds

Maxr1998:
### Change

This change makes it possible to set an alternative keystore for debug builds. By default, debug builds use an auto-generated keystore that's managed by the Android Gradle plugin. To use your own keystore, you need to supply its path in the property `debugKeystoreFile` in `key.properties` and also set the `debugKeystorePassword`, `debugKeyAlias` and `debugKeyPassword` properties to their respective values.

Release builds already use a separate configuration.

### Motivation 

I usually sign my builds with my own constant keystore, makes it easier to sync it between computers. It also helps when testing the debug build with release data on my devices (commenting out the `applicationIdSuffix`).

Previously, I always had to exclude these local changes from commits.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,8 +17,11 @@ You can ignore the Chrome/Web component, since Finamp is not a web app.
 ### Building for Android
 
 You can build debug builds for Android right-away, but you will get an error if you try to build a release build (which is the default type).  
-To build release APKs, you need to set up a signing key for Android. To get that set up, follow [this guide](https://docs.flutter.dev/deployment/android#signing-the-app) from the Flutter documentation.  
-Note that Finamp is configured to use a different app ID for debug builds, so if you install a release build and a debug build, you will have two independent versions of Finamp installed on your device. This also means that you can keep using your regular install of Finamp (from the Play Store, etc.) while developing new features for Finamp, without breaking your ability to listen to music :)  
+To build release APKs, you need to set up a signing key for Android. To get that set up, follow [this guide](https://docs.flutter.dev/deployment/android#signing-the-app) from the Flutter documentation.
+Respectively, you can also set a custom keystore for debug builds by setting the `debugKeystoreFile`, `debugKeystorePassword`, `debugKeyAlias`, and `debugKeyPassword` variables in the same `android/key.properties` file.
+This should not be necessary for most users, but can be useful if you develop on multiple machines and easily want to share the same debug key.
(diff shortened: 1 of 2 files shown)


Chaphasilor:
Thanks for the PR!  
Would you mind sharing a bit more details about what this changes and how it can be used? 🤔

Chaphasilor:
Also, this sounds like a change that could use a sentence or two in the contribution guidelines :)

Maxr1998:
That is true. I'll add some readme notes later.

Maxr1998:
Updated the pull request description already 👍

Chaphasilor:
Ah, so by default a random per-device key is used by gradle for debug builds, and with this change you can point it to a static keystore file that you can share between multiple devices, so that debug builds from all these devices can be installed on top of each other.

And I'm guessing the `key.properties` file is gitignored, so one can customize it? Or at least the path within it is outside the repo?

Had trouble understanding the change because many relevant/referenced sections weren't shown in the PR diff ^^

Maxr1998:
> Ah, so by default a random per-device key is used by gradle for debug builds, and with this change you can point it to a static keystore file that you can share between multiple devices, so that debug builds from all these devices can be installed on top of each other.

Yup! You can also share the auto-generated one by copying `~/.android/debug.keystore` between your machines, but for the second use-case of quickly switching between debug and (custom signed) release builds, this is also helpful.

> And I'm guessing the `key.properties` file is gitignored, so one can customize it? Or at least the path within it is outside the repo?

Right, that file can already be used for release builds, as described in the document linked from `CONTRIBUTING.md`. I also just added a note about my changes to that file.

> Had trouble understanding the change because many relevant/referenced sections weren't shown in the PR diff ^^

Fair, the diff can sometimes be a bit difficult to understand.

Chaphasilor:
Wait, are you saying it's possible to install a debug build on top of a release build without losing data? 🤯  
But seems like a sensible change, and causes no issues for me, so lets hope nothing breaks for others :)

Chaphasilor:
Feel free to fill out the sticker form if you want to be sent some after the hackathon!  
https://app.formbricks.com/s/cm8tajvx13912s001l9it719v

Maxr1998:
> Wait, are you saying it's possible to install a debug build on top of a release build without losing data? 🤯

Yup! Super useful for testing things sometimes. For other Android apps I work on this is the default for me, for Flutter I mostly test on a separate app and just occasionally try debug builds with the release id, since the debug builds are just too slow for me for daily use.

> But seems like a sensible change, and causes no issues for me, so lets hope nothing breaks for others :)

Should not, since the defaults stay the same.

PR #1134, opened 2025-03-29
Allow clearing the queue from within the queue

Maxr1998:
This change adds a button in the queue sheet to clear the current queue (and stop playback). Currently, you can do this only by swipe-dismissing the now playing bar, but this only works while not in dual-pane mode (e.g. in landscape mode or on large devices) and isn't as obvious. Another alternative is dismissing the playback notification. The new button should provide a more explicit way to invoke this action.

<details>
<summary>Screenshots</summary>

</details>

Unfortunately, I currently get a flutter error when invoking the button, but only while on debug builds in portrait mode:

<details>
<summary>Error message</summary>

</details>

However, I can reproduce the same issue while being on the queue list and dismissing the playback notification, so it doesn't appear to be caused by my changes and instead seems to be a general bug within Finamp.

Currently, the queue is cleared immediately when tapping the button, we may or may not want to ask the user to confirm first.

### DIFF
--- a/lib/components/AlbumScreen/track_menu.dart
+++ b/lib/components/AlbumScreen/track_menu.dart
@@ -7,6 +7,7 @@ import 'package:finamp/components/PlayerScreen/sleep_timer_cancel_dialog.dart';
 import 'package:finamp/components/PlayerScreen/sleep_timer_dialog.dart';
 import 'package:finamp/components/delete_prompts.dart';
 import 'package:finamp/components/themed_bottom_sheet.dart';
+import 'package:finamp/l10n/app_localizations.dart';
 import 'package:finamp/models/finamp_models.dart';
 import 'package:finamp/screens/artist_screen.dart';
 import 'package:finamp/services/current_track_metadata_provider.dart';
@@ -15,7 +16,6 @@ import 'package:finamp/services/metadata_provider.dart';
 import 'package:finamp/services/music_player_background_task.dart';
 import 'package:finamp/services/queue_service.dart';
 import 'package:flutter/material.dart';
-import 'package:finamp/l10n/app_localizations.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
 import 'package:get_it/get_it.dart';
@@ -48,6 +48,7 @@ Future<void> showModalTrackMenu({
   BaseItemDto? parentItem,
   Function? onRemoveFromList,
   bool confirmPlaylistRemoval = true,
+  bool showClearQueue = false,
 }) async {
   final isOffline = FinampSettingsHelper.finampSettings.isOffline;
   final canGoToAlbum = item.parentId != null;
@@ -71,6 +72,7 @@ Future<void> showModalTrackMenu({
           canGoToGenre: canGoToGenre,
           onRemoveFromList: onRemoveFromList,
           confirmPlaylistRemoval: confirmPlaylistRemoval,
+          showClearQueue: showClearQueue,
           childBuilder: childBuilder,
           dragController: dragController,
         );
@@ -91,6 +93,7 @@ class TrackMenu extends ConsumerStatefulWidget {
     required this.canGoToGenre,
     required this.onRemoveFromList,
     required this.confirmPlaylistRemoval,
+    required this.showClearQueue,
     this.parentItem,
     required this.childBuilder,
     required this.dragController,
@@ -106,6 +109,7 @@ class TrackMenu extends ConsumerStatefulWidget {
   final bool canGoToGenre;
   final Function? onRemoveFromList;
   final bool confirmPlaylistRemoval;
+  final bool showClearQueue;
   final ScrollBuilder childBuilder;
   final DraggableScrollableController dragController;
 
@@ -583,6 +587,20 @@ class _TrackMenuState extends ConsumerState<TrackMenu> {
               },
             ));
       }),
+      Visibility(
+        visible: widget.showClearQueue,
+        child: ListTile(
+          leading: Icon(
+            Icons.clear_all,
+            color: iconColor,
+          ),
+          title: Text(AppLocalizations.of(context)!.clearQueue),
+          onTap: () async {
+            if (context.mounted) Navigator.pop(context);
+            await _queueService.stopPlayback();
+          },
+        ),
+      ),
     ];
   }
 
--- a/lib/components/AlbumScreen/track_menu.dart
+++ b/lib/components/AlbumScreen/track_menu.dart
@@ -594,7 +594,7 @@ class _TrackMenuState extends ConsumerState<TrackMenu> {
             Icons.clear_all,
             color: iconColor,
           ),
-          title: Text(AppLocalizations.of(context)!.clearQueue),
+          title: Text(AppLocalizations.of(context)!.stopAndClearQueue),
           onTap: () async {
             if (context.mounted) Navigator.pop(context);
             await _queueService.stopPlayback();
--- a/lib/components/AlbumScreen/track_menu.dart
+++ b/lib/components/AlbumScreen/track_menu.dart
@@ -591,7 +591,7 @@ class _TrackMenuState extends ConsumerState<TrackMenu> {
         visible: widget.showClearQueue,
         child: ListTile(
           leading: Icon(
-            Icons.clear_all,
+            TablerIcons.clear_all,
             color: iconColor,
           ),
           title: Text(AppLocalizations.of(context)!.stopAndClearQueue),
--- a/lib/components/PlayerScreen/player_buttons_more.dart
+++ b/lib/components/PlayerScreen/player_buttons_more.dart
@@ -44,6 +44,7 @@ class PlayerButtonsMore extends ConsumerWidget {
               showPlaybackControls: true, // show controls on player screen
               parentItem: inPlaylist ? queueItem!.source.item : null,
               isInPlaylist: inPlaylist,
+              showClearQueue: true,
             );
           },
         ),
--- a/lib/components/PlayerScreen/queue_list.dart
+++ b/lib/components/PlayerScreen/queue_list.dart
@@ -704,9 +704,10 @@ class _CurrentTrackState extends State<CurrentTrack> {
           _queueService.getQueueStream(),
           (a, b) => _QueueListStreamState(a, b)),
       builder: (context, snapshot) {
-        if (snapshot.hasData) {
-          currentTrack = snapshot.data!.queueInfo?.currentTrack;
-          mediaState = snapshot.data!.mediaState;
+        var data = snapshot.data;
+        currentTrack = data?.queueInfo?.currentTrack;
+        if (data != null && currentTrack != null) {
+          mediaState = data.mediaState;
 
           final currentTrackBaseItem = jellyfin_models.BaseItemDto.fromJson(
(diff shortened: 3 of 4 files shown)


Komodo5197:
I believe the error is downstream from a failed null check in CurrentTrack where we should be checking for an actual non-null track instead of just snapshot.hasData.   Regarding a confirmation, I'd lean towards not having one.  The logic for the existing swipe down method not having a confirm is that if you do clear the queue by accident, you can just restore it from the queue restore screen.  Hopefully the user actually knows that that exists.

Maxr1998:
> I believe the error is downstream from a failed null check in CurrentTrack where we should be checking for an actual non-null track instead of just snapshot.hasData.

Thanks! I was worried for a moment that my button callback was the issue, but after realizing it also happens when clearing from the notification, I indeed seem to have found an unrelated bug.
I'm not sure I can trust the logs because the stack traces were a bit different each time, but it might be related to the code that handles the two-panel layout.
Confusingly, the issue only happens in single-panel/portrait mode, when I test in landscape the error doesn't occur.

> Regarding a confirmation, I'd lean towards not having one.  The logic for the existing swipe down method not having a confirm is that if you do clear the queue by accident, you can just restore it from the queue restore screen.  Hopefully the user actually knows that that exists.

I agree with that. Maybe a toast notification could be shown exactly once to educate the user about the existence of that option.

Chaphasilor:
> Confusingly, the issue only happens in single-panel/portrait mode, when I test in landscape the error doesn't occur.
That's probably because landscape mode was implemented by @Komodo5197, who actually seems to know what they're doing, and not by me 😅

I think the button placement is fine, but I'm not sure about the wording. I would expect a "clear queue" button to only clear the queue (any upcoming tracks), but not to stop playback.  
I've wished for a way to clear the queue in the past, but wouldn't have wanted for playback to stop immediately.

Maybe we could have a "Clear queue" button within the queue panel, and a "stop" button somewhere on the player screen?  

Also, how are you dismissing the notification? That only works when playback is paused, right? It might also depend of the "low-priority mode when paused" setting?

Maxr1998:
> I think the button placement is fine, but I'm not sure about the wording. I would expect a "clear queue" button to only clear the queue (any upcoming tracks), but not to stop playback.

That's fair. Ideally, the text would be "stop playback and clear queue", but that's simply too long. Maybe there's enough space to add a "full", but I'm not sure if that's noticeably better at bringing the point across.

> I've wished for a way to clear the queue in the past, but wouldn't have wanted for playback to stop immediately.

I don't see that useful enough to be implemented.

> Maybe we could have a "Clear queue" button within the queue panel, and a "stop" button somewhere on the player screen?  

I originally wanted the stop button on the player screen while in dual-pane mode (for an X example in the top right), but was worried about accidental presses.
Having both doesn't seem as useful to me, as I said.

> Also, how are you dismissing the notification? That only works when playback is paused, right? It might also depend of the "low-priority mode when paused" setting? 

I'm doing that while having playback paused, yes. The low priority mode is turned off on my device.

Chaphasilor:
How about short press to clear, long press to stop? Like with the add to playlist / like button?

Maxr1998:
> How about short press to clear, long press to stop? Like with the add to playlist / like button?

You mean clearing everything except the current track? I find that a little unintuitive. Instead, I'd only allow it why playback is paused or stopped, and otherwise disable or hide the button.

If the playback is already paused/stopped, it should be obvious that clear queue clears the whole queue including the current track. It's the same for other music players.

Chaphasilor:
I really don't think it's all that unintuitive. And even Jellyfin Web supports this:

https://github.com/user-attachments/assets/3e6d4cf9-1351-4252-9a6c-98464fdc9df0

Here's a short survey of other apps on my phone:  
Spotify doesn't support stopping playback at all.  
YT Music has a "Dismiss Queue" button in the menu, but that simply closes the player screen.
Symfonium has a FAB for "Clear queue", with a confirmation, that stops playback, but also a menu option for "Clear remaining queue".
Musicolet has a "Remove this queue" option, which deletes the named queue and stops playback. In the menu for each track within the queue there's also a "Stop playback after this track" option. I really like this approach.

The "stop playback after this track" version would also tie in well with the "play last track until the end" option of the sleep timer, which is another open issue.

Maxr1998:
I see, that's a lot more complex than I expected then 😅 
My primary motivation/use case for this ticket was mapping the already available action of completely clearing the whole queue by swiping down on the now playing bar to the dual panel layout.

I generally use this option a lot due to my somewhat uncommon approach of listening to music: I manually queue most of my tracks, when I'm done with a specific queue, I'm very unlikely to listen to it again. After a queue finishes, Finamp (and generally every music player) moves back to the start of that queue. That's ok, but queuing new tracks adds them to the end of that queue (which I want), but requires me to scroll down to the first new track and listen from there. Thus, clearing the queue in advance and starting with a clean slate is quite useful for me. (I hope that explanation makes sense.)

Now, I wonder what the best approach for this PR would be, since there seem to be multiple concepts for this:

- remove all following tracks, just like you can clear next up
- remove all previous tracks
- remove all tracks from the queue, which (sometimes inadvertently) stops current playback as well
- stop playback after the current song, keep the queue (useful for sleep timers)

The last option definitely makes sense to be set onto each individual track, but feels out of scope for this PR.

Chaphasilor:
Your explanation makes total sense! Just some suggestions:  
- If you disable the "Start Instant Mix on Tap" setting or whatever it's called, you could easily create a brand new queue just by playing, without having to clear the old one. Adding a "Play" option to the menus would also make this possible, and was discussed yesterday somewhere.  
- How about an option to automatically stop/clear the queue when it is finished? Seems like that would solve your problem quite elegantly?

While we're talking about use cases, here's mine: I hate stopping a track partway through, so I tend to add tracks I'd like to listen to spontaneously into Next Up, and wait for tracks to finish before I pause. I also give dancing lessons where the students usually get a single track for practicing at a time, and sometimes forget to only add individual tracks to the queue. And in those cases I wish for a way to clear the queue similar to how it's possible to clear Next Up. So an option to "stop playback after this track" would work for both use cases!

For this PR, I'd suggest we keep it as a way to stop the queue, paused or not. If you feel the current button placement is fine, then leave it there, and we'll simply put the option to stop after each track into the track menu.  
But from a consistency standpoint, the Next Up section already has a "Clear Next Up" button in a similar spot as your new button, and it would be strange for them to behave too differently. I wouldn't know where else to put it though.

Maxr1998:
> If you disable the "Start Instant Mix on Tap" setting or whatever it's called, you could easily create a brand new queue just by playing, without having to clear the old one.

Right, but that doesn't work for individual tracks (since it queries the whole "collection")…

> Adding a "Play" option to the menus would also make this possible, and was discussed yesterday somewhere.

…but this would, and act as a really useful alternative. Especially since Jellyfin Web already supports that, it also makes sense for Finamp.

> How about an option to automatically stop/clear the queue when it is finished? Seems like that would solve your problem quite elegantly?

I also really like this. Since that's another thing Jellyfin Web does, having it as an option at least seems sensible. Nonetheless, there are reasons not to enable it, so it shouldn't be the default. And I hope we can justify adding another toggle for it…

> While we're talking about use cases, here's mine: I hate stopping a track partway through, so I tend to add tracks I'd like to listen to spontaneously into Next Up, and wait for tracks to finish before I pause. I also give dancing lessons where the students usually get a single track for practicing at a time, and sometimes forget to only add individual tracks to the queue. And in those cases I wish for a way to clear the queue similar to how it's possible to clear Next Up. So an option to "stop playback after this track" would work for both use cases!

Very interesting use case, and definitely another strong argument for the "stop playback after this track" setting. Thinking about it, it shouldn't be too hard to implement it either.

> For this PR, I'd suggest we keep it as a way to stop the queue, paused or not. If you feel the current button placement is fine, then leave it there, and we'll simply put the option to stop after each track into the track menu.  

Sounds good!

> But from a consistency standpoint, the Next Up section already has a "Clear Next Up" button in a similar spot as your new button, and it would be strange for them to behave too differently. I wouldn't know where else to put it though.

I'm not very happy with the current position, but I had a hard time finding any position in the first place. I initially tried putting it right next to the header, but that looked *terrible*. Maybe we should revisit the idea of only showing that button on the second pane in the first place, since single pane mode already has the drag to dismiss option.

Chaphasilor:
> Right, but that doesn't work for individual tracks (since it queries the whole "collection")…

It does on the tracks tab xD  
But yeah, I can see the problem there.  
For the "Play" menu entry, I'd also like to have a "Play from here" option as an equivalent to simply tapping a track within a collection then :) Just in addition to regular play

Stopping/clearing on end of queue can be an option. Either in the settings, or as a fourth `FinampLoopMode`, although the latter might be messy to implement. Would be neat if you could tackle that then!

> Maybe we should revisit the idea of only showing that button on the second pane in the first place, since single pane mode already has the drag to dismiss option.
I think we should make any action discoverable visually, not just with randomly hidden gestures. So a button or menu entry somewhere to stop playback is definitely a good thing to do, not just in split screen mode.  
But we could just put it into the player screen menu, and then have a long-press gesture on the play/pause button as a shortcut? Would that work?

Chaphasilor:
Did you know that it's not possible to link an issue that starts with `[0.9.3]`, or probably and number after the bracket? 😅  
Had to rename the issue first...

Chaphasilor:
Just to make sure we're on the same page: Are you planning to change the wording/placement of the new button, or are you considering this PR done?

Maxr1998:
Still planning to make some changes, but my week is stacked 😅 
I'll look into all the open tasks on the weekend.

Maxr1998:
(only rebased, please ignore)

Maxr1998:
Made the "clear queue" an item in the track model now. Less in your face, but still discoverable now. And hopefully clear enough. We also have more space now, so we could write "stop playback and clear queue", or add a hint as a subtitle for the `ListTile`.

<details>
<summary>Screenshots</summary>



</details>

Chaphasilor:
Alright, I think that looks good!  
Is that only for the current track / the player screen menu, or for any track menu?

This should also work well for adding an option to "Stop after this track" in combination with #1166 :D

Chaphasilor:
@Maxr1998 looks like clearing from the queue menu throws an assertion error. Clearing from the player menu works fine.  
And let's go with the "Stop Playback and Clear Queue" message :)

Chaphasilor (changes requested):
Looks like clearing from the queue menu throws an assertion error. Clearing from the player menu works fine.  
And let's go with the "Stop Playback and Clear Queue" message :)

Maxr1998:
> Is that only for the current track / the player screen menu, or for any track menu?

Only for those launched from the player screen or queue panel (on the active track only).

> Looks like clearing from the queue menu throws an assertion error. Clearing from the player menu works fine.

I fear that's the same issue I had in my original version. Would need to look into how I could fix it.

> And let's go with the "Stop Playback and Clear Queue" message :)

Even when playback is paused?

Chaphasilor:
> Even when playback is paused?

Yeah, just keep it consistent. Especially since pressing the "stop" button in the media notification/controls will have this exact effect. So we *are* stopping, at least in the way that Finamp stops.

Maxr1998:
> And let's go with the "Stop Playback and Clear Queue" message :)

Updated :white_check_mark:

> Looks like clearing from the queue menu throws an assertion error.

I *think* I fixed this? Seemed far too easy, so please try if you can still reproduce it.

Chaphasilor:
Yep, seems to be fixed!  

Thanks for the PR!

Issue #458, opened 2023-06-09
Create playlists feature


Y0ngg4n:
Add the feature to create a new playlist. Currently this is not supported.

Chaphasilor:
I could design this for the redesign, in case you want to implement it yourself 😁

Y0ngg4n:
@Chaphasilor  yes please :)

Chaphasilor:
@Y0ngg4n btw, it is possible to create a playlist when adding a song to a playlist. On the playlist selection screen, there's a "+" button on the bottom right. But we could also show that button on the playlist list as well I guess

pixel24:
> @Y0ngg4n btw, it is possible to create a playlist when adding a song to a playlist. On the playlist selection screen, there's a "+" button on the bottom right. But we could also show that button on the playlist list as well I guess

That's correct :-) However, the playlist is automatically set to "Public," and unfortunately, it can't be deleted in Finamp :-(

flloschy:
There is a hackathon planed soon and I plan on always allowing playlist deletion, no promises though :)
In the mean time if you are using the beta you can enable Server-side deletion of media in `Settings>Interactions>Allow deletion from server`, this will allow you to delete playlists.

flloschy:
@stephen304 Playlists can always be deleted from now on (if jellyfin think you are allowed to) :D
Well not now now, but next beta release

PR #1135, opened 2025-03-30
Always allow playlist deletion

flloschy:
This bypasses the "allow deletion from Server" setting for Playlists since if those get deleted the files are still there.

Request: https://github.com/jmshrv/finamp/issues/458#issuecomment-2751684686


Maxr1998 (approved):

Chaphasilor:
Awesome, thanks!  
I guess the issue should stay open, given that it's originally about having a way to create a new playlists without having to add an item to it right away :)

PR #1139, opened 2025-03-31
Automatic Offline Mode

flloschy:
Automatically enable Offline mode when loosing WiFi connection.
This also adds a new dependency: https://pub.dev/packages/connectivity_plus

When "Auto Offline Mode" is enabled and Wifi connection is lost offline mode will be activated. (and later disabled again of course)
When "Auto Offline Mode" is enabled and the user manually enabled offline mode the feature will stay paused until the user disables offline mode again.

Idk why there are so many changes... I figured Ill push them anyway 

I havent tested yet because I dont know where to initialize the `AutoOffline` Object, it needs to keep running in the background and have access to an `WidgetRef`.

Ill add Translations once this feature is done.


Chaphasilor:
Thanks for this!  
Haven't tried it yet, but have some general notes:

There's a `postLaunchHook`, currently in the `music_screen.dart`, will be moved to `home_screen.dart` soon, which you can use to pass the WidgetRef. I think something similar is done in #850, so take a look at that.

The added diff might be because you used `hive_ce`, I'm not sure if that's what we use in the rest of the app?  
Or maybe Hive had an update.

Overall this solution seems very nice! Maybe add a delay for the switching though, the `connectivity_plus` readme said that the events fired by the listener can be unreliable and contain false negatives, especially on iOS :)

flloschy:
How do you mean the delay? Like this?
If a network change is registered, wait for 15 sec, check again, still no connection -> toggle

lib/components/InteractionSettingsScreen/auto_offline_selector.dart
```diff
@@ -0,0 +1,25 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<Box<FinampSettings>>(
+      valueListenable: FinampSettingsHelper.finampSettingsListener,
+      builder: (_, box, __) {
+        return SwitchListTile.adaptive(
+          title: Text("Auto Offline Mode"),
+          subtitle: Text("Automatically turn on offline mode when switching to a cellular connection, turns off automatically when establishing a wifi connection. If you manually enable offline mode it wont be turned off automatically."),
+          value: FinampSettingsHelper.finampSettings.autoOffline,
```
Komodo5197, line 19: We can prevent some excess rebuilds here by taking out the valuelistenable, making AutoOfflineSelector a consumerWidget, and watching finampSettingsProvider.autoOffline here.

lib/services/auto_offline.dart
```diff
@@ -0,0 +1,49 @@
...
+    bool hasEth = result.contains(ConnectivityResult.ethernet);
+    bool isConnected = hasWifi | hasEth;
+    Logger("AutoOffline").info(isConnected);
+    AutoOffline.setOfflineMode(!isConnected);
+  });
+
+  AutoOffline(WidgetRef ref) {
+    bool featureEnabled = ref.watch(finampSettingsProvider.autoOffline);
+    // false = user overwrote offline mode
+    bool featureActive =
```
Komodo5197, line 24: This riverpod usage is invalid.  We are currently in initState for the music screen, so watching a provider causes an error, and regardless of that we have the music screen's ref so we are triggering unwanted builds of that while not rebuilding AutoOffline, which is created in initState.  I think what we want to do is make AutoOffline a provider itself, and then call ref.listenManual with an empty handler on it from the musicscreen or main to force it to stay built.  Also, make sure to cancel the connectivity listener when the provider rebuilds.

lib/services/auto_offline.dart
```diff
@@ -0,0 +1,49 @@
...
+
+class AutoOffline {
+  final _autoOfflineLogger = Logger("AutoOffline");
+  final StreamSubscription<List<ConnectivityResult>> listener =
+      Connectivity()
+          .onConnectivityChanged
+          .listen((List<ConnectivityResult> result) {
+    bool hasWifi = result.contains(ConnectivityResult.wifi);
+    bool hasEth = result.contains(ConnectivityResult.ethernet);
+    bool isConnected = hasWifi | hasEth;
```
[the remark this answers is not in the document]
flloschy: Ive done that now and generated all the files but I get this.

`HiveError: Cannot write, unknown type: AutoOfflineOption. Did you forget to register an adapter?`

As far as I can tell ive done everything just like the other multi choice options :/
Mind taking a look? Ive pushed my changes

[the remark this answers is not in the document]
Komodo5197: You need to register the adapter in setupHive() in main.  Or maybe switch us over to Hive.registerAdapters(); to do that automatically now that we've got that generating for us.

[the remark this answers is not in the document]
flloschy: Ahhhh thanks! 
I dont know how to setup auto generation though

lib/services/auto_offline.dart
flloschy: Yeah I dont know how to do that either. Looked at the files with a `.g.dart` counter part in services but they are either based on Hive having a Listener already or using another listener. The Riverpod documentation is also not really helpful. I tried around but cant get anything good enough, may also just be a bit late

lib/services/auto_offline.dart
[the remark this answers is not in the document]
Chaphasilor: There auto-generation is already there, that's the new file you were wondering about. You can now replace all the `Hive.registerAdapter(blabla)` calls in `setupHive()` with one single `Hive.registerAdapters()` call (with an s).  
At least that's what I'm understanding, haven't looked into it myself yet

lib/services/auto_offline.dart
Chaphasilor: I'm no expert either, but here's what I understood:  
1. The `WidgetRef` we get is used by Riverpod to rebuild the widget if something changes. So if we watch something using the ref from `MusicScreen`, then `MusicScreen` will rebuild any time the value watched in AutoOffline changes.
2. We can essentially convert any regular class into a Provider by annotating it with `@Riverpod`. Riverpod will then generate some code, and we should be able to use `ref` within the class. But we still need to ensure that the AutoOffline instance stays alive after the `postLaunchHook` completes, that's why we're adding a `ref.listenManual(() {})` on it, which keeps the scope (where we created the `AutoOffline` instance) active.
3. Canceling the connectivity listener should be done because we'd be adding a new listener every time the AutoOffline instance is reloaded through the internal Provider (remember the `@Riverpod` annotation?). So to prevent adding more and more listeners, you need to remove the old listener when the old instance of `AutoOffline` is destroyed (like the `onDispose()` handler for widgets, before you then re-attach the listener. I'm not sure how exactly to do that in this case though, since we're dealing with an annotated class, not a Widget. But maybe `onDispose()` is a Dart feature, not a Flutter one, and it will work exactly the same?

Chaphasilor:
Hmm, maybe there should be a primer in `CONTRIBUTING.md` then 🤔  
Let me whip something up quickly. Anything in particular that isn't clear?

flloschy:
I guess its less of a Finamp specific thing. I just still need to develop an understanding of Mobile Development. Quite different than web/"script" dev :D

flloschy:
Thats my attemp at Riverpod with the help of Chaphasilor's comment
I doesnt work, but hey no build or runtime errors :)

Komodo5197:
Regarding foreground/background, as far as I'm aware all the dart code continues working like normal for as long as the app is active, which includes any time where music is actively playing.  Basically the only thing that happens is widgets stop building, so code in build() doesn't trigger and animations don't progress, which also happens all the time normally when we put one screen on top of the other.  Dart code in an OS callback, like the queue service or the connectivity listener, and anything operating off a dart native timer/delay like Future.delayed() or Stream.periodic() will continue operating as normal.  I believe the app can also be completely paused without fully closing, but I don't think anything at all can happen in that state, so it doesn't matter much?

lib/services/auto_offline.dart
Komodo5197: 1. Correct.
2. Pretty much correct.  Riverpod providers are lazily evaluated and are only built when watched by an active widget/provider.  listenManual will force riverpod to immediately rebuild the provider every time it gets invalidated by the watched values changing, because the active music screen widget is depending on it.  The fact that we don't actually need the results of the provider in music screen means that this is a bit hacky, but whatever.  Notably, marking a provider as keepAlive prevents it from being disposed if no one is watching, but can't actually trigger a build, so in this case because the music screen should always be watching it doesn't really make a difference either way.
3. Correct.  When the watched values change, all riverpod does is re-run the build method.  So we need to clean up the old listener with ref.onDispose because we're about to create a new one in the next call to build().  Theoretically we could store the listener statically to avoid recreating it, but disposing it is more standard for providers to make sure we don't accidentally keep old state around.

Chaphasilor: Hey, 2.5/3, that's a win in my book :D  
Btw, I'm planning to add some sort of caching functionality, similar to what you did for with the cover images, for the sections on the new home screen. Let's see how much of the Riverpod stuff I'll be able to figure out 😅

lib/services/auto_offline.dart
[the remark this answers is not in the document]
flloschy: When I convert to registerAdapter**s** I just get an error that some Adapter with Id 43 already exists

flloschy:
Besides `Hive.registerAdapters()` and the gitignore thing this should be it :D

Maybe a final test on mobile? On linux it works like a charm

Chaphasilor:
I've switched to the `Hive.registerAdapters()` call, but that's giving me errors on startup about duplicate adapters. I haven't yet figured out why this is happening.

And I'll take a proper look at this PR tomorrow, when I'm fully awake...

Komodo5197:
It looks like theres' a hive annotation on ClientDiscoveryResponse with a duplicate number, but we never actually registered that type with hive, so it should be removed.  Also, running code gen deletes and doesn't recreate the hive_registrar.g.dart file for me?

Chaphasilor:
Thanks, removed it (and a second case). Not sure why I added it initially.  
I also wasn't aware that the typeIDs were shared across FinampModels and JellyfinModels, that's pretty messy.

Code gen works fine for me when running `dart run build_runner build --delete-conflicting-outputs` on Flutter 3.29.2 and `hive_ce` 2.10.1 with `hive_ce_generator` 1.8.2.

Regarding the PR, after some testing I think that the default value for `autoOfflineListenerActive` is incorrect. After a fresh install, the listener is not active, and will only become active after going and then leaving offline mode at least once. Before that only `Paused Automation` is logged when changing the mode, and nothing happens when losing connection.  
Once that bug is dealt with, everything seems to work nicely for me!

Chaphasilor:
@flloschy is this ready to merge from your side? If so, please try to resolve the merge conflicts :)

flloschy:
what do you mean github, I resolved everything :(

flloschy:
okay, still some issues...

flloschy:
There we go

flloschy:
Would this close these issue? https://github.com/users/jmshrv/projects/5/views/1?pane=issue&itemId=104112101&issue=jmshrv%7Cfinamp%7C981 https://github.com/users/jmshrv/projects/5/views/1?pane=issue&itemId=104112106&issue=jmshrv%7Cfinamp%7C1006 Not a 100% match but close

Chaphasilor:
> Though everyone on the cutting edge might need to delete finamps data because I needed to change the IDs for the swipe settings

That's normal. As long as nothing that was already released to users is changed, it's fine :)

The issues aren't resolved yet, but you could try your hand at the prompts / snackbar messages?

I'll merge once the checks are done :)

flloschy:
Was kinda expecting it to fail, the github editor kinda sucks 

What do you mean with snackbar messages?

Chaphasilor:
You called it "notification" last time, I think. When you call `GlobalSnackbar.message()`. The snackbar is the small message bar that pops up at the bottom of the screen.  
On of the issues requested that such a bar should be shown if the downloads are queued but won't start because WiFi only downloads are enabled but WiFi is disconnected :)

Chaphasilor:
🎉

Thanks for the PR! <3

flloschy:
Where even is this require wifi option, cant find it

Chaphasilor:
It's called `requireWifiForDownloads`.

And no, the message should appear when adding a new download, as a reminder that it won't start due to the current settings :)

Issue #1132, opened 2025-03-29
redesign: playback fails to resume from media notification after app is killed (android)


corngood:
Steps:

- install finamp debug apk (on pixel 8a using grapheneos)
- make sure the "enter low-priority state on pause" setting is enabled
- start playback
- pause playback and exit to launcher
- kill the app `adb shell am kill com.unicornsonlsd.finamp.debug`
- press the play button on the system media notification

At this point nothing will happen.  Attempting to open the app through the launcher result in a black screen until it's force stopped.

The same thing happens with the beta release apk if I leave it running long enough for the OS to kill it instead of killing it manually.

I tracked the hang to this code (I've added the extra logs):

```
Future<void> _setupEdgeToEdgeOverlayStyle() async {
  if (Platform.isAndroid) {
    _mainLog.info("A");
    await SystemChrome.setEnabledSystemUIMode(SystemUiMode.edgeToEdge);
    _mainLog.info("B");
```

"A" will be printed, but not "B".

If I comment the call to `setEnabledSystemUIMode`, the app will start correctly using these steps, however playback still doesn't resume (the notification shows a loading spinner in the play button) without opening the app and doing it manually.  In this case I've found that the resume behaviour may be be broken.  For example it may restart the song from the start.

Commenting `setEnabledSystemUIMode` obviously breaks the chrome.  I just thought I'd give it a try for the sake of investigation.

I'm new to all of this (flutter, etc), so still trying to wrap my head around the AudioService lifecycle.  The only thing I see in the logs about how the app gets started from the play button is:

```
03-29 11:53:27.143  1371  1458 I ActivityManager: Start proc 28129:com.unicornsonlsd.finamp.debug/u0a234 for service {com.unicornsonlsd.finamp.debug/com.ryanheise.audioservice.AudioService}
```

Chaphasilor:
Also fixes it for me, nice! The adb command you mentioned in the issue really comes in handy, thanks for letting me know about that! Makes debugging much faster...

As for resuming, it actually works already! You just need to press play in the notification, and then wait a few seconds. During these seconds the app starts up and restores the queue. Before that, no playback can happen.  
I think the starting the track of the start it just the regular queue restore behavior, since we only take a snapshot every few seconds, so the progress won't always be up-to-date. Try letting the track play for a minute or so first, then it should restore further along :)

[...]

PR #1142, opened 2025-04-01
Fix hang on splash screen when resuming app via notification

corngood:
When finamp is started as a service, there's no PlatformChannel handler, so the Future<> returned by setEnabledSystemUIMode is never completed.

This could use some more investigation, but making it fire-and-forget seems harmless.  There are samples in flutter that do the same thing.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -39,7 +39,7 @@ android {
     defaultConfig {
         applicationId "com.unicornsonlsd.finamp"
         minSdkVersion 21
-        targetSdkVersion 34
+        targetSdkVersion 35
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
     }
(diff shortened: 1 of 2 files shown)


corngood:
This is relevant to #1132, but I decided not to mark it fixed since there are still issues with playback resuming.  This just fixes the 'black screen'.

Chaphasilor:
Meant to comment in the PR, but for reference here's my comment: https://github.com/jmshrv/finamp/issues/1132#issuecomment-2770601751

Chaphasilor:
I've bumped the target SDK to level 35 now, since you found out that the full screen system UI is applied for that. Initial testing seems to work. We can keep the existing calls for now, just in case :)

corngood:
This all looks good to me.

Chaphasilor:
Awesome, then it's in! Thanks for figuring this out!  
Since this is your first contribution of the Hackathon, make sure to fill out this form if you want to receive some stickers afterwards: https://app.formbricks.com/s/cm8tajvx13912s001l9it719v

Issue #572, opened 2024-01-05
Show "Appears On" section for artists


Chaphasilor:
Jellyfin supports differentiating between albums where the artist is also an album artists and albums/compilations where the artist simply has a song within an album released by another artist.  
This differentiation is very helpful when trying to manage multiple artists per track, as well as dealing with release compilation albums (e.g. Monstercat, or soundtrack albums).

Screenshot:
![image](https://github.com/jmshrv/finamp/assets/18015147/bd8182de-21fc-40cd-8a2e-7d11ec91d140)

Finamp should support and show this section as well, so that one can still link artists with all the albums they appear on, but without requiring them to be an actual album artist for that album.  
This would also be helpful for downloading artist collections, since all direct albums could be fully downloaded, and for albums where they only appear on the metadata and their own tracks would be downloaded, but not the tracks by other artists.

bsfreq:
This would be very cool. 
I also wish you can add the option to easily add all artist songs (including appears on songs/albums) to next up / queue / playlist etc. 
I often search for an artist and I’d like to be able to shuffle all artist songs+appers on songs/albums to queue. Currently I have to tap+hold each search result separately. Perhaps you could add a new search result (Artist + appears on) that we can tap+hold to easily add all those to queue?

Chaphasilor:
Well right now you can add all artist album tracks to the queue, this would probably simply be extended to include the tracks from appears on as well...  
Or do you mean something else?

bsfreq:
> ll artist album tracks to the queue, this would probably simply be extended to include the tracks from appears on as well...
> Or do you mean something else?

This exactly. An easy way to add everything from an artist, including appears on tracks..

ravtakhar:
The image linked above is the artists page. This would be very cool on the Finamp artist page allowing users to quickly navigate between all of their songs.

lukaslindnermusic:
I will include this in PR https://github.com/jmshrv/finamp/pull/1145 😊

bsfreq:
> I will include this in PR [#1145](https://github.com/jmshrv/finamp/pull/1145) 😊

Beautiful, thank you so much!

lukaslindnermusic:
No problem 😊 
Will look like this:
![Image](https://github.com/user-attachments/assets/c1c85381-03cc-43a3-8bb2-4493a4406fc9)

If there are a lot of albums, the Appears On section obviously disappears down the list.
What do you think would be a good solution to reach it quickly? 
[...]

PR #1145, opened 2025-04-02
Album Artists and Performing Artists on Artist List and Artist Screen

lukaslindnermusic:
Work in Progress, I hope this is a good starting point.

Already implemented:
- Filter Chips on Artists List to toggle between "Album Artists" (default) and "Performing Artists"
- Remember Decision in FinampSettings
- Call the jellyfin api artists and albumartists endpoints depending on setting

ToDo:
- Implement Album Artists in Offline-Mode.
(Temporary, when switching to Offline Mode, the setting currently gets changed to performing artists and album artists gets disabled until back online)

I have to get more familiar with the download system in order to implement album artists in order to implement this.
I will dig into it either today or on the weekend.


lukaslindnermusic:
Screenshot 2025-04-02 at 16 25 19
Screenshot 2025-04-02 at 16 25 58

lukaslindnermusic:
Now everything should work. 😊 

There is a special case:
When there is an album that has an album artist assigned which is NOT an (performing) artist within that album, this album will currently not get listed on the dedicated artist screen in offline mode.
But this is out-of-scope for this PR and (as Komodo has suggested) this should be part of the ArtistScreen redesign.
(I think, implementing Jellyfins "Appears On" for albums where the artist is only artist, and for "Albums" only showing albums where assigned as album-artist would be a good idea)

Thank you to @Komodo5197 for the huge help regarding understanding (parts of) the new download system. 😊

lukaslindnermusic:
Just an additional thought for a possible future expension: In case Jellyfin adds more support for a more comprehensive differentiation of people (additional tags composers, producers, mixing engineers, conductors...) in the future aka. makes them more prominent, we could add those as filter chips here as well and call the tab "People" instead.

Chaphasilor:
Awesome, I'll try to test this tomorrow! 😁

Chaphasilor:
Okay, just gave it a try,

In online mode everything works fine, great!

In offline mode I experienced a few issues:
- A bunch of images wouldn't download. A repair only resolved half of the errors. Not sure if that is related to the PR though, that would need further testing
- Some download items showed up as "Syncing..." after all downloads were finished. Possibly due to the image errors, but when I checked they didn't seem to be related.
- When I went into an album that crossed library boundaries (by going to the album artist first), I got a "Duplicate key" error for the artist chips. See screenshot below.
- It seems like only albums where the artist is a *performing* artist show up on the artist page. It should be the other way around (and purely performing artists should have the albums in the "Appears On" section, which isn't yet implemented, see https://github.com/jmshrv/finamp/issues/572)




Logs:  
finamp-logs.txt

lukaslindnermusic:
> * A bunch of images wouldn't download. A repair only resolved half of the errors. Not sure if that is related to the PR though, that would need further testing
It looks to me like this is not related. In my testing, every image downloaded successful. 🤔

> * Some download items showed up as "Syncing..." after all downloads were finished. Possibly due to the image errors, but when I checked they didn't seem to be related.
This was fixed with commit [957bd98] 😊 (https://github.com/jmshrv/finamp/pull/1145/commits/957bd987381c7b3612bda2af27af4fc75e72ee55)

> * When I went into an album that crossed library boundaries (by going to the album artist first), I got a "Duplicate key" error for the artist chips. See screenshot below.
Hmm... I could not reproduce this. 👀 But I only tried after the "Syncing..." issue was fixed. Maybe this was fixed as a side-effect? Or it has to do with your other problem with the missing images? Could you please try it again with the current state of this PR to check if it is fixed with the newer commits?

> * It seems like only albums where the artist is a _performing_ artist show up on the artist page. It should be the other way around (and purely performing artists should have the albums in the "Appears On" section, which isn't yet implemented, see [Show "Appears On" section for artists #572](https://github.com/jmshrv/finamp/issues/572))
Is already fixed, but not yet commited. I will now work on the Appears On section on the artistScreen and commit those once they are working. 👍

Chaphasilor:
I'll test again :)  
The chips are fine, I'll try scrolling and see how it looks. But don't stress about this too much, the current design is being phased out anyway. It's just about the functionality for now.

Chaphasilor:
One thing you could maybe try is keeping both artist tabs loaded, so that switching between them doesn't reload everything, unless the user manually refreshes by swiping down.  
So just like it already works for the top-level tabs.

Downloading worked fine this time, no errors.

The items scrolling below the chips looks absolutely fine. I'd prefer merging this PR over adding fancy fade-out masks like we have on the queue panel, but theoretically that's an option.

The synching... issue is resolved, but now when going to the downloads screen I can't scroll of the first couple of seconds (until the sizes have loaded), and if I try to I get an "App not responding" dialog. I think I saw you use a `yadaydaSync` call, maybe you should use an async one instead to not block the main thread?

The "0 tracks downloaded" is still an issue, and consequently the play/shuffle buttons don't work either.

I'm still getting the duplicate key error. I think the error is here:  
https://github.com/lukaslindnermusic/finamp/blob/6cd3932ac77c04f2388f03d514c318f5e0cd1709/lib/components/PlayerScreen/artist_chip.dart#L69-L81
We should probably filter the list of `artists` to make sure its a set, i.e. no duplicates.  
Might not even originate in this PR, but I think it's in a good position to fix it.

While looking at that I also noticed we still have the `Use Album Artists` setting and use it for e.g. the artist chips. Maybe it's time to rework this with the new change? Maybe it's too early, and we need a different design or a better "Peoples" section first?

The Appears On section works great!

Chaphasilor:
Okay, so the issue with the artist chips is fixed now, thanks!  
Everything else seems to work too. Once the conflicts are resolved we can merge this I think :)

lukaslindnermusic:
> Thanks for the PR! Really a nice change, not only the proper distinction, but also making it consistent across offline and online mode! What was it again that you wanted to work on next? The genre screen and using providers for the tabs, right? Just so I can add it to the hackathon project and we don't forget about it in the coming days :)

Thank you 🤗
Yes, that inconsistency was a huge pain, glad we could fix it 😊 

Yes, if thats okay, I want to tackle the genre screen and can work on it in my free time over the next few days/weeks. 😊 
Because similar to the person that mentioned it on Github, I'm a genre-browsing person myself and therefore have a lot of motivation to fix this :P 
I can do a draft PR so that you can track the progress if you want. 😊

Chaphasilor:
Alright, I'll add it to the project board! Feel free to open a draft if you have a proof of concept working, but no need to do it right-away :)

lukaslindnermusic:
Also of course:
A huge shoutout and thank you to https://github.com/Komodo5197 who helped a great deal implementing this stuff in the download system!

lukaslindnermusic:
Found a bug where the TrackCount (more specifically "allTracks", used for track count and direct artist playback) of a PerformingArtist in OfflineMode was inconsistent from OnlineMode.

If an artist only was on one track of an album, the performing-artist screen in online mode displayed 1 Track (how it should be). 
If I now would download that album the performing-artist is on, and would go to the performing-artist screen in offline mode, it would actually include all album-tracks, not just the one the artist is on.

I fixed this already 👍
But just to document it here: the fix-commit is in my current working branch for the GenreScreen and once that hopefully gets merged, it will be fixed here as well 😊

PR #1148, opened 2025-04-03
fix: playback fails to resume from media notification after app is killed (android)

corngood:
Fixes: #1132

When the app is restarted from the notification to continue playback, `AudioService.onPlay` will be called, and the play state propagates through various async queues.

This change attempts to avoid anything cancelling out that play state.

- `initializePlayer` could do that by playing an empty queue, which immediately stops.
- `_replaceWholeQueue` could do that via its `pause` and `stopPlayback` calls.

I'm worried the potential problems this could cause, but I haven't found anything obviously broken on android yet.

Qs

- is there anything in `stopPlayback` that still needs to be done in `_replaceWholeQueue`?

```
  Future<void> stopPlayback() async {
    queueServiceLogger.info("Stopping playback");

    archiveSavedQueue();
    if (_savedQueueState == SavedQueueState.pendingSave) {
      _savedQueueState = SavedQueueState.saving;
    }

    await _queueAudioSource.clear();

    _queueFromConcatenatingAudioSource();

    await _audioHandler.stopPlayback();

    // await _audioHandler.initializeAudioSource(_queueAudioSource,
    //     preload: false);

    return;
  }
```

- what are the consequences of not calling `initializePlayer`?

Playback still seems to work when queue loading is completely disabled.

TODOs

- [ ] testing
- [x] better commit messages

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -388,7 +388,6 @@ Future<void> _setupPlaybackServices() async {
   GetIt.instance.registerSingleton<MusicPlayerBackgroundTask>(audioHandler);
   var queueService = QueueService();
   GetIt.instance.registerSingleton(queueService);
-  await GetIt.instance<QueueService>().initializePlayer();
   GetIt.instance.registerSingleton(PlaybackHistoryService());
   GetIt.instance.registerSingleton(AudioServiceHelper());
 
--- a/lib/services/queue_service.dart
+++ b/lib/services/queue_service.dart
@@ -501,9 +501,6 @@ class QueueService {
         }
       }
 
-      //await stopPlayback(); //TODO is this really needed?
-      // await _audioHandler.initializeAudioSource(_queueAudioSource);
-      await _audioHandler.stopPlayback();
       await _queueAudioSource.clear();
 
       List<AudioSource> audioSources = [];
@@ -547,8 +544,6 @@ class QueueService {
       if (beginPlaying) {
         // don't await this, because it will not return until playback is finished
         unawaited(_audioHandler.play(disableFade: true));
-      } else {
-        unawaited(_audioHandler.pause(disableFade: true));
       }
 
       _audioHandler.nextInitialIndex = null;
--- a/lib/services/queue_service.dart
+++ b/lib/services/queue_service.dart
@@ -146,9 +146,6 @@ class QueueService {
     // );
   }
 
-  Future initializePlayer() =>
-      _audioHandler.initializeAudioSource(_queueAudioSource, preload: true);
-
   void _queueFromConcatenatingAudioSource({
     bool logUpdate = true,
   }) {


Chaphasilor:
Seems like playback reporting is broken on 0.9.15 anyway, when starting a new queue the old track is never reported as stopped. So I'll have to fix that anyway.  

Testing the restore required me to manually apply the fix from #1142, so now that it's merged you should definitely rebase this PR :)
But it seemed to work just fine!  
I'm thinking about if there's a way to somehow show a loading state in the notification (either the buffer animation, or simply showing some text like "Restoring queue...") in case the restore takes a bit longer.

corngood:
> Git blame indicate it was initially added to make the queue restore on startup work, so that needs to be tested I guess. If it still works, we can get rid of the call.

`initializePlayer` just sets the audio source, and I'm pretty sure that happens when you change the queue later in order to play something.  

One reason for doing it early might be to warm up the playback systems, but I didn't notice an obvious performance problem, and it was being done synchronously so it would probably just increase start time.

That being said I've only tested android, and these audio plugins are huge complicated abstractions.

> This (playback reporting) might be the biggest problem, since it's rather fragile.

I'm not familiar with this.  Is this reporting to the jellyfin server?  If you have any ideas for testing I'll give it a go.

I'm currently set up with local copies of `just_audio` and `audio_service` I've loaded with debug logs, so I can investigate anything in that area pretty easily (on android, and maybe linux desktop).

>  I'm thinking about if there's a way to somehow show a loading state in the notification (either the buffer animation, or simply showing some text like "Restoring queue...") in case the restore takes a bit longer.

I can take a look at this.  It definitely has that loading/buffering animation state.  When I was seeing it before it was because the media session was inactive, which isn't what we want, but there's probably another way to show it.


If anyone could test this on iOS it would be very much appreciated, assuming we don't find any other problems first.

Chaphasilor:
Yeah if someone on iOS could test this that would be great. But that should probably be done at the end of this hackathon anyway.  

You don't need to handle the playback reporting stuff yourself, but if you want to look into it, I recommend going to your server's admin dashboard, and looking at the activity section there. It should ideally say "Started playing xyz" when a new track starts playing (or the current one loops), and then say "Stopped playing xyz" before the next "Started playing".  
And yes, it's powered by simple HTTP requests to the server.

If you can figure out the loading state, nice. Might be problematic to work around the audio plugin, but who knows :)

corngood:
I've rebased on latest (there was also a conflict) and improved the commit messages.

> Seems like playback reporting is broken on 0.9.15 anyway, when starting a new queue the old track is never reported as stopped. So I'll have to fix that anyway.

Do you think this change is likely to make the problem worse?  Playback reporting might be a little tricky to.  When resuming from the notification a 'play' message is sent to the `audio_player` before a queue has been created, and then setting the queue later on is what actually allows playback to start.  So there's no way to know what's going to play when playback is requested.

However, maybe it shouldn't even be reporting playback in this case?  It doesn't report pause/play transitions, does it?  Maybe resuming a queue after the app has been killed should be the same as if the app _wasn't_ killed (i.e. just a resume from pause).

Chaphasilor:
We only report plays when the `AudioPlayer` indicates a track change, so it shouldn't make things worse. Robustifying playback reporting is a different issue :)  
And we do report play/pause transitions, but not as a "started playing" or "has finished playing" event, but as an "is playing" event, which just updates the state in Jellyfin without starting a new playback event. Resuming after the app was killed should start a new play though, because otherwise there would be hour-long plays. It's also not possible to prevent, since the server will time out a play after around 5 minutes without updates. That's why we periodically actively update the current state while the app is running, even when paused. But we can't do that if the app is killed ^^

From my side this is ready to merge! What about you?

corngood:
> From my side this is ready to merge! What about you?

All good here.  I'm still a bit worried about other platforms or edge-cases, but I guess we'll see when it gets more testing.

Chaphasilor:
Yeah that's what the beta is for. But I think aside from playback reporting nothing should break :)  
Thanks for the PR!

Issue #785, opened 2024-06-19
Album Art View


sinth0s:
It would be very nice if you could enlarge the album cover by tapping on the image within the album itself. It would only need to be as big as the album art is when you're playing a song in the beta UI. Of course, I could just play a song to see the album art, but it would be a nice QOL feature.

Chaphasilor:
You're talking about tapping the cover when viewing an album, right?

sinth0s:
Yes, exactly.

fridaynext:
I'll take this one! Still setting up my dev environment and familiarizing myself with Flutter, so this seems like a good first ticket for me 👍🏻

Chaphasilor:
Awesome! Let me know if you have any questions!

fridaynext:
Had a great breakthrough today studying Flutter, and I think I should have a PR for review by tomorrow!

@Chaphasilor question for you - does this repo consider it bad from to use extra libraries, vs doing everything from scratch? I found [this library](https://pub.dev/packages/insta_image_viewer) that fits the bill quite nicely, but I didn't want to add a new library if that's not efficient enough, etc.

Chaphasilor:
@fridaynext that's great news! Generally there's no issue with adding new dependencies, especially if they are easy to replace if needed and offer additional value over a custom implementation.  
So you could use it, but using `InteractiveViewer` seems almost as simple and is definitely cleaner and easier to extend in the future :)

Looking forward to the PR!

PR #1150, opened 2025-04-03
Tap to enlarge `AlbumImage` on `AlbumScreen`

fridaynext:
## Album Art Tap to Magnify
The goal is to be able to tap the small album art image in the FlexibleSpaceBar header to reveal a larger version of the album art that is full-width on a mobile device.

### Objectives
- reuse the same image (ideally cached) rather than pulling over the network again
- incorporate a smooth animation as the image magnifies
- keep the area above and below the enlarged image transparent, to view the AlbumScreen behind
- use a "throw away" gesture to get rid of the album art
  - also tapping outside the bounds of the image will dismiss it

### Related Issue
- https://github.com/jmshrv/finamp/issues/785

### DIFF
--- a/lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
+++ b/lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
@@ -232,7 +232,8 @@ class AlbumScreenContentFlexibleSpaceBar extends StatelessWidget {
                   children: [
                     SizedBox(
                       height: 125,
-                      child: AlbumImage(item: parentItem),
+                      child: TapToZoomImage(
+                          albumImage: AlbumImage(item: parentItem)),
                     ),
                     const Padding(
                       padding: EdgeInsets.symmetric(horizontal: 4),
--- a/lib/components/album_image.dart
+++ b/lib/components/album_image.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 import 'dart:math';
+import 'dart:ui';
 
 import 'package:finamp/components/PlayerScreen/player_split_screen_scaffold.dart';
 import 'package:finamp/models/finamp_models.dart';
@@ -265,6 +266,7 @@ class _TapToZoomImageState extends State<TapToZoomImage> {
         child: Hero(
           tag: zoomedImageHeroTag,
           child: widget.albumImage,
+          placeholderBuilder: (context, heroSize, child) => widget.albumImage,
         ),
       ),
     );
@@ -287,6 +289,10 @@ class ZoomedImage extends StatelessWidget {
           child: IgnorePointer(
             child: Container(
               color: Colors.black.withOpacity(0.5),
+              child: BackdropFilter(
+                filter: ImageFilter.blur(sigmaX: 5.0, sigmaY: 5.0),
+                child: Container(),
+              ),
             ),
           ),
         ),
--- a/lib/components/album_image.dart
+++ b/lib/components/album_image.dart
@@ -4,9 +4,11 @@ import 'dart:math';
 import 'package:finamp/components/PlayerScreen/player_split_screen_scaffold.dart';
 import 'package:finamp/models/finamp_models.dart';
 import 'package:finamp/services/finamp_settings_helper.dart';
+import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_blurhash/flutter_blurhash.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
 import 'package:octo_image/octo_image.dart';
 
 import '../models/jellyfin_models.dart';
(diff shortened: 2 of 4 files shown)


Chaphasilor:
Awesome, happy to see this!  
Regarding the animation you should definitely take a look at the [`Hero` widget](https://docs.flutter.dev/ui/animations/hero-animations), which should cover this use case pretty well with almost no code needed :)

fridaynext:
@Chaphasilor definitely working toward using a Hero for this! I'm sure I'm over-doing it for the most part, but just want to get it to a working state, and then trim away the unnecessary stuff. Right now I have it all working except for the throwaway gesture. I'm thinking that has to do with the way I wrapped the containers.

Def learning a _lot_ about Flutter/Dart!

Chaphasilor:
That sounds good! I'm looking forward to future contributions then :P

Just tested it briefly, it seems to work. I can't tap to dismiss it though, not sure if you included that with the "throw to dismiss" gesture or not.

I'm also thinking if it would be possible to directly integrate this into the `AlbumImage` widget, so that it works *everywhere* right-away, as well as for other image types (artists, playlists, etc.). If that's possible to do, we should of course add a flag to suppress this behavior, like for any cases where tapping the cover already does something (player screen, now playing bar).  
What do you think? Is it possible to do? Is it easy to do it? If not, we'll just continue with this approach :)

fridaynext:
Your approach just clicked in my brain, and yes absolutely I'll do it that way. Should actually be *less* work than the way I'm doing it now 😅

fridaynext:
Made more progress, but calling it a night - I think I've got a much better grasp on how to integrate with the framework though, which will be good for future contributions! 😎

fridaynext:
@Chaphasilor would you mind taking a look at this? I'm doing all my debugging on a P9XL, and whether WiFi7 or USB4, debugging Flutter seems suuuuper slow. I'm not solid on how to build a full APK with AndroidStudio yet, so figured you'd have quicker access.

Let me know if you see anything that could be changed!

lib/components/album_image.dart
```diff
@@ -219,3 +219,75 @@ class _AlbumImageErrorPlaceholder extends StatelessWidget {
...
+  final bool isDisabled;
+
+  const TapToZoomImage({
+    Key? key,
+    required this.albumImg,
+    this.isDisabled = false,
+  }) : super(key: key);
+
+  @override
+  _TapToZoomImageState createState() => _TapToZoomImageState();
```
Chaphasilor, line 234: getting a linter warning here. try removing the underscores

lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
```diff
@@ -233,7 +233,7 @@ class AlbumScreenContentFlexibleSpaceBar extends StatelessWidget {
                   children: [
                     SizedBox(
                       height: 125,
-                      child: AlbumImage(item: parentItem),
+                      child: TapToZoomImage(albumImg: AlbumImage(item: parentItem)),
```
Chaphasilor, line 236: well that's much cleaner than before! :D

lib/components/album_image.dart
```diff
@@ -219,3 +219,75 @@ class _AlbumImageErrorPlaceholder extends StatelessWidget {
     );
   }
 }
+
+class TapToZoomImage extends StatefulWidget {
+  final AlbumImage albumImg;
```
Chaphasilor, line 224: let's try to avoid abbreviations. Make widget and argument names as long and descriptive as needed :)  

But I do like the wrapper approach, that may be a bit nicer than always using boolean flags ^^

lib/components/album_image.dart
```diff
@@ -219,3 +219,75 @@ class _AlbumImageErrorPlaceholder extends StatelessWidget {
...
+      onTap: () => Navigator.of(context).push(
+        PageRouteBuilder(
+          opaque: false,
+          barrierDismissible: true,
+          transitionDuration: const Duration(milliseconds: 300),
+          pageBuilder: (
+            BuildContext context,
+            Animation<double> animation1,
+            Animation<double> animation2
+          ) {
```
Chaphasilor, line 254: Can be blur the background here, or put a semi-transparent overlay behind the zoomed image?  

And is it possible to keep the original album image in place when opening the zoomed version, or does that break the hero transition? The empty space looks a bit weird.

lib/components/album_image.dart
```diff
@@ -219,3 +219,75 @@ class _AlbumImageErrorPlaceholder extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    if (widget.isDisabled) {
+      return widget.albumImg;
+    }
+
+    return GestureDetector(
+      onTap: () => Navigator.of(context).push(
+        PageRouteBuilder(
+          opaque: false,
+          barrierDismissible: true,
```
Chaphasilor, line 248: The dismiss gesture works, but it takes half a second for the original cover to re-appear. Not hiding the original cover could help to mitigate that, or if that isn't an option maybe you can play with the transition durations?

Chaphasilor (commented):
Seems to work pretty well! Found a few minor oddities that you could address, but that's it :)

But now that you've go the album screen down, how about you wrap a few more instances of the `AlbumImage` with your shiny new zoom helper? :D  
Good candidates would be the artist screen (which currently also serves as the genres screen, and the track menu. Should be easy to add, and make things more consistent!

fridaynext:
Will do!

fridaynext:
@Chaphasilor nah, just life stuff taking up time. I should be able to get the fixes in by tomorrow! 👍🏻

fridaynext:
@Chaphasilor made some more progress, but I borked the branch with a bad `git rebase`, so I'll just open a new PR to clean it up.

Chaphasilor:
@fridaynext haha it's alright! No need to rebase, just merge and we can squash-merge the PR at the end :)

Chaphasilor:
@fridaynext looks like you never opened a new PR 😅  

If you want me to finish this, just let me know :)

Chaphasilor:
@fridaynext sorry to bother you again. Do you still have the original commit somewhere? It looks like this PR doesn't include any of your changes anymore.  
I could try digging through the reflog, but I'm not sure I'll find everything there...

fridaynext:
@Chaphasilor no worries - let me look through some of my snapshots to see if I can find it (that's the same problem I was having after the bad rebase).

Chaphasilor:
@fridaynext yep, that's all I needed, thanks!  
I've finished up the PR now, have a look :)

Chaphasilor:
@fridaynext thanks for the PR! Everything works now, have fun using it in the next beta release!

lukaslindnermusic:
image

We should use a SafeArea widget, though. 🤔

lukaslindnermusic:
https://github.com/user-attachments/assets/e344964c-c8ff-4754-a7e3-9421a85c4486

Also, I now get this Image-Animation when navigating from an album to an artist. The otherway around, everything is fine.

PR #1154, opened 2025-04-05
Wrong Navigation when Deleting

flloschy:
>  I just noticed in the code that deleting something from the tracks tab will open the parent album instead of refreshing the tab. I think we might want to change that?
Also it seems like there's an error when deleting a track from the tracks tab. The deletion goes through, but the menu doesn't close and (flawed) navigation doesn't happen.
If you could take a look that would be great! Should be a simple fix :)

\- Chaphasilor
___

I believe this is fixed now. Im not a 100% sure if my logic here is sound though.

### DIFF
--- a/lib/components/AlbumScreen/album_screen_content.dart
+++ b/lib/components/AlbumScreen/album_screen_content.dart
@@ -1,3 +1,4 @@
+import 'package:finamp/components/MusicScreen/music_screen_tab_view.dart';
 import 'package:flutter/material.dart';
 import 'package:finamp/l10n/app_localizations.dart';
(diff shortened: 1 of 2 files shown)


Chaphasilor:
That's a clever solution! We should've done it like this from the start ^^  
I think we can just merge this right-away, I'm guessing you tested if it actually works (i.e. refreshes the album contents after delete)

flloschy:
I did test but I dont know If i tested all cases because the navigation code that was there before never should exist in the first place but there must have been a reason

Chaphasilor:
The navigation code was my attempt at re-loading the album after deleting a track form the album ^^

PR #1159, opened 2025-04-06
Redesign Remaining Menus

Chaphasilor:
This PR will make the menus for all remaining item types consistent with the track menu.  
The design is the same, and the options are as similar as possible.  

I'm also planning to add a proper "Play" option to the menu, as this came up a few times now.

TODOs:

- [x] Port all menus
- [x] Add options for play/shuffle
- [x] Make sure all old menu options are still available (and possibly new ones)
- [x] Figure out a way to improve code sharing
  - This could work well by just having an enum for the different options, and then having a custom list consisting of these enums for each type of menu, along with custom handlers for the options (maybe combined into one as a map?). Just the player screen menu might be a bit more complicated due to the playback actions
- [x] Customize ItemInfoHeader for different menu types
- [x] Change swipe animation speed
- [x] Replace strings with translation tokens


Komodo5197:
My thoughts on the potential for code sharing - I'd just turn all the options into widgets.  So we have a PlayNextMenuOption(BaseItemDto), and an AddToQueueMenuOption(BaseItemDto), and a DownloadMenuOption(BaseItemDto) and so on.  You'll want a helper method that takes a baseItemDto and spits out the list of child songs from either the server or downloads as appropriate.  Also, the buildWrapper option for the themed bottom sheets is mostly needed due to the complications of the track menu.  I think most of the others could use the buildSlivers option where you just hand it a sliver list.

Chaphasilor:
Missed your comments, got lost in notifications. I'd never have considered using widgets for menu items, but it makes a lot of sense for sure. It's also a great way to keep the functionality/logic tied the option, and allows easily extending the functionality for specific options (showing favorite state, handling long presses, etc.).  
So I'll probably do just that.

For the helper method, I'll look into it. But I'm not sure we need the same list of tracks in all cases, so the logic might become more complicated.

I'll also take a look at replacing the buildWrapper. I'm guessing you'll let me know if I break something there...

Chaphasilor:
I also think I understand your comment about the helper function now. We add one function per menu entry widget, and that function handles the various `BaseItemDto` types?  
I'll try to do that, but maybe handling the cases right within the onTap function works even better.

Komodo5197:
I was imagining a single global helper function.  All the play options should be doing the same thing.  So the play, add to queue, and add to next up options for, say, an artist always fetch the same set of tracks using the same logic.  They just add it to the queue differently.   I can't think of a reason we would want a different set of tracks for the same item - that would be really confusing and unexpected for the user?  Even the play/shuffle options on the artist page itself should want the same set of tracks,  I'm not sure those could be easily or usefully unified.  But I was imagining a single function which switch between multiple sets of logic depending on the given BaseItemDto type.  And then all the menu option widgets just use that function and feed the returned tracks into the correct queue calls in their tap handler.  At that point, the difference between the same menu entry, such as play next, across the different item types is hopefully fairly minimal?

Chaphasilor:
Yeah I think it could work. Although essentially this is was `JellyfinApiHelper.getItems()` is supposed to be, isn't it? But maybe this helper function could be slightly less abstract, and pre-apply some filters based on the current settings (i.e. global favorite filter and such).

Chaphasilor:
@Komodo5197 I've moved a good chunk of the menu into reusable widget that can be composed into the different menus.  
I'm guessing I could squeeze out a bit more code sharing, but I don't know if that makes a lot of sense? We could have a generic `showModalMenu()` function that accepts the content from all specific instances of the menu. It would make things less flexible, but maybe that's what we want to improve maintainability? Would appreciate your thoughts on the matter :)

Oh and I made the item request helper method (`loadChildTracks` in `item_helper.dart`) a provider. Doesn't add much complexity, and I think network requests are really what providers are made for? Is there any downside to doing it like this?

Komodo5197:
I don't think a showModalMenu function is needed, it feels like the code reuse balance is in a pretty good place as-is.  Although I'm questioning why either argument to PlaybackActionRow exists.  Why would we ever want to show a different set of options?  That would also open up more avenues to make the section more adaptive and/or customizable.  The other code reuse thought is to just see how much of the track menu can be built out of the new widgets.  And it might make more sense to pack all the download menu options into one widget that returns the correct state rather than having to always add all three to the menu.

For loadChildTracks, I don't think putting it in a provider is really that great in this scenario.  providers are great whenever you need to make a network request while building a widget, but they aren't really helpful when making a network request from a tap handler.  All they do is add the possibility of accidentally not waiting for the future to complete when you need its value.  I'd make it a regular function unless you have a specific goal in mind that needs the caching or rebuilding.

Chaphasilor:
We will actually already have a different set of options for the artist menu, there currently is a "shuffle artist" (all tracks shuffled" and "shuffle artist albums" (albums shuffled, but tracks in album order). That's why I made it flexible.  
Bundling the download options is a good idea.

The provider was mostly meant to be consistent with the home screen PR (where I responded to your comment btw), since I'm also using a provider there. But there the caching is more important too, so yeah, maybe the provider is overkill.

Komodo5197:
If we already have variations, that does make more sense, although I'd still lean towards having that handled internally because its just based off the item type in the same way the child track fetch is.  But I think part of that is because I don't actually like the precise layout of the actions, so I'm already thinking of customizations.  I can see the logic of play and shuffle being the two categories, but the most used buttons are probably play and shuffle, and this puts them on seperate pages so even if we save the position there's a good chance you need to switch.  It might be nice to have all the next up related items on one page, and the other options together on the other.  Or maybe it could even be customizable like the filter chips or music screen tabs.  There's also the idea of just showing both rows at once on larger devices.  But this might all be me overthinking, I always just use the main buttons on the album screen anyway, and it's not like a swipe or two is that bad.

lukaslindnermusic:
@Chaphasilor So, I just experimented with the childCount for Artists (regarding the displayal of the albumCount as subtitle on the artist list), as you've mentioned that you are also planning on integrating it in the new menus.

Unfortunately, I got a lot of album-artists that displayed my "?? 0" fallback of "0 Albums", which shouldn't be possible because they definitely have albums - I checked manually - and otherwise they wouldn't even show up in the list in the first place. 

I think I found the answer for that oddity. The problem I discovered: 
For an album, Jellyfin only increases the childCount for ONE album artist. If you have an album that has two or more albumartists assigned, the childCount for only one of those will be increased, resulting in incorrect data for the others. 🤡

Unfortunately, it seems that there is also no valid count for trackChilds (which I planned to display as subtitle when "Performing Artists" is selected, and as getting both values accurately would require additional queries at the moment, I vote that we remove the albumCount for now (since it displays incorrect data) but leave the functionality in the code, commented out. 
We then riot on the Jellyfin Server issue board and as soon as they fixed this server-side, we can enable it on our side again.

lib/components/AlbumScreen/genre_menu.dart
```diff
@@ -0,0 +1,674 @@
...
+              if (widget.isOffline) {
+                genreTracks = await downloadsService
+                    .getCollectionTracks(widget.item, playable: true);
+              } else {
+                genreTracks = await _jellyfinApiHelper.getItems(
+                  parentItem: widget.item,
+                  sortBy: "ParentIndexNumber,IndexNumber,SortName",
+                  includeItemTypes: "Audio",
+                );
+              }
```
lukaslindnermusic: This will include items from other libraries.
You will have to pass the current library as parent and the genre as "genreFilter".
And for offline you should use getAllTracks instead, and pass the library as "viewfilter" and the genre as genreFilter as well.

Or, you can just use the genre_provider to get all the tracks. 😊

Chaphasilor:
Haha yes I also had similar problems with those counts. Good catch we with the album artist counts, I guess that really is something they should be able to fix on the server :)  
I guess we should wait until the next release though, since they are already behind schedule and probably super busy right now.

lib/services/item_helper.dart
```diff
@@ -0,0 +1,113 @@
...
+      case BaseItemDtoType.genre:
+        newItemsFuture = jellyfinApiHelper.getItems(
+          parentItem: baseItem,
+          includeItemTypes: [
+            BaseItemDtoType.track.idString,
+          ].join(","),
+          sortBy: sortBy?.jellyfinName(null) ?? SortBy.album.jellyfinName(null),
+          sortOrder: sortOrder?.toString(),
+          // filters: settings.onlyShowFavourites ? "IsFavorite" : null,
+        );
```
lukaslindnermusic, line 60: This is wrong. It includes items from all libraries, not just the selected one.
Additionally, the artist one will not return the "Appears On" tracks.

For genres, it should be:
```
newItemsFuture = jellyfinApiHelper.getItems(
        parentItem: _finampUserHelper.currentUser?.currentView,
       includeItemTypes: [
            BaseItemDtoType.track.idString,
          ].join(","),
          sortBy: sortBy?.jellyfinName(null) ?? SortBy.album.jellyfinName(null),
          sortOrder: sortOrder?.toString(),
        genreFilter: baseItem,
      );
 ```
      
And for artists, there has to be some filtering logic. I think it would be best to just use the allTracksProvider from `artist_screen_provider.dart`:
```
newItemsFuture = ref.watch(
        getAllTracksProvider(baseItem, _finampUserHelper.currentUser?.currentView).future;
```

File: lib/services/item_helper.dart [hunk 3]
```diff
@@ -0,0 +1,113 @@
...
+  switch (BaseItemDtoType.fromItem(baseItem)) {
+    default:
+      items = await downloadsService.getCollectionTracks(
+        baseItem,
+        playable: true,
+      );
+      break;
+  }
+
+  return (limit != null ? items.take(limit) : items).toList();
```
lukaslindnermusic, line 112: Same here. 
For artists, you can use the exact same provider again and it will automatically return the offline tracks.
For genres, use:
```
items = await downloadsService.getAllTracks(
          viewFilter: _finampUserHelper.currentUser?.currentView,
          nullableViewFilters: settings.showDownloadsWithUnknownLibrary,
          genreFilter: baseItem,
          playable: true,
      );
```

lukaslindnermusic (commented):
The actions to fetch artist and genre tracks currently are not yet correct.
I think they would include items from all libraries, not just the selected one. 
And artists won't include the "Appears On" tracks, only "album artist"-tracks.
I've left some comments how I think it should work 😊

lib/services/item_helper.dart
lukaslindnermusic: or probably ref.read, as we only care about getting those items once?

lukaslindnermusic: and probably, since we are using this provider in different places now, it would be good to rename it to `getAllArtistTracks`?

lib/menus/components/menuEntries/clear_queue_menu_entry.dart
```diff
@@ -0,0 +1,36 @@
...
+  const ClearQueueMenuEntry({
+    super.key,
+    required this.baseItem,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final queueService = GetIt.instance<QueueService>();
+
+    final canGoToGenre = (baseItem.genreItems?.isNotEmpty ?? false);
```
Komodo5197, line 22: I think you messed up when copying.

lib/menus/components/menuEntries/go_to_album_menu_entry.dart
```diff
@@ -0,0 +1,61 @@
...
+  const GoToAlbumMenuEntry({
+    super.key,
+    required this.baseItem,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final jellyfinApiHelper = GetIt.instance<JellyfinApiHelper>();
+
+    final canGoToAlbum = baseItem.parentId != null;
```
Komodo5197, line 26: Shouldn't this be albumId?  That's what the rest of the widget uses.

lib/menus/track_menu.dart
```diff
@@ -765,9 +405,16 @@ class _TrackMenuState extends ConsumerState<TrackMenu> {
       MenuMask(
         height: playActionRowHeight + 8.0,
```
Komodo5197, line 385: This should be the same as the other mask, 135.

lib/menus/components/playbackActions/playback_action.dart
```diff
@@ -0,0 +1,65 @@
...
+          ],
+        ),
+        onPressed: () {
+          FeedbackHelper.feedback(FeedbackType.selection);
+          onPressed(ref);
+        },
+        visualDensity: VisualDensity.compact,
+        padding: const EdgeInsets.only(
+            top: 10.0, left: 8.0, right: 8.0, bottom: 12.0),
+        tooltip: tooltip,
```
Komodo5197, line 61: Why do we have the text visible and as a tooltip?

lib/menus/components/menuEntries/go_to_album_menu_entry.dart
Chaphasilor: Could be. The original track menu used the parentId check, so I simply moved that check here.  
I haven't properly tested things yet, so I'm not even sure if everything still works.

lib/menus/components/playbackActions/playback_action.dart
Chaphasilor: I'll have to check on my laptop, but I think there was an instance where the text and tooltip were different at some point, and this is just what remains of that. I'll confirm that TalkBack treats it equally if I omit the tooltip.

lib/menus/track_menu.dart
Komodo5197: It specifically masks out any portion of its contents that gets withing X pixels of the top of the scrollview.  We need to do this because the iteminfoheader is transparent and allows you to see the background, so when the other menu items are slid behind the header as you scroll you can still see them.  This can't be done in the header because it doesn't have control of how the other items in the list are rendering, or at least I don't think it does.  It feels overly specific to put into the themedbottom sheet.  We could put all the other items into a single instance of menumask, but it should work fine with multiple, and I'm not sure if either way any performance implication.  Either way is probably fine.

Chaphasilor: Okay. When I tried to wrap the StreamBuilder + row into a MenuMask it was completely invisible as long as I had it at the top. Now even with the mask it works fine because the stuff above the header isn't sticky.  
But I'll change the height to match the other mask!

lib/menus/components/menuEntries/go_to_genre_menu_entry.dart
```diff
@@ -0,0 +1,58 @@
...
+                  .getItemById(baseItem.genreItems!.first.id);
+            }
+          } catch (e) {
+            GlobalSnackbar.error(e);
+            return;
+          }
+          if (context.mounted) {
+            Navigator.pop(context);
+            await Navigator.of(context)
+                .pushNamed(ArtistScreen.routeName, arguments: genre);
```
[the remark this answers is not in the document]
Chaphasilor: A sub-menu could work, that's what Spotify does. Although I would prefer to include the different genres (or artists) directly in the menu list, this can cause issues with bad metadata. I have tracks where Jellyfin assigned a ton of random genres, and the menu in Symfonium takes ages to scroll, so I'd like to avoid that.  
We also should have the chips in the ItemInfoHeader, I think we could include genres there as well. There it isn't an issue since we scroll horizontally...

Chaphasilor:
I believe this is done now :)  
Though I'm pretty sure I've missed *something*...

lib/menus/components/menuEntries/adaptive_download_lock_delete_menu_entry.dart
```diff
@@ -0,0 +1,46 @@
...
+  const AdaptiveDownloadLockDeleteMenuEntry({
+    super.key,
+    required this.baseItem,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final downloadsService = GetIt.instance<DownloadsService>();
+
+    final downloadStatus = downloadsService.getStatus(
```
Komodo5197: We should probably use the statusProvider to make this reactive.

lib/menus/album_menu.dart
```diff
@@ -0,0 +1,90 @@
...
+      DeleteFromServerMenuEntry(baseItem: baseItem),
+    ];
+  }
+
+  (double, List<Widget>) getMenuProperties(BuildContext context) {
+    var stackHeight = infoHeaderFullHeight + playActionRowHeight;
+    var menuEntries = getMenuEntries(context);
+    stackHeight += menuEntries
+            .where((element) =>
+                switch (element) { Visibility e => e.visible, _ => true })
```
Komodo5197: There seem to be a number of new and outstanding issues with how this stackheight is calculated.
- Switching on widget type no longer works because the menu widgets aren't built yet.  We probably need to add an isVisible method to them all or something.  We also hide some of them via Sizedbox.shrink instead of visibility, so that needs to be accounted for.
- The per-entry height is only 48 on desktop, where visualdensity is compact, instead of the 56 from mobile.
- infoHeaderFullHeigh is the 120 of the internal height but should be the 150 of the used extent.
- playActionRowHeight is missing the height of the page indicator.
- ThemedBottomSheet should probably be increasing the stackheight by MediaQuery.paddingof(context).bottom before using it to account for the padding it adds to get around system ui elements.

These issues apply to all the different copies of this calculation.  Maybe we should be centralizing it into a helper method?

lib/menus/components/menuEntries/adaptive_download_lock_delete_menu_entry.dart
```diff
@@ -0,0 +1,46 @@
...
+        null);
+
+    if (!ref.watch(finampSettingsProvider.isOffline) &&
+        downloadStatus == DownloadItemStatus.notNeeded) {
+      return DownloadMenuEntry(baseItem: baseItem);
+    } else if (downloadStatus.isRequired) {
+      return DeleteFromDeviceMenuEntry(baseItem: baseItem);
+    } else if (!ref.watch(finampSettingsProvider.isOffline) &&
+        downloadStatus.isIncidental) {
+      return DownloadMenuEntry(baseItem: baseItem);
```
Komodo5197: This needs to be using the lock download entry.

lib/menus/components/menuEntries/delete_from_device_menu_entry.dart
```diff
@@ -0,0 +1,48 @@
...
+  const DeleteFromDeviceMenuEntry({
+    super.key,
+    required this.baseItem,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final downloadsService = GetIt.instance<DownloadsService>();
+
+    final downloadStatus = downloadsService.getStatus(
```
Komodo5197: This should use the status provider, or maybe skip the check and assume the adaptive provider has checked the status correctly.

lib/menus/components/menuEntries/download_menu_entry.dart
```diff
@@ -0,0 +1,52 @@
...
+  const DownloadMenuEntry({
+    super.key,
+    required this.baseItem,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final downloadsService = GetIt.instance<DownloadsService>();
+
+    final downloadStatus = downloadsService.getStatus(
```
Komodo5197: This should use the status provider, or maybe skip the check and assume the adaptive provider has checked the status correctly.

lib/menus/components/menuEntries/adaptive_download_lock_delete_menu_entry.dart
```diff
@@ -0,0 +1,46 @@
...
+    required this.baseItem,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final downloadsService = GetIt.instance<DownloadsService>();
+
+    final downloadStatus = downloadsService.getStatus(
+        DownloadStub.fromItem(
+            type: BaseItemDtoType.fromItem(baseItem) == BaseItemDtoType.track
```
Komodo5197: This logic will use the all library downloads of artists/genres instead of the per-library collections.  You may want to switch this logic to use those, and change the child download widgets to take downloadStubs instead of baseItems to consolidate the logic.

lib/menus/artist_menu.dart
```diff
@@ -0,0 +1,88 @@
...
+}) async {
+  // Normal menu entries, excluding headers
+  List<Widget> getMenuEntries(BuildContext context) {
+    return [
+      AddToPlaylistMenuEntry(baseItem: baseItem),
+      InstantMixMenuEntry(baseItem: baseItem),
+      AdaptiveDownloadLockDeleteMenuEntry(baseItem: baseItem),
+      ToggleFavoriteMenuEntry(baseItem: baseItem),
+      GoToGenreMenuEntry(baseItem: baseItem),
+      DeleteFromServerMenuEntry(baseItem: baseItem),
```
Komodo5197: Not much reason to have this entry in the list when artists can't be deleted from the server.

lib/menus/genre_menu.dart
```diff
@@ -0,0 +1,85 @@
...
+  required BaseItemDto baseItem,
+}) async {
+  // Normal menu entries, excluding headers
+  List<Widget> getMenuEntries(BuildContext context) {
+    return [
+      AddToPlaylistMenuEntry(baseItem: baseItem),
+      InstantMixMenuEntry(baseItem: baseItem),
+      AdaptiveDownloadLockDeleteMenuEntry(baseItem: baseItem),
+      ToggleFavoriteMenuEntry(baseItem: baseItem),
+      DeleteFromServerMenuEntry(baseItem: baseItem),
```
Komodo5197: Not much reason to have this entry in the list when genres can't be deleted from the server.

lib/menus/components/menuEntries/delete_from_server_menu_entry.dart
```diff
@@ -0,0 +1,48 @@
...
+        ref.watch(jellyfinApiHelper.canDeleteFromServerProvider(baseItem));
+
+    return Visibility(
+      visible: canDelete,
+      child: MenuEntry(
+        icon: TablerIcons.trash_x,
+        title: AppLocalizations.of(context)!
+            .deleteFromTargetConfirmButton("server"),
+        onTap: () async {
+          var item = DownloadStub.fromItem(
```
Komodo5197, line 37: This needs to be updated to handle tracks.

lib/menus/components/menuEntries/lock_download_menu_entry.dart
```diff
@@ -0,0 +1,69 @@
...
+  const LockDownloadMenuEntry({
+    super.key,
+    required this.baseItem,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final downloadsService = GetIt.instance<DownloadsService>();
+
+    final downloadStatus = downloadsService.getStatus(
```
Komodo5197: This should use the status provider, or maybe skip the check and assume the adaptive provider has checked the status correctly.

lib/menus/components/menuEntries/remove_from_current_playlist_menu_entry.dart
```diff
@@ -0,0 +1,47 @@
...
+    this.confirmRemoval = false,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return Visibility(
+        visible:
+            parentItem != null && !ref.watch(finampSettingsProvider.isOffline),
+        child: MenuEntry(
+          icon: TablerIcons.playlist_add,
```
Komodo5197: This icon seems incorrect.

lib/menus/components/menu_item_info_header.dart
```diff
@@ -0,0 +1,449 @@
...
+    return switch (BaseItemDtoType.fromItem(item)) {
+      BaseItemDtoType.track => TrackInfo(
+          item: item,
+        ),
+      BaseItemDtoType.album => AlbumInfo(
+          //TODO don't show parent album
+          item: item,
+        ),
+      BaseItemDtoType.playlist => PlaylistInfo(
+          //TODO show track count instead of artist+album info
```
Komodo5197: I believe these TODOs have been completed.

lib/menus/components/menu_item_info_header.dart
```diff
@@ -0,0 +1,449 @@
...
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return ItemInfo(
+      item: item,
+      condensed: condensed,
+      shape: RoundedRectangleBorder(
+        borderRadius: BorderRadius.only(
+            topLeft: Radius.circular(9999),
+            bottomLeft: Radius.circular(9999),
+            // circular radii influence each other. this should look roughly the same as BorderRadius.circular(12)
```
Komodo5197: Is there not a way to do this without arbitrary large numbers?

lib/menus/components/menu_item_info_header.dart
```diff
@@ -0,0 +1,449 @@
...
+          ),
+          overflow: TextOverflow.ellipsis,
+          softWrap: true,
+          maxLines: 2,
+        ),
+        const SizedBox(height: 4),
+        if (!condensed)
+          ItemAmountChip(
+            baseItem:
+                item, //FIXME fetch extended item when menu is opened to include BaseItemDto.albumCount
```
Komodo5197: This is showing 0 for all artists still.

lib/menus/components/menu_item_info_header.dart
```diff
@@ -0,0 +1,449 @@
...
+          clipBehavior: Clip.antiAlias,
+          decoration: ShapeDecoration(
+            color: Theme.of(context).brightness == Brightness.dark
+                ? Colors.black.withOpacity(0.25)
+                : Colors.white.withOpacity(0.15),
+            shape: widget.shape,
+          ),
+          child: Stack(
+            children: [
+              Positioned(
```
Komodo5197: This is being shown on track menus where it has no functionality.

lib/menus/components/playbackActions/playback_action.dart
```diff
@@ -0,0 +1,64 @@
...
+                  style: const TextStyle(
+                    fontSize: 12,
+                    height: 1.4,
+                    fontWeight: FontWeight.w300,
+                  ),
+                ),
+              ),
+            ),
+          ],
+        ),
```
Komodo5197: This makes the button elements awkwardly cluster upwards for me.
```suggestion
              icon: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          crossAxisAlignment: CrossAxisAlignment.stretch,
          spacing: 9,
          children: [
            Icon(
              icon,
              color: iconColor,
              size: 35,
              weight: 1.0,
            ),
            Text(
              label,
              textAlign: TextAlign.center,
              maxLines: 2,
              overflow: TextOverflow.ellipsis,
              style: const TextStyle(
                fontSize: 12,
                height: 1.4,
                fontWeight: FontWeight.w300,
              ),
            ),
          ],
        )
```

lib/menus/components/playbackActions/playback_action_page_indicator.dart
```diff
@@ -0,0 +1,60 @@
...
+                    duration: Duration(milliseconds: 750),
+                    curve: Curves.easeOutCubic,
+                  );
+                },
+                child: Container(
+                  margin: const EdgeInsets.symmetric(horizontal: 4.0),
+                  padding: const EdgeInsets.symmetric(
+                      horizontal: 8.0, vertical: 4.0),
+                  decoration: BoxDecoration(
+                    color: pageController.page?.round() == index
```
Komodo5197: pageController.page is still null during the initial build before the theme calculations trigger a rebuild.
```suggestion
                    color: (pageController.page ?? pageController.initialPage).round() == index
```

lib/menus/components/playbackActions/playback_action.dart
```diff
@@ -0,0 +1,64 @@
...
+    required this.icon,
+    this.value,
+    required this.onPressed,
+    required this.label,
+    required this.iconColor,
+  });
+
+  final IconData icon;
+  final String? value;
+  final void Function(WidgetRef ref) onPressed;
```
Komodo5197: Don't give the tap handler the WidgetRef.  There isn't anything useful they should be doing with it.

lib/services/item_helper.dart
```diff
@@ -0,0 +1,146 @@
...
+import 'package:finamp/services/finamp_user_helper.dart';
+import 'package:finamp/services/jellyfin_api_helper.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:get_it/get_it.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'item_helper.g.dart';
+
+@riverpod
+Future<List<BaseItemDto>?> loadChildTracks(
```
Komodo5197: I don't think this makes much sense as a provider when we only use the future and not the asyncValue.  These should just be regular methods.

lib/menus/album_menu.dart
```diff
@@ -0,0 +1,90 @@
...
+            .length *
+        56;
+
+    final pageViewController = PageController();
```
Komodo5197, line 44: It might make sense to remember the last position of this slider in hive instead of always starting on play.

lib/components/MusicScreen/item_collection_wrapper.dart
```diff
@@ -173,10 +150,7 @@ class _ItemCollectionWrapperState extends ConsumerState<ItemCollectionWrapper> {
                 item: widget.item
```
Komodo5197: We should remove all these unused variables.

lib/components/MusicScreen/item_collection_wrapper.dart
```diff
@@ -190,614 +164,29 @@ class _ItemCollectionWrapperState extends ConsumerState<ItemCollectionWrapper> {
...
           : EdgeInsets.zero,
       child: GestureDetector(
-        onLongPressStart: (details) => menuCallback(
-            localPosition: details.localPosition,
-            globalPosition: details.globalPosition),
-        onSecondaryTapDown: (details) => menuCallback(
-            localPosition: details.localPosition,
-            globalPosition: details.globalPosition),
+        onLongPressStart: (details) => menuCallback(),
+        onSecondaryTapDown: (details) => menuCallback(),
```
Komodo5197, line 159: You should add an onTapDown handler that begins theme precalculation like the TrackListItem has here.

lib/menus/album_menu.dart
```diff
@@ -0,0 +1,90 @@
...
+
+    List<Widget> menu = [
+      SliverPersistentHeader(
+        delegate: MenuItemInfoHeader(
+          item: baseItem,
+        ),
+        pinned: true,
+      ),
+      MenuMask(
+        height: playActionRowHeight + 8.0,
```
Komodo5197: This is incorrect.  All the menumasks need to have the height of the pinned header bars, which is MenuMask.defaultHeight, or the play actions clip into the item info header when scrolling.

lib/menus/artist_menu.dart
```diff
@@ -0,0 +1,88 @@
...
+
+    List<Widget> menu = [
+      SliverPersistentHeader(
+        delegate: MenuItemInfoHeader(
+          item: baseItem,
+        ),
+        pinned: true,
+      ),
+      MenuMask(
+        height: playActionRowHeight + 8.0,
```
Komodo5197: This is incorrect.  All the menumasks need to have the height of the pinned header bars, which is MenuMask.defaultHeight, or the play actions clip into the item info header when scrolling.

lib/menus/genre_menu.dart
```diff
@@ -0,0 +1,85 @@
...
+
+    List<Widget> menu = [
+      SliverPersistentHeader(
+        delegate: MenuItemInfoHeader(
+          item: baseItem,
+        ),
+        pinned: true,
+      ),
+      MenuMask(
+        height: playActionRowHeight + 8.0,
```
Komodo5197: This is incorrect.  All the menumasks need to have the height of the pinned header bars, which is MenuMask.defaultHeight, or the play actions clip into the item info header when scrolling.

lib/menus/playlist_menu.dart
```diff
@@ -0,0 +1,88 @@
...
+
+    List<Widget> menu = [
+      SliverPersistentHeader(
+        delegate: MenuItemInfoHeader(
+          item: baseItem,
+        ),
+        pinned: true,
+      ),
+      MenuMask(
+        height: playActionRowHeight + 8.0,
```
Komodo5197: This is incorrect.  All the menumasks need to have the height of the pinned header bars, which is MenuMask.defaultHeight, or the play actions clip into the item info header when scrolling.

lib/menus/track_menu.dart
```diff
@@ -0,0 +1,420 @@
...
+    ];
+  }
+
+  // All track menu slivers, including headers
+  List<Widget> menu(BuildContext context, List<Widget> menuEntries,
+      MetadataProvider? metadata) {
+    var iconColor = Theme.of(context).colorScheme.primary;
+
+    return [
+      if (widget.showPlaybackControls)
```
Komodo5197, line 242: I don't feel like these controls look very good up here?

lib/menus/components/menu_item_info_header.dart
```diff
@@ -0,0 +1,449 @@
...
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return ItemInfo(
+      item: item,
+      condensed: condensed,
+      shape: RoundedRectangleBorder(
+        borderRadius: BorderRadius.circular(12),
+      ),
+      featureImage: AlbumImage(
```
Komodo5197, line 399: It looks like the tapToZoom animation doesn't work for any of the menus.  This appears to be a limitation of the Hero widget inside modal routes.  It probably isn't worth trying to fix this.

lib/services/item_helper.dart
```diff
@@ -0,0 +1,146 @@
...
+  Ref ref, {
+  required BaseItemDto baseItem,
+  SortBy? sortBy,
+  SortOrder? sortOrder,
+  String? Function(BaseItemDto)? groupListBy,
```
Komodo5197, line 20: It's generally a bad idea to pass objects without a sensible equals method to providers because it prevents them from ever being reused.  This doesn't matter if you make this a regular function, of course.

lib/services/item_helper.dart
```diff
@@ -0,0 +1,146 @@
...
+      baseItem: baseItem,
+    ).future);
+  } else {
+    switch (BaseItemDtoType.fromItem(baseItem)) {
+      case BaseItemDtoType.track:
+        newItemsFuture = Future.value([baseItem]);
+        break;
+      case BaseItemDtoType.album:
+      case BaseItemDtoType.playlist:
+        newItemsFuture = jellyfinApiHelper.getItems(
```
Komodo5197: We should reuse the providers for the album, playlist, and genre screens to minimize reloads.  Also, it would be more consistent if the playback options respected the sort order when accessed from the overflow menu in the same way that the main play/shuffle buttons do.

lib/menus/components/menu_item_info_header.dart
```diff
@@ -0,0 +1,449 @@
...
+                    child: widget.featureImage,
+                  ),
+                  Expanded(
+                    child: Container(
+                      padding: const EdgeInsets.symmetric(horizontal: 12.0),
+                      child: Column(
+                        mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+                        crossAxisAlignment: CrossAxisAlignment.start,
+                        mainAxisSize: MainAxisSize.min,
+                        children: widget.infoRows,
```
Komodo5197, line 533: The top info row can overlap with the jump to item button if its long enough, which looks a bit awkward.

Chaphasilor:
Thanks for taking a look!

1. Noted, I'll remove that
2. Only edited playlists, didn't touch albums and artists. I can add the genres there. I guess we could get rid of the "go to xyz" buttons then.
3. Hmm. Nothing in the logs?
4. Yeah probably due to the two lines. Maybe I can center it, but it's not really a problem I guess...
5. Probably a copy-paste error. There should be no message for play/shuffle now
6. I could increase the distance, but I want to avoid losing the visual grouping. So I'd have to increase the distance at the top as well?
7. I guess we could apply the limit, or remove the entire section from the menu. I'm fine with either. Do you have preferences?

lib/menus/components/menuEntries/delete_from_device_menu_entry.dart
Chaphasilor: I wanted to keep the individual menu items working, so I'll just change the status stuff ;)

lib/services/item_helper.dart
Chaphasilor: I originally aimed for caching of similar requests, but that's disabled anyway.  
Should I try using the AsyncValues instead? Or is there no potential use for a provider at all?

lib/menus/album_menu.dart
Chaphasilor: Hmm. Per menu type or globally?

If we do it globally we could even apply it to the play buttons in the flexible section as well, so that we only have one play button that can play/shuffle/shuffle albums.

lib/menus/track_menu.dart
Chaphasilor: I'm not super happy with them either. But the position works well with the special semantics of the player screen menu I think, where these actions are related to playback and not the track menu.

Open for alternative suggestions!

lukaslindnermusic:
> 3. Hmm. Nothing in the logs?
I haven't noticed anything, but I will check again tomorrow and let you know if I find something... couldn't reproduce it yet.


> 6. I could increase the distance, but I want to avoid losing the visual grouping. So I'd have to increase the distance at the top as well?

Yeah totally, I get that and I think it makes sense, so that you understand that the "category-buttons" are affecting the actions. So maybe sligthly increase all distances?

> 7. I guess we could apply the limit, or remove the entire section from the menu. I'm fine with either. Do you have preferences?
I think, a limit to Play All does not make sense... Shuffle yes, because you don't care about the order. but If you apply a limit to the normal "Play", it will always play the same tracks and stop at everything afterwards. So I personally would prefer to mimick how we are doing it with the blue Play/Shuffle Buttons, aka. hide or disable it for now for the genre until we can properly handle large queues. 😊

lib/services/item_helper.dart
Komodo5197: I don't think there's much value.  The two big benefits of providers are caching and rebuilds.  We don't benefit from the rebuilds because this isn't used by any widgets, but instead is in the async handlers that just wait until the future completes once.  The caching probably isn't doing much for us here.  We can get some caching/deduplication of requests by consistently using the same providers used by the collection screens, though, but that doesn't require this to be a provider.

lib/menus/album_menu.dart
Komodo5197: I'm not sure.  I've also imagined that allowing for the customization of the flexible buttons to any of these types might be nice, but I pretty much just play/shuffle myself so I don't have direct insight on how useful any of this would really be, or whether its not worth thinking about.

lib/menus/track_menu.dart
Komodo5197: One option would be to put the playback options back into the menu list if the controls are shown, because there's only 3 and I can't imagine they are used that much.  That does seem like a lot of special case widgets though.  It might look better if you put a divider line in between the playback controls and the track info?  Also, do you think it would look better if the playback controls were using the theme color instead of white?

lib/services/item_helper.dart
Chaphasilor: You mean for playlists specifically? Because I'm not sure where else there would be a sorting applied to the entire contents of the screen?

Komodo5197: Yeah, I'm just talking about the new playlist sorting option.

lib/menus/components/playbackActions/playback_action.dart
Chaphasilor: Thanks. Not sure what happened here, but it's probably also why the page indicators were closer that they should've been...

lib/services/item_helper.dart
Chaphasilor: Can I just use the `ProviderScope` to read the `getArtistTracksProvider` then? Or is there anything I need to keep in mind?

lib/components/MusicScreen/item_collection_wrapper.dart
Chaphasilor: huh, that were more than expected. Would be nice if the linter could check dependencies to detect such cases of unused variables referencing other unused variables...

lib/menus/track_menu.dart
Chaphasilor: Now that you mention it, the playback actions (non-player-menu) should be white too, because they get colorful in an active state (e.g. shuffle active, sleep timer active, playback speed modified.  

Moving them into an extra page in the PlaybackActionsRow is indeed an option, even if messy.  
I've tried the divider approach for now, take a look :)

lib/services/item_helper.dart
Komodo5197: Use GetIt.instance<ProviderContainer>().  It's less messy.  We should be able to replace all the existing background services that get the snackbar providercontext with this.

lib/menus/album_menu.dart
Chaphasilor: If we always need to use the same height, should we just get rid of the `height` parameter altogether?

Chaphasilor: Or we could make the parameter optional, since a different height is used in cases like the output or playlist actions menu.  
Alternatively the `MenuItemInfoHeader`, and any other headers used in bottom sheets, could have a default height configured as a static parameter, instead of putting it into the `MenuMask`.  
Or we have multiple heights for all different header types in the `MenuMask`.

lib/menus/components/menu_item_info_header.dart
Chaphasilor: It requires knowing the height, but we do know that already, yes :)

lib/menus/album_menu.dart
Chaphasilor: Let's wait for user feedback on this. If people start complaining that it isn't remembered then we can try to figure out how much to remember.

lib/menus/track_menu.dart
Chaphasilor: The approach with the separate page on the PlaybackActionsRow might also cause issues with potentially remembering the last-used page, since I'd to show these special actions by default.  
Just let me know how you like the divider :)

Komodo5197: I think it looks at least a little better.  You forgot to consider light mode though.

lib/menus/album_menu.dart
Komodo5197: Attaching the mask height variable to the thing that actually has that height does make sense to me.

lib/components/PlayerScreen/queue_list.dart
```diff
@@ -502,7 +502,7 @@ class _NextUpTracksListState extends State<NextUpTracksList> {
   @override
   Widget build(context) {
     return MenuMask(
-      height: 131.0,
+      height: MenuMask.defaultHeight,
```
Komodo5197: Why'd you change this?  The header size in the queue doesn't have anything to do with the item info header size.

lib/menus/album_menu.dart
Chaphasilor: I (believe I) managed to fix all but the first of these issues. Could you please elaborate on the `isVisible` property?

lib/components/PlayerScreen/queue_list.dart
Chaphasilor: Ah true. I wanted to consolidate it because ~~it said 131 here and 132 somewhere else~~ I'm imagining things, but I'll create a separate height variable for it

lib/menus/album_menu.dart
Komodo5197: This code attempts to count the amount of actually visible menu entries, but I don't think it's working.  Even if the favorite entry is hidden because we're in offline mode, as an example, it's height still gets added because it's not a disability widget, it's a togglefavoritemenuentry.  So the menu ends up way to tall.  I think we make all the menu entries implement an interface with an isvisible function or something so that we can use that to determine if they are visible and their height can be counted.

lib/menus/album_menu.dart
Chaphasilor: Okay, I've tried to do that now. Curiously the new heights don't apply without a hot restart, I believe Flutter's hot reload doesn't support updating global variables or static properties. Just something to keep in mind...

lib/menus/album_menu.dart
Chaphasilor: Okay, I get it now. Doesn't the `Visibility` widget expose such an interface already? Or is that hidden by our custom widget?

lukaslindnermusic:
I also wrote to you on discord for the genre album count 😊 that would be nice, too, if you want to add that 😊 
Just to have it as reference here on github as well:

**Genres**

Online Mode:
```
final fetchedItems = await jellyfinApiHelper.getItemsWithTotalRecordCount(
    parentItem: library,
    genreFilter: parent, 
    limit: 1,
    includeItemTypes: "MusicAlbum",
  );

final genreAlbumCount = fetchedItems?.totalRecordCount ?? 0;
```
offline mode:
```
final List<DownloadStub> fetchedItems = await downloadsService.getAllCollections(
          baseTypeFilter: BaseItemDtoType.album,
          fullyDownloaded: ref.watch(finampSettingsProvider.onlyShowFullyDownloaded),
          viewFilter: library?.id,
          nullableViewFilters: ref.watch(finampSettingsProvider.showDownloadsWithUnknownLibrary),
          genreFilter: parent);
  var items = fetchedItems.map((e) => e.baseItem).nonNulls.toList();
  var itemCount = items.length;
```

**Artists**

Online Mode:
```
    final QueryResult_BaseItemDto? artistAlbumsQuery = await jellyfinApiHelper.getItemsWithTotalRecordCount(
        libraryFilter: library,
        parentItem: parent,
        genreFilter: genreFilter,
        includeItemTypes: "MusicAlbum",
        limit: 1,
        artistType: ArtistType.albumartist);
    final artistAlbumsCount = artistAlbumsQuery?.totalRecordCount ?? 0;
```
offline mode:
use the `getArtistAlbums` provider and `.length`.

if we get 0 Albums, the artist is a performing artist only so we could display a track count instead if we want.
For online mode, run the same query again with `includeItemTypes: "Audio"` and `artistType: ArtistType.artist`.
For offline mode, use the `getArtistTracksProvider` and `.length`.

lib/menus/album_menu.dart
Komodo5197: We can't see the inner visibility widgets.  The build() method of the parent is called before the children, so the visibility widgets inside the menu entry widgets haven't been created yet.  That's why we need the menu entries to directly expose their visibility before they get built.

lib/menus/components/menu_item_info_header.dart
```diff
@@ -0,0 +1,549 @@
...
+          GenreIconAndText(
+            parent: item,
+          ),
+        ]
+      ],
+    );
+  }
+}
+
+class AlbumInfo extends ConsumerWidget {
```
Komodo5197, line 195: I'm getting 4px of clipping off the bottom for albums with multi-line titles.

lib/menus/album_menu.dart
Komodo5197: You are correct about the reloading.  That's a bit irritating.  I've noticed that we can make all these variables const, and then reloads work normally.

lib/menus/components/menuEntries/adaptive_download_lock_delete_menu_entry.dart
```diff
@@ -0,0 +1,59 @@
...
+          type: FinampCollectionType.collectionWithLibraryFilter,
+          library: library,
+          item: baseItem,
+        )),
+      _ => DownloadStub.fromItem(
+          type: DownloadItemType.collection, item: baseItem),
+    };
+
+    final DownloadItemStatus? downloadStatus =
+        ref.watch(downloadsService.statusProvider((downloadStub, null))).value;
```
Komodo5197: I think we can make this provider non-async by replacing it with

```
 late final statusProvider = Provider.family
      .autoDispose<DownloadItemStatus, (DownloadStub, int?)>((ref, record) {
    var (stub, childCount) = record;
    assert(stub.type != DownloadItemType.image &&
        stub.type != DownloadItemType.anchor);
    // Refresh on addDownload/removeDownload as well as state change
    ref.watch(_anchorProvider);
    var sub = _isar.downloadItems
        .watchObjectLazy(stub.isarId).listen((_){
          ref.state=getStatus(stub, childCount);
    });
    ref.onDispose(sub.cancel);
    return getStatus(stub, childCount);
  });
```

This will prevent us from ever having to deal with nulls.

lukaslindnermusic:
If you want to, you can also check which artist Type currently is selected and decide based on that if we show Album or TrackCount.

or a combination:
default to first check album and if 0 then track count.
But if normal artists are selected, skip the album count.

lib/menus/components/menuEntries/adaptive_download_lock_delete_menu_entry.dart
Chaphasilor: Hmm, that seems really complicated just to get rid of null values. I don't think the null handling is such a huge issue?  
But if you prefer it like this, I can add the change :)

Komodo5197: I don't think its noticeably more complicated than the existing code for this provider.  And dealing with a bunch of awaits and nulls for something that's doing all of its calculations synchronously seems silly.  I'd prefer to replace it.

Chaphasilor: I initially thought you wanted to add the code in `adaptive_download_lock_delete_menu_entry.dart` - that's what provoked my original comment. The changes to the actual provider implementation is indeed very minor, and improves the usage globally. So thanks for the suggestion!

lib/components/DownloadsScreen/downloaded_items_list.dart
```diff
@@ -178,20 +178,9 @@ class _DownloadedChildrenListState
...
-                if (snapshot.hasData && snapshot.data!) {
-                  return IconButton(
-                    icon: const Icon(Icons.delete),
-                    onPressed: () =>
-                        askBeforeDeleteDownloadFromDevice(context, stub),
-                  );
-                }
-                return const SizedBox.shrink();
-              },
+            trailing: IconButton(
```
Komodo5197: You need an if(_downloadsService.statusProvider((stub, null)).isRequired) to hide this on incidental downloads.

lib/components/PlayerScreen/item_amount_chip.dart
```diff
@@ -1,69 +1,167 @@
...
 
-    final itemCount = switch (BaseItemDtoType.fromItem(baseItem!)) {
-      BaseItemDtoType.album => baseItem?.childCount ?? 0,
-      BaseItemDtoType.artist => baseItem?.albumCount ??
-          0, //FIXME this doesn't necessarily reflect the actual number of albums, for that a separate API request would be needed. But that should be handled by a provider down the road (or, even better, fixed server-side)
-      BaseItemDtoType.genre => baseItem?.albumCount ??
-          0, //FIXME this is not requested by default, as it takes a lot of time to load
-      BaseItemDtoType.playlist => baseItem?.childCount ?? 0,
-      _ => baseItem?.childCount ?? 0,
+    final Future<int> itemCountFuture = switch (itemType) {
```
Komodo5197: Do not make network requests in build.  You need to create this future in initState.

Chaphasilor: Hmm, like I did in 02d70415bf8bb65388d31f1f0afb142137ba5390? That way we lose all reactivity I guess?  
Even if I manually `.listen()` on the relevant providers, `initState()` will only be called once 🤔

Komodo5197: Yeah, that should work.  Wouldn't listen on the relevant providers get you just as much reactivity?  The code would be a mess, though.  If you want the future to rebuild if offline mode changes or whatnot, the easiest method is probably to just convert it into a provider.  If you do that, put the watch of the artist type inside the switch block so that non-artists don't end up watching that setting for no reason.

Chaphasilor: I did that now in 0b92229586fa043ce3aa32bb868cffbb502f8279, and it seems to be working. There is a warning about nested providers (`provider_dependencies`), but I'm not sure if that's an actual issue here.

Komodo5197: The provider usage seems good.  I think it dislikes how you're dynamically choosing between the artist tracks and albums provider inside ref.watch, so the linter can't figure out which is being used.  I split that apart into two watch statements and it seems happy.  I also ended up refactoring it to use more awaits instead of all the then() blocks, and to limit the watch() of the defaultArtist setting to only artist types because that bugged me, even though the amount of extra rebuilds that setting would cause is probably ~0.  hopefully you think this looks a little more readable?  I'm noticing that the services folder is really starting to fill up with all these providers.  We should probably move them to their own folder at some point.

Chaphasilor: Thanks a ton for the improvements. The await stuff makes absolute sense. If the tighter watches make you happy that's absolutely fine with me.  

Regarding the service directory, yes, I'm planning some house keeping where all files are re-grouped into new directories. The components are all over the place; the screen-related directories essentially lost their entire purpose.  
I've started that here by moving all menus together, but there's much more to do.  
I'd be happy to discuss options for organizing the files on Discord with some of the other contributors :)

PR #1161, opened 2025-04-07
Added Loading Indicator to ArtistScreen + Performance Improvements

lukaslindnermusic:
Just a little fix for the ArtistScreen that I missed in my previous PR.
The ArtistScreen showed the "emptyFilteredListTitle" even if the info was not yet loaded, wich gives a wrong impression.
I added a Loading Progress Indicator that gets shown until the futures completed for the first time.

### DIFF
--- a/lib/components/ArtistScreen/artist_screen_content.dart
+++ b/lib/components/ArtistScreen/artist_screen_content.dart
@@ -37,6 +37,7 @@ class _ArtistScreenContentState extends State<ArtistScreenContent> {
   bool _showTopTracks = true;
   bool _showAlbums = true;
   bool _showAppearsOnAlbums = true;
+  bool _isLoading = true;
 
 
   StreamSubscription<void>? _refreshStream;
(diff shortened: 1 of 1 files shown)


lukaslindnermusic:
I found a potential performance bottleneck that I want to fix, please wait before you merge this 😊

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -209,6 +208,7 @@ class _ArtistScreenContentState extends State<ArtistScreenContent> {
     return FutureBuilder(
         future: futures,
         builder: (context, snapshot) {
+          _isLoading = (_isLoading) ? (snapshot.connectionState == ConnectionState.waiting) : false;
```
Chaphasilor, line 211: I don't really understand this line. Why not just use the connectionState?

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -248,143 +248,175 @@ class _ArtistScreenContentState extends State<ArtistScreenContent> {
...
+            showPlayCount: true,
+            isOnArtistScreen: true,
+            parent: widget.parent,
+          ),
+      ],
+    );
+  }
+}
+
+class BuildAlbumSection extends StatefulWidget {
```
Chaphasilor, line 363: Widgets shouldn't be called "Build...". You use "build..." when it's a function returning a widget :)

Chaphasilor (commented):
Definitely better than before! Left a few minor comments you should address before merging

Chaphasilor:
Awesome, thanks for the fixup! :D

Issue #981, opened 2024-12-21
Suggestion: Turn on "Require WiFi when downloading" by default


Whoever4976:
I was recently quite annoyed when the app used up my entire mobile data quota for the month in a singular day while attempting to finish a 44 GB download in the background which i had started a couple of hours earlier while still inside my home WiFi-network.
I know it is not a written rule to design mobile apps with safeguards against excessive data usage but i do feel like this is something that most users expect from any app they install on their phone. I only noticed the relevant setting afterwards which, by default, allows for downloads of any(!) size using mobile data.   

I know that this is also a user error on my part because i haven't checked the relevant setting to be sure, but quite frankly, it just didn't cross my mind that any modern app would allow for such behaviour and i don't think that i am in the minority with this assumption.

Luckily i was using a Prepaid SIM card and the cost was limited to a few euros and my quota just ran out. But if i had been using another contract, downloading 44GB on mobile data could have been quite an expensive mistake which could have been avoided if i checked the setting beforehand **or** if the app had, in my opinion, better defaults

So i suggest that, at the very least, the setting "Require WiFi when downloading" is turned on by default. An additional feature could be a prompt which appears when a download is pending but no wifi is connected. Something like "No Wifi connected. Do you want to use mobile data to continue the download anyway?"

Chaphasilor:
I agree, will update the default! Sorry about the rogue download :(  
For now, there won't be any notification or size limit, since that's a bit more complex to implement...

flloschy:
Uhh yeah we talked about it but I dont think I implemented it for a reason I cant recall... Let me check

flloschy:
Im unsure on how to trigger the notification though, preferably id like to get an update from the downloader package when its paused but that doesnt seam to be get-able *(does the downloader stop and resume? I tried to understand the code but there isnt really a pause&resume I could find in relation to wifi state inside the package?)*
Alternatively I could add another listener to the current Connectivity listener and check if the download queue is empty or not but that doesnt feel quite right

flloschy:
I ended up going the second route: https://github.com/jmshrv/finamp/pull/1162/commits/cbe804e6859030618097f1c6015d0c2cdc41915b

Issue #620, opened 2024-03-03
[Feature Request] Automatically Switch to Offline Mode


lukaslindnermusic:
Hey there 😊 

Currently, when I am not within my Home Network, my VPN is disabled and I am opening the app while Offline Mode is disabled (or when I disable it directly within the app), I get a few of these (sorry, but) ugly Error messages:
![IMG_4565 Kopie](https://github.com/jmshrv/finamp/assets/8149031/a8688dfd-5f2e-4251-9921-ebf95f40c8cb)

Would it be possible to add a connection progress message instead, directly with the option to switch into Offline Mode right away?
Something like this maybe?
![IMG_4565_mockup](https://github.com/jmshrv/finamp/assets/8149031/805b78cf-db42-4deb-a6ba-02e63bfac661)

Or "We could not find your Jellyfin server. Please make sure...."

What do you think? 😊

lukaslindnermusic:
Not my personal use-case, but if you plan to add functionality to the new Redesigned Version for remembering different servers, you could also add a button to try a different connection.

lukaslindnermusic:
Or maybe something like that:
![IMG_4565_2](https://github.com/jmshrv/finamp/assets/8149031/b46d76c5-f30d-4aaa-86ee-cbb0a9a4c269)

For me, I use Tailscale VPN to connect to my NAS which runs Jellyfin. I set up the server connection with the machine name instead of an ip address, because Tailscale has this amazing feature called Magic DNS which automatically links the machine name to the VPN ip address instead of the home network ip address when Tailscale is connected.
Therefore, when I'm out of home, I just have to connect to Tailscale (you even can automate this with Apple Shortcuts) and everything is working. However, I mostly use Offline Mode to avoid latency that depends on my current network connection.

Sometimes, I forget to enable the Tailscale VPN connection before I disable Offline Mode and then I get bombarded with these error messages 😅Therefore, I think such a message like in my photoshop mockup would be a nice-to-have improvement.

Chaphasilor:
Thanks for the suggestion! Yeah, I also though about some dialog that offers to enable offline mode in the past, but I haven't gotten around to it yet.  
Multi-server support is planned (although I'm not sure how much work that will be with regards to the new download system), so the "switch server" option is a nice idea!  

[...]

PR #1162, opened 2025-04-07
Prefer Local Address when in Home Network 

flloschy:
This creates a new Settings Tab "Network", moved the Auto Offline Setting to there. Created new Settings

1. enabled/disabled this feature
2. change the Address of the Server inside the Network

Also added a Snackbar message when downloads are active without wifi (#981) and tweaked the auto offline code a bit

 
~~**Before Release: https://github.com/jmshrv/finamp/pull/1162#issuecomment-2813696566**~~ The Opposite actually now, Need to remove location permissions again for ios


Chaphasilor:
How about you change the way the baseUrl is retrieved inside `JellyfinApi`, and then add a new field to the `FinampUser` object for storing the `localBaseUrl`? Then you can use either of the two in `JellyfinApi`, depending on the current setting in `FinampSettings`.  
We can handle the UI along with the general redesign of the settings :)

Unreachable servers should be handled the same way the regular URL is handled when unreachable (so currently not at all, but that's an open issue, see #620)

Chaphasilor:
@flloschy try `FinampUserHelper.saveUser()` (and then delete the old user?). It's a bit hacky, but at the moment the baseURL is tied to the user object.  
Or maybe you can add a few more methods in `FinampUserHelper` to modify the current user object in Isar?

flloschy:
I think this works, though on mobile it cant get the network name for some reason. I guess finamp needs to request certain permissions?

Chaphasilor:
Network name probably requires location permission, just like for Immich. Since that can be (and in this case is) used to determine where the user is.

Chaphasilor:
https://pub.dev/packages/permission_handler

We currently have v11 as a dependency, but I think you can update to v12. Just make sure to look at the breaking changes first!

flloschy:
Ive not done the ios configuration for network permissions since I cant test it. Else this feature is feature complete xd

edeuss:
@Chaphasilor @jmshrv You will need to enable the `Access Wi-Fi Information ` in the `App ID Configuration` for the app in the Apple Certificates portal.

Screenshot 2025-04-17 at 20 07 26

lib/components/ensure_location_permission_prompt.dart
```diff
@@ -0,0 +1,64 @@
...
+Future<bool> ensureLocationPermissions(BuildContext context) async {
+  // Desktop doesnt need permissions
+  if (!(Platform.isAndroid || Platform.isIOS)) return true;
+
+  PermissionStatus permission = await Permission.locationWhenInUse.status;
+
+  // Permission wasnt requested yet
+  if (permission.isDenied) {
+    // Sorta simulate the ios dialog info text thing
+    if (Platform.isAndroid && !await _infoPrompt(context) ) return false;
```
flloschy, line 52: To confirm my logic here, if `isAndorid = false` the `_infoPrompt` wont be called right?

flloschy:
That should be it

lib/l10n/app_en.arb
```diff
@@ -2030,5 +2030,36 @@
...
+  "preferHomeNetworkActiveAddressInfoText": "Active Address",
+  "@preferHomeNetworkActiveAddressInfoText": {
+      "description": "Shows the active address inside the network settings screen so the user can see if this feature works"
+  },
+  "preferHomeNetworkTargetAddressLocalSettingTitle": "Home Network Address",
+  "preferHomeNetworkTargetAddressLocalSettingDescription": "Target address of the Jellyfin server inside your home network",
+  "preferHomeNetworkSettingNetworkNameTitle": "Home Network Name",
+  "preferHomeNetworkSettingNetworkNameDescription": "The name of your home network will be used to determine if you are at home or not",
+  "preferHomeNetworkEnableSwitchTitle": "Prefer Home Network",
+  "preferHomeNetworkEnableSwitchDescription": "Whether or not to use a different address when you are at home",
```
Komodo5197, line 2043: You should probably add some sort of description about the limitations of switching URLs with regards to the queue and whatnot.

lib/services/auto_offline.dart
```diff
@@ -88,3 +98,42 @@ Future<bool> _shouldBeOffline(List<ConnectivityResult>? connections) async {
...
+  Logger("Network Switcher").info("Changed active network to public address");
+  FinampUserHelper()
+    .currentUser!
+    .changeUrl(FinampSettingsHelper
+        .finampSettings.publicAddress);
+  currentAddressTextUpdateStream.add(null);
+}
+void _useHomeAddress() {
+  Logger("Network Switcher").info("Changed active network to home address");
+  FinampUserHelper()
```
Komodo5197, line 112: Do not call FinampUserHelper().  Use Use GetIt.instance<FinampUserHelper>() instead.  Also, maybe we should use currentUser?.changeUrl to avoid throwing errors when logged out?

lib/services/auto_offline.dart
```diff
@@ -88,3 +98,42 @@ Future<bool> _shouldBeOffline(List<ConnectivityResult>? connections) async {
...
+Future<void> changeTargetUrl({bool? isLocal}) async {
+  if (isLocal != null && isLocal) {
+    return _useHomeAddress();
+  }
+  else if (isLocal != null && !isLocal) {
+    return _usePublicAddress();
+  }
+
+  if (!FinampSettingsHelper.finampSettings.preferHomeNetwork) {
+    return changeTargetUrl(isLocal: false);
```
Komodo5197, line 128: I don't believe this is needed.  We explicitly call changeTargetUrl with isLocal: false when disabling the setting, so this only fires when changing connection or starting up with the setting already disabled, where this just causes needless re-saving of the user.

lib/services/auto_offline.dart
```diff
@@ -3,49 +3,59 @@ import 'dart:core';
...
-});
+    .listen(_onConnectivityChange);
+final currentAddressTextUpdateStream = StreamController<void>.broadcast();
 
 @riverpod
 class AutoOffline extends _$AutoOffline {
   @override
   void build() {
-    listener = Connectivity()
+    _listener = Connectivity()
```
Komodo5197, line 24: This should be a local variable, as updated in the main redesign branch.

lib/models/finamp_models.dart
```diff
@@ -575,6 +586,18 @@ class FinampSettings {
...
   ArtistType artistListType;
 
+  @HiveField(93, defaultValue: DefaultSettings.preferHomeNetwork)
+  bool preferHomeNetwork;
+
+  @HiveField(94, defaultValue: DefaultSettings.homeNetworkName)
+  String homeNetworkName;
+
+  @HiveField(95, defaultValue: DefaultSettings.homeNetworkAddress)
+  String homeNetworkAddress;
```
Komodo5197, line 596: I would prefer a solution where the two addresses and the home network name are stored in the user, and the baseUrl is a getter that selects between them.  I believe the url is stored with the user due to the idea that we may eventually have multi-server support, which theoretically is still on the roadmap?

lib/l10n/app_en.arb
flloschy: What limitations?

Komodo5197: As far as I'm aware, all the items in the queue get their full url calculated when they are added and will not respond to baseURL changes.  A similar thing happens should apply for any active downloads on ios, which all get submitted to the OS at once.

Komodo5197: I think this sort of URL lag will also break our log censoring, which just scrubs any instance of the base url it finds when writing log lines, so you may want to update that.

lib/components/NetworkSettingsScreen/active_network_display.dart
```diff
@@ -0,0 +1,48 @@
...
+    super.dispose();
+    if (_listener != null) {
+      _listener.cancel();
+      _listener = null;
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    String address = FinampUserHelper().currentUser!.baseUrl;
```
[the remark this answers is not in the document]
flloschy: There already is a static Provider in `FinampUserHelper` :D
But `ref.watch(FinampUserHelper.finampCurrentUserProvider.baseUrl)` throws `The getter 'baseUrl' isn't defined for the type 'AutoDisposeStreamProvider<FinampUser?>'.`

lib/components/NetworkSettingsScreen/prefer_home_network_address_selector.dart
```diff
@@ -0,0 +1,54 @@
...
+
+class HomeNetworkAddressSelector extends ConsumerWidget {
+  const HomeNetworkAddressSelector({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    bool preferLocalNetwork = ref.watch(finampSettingsProvider.preferHomeNetwork);
+    String localNetworkAddress = ref.watch(finampSettingsProvider.homeNetworkAddress);
+
+    final _controller = TextEditingController(
```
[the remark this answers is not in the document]
flloschy: How though? ConsumerWidget doesnt have a dispose function

lib/components/NetworkSettingsScreen/prefer_home_network_address_selector.dart
```diff
@@ -0,0 +1,54 @@
...
+          title: Text(AppLocalizations.of(context)!.preferHomeNetworkTargetAddressLocalSettingTitle), // TODO TRANSLATION
+          subtitle: Text(AppLocalizations.of(context)!.preferHomeNetworkTargetAddressLocalSettingDescription),
+          trailing: SizedBox(
+            width: 200 * MediaQuery.of(context).textScaleFactor,
+            child: TextField(
+              enabled: preferLocalNetwork,
+              controller: _controller,
+              textAlign: TextAlign.center,
+              keyboardType: TextInputType.url,
+              onChanged: (value) {
```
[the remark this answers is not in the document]
flloschy: Didnt know this exists :)

lib/components/NetworkSettingsScreen/active_network_display.dart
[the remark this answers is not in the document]
Komodo5197: Whoops.  But anyway, the specific attribute watches are actually FinampSettings-specific and are built via codegen.  You can just watch the whole provider, e.g ref.watch(FinampUserHelper.finampCurrentUserProvider).valueOrull?.baseUrl;

lib/components/NetworkSettingsScreen/prefer_home_network_address_selector.dart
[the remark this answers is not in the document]
Komodo5197: I think you need to switch to a ConsumerStatefulWidget.

lib/l10n/app_en.arb
Komodo5197: Maybe?  There's a lot of complications for that though.  The whole queue with exact urls ends up submitted to the native audio player, so any solution runs into a bunch of issues relating to things like having to remove and replace upcoming tracks, or avoiding ever submitting tracks to the OS until just before we need them, and there's lots of complications related to preloading of upcoming tracks.  There's been discussion of this sort of thing in https://github.com/ryanheise/just_audio/pull/779 but it hasn't gone anywhere yet.

flloschy: Would it work to clear the OS queue without clearing the internal queue and then just loading the queue again? Same with downloads

lib/models/finamp_models.dart
flloschy: Would you say including the toggle to enable this feature? Would make sense imo

lib/models/finamp_models.dart
```diff
@@ -51,6 +59,21 @@ class FinampUser {
   @HiveField(5)
   Map<BaseItemId, BaseItemDto> views;
 
+  @HiveField(6)
+  String publicAddress;
```
Komodo5197, line 63: To make migrations work properly, this field should be annotated with @HiveField(1) and @ Name("baseUrl").  Additionally, you should provide default values in the annotations for homeAddress, homeNetworkName, isLocal, and preferHomeNetwork.

lib/services/auto_offline.dart
```diff
@@ -99,41 +111,31 @@ Future<bool> _shouldBeOffline(List<ConnectivityResult>? connections) async {
...
-  currentAddressTextUpdateStream.add(null);
-}
-
 Future<void> changeTargetUrl({bool? isLocal}) async {
-  if (isLocal != null && isLocal) {
-    return _useHomeAddress();
-  }
-  else if (isLocal != null && !isLocal) {
-    return _usePublicAddress();
+  if (isLocal != null) {
```
Komodo5197, line 115: We could also check isLocal!= currentUser.isLocal here to reduce unneeded updates.

lib/components/NetworkSettingsScreen/prefer_home_network_selector.dart
```diff
@@ -1,41 +1,34 @@
...
+    return SwitchListTile.adaptive(
+      title: Text(AppLocalizations.of(context)!.preferHomeNetworkEnableSwitchTitle), // TODO TRANSLATION
+      subtitle: Text(AppLocalizations.of(context)!.preferHomeNetworkEnableSwitchDescription),
+      value: preferLocalNetwork,
+      onChanged: (value) async {
+        if (value) {
+          value = await ensureLocationPermissions(context);
+        }
+        GetIt.instance<FinampUserHelper>().currentUser?.update(newPreferHomeNetwork: value);
+        await changeTargetUrl();
```
Komodo5197, line 29: I don't believe we need to call this on any settings update but home network name.  Maybe this call to changeTargetUrl should be baked into the update() function?

lib/models/finamp_models.dart
```diff
@@ -34,8 +39,11 @@ class FinampUser {
 
   @HiveField(0)
   String id;
-  @HiveField(1)
-  String baseUrl;
+
+  // @HiveField(1)
+  // String baseUrl;
+  String get baseUrl => isLocal ? homeAddress : publicAddress;
```
Komodo5197, line 45: Maybe we should use isLocal&&preferHomeNetwork as a sanity check?

lib/models/finamp_models.dart
Komodo5197: I would have instinctively kept it in settings, but I think you may be right that it actually makes more sense to associate it with the user like this.

lib/l10n/app_en.arb
Komodo5197: Our internal queue is constantly rebuilt to match the OS one, and doesn't really matter.  The OS player is what actually preloads the tracks and plays audio and whatnot, so messing with it is what incurs all the complication and restrictions.  There's a lot of possibility of getting bad behavior by messing with it in weird ways.  I think @Chaphasilor has thought about this sort of thing more than I have?

lib/components/NetworkSettingsScreen/prefer_home_network_selector.dart
flloschy: Ive thought about that but it would result in a loop since `changeTargetUrl` calls the `update` function which then would call `changeTargetUrl` again and so on

Komodo5197: I think if we only called changeTargetUrl when updating the home network name it would work, because chagneTargetUrl only updates islocal.  And as I said, I think that's the only update that needs the call because all others should apply instantly via the getter.  This is assuming we don't end up with any fancy queue modification work when switching URLs that needs to be done, though.

flloschy: Ah yeah okay true. Already forgot its all provider based now haha

lib/l10n/app_en.arb
Chaphasilor: Yeah Komodo is absolutely right here, but what I think we can do is show a snackbar with a "reload queue" button that will just re-build the queue if offline mode is enabled or the baseUrl changes. That should be fairly easy to implement, and would only lead to a brief interruption.

flloschy: Why not automatically reload the queue though  
Especially since the address will switch even if the app isnt in forground

flloschy:
Readding the @Name causes this when doing code generation:
```
[SEVERE] isar_generator on lib/models/finamp_models.dart:

Two or more properties have the same name.
package:finamp/models/finamp_models.dart:26:7
   ╷
26 │ class FinampUser {
   │       ^^^^^^^^^^
```

Chaphasilor:
I think the local server scanning during the login flow is broken right now.

About the settings:
- Requesting permissions and obtaining WiFi name worked great
- The pre-filled address could be improved. There should be a scheme (http://) in front. The port should be Jellyfin's default (`8096`) instead of `1234`. We could also try a local network scan to discover Jellyfin servers on the network (we can get the local address from the UDP source IP), but that's probably out-of-scope. The functionality is however already implemented for the login flow.
- I first clicked the "Use current network" button before enabling the home network toggle, and while that worked, it didn't update the "Active Address" even after entering the local address. Leaving and re-entering the settings page also didn't help. I had to click the "Use current network" button again for the "Active Address" to update. Maybe that could be improved?
- A way to test the connection would be good. Maybe add a "Test connection" button that tries to send a simple request to the server (which uses authentication, so not /Public/System/Info!), and show a snackbar with the result? You could also compare the server ID to make sure we're connected to the same server via both URLs

About the functionality:  
It doesn't appear to be working for me? Settings say "Active address: <my public host>", but then I click the pause button and check the logs, and there's a request to `POST <my local host>/Sessions/Playing/Progress`. Tried with different URLs, domain-based and IP-based, and also tried with and without a scheme. Nothing worked :(  
Any tips what I might be doing wrong?

Chaphasilor:
@flloschy there error is because right below your annotation there already is a property called `baseUrl`. So you'd have to rename that as well I guess? I'm not totally sure what the annotation is for, I'm guessing to migrate existing baseUrls to be the default publicUrl instead?

flloschy:
Okay thats surprising, my used url does change 
image

Komodo5197:
I thought the getter with no setter would be ignored, but I guess not.  You can annotate baseUrl with @ ignore if you don't want to rename it.  The name tag is indeed to allow migration.  Regarding the address not updating, one possible issue is that I think the baseURLTemp from login has precedence over the real URL.  Could that be the issue?

flloschy:
Okay I got an Idea to circumvent the location requirement, the endpoint I use to test the connection (/System/Endpoint) returns two values one of which is if you are on the same network as the server. But I need to test that first if that works through the domain, currently not on my servers network but soon:tm:

Chaphasilor:
Ah yes, it's probably because I just logged in an the `baseUrlTemp` was still set. We probably need to clear that after login.  
If that's the case, restarting the app once should make everything work. I'll give that a try.

Chaphasilor:
Yeah it's working now! :D  
Thanks for adding the test button, I'll just change the wording a bit. The difference between "can" and "cannot" in a sentence with 7+ words is a bit too subtle ^^  

We should probably also restart the Play On service when switching between addresses, at least when switching from private to public. I'll add some commits for that in a bit.

Chaphasilor:
Okay, made some changes. I noticed the downloads paused warning was also shown when nothing was currently downloading, so I fixed that as well.  
From my side this is ready to merge, what about you? :)

flloschy:
well the android build failed, so there is that
and also this isnt resovled yet https://github.com/jmshrv/finamp/pull/1162#discussion_r2051788502

flloschy:
okay this is awkward now, Im moving all my server stuff from my raspberry to my old pc, sooo I dont have jellyfin for the time being haha
I also get the build error:
```
This will generate a JSON format file containing all messages that 
need to be translated.
Font asset "tabler-icons.ttf" was tree-shaken, reducing it from 2472096 to 17640 bytes (99.3% reduction). Tree-shaking can be disabled by providing the --no-tree-shake-icons flag when building your app.
Font asset "MaterialIcons-Regular.otf" was tree-shaken, reducing it from 1645184 to 9904 bytes (99.4% reduction). Tree-shaking can be disabled by providing the --no-tree-shake-icons flag when building your app.
Caught exception: Already watching path: /home/floschy/Projects/finamp-hackathon/android

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:packageRelease'.
> A failure occurred while executing com.android.build.gradle.tasks.PackageAndroidArtifact$IncrementalSplitterRunnable
   > SigningConfig "release" is missing required property "storeFile".

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 8m 43s
Running Gradle task 'assembleRelease'...                          523.8s
Gradle task assembleRelease failed with exit code 1
```
running `flutter build apk --debug` works though? interesting

Chaphasilor:
@flloschy you're a contributor, so why haven't you read [`CONTRIBUTING.md`](https://github.com/jmshrv/finamp/blob/main/CONTRIBUTING.md) in its entirety? :P

> ### Android Keys
> 
> To build release APKs, you need to set up a signing key for Android. To get that set up, follow [this guide](https://docs.flutter.dev/deployment/android#signing-the-app) from the Flutter documentation. Note that if you have Finamp installed already, your phone may panic because the key is different.

Chaphasilor:
You can ignore the Android stuff. Just gradle being grade. I had to make some changes to get the build working for me after an update, which in turn broke the GitHub build...  
And about the queue issues, yeah maybe it's time to tackle this now 😅

Chaphasilor:
Well you *are* running an APK. That's how Finamp ends up on your phone ^^  
I'm working on the queue stuff already. The issue is that we can't modify stuff in the queue after it has been added, so we need to clear the queue completely and replace it with newly-generated queue items that use the new source. Which causes an interruption.  
I almost got a snackbar with button working now, and I might also try adding a setting to make this automatic. However, the latter might not work reliably, because when we're replacing the queue, we are briefly stopping playback, which allows the system to kill the app in the background. This is currently also an issue when trying to play something new via Play On...

Chaphasilor:
Yeah that's one option, but it's messy. Longer/HQ tracks might not be fully cached, and then we got mixed sources and such. We would also need to replace all previous tracks (because they're not previous tracks any more if the user shuffles the queue), but removing tracks before the current tracks messes with IDs, which causes issues.  
So far the queue reload seems to be quick. If we get issues with it we can always try to make more messy changes.  
And actually playback isn't stopped currently when replacing the queue (changed in the last one or two updates), so we might not run into that issue at all...

Chaphasilor:
Okay, I'm relatively happy with how it works now. A prompt will be shown if the data source changes (new baseUrl or offline mode toggle) *and* the current queue would actually be affected. Clicking the button on the prompt will reload the queue, re-using the existing `BaseItemDto`s to speed things up significantly.   
What's not working right now is checking how many tracks have been buffered already to prevent removing them from the queue. But if we don't remove them, we could get problems later on (e.g. if the user toggles shuffle and the tracks then up being loaded once again, via the old source).

Later on an option to fully reload the queue, including all `BaseItemDto`s, could be added, although I currently don't see a need for it.

lib/services/queue_service.dart
```diff
@@ -403,7 +414,8 @@ class QueueService {
       await _replaceWholeQueue(
           itemList: items["previous"]! + items["current"]! + items["queue"]!,
           initialIndex: items["previous"]!.length,
-          beginPlaying: false,
+          beginPlaying:
+              _audioHandler.playbackState.valueOrNull?.playing ?? false,
```
Komodo5197, line 418: I think this should also use the original behavior when not reloading.  I'm mostly thinking of restoring queues from the restore queue screen, where we could be playing and probably want to start paused.

lib/services/queue_service.dart
```diff
@@ -568,6 +580,37 @@ class QueueService {
...
+
+    await _saveCurrentQueue(withPosition: true);
+    archiveSavedQueue();
+
+    var info = _queuesBox.get("latest");
+    if (info != null) {
+      final Map<jellyfin_models.BaseItemId, jellyfin_models.BaseItemDto>
+          existingItems = {};
+      final queueInfo = getQueue();
+      for (var item in (FinampSettingsHelper.finampSettings.isOffline
```
Komodo5197, line 598: It might make sense to just not supply any items when offline, so that we re-fetch them from the downloads service with the latest status.  It should be pretty quick, I think?

lib/services/queue_service.dart
```diff
@@ -568,6 +580,37 @@ class QueueService {
...
 
+  Future<void> reloadQueue() async {
+    _queueServiceLogger.info("Reloading queue");
+
+    if (_queueAudioSource.length == 0) {
+      return Future.error("Queue is empty, cannot reload!");
+    }
+
+    await _saveCurrentQueue(withPosition: true);
+    archiveSavedQueue();
```
Komodo5197, line 591: We can probably avoid archiving the queue in all cases except toOffline, because all the other cases should be able to rebuild every track so only toOffline risks dropping items.  This reduces duplicate queues being stored.

lib/components/global_snackbar.dart
```diff
@@ -128,4 +128,29 @@ class GlobalSnackbar {
       ),
     );
   }
+
+  /// Show a localized action to the user using the global context
+  static void action(
+    String Function(BuildContext scaffold) message, {
+    required SnackBarAction Function(BuildContext scaffold) action,
```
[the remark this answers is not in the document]
Chaphasilor: I guess we could. It's just a semantic difference.  
Having it separate gives us more freedom to deviate, but I'm not sure if that's needed.

lib/services/queue_service.dart
```diff
@@ -252,9 +250,7 @@ class QueueService {
     _audioHandler.queueTitle.add("Finamp");
 
     if (_savedQueueState == SavedQueueState.saving) {
-      FinampStorableQueueInfo info =
-          FinampStorableQueueInfo.fromQueueInfo(getQueue(), null);
-      _queuesBox.put("latest", info);
+      final info = _saveCurrentQueue(withPosition: false);
```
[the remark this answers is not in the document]
Chaphasilor: The variable is used, but defaults to `false`. I just wanted to make things more explicit/obvious.  
Is that a bad practise?

lib/services/queue_service.dart
```diff
@@ -364,13 +372,16 @@ class QueueService {
       }
 
       // Get list of unique ids that do not yet have an associated item.
-      List<jellyfin_models.BaseItemId> missingIds =
-          allIds.toSet().difference(playlistIds).toList();
+      List<jellyfin_models.BaseItemId> missingIds = allIds
+          .toSet()
+          .difference(playlistIds)
```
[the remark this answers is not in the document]
Chaphasilor: I believe it is only inclusive if we're fetching a playlist.  
I didn't want to mess with `playlistIDs` for non-playlist sources.  
(On my phone right now, could be wrong since I don't see the entire context)

lib/services/data_source_service.dart
```diff
@@ -0,0 +1,96 @@
...
+
+Future<void> _onDataSourceChange(SourceChangeType event) async {
+  final QueueService queueService = GetIt.instance<QueueService>();
+  _dataSourceServiceLogger
+      .finest("Connectivity Change Triggered, event is '$event'");
+
+  final queueInfo = queueService.getQueue();
+
+  if (queueInfo.trackCount > 0) {
+    if ((event == SourceChangeType.toOffline &&
```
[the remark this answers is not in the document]
Chaphasilor: Was trying to future-proof, and make a habit of checking "event type" + "event conditions". Theoretically that check would be enough, yes. Maybe I can shorten the condition using a `List.contains()`.  

The reload not happening on offline -> online is fine I think, since the sources wouldn't change anyway. Only some metadata (like play counts) would potentially change, but since we're currently not re-fetching the BaseItems, no reload is needed.

lib/services/data_source_service.dart
```diff
@@ -0,0 +1,96 @@
...
+        (event == SourceChangeType.toLocalUrl &&
+            queueInfo.undownloadedTracks > 0) ||
+        (event == SourceChangeType.toRemoteUrl &&
+            queueInfo.undownloadedTracks > 0)) {
+      if (FinampSettingsHelper.finampSettings.autoReloadQueue) {
+        await queueService.reloadQueue();
+      } else {
+        GlobalSnackbar.action(
+          (context) =>
+              "${AppLocalizations.of(context)!.autoReloadPrompt}${event == SourceChangeType.toOffline && queueInfo.undownloadedTracks > 0 ? ". ${AppLocalizations.of(context)!.autoReloadPromptMissingTracks(queueInfo.undownloadedTracks)}" : ""}",
```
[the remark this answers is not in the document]
Chaphasilor: I'll try.

What about baking is concerning you? :D

lib/services/data_source_service.dart
```diff
@@ -0,0 +1,96 @@
...
+class DataSourceService extends _$DataSourceService {
+  StreamSubscription<String>? _baseUrlListener;
+
+  @override
+  void build() {
+    final FinampUserHelper finampUserHelper =
+        GetIt.instance<FinampUserHelper>();
+
+    ref.listen(finampSettingsProvider.isOffline, (_, value) {
+      final offlineEnabled = value;
```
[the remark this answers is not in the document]
Chaphasilor: Yeah I was still figuring out which argument is what. Thought it's `(value, previousValue)` at first, but it's the other way around, right?  
Function signature isn't super clear on this.

I can change it to `(wasOfflineBefore, isOffline)`?

lib/services/queue_service.dart
```diff
@@ -412,9 +424,9 @@ class QueueService {
                   id: "savedqueue"));
 
       Future<void> seekFuture = Future.value();
-      if ((info.currentTrackSeek ?? 0) > 5000) {
```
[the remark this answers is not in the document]
Chaphasilor: What's your reasoning for it? Only thing I can think of right now is using it in combination with the fade out or sleep timer to avoid overshooting the beginning of the track?

lib/services/queue_service.dart
Chaphasilor: I think there's an ID that we can compare here, right? To see if it's the same queue or not?

I've been meaning to make this less exact anyway, so that we can build a "multi-queue" feature that lets the user quickly switch between different started playlists or podcast/audiobook tracks.  
Do you think that could work?

lib/services/queue_service.dart
Chaphasilor: That could work. What's the benefit? To rely on background syncs for more up-to-date metadata, like favorite status?

lib/services/queue_service.dart
Chaphasilor: Ah, so `archiveQueue` enforces a copy to be created that the user can restore?  
That's a good idea then.

lib/services/queue_service.dart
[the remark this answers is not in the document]
Komodo5197: I'm talking about the return result, info.  We don't use the variable.

lib/services/data_source_service.dart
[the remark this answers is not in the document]
Komodo5197: I guess you're right that offline to online wouldn't do anything, so it can be skipped.  I'm not sure how doing the logic like this makes it more future proof?  Anyway, regarding future code, my preference for things like this, especially with an enum, is to always try to use a switch statement.  It makes things more explicit and the exhaustiveness checks are nice, especially if any values ever get added to the enum later.

lib/services/queue_service.dart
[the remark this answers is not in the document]
Chaphasilor: Ah, you mean the other way around, only remove what's in `idMap`, and ignore `playlistIDs`? I was thinking that re-fetching the playlist could contain newly-added items, so I didn't want to mess with it.  
So you think we should not add any new playlist tracks to the queue?

lib/services/data_source_service.dart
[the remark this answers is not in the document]
Chaphasilor: Okay, yes that could work. I'll just use a fall-through then, that should be obvious enough.  
Future-proofing was about ensuring the switch-case nature of the check, to always have clear conditions based on the event, and not mix different conditions from different events. Just a readability thing.  
But a proper switch-case is indeed better.

lib/services/data_source_service.dart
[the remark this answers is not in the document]
Komodo5197: Apparently I accidentally typed markdown? Anyway, I was just vaguely concerned about baking sentence structure into the code separate from the localization.

lib/services/data_source_service.dart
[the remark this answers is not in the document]
Komodo5197: void listener(T? previous, T next ), apparently.  I don't think I've ever actually added a real listener to a provider before.  (_, isOffline) should be fine.  I just thought it was funny that you get the argument, then immediately rename it for no reason.

lib/services/queue_service.dart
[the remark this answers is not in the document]
Komodo5197: I'd just like it to snap to the track start with reasonable aggression when restarting the app.  If you come back a day or two later, its nice to actually start at the beginning of a song instead of a 3 seconds into it.  For the 1 second rewind on restore, I think I just wanted to give a bit of context?  I'm not sure how noticeable that sort of thing really is though.  And of course, neither of these are things you want when loading the same queue as is currently active.

lib/services/queue_service.dart
Komodo5197: I was mostly imagining that reloadQueue would just give this function a flag.  I don't believe queues have an ID, per se, but they do have a creation timestamp, if you want to use that for something.  Regarding multi-queue, what exactly do you mean by 'less exact'?

lib/services/queue_service.dart
Komodo5197: Providing nothing is simpler than filtering like this, and theoretically someone could have deleted some of the items since the queue was created.  Honestly, I don't think either choice really matters much.

lib/services/queue_service.dart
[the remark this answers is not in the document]
Chaphasilor: Okay, sure. I was just wondering how often the queue would save between 0.5 and 5 seconds of progress, since we're saving right after track change and periodically with a minimum interval of 10s. It *could* happen, but should be rare.

I can re-enable it though, jt shouldn't hurt.


And I also noticed that I need to disregard the seek position if the current track way regarded because it's un-downloaded. Anything else is very strange ^^

lib/models/finamp_models.dart
```diff
@@ -52,6 +63,18 @@ class FinampUser {
   @HiveField(5)
   Map<BaseItemId, BaseItemDto> views;
 
+  @HiveField(7)
```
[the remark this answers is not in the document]
flloschy: Like [so](https://github.com/jmshrv/finamp/pull/1162/commits/27c640f71749d97b0fcf280f83a9dbc72f2a84c9)?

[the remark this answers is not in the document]
Komodo5197: No, that doesn't help.  The core issue is that when hive deserializes the old objects it doesn't have values for these fields, but the constructer requires them.  There's basically two possible paths here.  One is to give hive defaults to pass into the constructer, e.g. @HiveField(2, defaultValue: DefaultSettings.transcodeBitrate).  The other is to add defaults like you did, but then also modify the constructor to not require these fields.  If you go that route, I think you can remove the hive annotations for all the new fields entirely, because after the migration this class is only stored via Isar.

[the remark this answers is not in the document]
flloschy: Ahh, those defaults :D
I went the `defaultValue` route since its used everywhere

lib/models/finamp_models.dart
```diff
@@ -63,16 +62,13 @@ class FinampUser {
   @HiveField(5)
   Map<BaseItemId, BaseItemDto> views;
 
-  @HiveField(7, defaultValue: DefaultSettings.homeNetworkAddress)
+  @HiveField(6, defaultValue: DefaultSettings.homeNetworkAddress)
```
Chaphasilor, line 65: Why was id 6 unused before? Was it used in the past?

lib/screens/network_settings_screen.dart
```diff
@@ -44,45 +38,31 @@ class _NetworkSettingsScreenState extends State<NetworkSettingsScreen> {
...
                 .then((available) {
                   if (available) {
-                    GlobalSnackbar.message((context) => AppLocalizations.of(context)!.pingSuccessful);
+                    GlobalSnackbar.message((context) => AppLocalizations.of(context)!.pingSuccessful(active));
                   } else {
-                    GlobalSnackbar.message((context) => AppLocalizations.of(context)!.pingFailed);
+                    GlobalSnackbar.message((context) => AppLocalizations.of(context)!.pingFailed(active));
                   }
                 });
+              if (active != local) {
```
Chaphasilor, line 56: So we're pinging twice now? Maybe have a single, more customized message for the result? Something like "Both addresses valid", "Local address invalid", "Remote address invalid"?

lib/services/jellyfin_api_helper.dart
```diff
@@ -835,8 +834,42 @@ class JellyfinApiHelper {
...
+    FinampUser? user = GetIt.instance<FinampUserHelper>().currentUser;
+    if (user == null) return false;
+
+    final client = ChopperClient(
+        baseUrl: Uri.tryParse(user.homeAddress),
+        client: http.IOClient(HttpClient()
+              ..connectionTimeout = const Duration(
+                  seconds: 3)),
+        interceptors: [
+          jellyfin_api.JellyfinLocalInterceptor()
```
Chaphasilor, line 848: Can't we also add the `HttpAggregateLoggingInterceptor` interceptor here?

lib/services/network_manager.dart
```diff
@@ -76,9 +94,9 @@ Future<void> _onConnectivityChange(
   _networkAutomationLogger.finest(
       "Network Change: ${connections?.map((element) => element.toString()).join(", ") ?? "None (likely a manual function call)"}");
   connections ??= await Connectivity().checkConnectivity();
-  final [offlineModeActive, baseUrlChanged] = await Future.wait([
+  final [_, baseUrlChanged] = await Future.wait([
     _setOfflineMode(connections),
-    changeTargetUrl(),
+    changeTargetUrl()
```
Chaphasilor, line 99: why did you remove the trailing comma? :)

lib/services/network_manager.dart
```diff
@@ -76,9 +94,9 @@ Future<void> _onConnectivityChange(
   _networkAutomationLogger.finest(
       "Network Change: ${connections?.map((element) => element.toString()).join(", ") ?? "None (likely a manual function call)"}");
   connections ??= await Connectivity().checkConnectivity();
-  final [offlineModeActive, baseUrlChanged] = await Future.wait([
+  final [_, baseUrlChanged] = await Future.wait([
```
Chaphasilor, line 97: Did the linter yell at you because you're not using the variable? If so, just ignore it. I think it's good to know what a returned value is/represents.

lib/services/network_manager.dart
```diff
@@ -89,12 +107,12 @@ Future<void> _onConnectivityChange(
 }
 
 /// Sets the offline mode based on the current connectivity and user settings
-/// Returns true if offline mode is now enabled
+/// Always returns false for proper typing in the above Future.wait
```
Chaphasilor, line 110: I don't really understand this

lib/services/network_manager.dart
```diff
@@ -217,3 +223,38 @@ void reconnectPlayOnService(List<ConnectivityResult> connections) async {
...
+
+//   String? currentWifi = await NetworkInfo().getWifiName();
+//   if (currentWifi == null) return changeTargetUrlNetworkBased(isLocal: false);
+
+//   // Android returns wifi name with quotes
+//   currentWifi = currentWifi.replaceAll("\"", "");
+//   _networkAutomationLogger.finest("Wifi Name detected: $currentWifi");
+
+//   return await changeTargetUrlNetworkBased(isLocal: currentWifi == targetWifi);
+// }
```
Chaphasilor, line 260: Why not remove it fully?  
You can get it back via Git if needed :)

Chaphasilor (changes requested):
Didn't expect you to discard all the location logic 😅  
Would be nice to keep it around just in case it turns out to be more reliable. But we can do that later on, if needed. Getting rid of the permissions is a good idea for now :)  

Haven't actually tested this yet, this is purely diff-based.

lib/services/network_manager.dart
flloschy: Dart said The `final [offlineModeActive, baseUrlChanged] =` is invalid when returning a void since its not all the same datatype

lib/services/network_manager.dart
flloschy: because I way playing with the idea of being able to toggle between ping and location based detection

lib/services/network_manager.dart
flloschy: Artifact from trying to make offlineModeActive return a void

lib/services/data_source_service.dart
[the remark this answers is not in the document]
Chaphasilor: I think since these are two separate sentences it shouldn't be dramatic. But at some point the snackbar could be replaced with something similar to [this](https://github.com/jmshrv/finamp/issues/620#issuecomment-1975134876), which should make the translation cleaner.

lib/services/data_source_service.dart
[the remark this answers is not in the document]
Chaphasilor: > I just thought it was funny that you get the argument, then immediately rename it for no reason.  

Yeah I have no idea what I did there 😂  
Thanks Floschy!

lib/services/queue_service.dart
Chaphasilor: Okay maybe the saved queues don't but, the current queue has a UUID, `getQueue().id`. That could be saved alongside the persisted queues, and then we could use that for checking if it's "the same" queue, not only when restoring.  
By "less exact" I mean that a queue stays the same even if some tracks are added or removed. So it's not an exact comparison, instead we just compare the ID. I think this is currently handled via the `archiveSavedQueue()` method, which creates a new distinct queue for some changes, but not others.

For now, I'll add a simple flag.

lib/services/queue_service.dart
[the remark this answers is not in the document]
Chaphasilor: I've completely removed the `playlistIds` variable now, since that was its only purpose...

lib/services/network_manager.dart
Chaphasilor: Ah right, since it's not a future

Chaphasilor:
@flloschy if you give me the go, I'll merge this later!  
I will mark the "Auto-Reload Queue" settings as experimental because it will almost definitely remove all non-downloaded tracks if the app was in the background for any duration, which is a bit frustrating. We'll need to see how to make the offline mode more "reliable", so that it's only enabled if connectivity is lost *and* is actually needed right now. But that's a separate issue.

The snackbar prompts can also pile up; it seems like the `duration` option doesn't apply when there's an `action` configured? But the notification system needs an overhaul anyway, I'll create a tracking issue for that :)

Chaphasilor:
We'll see ^^  
I'll handle the conflicts

PR #1165, opened 2025-04-09
feat: switch to using gaimon for feedback

edeuss:


lib/components/AlbumScreen/speed_menu.dart
```diff
@@ -78,7 +77,7 @@ class _SpeedSliderState extends State<SpeedSlider> {
               setState(() {
                 _dragValue = value;
               });
-              FeedbackHelper.feedback(FeedbackType.impact);
+              FeedbackHelper.feedback(FeedbackType.heavy);
```
Chaphasilor, line 80: I'll have to test how this feels, but it should definitely only be a short feedback

Chaphasilor:
Okay, I've made some small adjustments to avoid feedback being too strong in some cases, but nothing major.  
I also noticed that in debug builds the feedback is sometimes really delayed, especially in the plus/minus buttons of the speed menu, but that seems to be a general issue and also happens on different branches from my testing :)

Chaphasilor:
So yeah, thanks for the quick PR! :D

Issue #941, opened 2024-11-03
Add an option for "end of track" in the sleep timer


george-dk:
I have been using the redesign for a while and I really like it. However, it would be nice to have this option as many other music players have this, could also potentially add an input box specifying the amount of tracks before stopping playback rather than having a single "end of track" option. Sorry if it was already suggested before, I already searched all the issues regarding the sleep timer function and saw no one suggesting it.

matty-r:
@Chaphasilor commenting to give this one a go

Issue #769, opened 2024-06-06
Volume Normalization short delay noticeable


Emporea:
The moment a song is finished and a new one begins finamp takes around 0.1~ seconds to adapt to the new normalization dB.
This is especially noticable if the following song has higher volume than the first one (eg from -10dB to -4dB) because then the first 0.1-0.2 seconds are very loud, thus "ruining" gapless play.

Sorry if I mixxed things up with the dB values. I am always confused about which value means a lower volume, and which one is louder.

Chaphasilor:
Is this on Android or iOS?

I can try to change when/how the new gain is applied, but it will probably be hard to time it exactly. Maybe a gradual fade over a short duration would work better 🤔

Chaphasilor:
> simply setting the new volume at the moment the next track begins should solve the issue

Well, in a way that's exactly what we're currently doing. We wait for an event that tells us that a different track is playing. It seems like the event might be slightly delayed in some cases, so maybe we can improve the timing somehow.

Which phone are you using?

solidsnake1298:
I noticed the same issue on my Pixel 6a.  But only for tracks with vastly different normalization.  In my case going from a +5dB song to a -4dB song that starts right at the beginning of the file.

Titaniumtown:
I can reproduce this on Pixel 7 with 0.9.12 It is very annoying and distracting.

Chaphasilor:
Is the delay always the same? Or does it depend on what you're playing or maybe how big the queue is?  
When testing out the limits of the queueing system (queueing up thousands of tracks), I noticed that the delay was becoming greater, but that could've been a general slowdown of the app.

solidsnake1298:
My playlist is a sort of "living playlist" and hovers around ~130 songs.  I only really notice the delay in normalization if the song I'm currently listening to has a really high normalization value (the unmodified file is quiet) and the next song in the playlist has a small normalization value AND the music starts _**immediately**_ at the beginning of the file.  I'd say the delay is less than 250ms.

Titaniumtown:
> Is the delay always the same? Or does it depend on what you're playing or maybe how big the queue is? When testing out the limits of the queueing system (queueing up thousands of tracks), I noticed that the delay was becoming greater, but that could've been a general slowdown of the app.

[...]

PR #1166, opened 2025-04-09
Issue #941 - Better sleep timer options

matty-r:
This fixes/implements a better sleep timer for Issue #941 - We have a SleepTimer object with either a tracks or duration type. And we can either select a duration (in minutes) with the option to play until the end of the track, or we can select a number of tracks, or simply until the end of the current track:

[Screencast_20250410_060658.webm](https://github.com/user-attachments/assets/0d46bbdb-6b0e-4ab0-90fa-a62551301f78)


Chaphasilor:
Awesome, I'll give this a test tomorrow!  
The UI looks a bit out of place at the moment. Should I handle that, or do you want some pointers from my side instead?

matty-r:
> Awesome, I'll give this a test tomorrow! The UI looks a bit out of place at the moment. Should I handle that, or do you want some pointers from my side instead?

Let me know what you'd like changed and I can fix it up :)

Chaphasilor:
I think since at the moment we can only activate the sleep timer via the player screen menu, it would be nice to move the sleep timer options into an expandable menu within the menu, just like the playback speed options.  
The time and track selection could work very similar to setting the playback speed, by having a bunch of presets and an input field.  
I also still think having tabs or chips to switch between for the minutes vs tracks options would make sense.  
The "after current track" could be a third tab/chip instead of being an option, just to be quick and easy to select.

For a simpler change, you could simply tweak the dialog theme a bit to be less gray and bold?

What do you think?

matty-r:
Sounds like a decent idea, i'll give that a go

Chaphasilor:
Nice. I just noticed that there is no input field for the speed selector, just the slider, but maybe we can use your fancy scrollable slider (that is infinite, right?) and but a number field next to it or below it?

Chaphasilor:
Sure, if you prefer that! I'll give it a brief test tomorrow morning and then merge it :)

Chaphasilor:
> tomorrow morning  
well, so much for that. Sorry, life got in the way.  

Took a look at the code just now and there are a few things I'll tweak before merging:
- [ ] Remove some stray imports
- [ ] Add missing translation keys instead of using hard-coded strings
- [ ] Add the remaining time calculation
- [ ] Switch to using a different hive field. We can't change the data type of an existing field AFAIK, so we'll just create a new one.
- [ ] Change the ID of the new HiveType. We can't re-use old IDs since that can cause issues with older installations.
- [ ] Handle the logging differently. Overwriting the callback would get rid of the logging as well, which isn't what we want.
- [ ] Make `getSleepTimer()` an actual getter for brevity
- [ ] Pre-select the "Minutes" option for the dialog to provide a default and make sure at least one input field is enabled when the dialog is first shown

In general I'm also not sure if it's a good idea to have such logic-rich classes as a HiveType; maybe it would be better to instead convert this into a sleep timer service class, and then only have specific configuration settings as a type that are restored into the service on startup.

But in general I like the implementation, and how you organically interleaved the two timer types to offer additional functionality (i.e. for the "Play last track until the end" stuff)!

As for the actual functionality, I noticed that the dialog is a bit broken on the light theme, probably because you hard-coded the colors? The checkboxes are invisible, and the input fields are therefore disabled, which is confusing.  
And there is a short delay between the track ending an playback actually pausing, which leads to the next track playing very briefly. This is probably related to the same issue as #769: https://github.com/jmshrv/finamp/pull/1167#issuecomment-2798007303, so not your fault in any way :)

Chaphasilor:
Functionality-wise this should be working now. Just adding the translation tokens and maybe refactoring some things is missing.  
I'll probably leave this as a dialog for now to get the functionality in, and convert it to a submenu (like the playback speed controls) later on.

Komodo5197:
Is this somethinfg like what you were looking for?

```
SliverToBoxAdapter(
        child: AnimatedSize(
          duration: trackMenuDefaultAnimationDuration,
          curve: Curves.easeInOut,
          clipBehavior: Clip.none,
          alignment: Alignment.topCenter,
          child: SizedBox(
            height: menuHeight,
            child: OverflowBox(
              maxHeight: double.infinity,
              fit: OverflowBoxFit.deferToChild,
              alignment: Alignment.topCenter,
              child: AnimatedSwitcher(
                duration: trackMenuDefaultAnimationDuration,
                switchInCurve: Curves.easeInOut,
                switchOutCurve: Curves.easeInOut,
                layoutBuilder: (currentChild, previousChildren) {
                  return Stack(
                    alignment: Alignment.topCenter,
                    children: <Widget>[
                      ...previousChildren,
                      if (currentChild != null) currentChild,
                    ],
                  );
                },
                transitionBuilder: (child, animation) {
                  if (MediaQuery.of(context).disableAnimations) {
                    return child;
                  }
                  // Determine if this is the incoming or outgoing child
                  final isSpeedMenu = (child.key is ValueKey &&
                      (child.key as ValueKey).value == 'speed');
                  final isSleepMenu = (child.key is ValueKey &&
                      (child.key as ValueKey).value == 'sleep');
                  // Slide in from right for speed, left for sleep
                  final Offset beginOffset =
                      previousMenu == null || activeMenu == null
                          ? Offset.zero
                          : (isSpeedMenu
                              ? const Offset(1, 0)
                              : isSleepMenu
                                  ? const Offset(-1, 0)
                                  : Offset.zero);

                  final Offset endOffset = Offset.zero;
                  return FadeTransition(
                    opacity: animation,
                    child: SlideTransition(
                      position: Tween<Offset>(
                        begin: beginOffset,
                        end: endOffset,
                      ).animate(animation),
                      child: child,
                    ),
                  );
                },
                child: switch (activeMenu) {
                  SubMenu.speed => SpeedMenu(
                      key: const ValueKey('speed'),
                      iconColor: iconColor,
                    ),
                  SubMenu.sleepTimer => SleepTimerMenu(
                      key: const ValueKey('sleep'),
                      iconColor: iconColor,
                      onStartTimer: () {
                        toggleSleepTimerMenu();
                      },
                    ),
                  _ => null,
                },
              ),
            ),
          ),
        ),
      ),
```

Chaphasilor:
That worked perfectly. I was also able to transition between the different submenu heights when switching between sleep timer types.

PR #1167, opened 2025-04-10
[Redesign] Hackathon Cleanup

Komodo5197:
Various changes to things that bugged me from the hackathon.

- Attempted to switch as many uses of FinampSettings as possible to riverpod watches of specific settings.  This might remove some unneeded rebuilds and increase responsiveness to settings changes.
- Fix a minor bug with theme precalculation.
- Fix log censoring prioritizing TEMP replacement text when both are possible
- Refactor displayFeaturesProvider and brightnessProvider to be cleanly set up in main() with no double rebuilds.
- Fix issues with registerAdapter and hive type 41.
- Remove postLaunchHook from music screen.  Playlist metadata download moved to dedicated user login hook.
- Added library downloads into downloads screen 'special downloads' category.
- Updated download repair link pruning to account for new album->album artist info links.
- Moved download file size calculation to background isolate.
- Refactored audio fade in/out code to separate replayGain and fade volumes.
- Prevented replayGain reapplying with every settings change.
- Pause player on startup on desktop because it seems to default to playing.
- Removed some plugin files that were in gitignore.
- Switch isar_generator to a fork with analyzer 7.0.0 compatibility.  This allowed the riverpod linter to begin working again.  It seems to produce identical generated files.

I'll be honest, this isn't the most thoroughly tested PR ever.  But it probably all works?


Chaphasilor:
Awesome, a lot of good stuff in there!  
I probably won't be able to test everything individually, but maybe I'll gloss over it tomorrow. Although some things don't even need testing I guess, e.g. just adding a select to a more general watch of the settings.

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -269,12 +270,17 @@ class TrackListTile extends StatelessWidget {
...
+                    .watch(finampSettingsProvider.itemSwipeActionLeftToRight)),
                 color: Theme.of(context).colorScheme.secondary,
                 size: 40,
               ),
               const SizedBox(width: 4.0),
-              Text(currentSettings.itemSwipeActionLeftToRight.toLocalisedString(context),
+              Text(
+                ref
+                    .watch(finampSettingsProvider.itemSwipeActionLeftToRight)
+                    .toLocalisedString(context),
```
Chaphasilor, line 283: You're using the static settings to get the followUpAction, but then watch the settings to render text & icon? I don't see the benefit in this case.

lib/services/display_features_helper.dart
Maxr1998, line 1: File should be renamed to include new behavior

lib/services/display_features_helper.dart
```diff
@@ -1,50 +1,84 @@
...
-  bool get halfOpened => _halfOpened;
+final brightnessProvider = StateProvider((ref) => Brightness.dark);
 
-  void attach(BuildContext context) {
-    _context = context;
+final halfOpenFoldableProvider = Provider((ref) => ref.watch(
+    displayFeaturesProvider.select((x) => x.any((d) =>
+        (d.type == DisplayFeatureType.fold ||
+            d.type == DisplayFeatureType.hinge) &&
+// Flip-style foldable, top == bottom, height == 0
```
Maxr1998, line 21: Comment indentation seems off (but might just be the GitHub app)

Maxr1998 (changes requested):

Chaphasilor:
Oh, I didn't even know just watching specific settings was as easy as `ref.watch(finampSettingsProvider.xyz)`! I thought it required some form of `.select(x => x.y)`...  
I'll try to use that more often then :)

I would prefer to make library downloads a separate category; special downloads are semantically different imo.

I'll have to take a look how to integrate your new VolumeController with the internal volume setting introduced in #850, but that should be doable.  
Do you know if your changes to how/when ReplayGain is applied also help with #769 ? Or could you imagine where the problem might be, given that you're most up-to-date with the code right now?

If there's a fork of isar_generator, is there also a fork of isar already? I've heard that the hive/isar projects are unmaintained, and hive_ce is already a nice improvement in certain regards.

Thanks for the changes, especially the major refactor of how settings are accessed, much appreciated!

Maxr1998:
> Refactor displayFeaturesProvider and brightnessProvider to be cleanly set up in main() with no double rebuilds.

Thanks for handling this one! I was planning to look into it still, but didn't have the time yet. The proposed solution looks really clean.

lib/components/AlbumScreen/track_list_tile.dart
Komodo5197: The static usage is inside the callback, so the value used is always the most current when the swipe actually happens, not what it was when the widget was built.  If we'd calculated this outside in build() and then just reused that value inside the callback it would have been important to use watch(), but for actually getting values inside an asynchronous callback this is fine.

Komodo5197:
The easy watching of specific settings is the result of the new code generation.  finampSettingsProvider.xyz is a generated getter for finampSettingsProvider.requireValue.select((x)=>x.xyz).  For replaygain, I imagine the issue is just that there's a good amount of lag from track change in native player -> dart code responds and changes volume -> native player responds to volume change request.  I've moved the call a bit earlier in the event stream propagation, but this probably doesn't make any difference at all.

Komodo5197:
I was thinking about the new volume setting when making this.  You should be able to just add a third setting to the controller, resolving the conflict with the fade volume I believe that PR currently has.  There is indeed a community fork of isar to go with isar generator.  It seems like they haven't finished refactoring all the package name references to get it onto pub.dev yet though, and the generator I linked for analyzer 7.0.0 is an unmerged fork, so I decided we might want to wait a bit.

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -19,25 +20,24 @@ import '../favourite_button.dart';
 import '../padded_custom_scrollview.dart';
 import 'artist_screen_content_flexible_space_bar.dart';
 
-
-/// ToDo for proper Albums and Appears On Section + reasonable playback queue: 
+/// ToDo for proper Albums and Appears On Section + reasonable playback queue:
```
lukaslindnermusic, line 23: I already implemented this, I just forgot to delete my comment in the end, sorry. 
So if you want, you can remove these comments.
Otherwise I will do it with my next PR.

Thanks 😊

lukaslindnermusic (commented):
Forgot to delete a comment. ooops :P

Chaphasilor:
Yeah I also fear that the delay is simply due to how the track changes are handled, and/or because the gain control on Android is slower to respond than a simple volume change. The latter theory would be supported by not having seen any reports of volume change delays on iOS so far.  

Thanks for taking care of the other comments, and planning ahead for the internal volume stuff. Much appreciated!  
That isar fork also sounds good, but we are in no rush so we can wait for things to settle down and be finalized before switching over.

Didn't respond earlier since I was on vacation ^^

Maxr1998 (approved):

Issue #1029, opened 2025-01-22
Ability to see Artists, Tracks in the Genre tab, not just albums


neeleysc:
I meticulously use genres to organize my music and many players support browsing by Genre, Album Artist, then Album. This is my preferred way to choose music, but Finamp goes straight from Genre to Albums. This results in less granular browsing and an overwhelming number of albums to sort through. Could this be an option so users cold have it which every way they prefer?

Chaphasilor:
Just tried it, the Jellyfin app has an option to filter artists by genres, which is means it should be doable.  
In that case, might be able to consider your request in the redesign :)

neeleysc:
Thanks for looking into it! It's the thing I miss most about Google Play Music, so I use VLC mostly for music listening on the phone. I do agree that artists can transcend genres, so a per-album basis makes most sense when browsing Genre-Album Artist-Album.

Chaphasilor:
> so a per-album basis makes most sense when browsing Genre-Album Artist-Album

Could you explain what you mean by "browsing Genre-Album Artist-Album" please?  
I was so far thinking of letting users choose if they want to see albums or artists within that genre, would that work?

neeleysc:
Yeah, that's what I mean. Going to the Genre tab, choosing the artist or more specifically the album artist, then choosing the album from that artist associated with the genre.

Chaphasilor:
Ah okay, so we are talking about two different things. But I just gave it a try, we can also filter albums by (album) artist **and** genre, so it seems doable, but becomes a bit harder to implement...

neeleysc:
Here is a video example from Symfonium:

https://github.com/user-attachments/assets/69a6a613-05a1-4a98-bc68-65d421af4e70

mist2t:
I second this feature request 🙂

An "aggregated" genre page that also shows artists for that genre will be amazing.

Maybe have various "sections" on the genre page: album artists, albums, songs ... with few items displayed for each and the ability to navigate further into all album artists for that genre, all albums, all songs etc.

Chaphasilor:
@lukaslindnermusic would you mind commenting here so I can assign you to the issue? :D

lukaslindnermusic:
Yes, sure 😊 

Just so that the current plan is also here on Github and not just on Discord:

We're planning to show Top Albums, Top Artists and Top Tracks on the genre screen (maybe something like the top 5 of each category). 
[...]

Issue #254, opened 2022-06-15
[Feature Request] Ability to shuffle play songs within a Genre


MagicTrevor:
Just like on the songs list it would be nice if you could have a shuffle play button within the song list of a genre.

faizzzzz:
A good work around is to use the "Instant Mix" feature in the songs tab. 
Long press on a song of a genre that you want to shuffle and tap on the "Instant Mix" option. It will shuffle all the songs of the genre of the song that you selected. For example, if you selected a song that is in the "Alternative" genre, starting an "Instant Mix" from it will shuffle and play all the songs of the "Alternative" genre.
![image](https://user-images.githubusercontent.com/27871755/174140285-c5581118-1d86-46ca-aa8b-f9834b51ed89.png)

4o66:
Is the Instant mix shuffling every song of all genres of the originating track?
What are the specific rules for this please?

jmshrv:
It'll use stuff like the genre to shuffle, I'm not sure how the algorithm works specifically but it'd probably give you music of the same genre

lukaslindnermusic:
But as this issue only specifically is refering to a shuffle button on the list of genre tracks similar to the one on the tracks tab, I think we can consider this issue closed @Chaphasilor ?

Issue #182, opened 2022-01-18
[Feature Request] Larger Album/Artist View and Show Album Genres


seniorm0ment:
Requesting genres be shown on album page to click.

Also an option for an expanded album page, and artist page (which I mentioned in another issue) where I can see the artwork in big at the top of the page, and the tracks below it and any information or metadata.
Similar to the Jellyfin app and web gui.

UnicornsOnLSD:
That should be pretty easy to do. I could think of ways to better show album art in the album screen since right now its a bit small.

Chaphasilor:
This has now been resolved as part of [beta version 0.9.18](https://github.com/jmshrv/finamp/releases/tag/0.9.18-beta)!

PR #1176, opened 2025-04-20
New Genre Screen

lukaslindnermusic:
This is a PR for https://github.com/jmshrv/finamp/issues/1029. Please check the issue for details.

I added a dedicated GenreScreen and separated Genres from the ArtistScreen.
On the new GenreScreen, you first get <= 5 Picks of Tracks, Albums and Artists of that Genre. 
You can choose in the Settings for the GenreScreen, how those Picks will get curated: Most Played, Random Favorites only, Random with preferred Favorites, Random.

Next to each section, there is a "See All" Button that takes you to a genre-filtered list of every track|album|artist of that genre. I re-used the MusicScreen for this, as everything important (Search, Sorting,...) is already there. 

The genre-filter propagates to the ArtistScreen, so when you choose an artist from within a genre, you only get the albums and tracks that are assigned to that genre... there is a clear indication of that filter right next to the artist image and you easily can remove it to display every album/track.

The current state is already useable, but there is still some stuff to do:
- In the future, it is planned to switch the style of the curated Picks to the style that will be used on the HomeScreen.
- Also, I noticed that in the old implementation, the ArtistScreen hid the Play and Shuffle Buttons if the displayed item was a Genre. I'm not quite sure why, but I guess it was because loading a huge genre directly into the queue in one go is too laggy? 
I left it like that for now, but I'm planning on finding a way to bringing those buttons back. I can only speak for myself: I often browse genres and then just hit Shuffle on the genre I want to listen too.
- Something that could be nice as well is the option to edit the order of the items on the GenreScreen... (some people want the Artists to be on top, while others prefer Albums etc.). What do you think? Should I add this?

(This PR also fixes a bug with Performing Artists on the ArtistScreen, see here: https://github.com/jmshrv/finamp/pull/1145#issuecomment-2816940784)

Some Screenshots:
Simulator Screenshot - iPhone 15 Pro - 2025-04-20 at 05 21 31

Simulator Screenshot - iPhone 15 Pro - 2025-04-20 at 05 21 59

Simulator Screenshot - iPhone 15 Pro - 2025-04-20 at 05 22 33


lukaslindnermusic:
Another thing that I still want to add is that the GenreScreen doesn't refresh when you navigate back to it from a "See All" section.

lib/services/downloads_service.dart
```diff
@@ -1678,7 +1699,8 @@ class DownloadsService {
     }
     int childCount;
     if (stub.baseItemType == BaseItemDtoType.album ||
-        stub.baseItemType == BaseItemDtoType.playlist) {
+        stub.baseItemType == BaseItemDtoType.playlist ||
+        stub.baseItemType == BaseItemDtoType.genre) {
```
Komodo5197, line 1703: Why'd this change?  I believe the previous logic was correct.

lib/services/downloads_service.dart
```diff
@@ -1329,7 +1329,7 @@ class DownloadsService {
   /// album/playlist screen.  Can return all tracks in the album/playlist or
   /// just fully downloaded ones.
   Future<List<BaseItemDto>> getCollectionTracks(BaseItemDto item,
-      {bool playable = true}) async {
+      {bool playable = true, String? genreFilter}) async {
```
Komodo5197, line 1332: genrefilter should be a BaseItemDto, or at least a BaseItemId.

lib/services/downloads_service.dart
lukaslindnermusic: Because we now don't have all the albums of a genre from the genreScreen itself... we only have a selection (top albums, or random, or favorites, etc..). This is a) cleaner, you first get a selection and if you want to see all albums you can browse further, which then utilizes the pagination from MusicScreen, and b) this also improves loading time on genres with a lot of items.
and therefore, we don't have the correct itemcount for the status.
We do however have all the tracks, for playback (which I will add later). they are loaded detached from the initial UI rendering, so until they are there, we pass null to the status, and when they are there, we check all the tracks, and therefore, we have to move the BaseItemDtoType.genre to the check where it is looking for tracks childs.
As far as I understand it, this basically is the same, as genres are defined by the track-metadata anyway?
With my testing so far, it always showed the correct syncStatus.
Do I don't understand something or is this correct?

lib/services/downloads_service.dart
lukaslindnermusic: ahh yes, thanks 😊 
I first used the idString for everything, and later switched to BaseItemDto, but apparently I forgot to change this one. 😊 will fix it in the next commit.

lib/services/downloads_service.dart
lukaslindnermusic: I think, if you want to keep the previous logic, we could also load all Albums delayed, instead of allTracks. But then we have to fetch the tracks of each album once the user wants to start playback directly from the genre...

I also tested albums that have tracks with different genres. When I download a genre, it wil only download the tracks of that album that have the genre, not the whole album. So using the track-children should work as well, and for now, it always worked in my testing.

I'm still thinking of how we can improve the lagging when pressing shuffle or play on large genres etc.... I'm having the idea to first only add a limited number of items to the queue to quickly start playback, and then after x seconds add more items etc.

lukaslindnermusic:
I'm still working on:

directly related to the Genre Screen:
- [ ] Remove "Random (Favorites First)" Filter option
- [ ] Let the GenreScreen keep its state remembered when re-accessing it again with the same genre, only reload once a new genre gets accessed
- [ ] Kill the allTracks request early if its still ongoing when leaving the GenreScreen to improve performance hickups in other parts of the app (this will also be great for the ArtistScreen)
- [ ] Bring back Play / Shuffle all. Come up with a solution to do those on genres with a lot of tracks without lagging and app freezing (maybe by first adding only a limited amount of tracks to the queue to immediately start playback, and then after x seconds add more tracks in the background as you go?)

various other things:
- [ ] Fix SortOrder options in MusicScreen for Genres not working
- [ ] consistent icon for Albums on ArtistScreen Info Chips
- [ ] maybe: Offline Top Tracks on ArtistScreen? or later with another PR where we also properly update the Playcounts once back online?

lib/services/downloads_service.dart
lukaslindnermusic: Ok @Komodo5197, I have to take my previous messages back. Of course you were right!
The previous logic was correct! I'm sorry!

Here is what has happend:
The allTracks were fetched outside of the futures variable that is attached to the futurebuilder. The request was made but the children actually never made it to the DownloadButton widget, that always got null instead and therefore did not show an outdated sync status.

I've now added the option to get the totalRecordCount of jellyfin-api item requests, and modified DownloadButton so that it alternatively directly takes an integer (it still takes BaseItemDto-children if provided).

this means, that there is no need for an allTracks api request anymore, and we can fetch the totalAlbumCount even if we are requesting only 5 items. Only when filtering for favorites, we have to make an extra request, but with a limit:1, that happens really fast.

Now I think everything should work?

lib/services/jellyfin_api_helper.dart
```diff
@@ -283,9 +362,7 @@ class JellyfinApiHelper {
           isFavorite: isFavorite,
         );
       }
-
-      return QueryResult_BaseItemDto.fromJson(response as Map<String, dynamic>)
-          .items;
+      return response;
```
Komodo5197, line 365: you don't need to be directly exposing all the mess of dynamic types to outside methods here, and also we want the primary JSON parsing of the items list to be in the background, so you should just return the QueryResult_BaseItemDto here, and then just return that class from getItemsWithTotalRecordCount.

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -88,7 +89,8 @@ class _GenreScreenContentState extends ConsumerState<GenreScreenContent> {
       final sortBy = 
         [sortByMain, if (sortBySecondary != null && sortBySecondary.isNotEmpty) sortBySecondary]
         .join(',');
-      final items = await jellyfinApiHelper.getItems(
+
+      final fetchedItems = await jellyfinApiHelper.getItemsWithTotalRecordCount(
```
Komodo5197, line 93: My recommended solution has a QueryResult_BaseItemDto, but if you end up with something similar in the future this is a great case for pattern matching.  You can do something like 
`final {'items': List<BaseItemDto>? items, 'totalRecordCount': int? totalrecordCount} = await jellyfinApiHelper.getItems...`

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -102,12 +104,32 @@ class _GenreScreenContentState extends ConsumerState<GenreScreenContent> {
         artistType: (baseItemType == BaseItemDtoType.artist)
             ? artistListType : null,
       );
+      // Set the Total Record Count
+      if (setTotalRecordCount && genreCuratedItemSelectionType == GenreCuratedItemSelectionType.favorites) {
+        // When we filter the favorites, we have to make an additional request to get the correct number
+        // otherwise we would only get the totalRecordCount of Favorites of that genre
+        final fetchedItemCountWithoutFavorites = await jellyfinApiHelper.getItemsWithTotalRecordCount(
```
Komodo5197, line 111: I'm not 100% sure its actually worthwhile to make a request just for the record count instead of just using null.  It probably can't hurt too much, I guess?

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -102,12 +104,32 @@ class _GenreScreenContentState extends ConsumerState<GenreScreenContent> {
...
+          genreFilter: (baseItemType == BaseItemDtoType.artist)
+              ? widget.parent : null,
+          limit: 1,
+          includeItemTypes: baseItemType.idString,
+          artistType: (baseItemType == BaseItemDtoType.artist)
+              ? artistListType : null,
+        );
+        totalRecordCount = fetchedItemCountWithoutFavorites['totalRecordCount'] as int;
+      } else if (setTotalRecordCount) {
+        totalRecordCount = (setTotalRecordCount) ? fetchedItems['totalRecordCount'] as int : totalRecordCount;
```
Komodo5197, line 123: Isn't this ternary pointless due to the if?

lib/components/GenreScreen/genre_screen_content.dart
lukaslindnermusic: I now added a section for the track, album and artist counts at the top, because I think it still would be nice to quickly see how many items you have in one genre. 
But if that section gets disliked, then I will let it be null if favorites are selected as filter and forgo that additional request 👍

lib/services/jellyfin_api_helper.dart
Komodo5197: The change to return the Queryresult_BaseItemDto is better, but you still need to get it into the background.  getItems works by passing a big lambda into runInIsolate, and you need this fromJson call to be inside that lambda because its the most compute heavy part of the call, so we want to get it out of the UI thread.

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -16,6 +17,166 @@ import '../../services/jellyfin_api_helper.dart';
...
+      genreCuratedItemSelectionType: genreCuratedItemSelectionType,
+      baseItemType: baseItemType,
+      artistInfoForType: (baseItemType == BaseItemDtoType.artist)
+          ? artistInfoForType
+          : null,
+    );
+    items = result.$1;
+    itemCount = result.$2;
+
+    final listener = GetIt.instance<DownloadsService>()
```
Komodo5197, line 51: It feels slightly cleaner to put this inside getCuratedItemsOffline instead of outside it?

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -17,6 +18,187 @@ import '../favourite_button.dart';
 import '../padded_custom_scrollview.dart';
 import 'artist_screen_content_flexible_space_bar.dart';
 
+part 'artist_screen_content.g.dart';
+
+@riverpod
+Future<(List<BaseItemDto>, List<BaseItemDto>, List<BaseItemDto>, List<BaseItemDto>)> getArtistItems(
```
Komodo5197, line 24: I think it would make sense to split this up into 4 separate providers because the returned values seem independent.

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -238,21 +302,31 @@ class _ArtistScreenContentState extends ConsumerState<ArtistScreenContent> {
...
+        final allTracks = ref.watch(
         getAllTracksProvider(widget.parent, widget.genreFilter).future,
       );
 
-    final isLoading = artistItems == null;
-    final topTracks = artistItems?.$1 ?? [];
-    final albumArtistAlbums = artistItems?.$2 ?? [];
-    final performingArtistAlbums = artistItems?.$3 ?? [];
-    final allPerformingArtistTracks = artistItems?.$4 ?? [];
+    final isLoading = [
```
Komodo5197, line 319: It might make more sense to not have a global isLoading variable, and for all the sections to just watch their own asyncValue.  Unless we intentionally want to wait for all the elements to load to decrease page reflows?  We should definitely split out the allPerformingArtistTracksAsync provider at a minimum.

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -21,7 +21,46 @@ import 'artist_screen_content_flexible_space_bar.dart';
...
+  // Get Items
+  if (isOffline) {
+    // In Offline Mode
+    // We already fetch all tracks for the playback, 
+    // and as in offline mode this is much faster, we can just
+    // sort and filter them.
+    final List<BaseItemDto> allArtistTracks = await ref.watch(
+      getAllTracksProvider(parent, genreFilter).future,
+    );
+    var items = sortItems(allArtistTracks, SortBy.playCount, SortOrder.descending);
```
Komodo5197, line 41: I'm still not really a fan of showing these in offline mode.  The play count will always be what it was when the item was downloaded and will never update.  So the play count could easily have more to do with download timing then the actual play values, and thus could even be negatively correlated with the true value if you download the best tracks first and then grab everything else later.  I also feel like showing this section implies that our data is more accurate then it really is and could be perceived as buggy.

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -54,50 +120,52 @@ Future<(List<BaseItemDto>, List<BaseItemDto>, List<BaseItemDto>, List<BaseItemDt
...
+  final jellyfinApiHelper = GetIt.instance<JellyfinApiHelper>();
+  final downloadsService = GetIt.instance<DownloadsService>();
+  final bool isOffline = ref.watch(finampSettingsProvider.isOffline);
+
+  // Get Items
+  if (isOffline) {
+    // In Offline Mode
+    // In Offline Mode, we are getting the Performaning Artist Tracks
+    // directly in the other provider
+    return <BaseItemDto>[];
```
Komodo5197, line 154: I'd prefer returning null or throwing to be more explicit.  Although I don't see why we don't do the work here?  Also, I noticed the sortedTracksIncludingAppearsOn variable in the other provider isn't used.

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -206,11 +206,37 @@ class _GenreScreenContentState extends ConsumerState<GenreScreenContent> {
   }
 
   void openSeeAll(TabContentType tabContentType) {
+    bool isFavoriteOverride = false;
+    SortBy? sortByOverride;
+    SortOrder? sortOrderOverride;
+
+    if (ref.watch(finampSettingsProvider.genreListsInheritSorting)) {
```
Komodo5197, line 213: openSeeAll seems to be an onTap handler, so I think these should be ref.read.

lib/screens/music_screen.dart
```diff
@@ -121,7 +137,7 @@ void _buildTabController() {
         onPressed: () async {
           try {
             await _audioServiceHelper.shuffleAll(
-                ref.watch(finampSettingsProvider.onlyShowFavourites));
```
Komodo5197, line 107: I think this should be ref.read.

lib/screens/music_screen.dart
```diff
@@ -393,11 +376,25 @@ void _buildTabController() {
...
                     IconButton(
-                      icon: ref.watch(finampSettingsProvider.onlyShowFavourites)
+                      icon: (ref.watch(finampSettingsProvider.onlyShowFavourites) || isFavoriteOverride)
                           ? const Icon(Icons.favorite)
                           : const Icon(Icons.favorite_outline),
-                      onPressed: () => FinampSetters.setOnlyShowFavourites(!ref
-                          .watch(finampSettingsProvider.onlyShowFavourites)),
+                      onPressed: () {
+                        if (isFavoriteOverride) {
+                          if (ref.watch(finampSettingsProvider.onlyShowFavourites)) {
```
Komodo5197, line 384: All the ref.watch in this onPressed should be ref.read too.

lib/components/ArtistScreen/artist_screen_content.dart
lukaslindnermusic: Yeah, I had it independently before, but it resulted in ugly page reflows, and especially when the server takes a little bit longer for the top tracks (which happened a lot... my server was busy with other tasks during my testing) and you are already browsing the albums, that becomes really annoying as the list jumps down during scroll.
It just looked way cleaner and smoother when there is just one loading animation. 

I do agree that we should exclude allPerformingTracks from the isLoading variable, though. That data is only relevant for the Download Sync Button. I will do that with the next commits 👍🏻

lib/components/ArtistScreen/artist_screen_content.dart
lukaslindnermusic: Hmm... I included it because I thought that offline PlayCounting and reporting it and syncing it when back online was planned. Aren't we syncing the downloads either way when back online? Isn't it possible to update the PlayCount during that task?
I see the concern, and of course thats not ideal. However, if the Offline PlayCount could at least be updated once in a while, I would assume that users are expecting that if you only listen offline for a longer period, you don't have the current server playCount anymore.
I think, the more consistent offline and online mode are, the better. And as I thought that implementing the syncing of the PlayCount can be done, I included it here already.
If that is not possible, we can alternatively show Favorite Tracks here in Offline Mode instead. However, then there is no consistency between the two modes again. But I would prefer having a Favorite Tracks section over having an entire section disappear entirely. It also could be a setting for Online Mode what to display (Most Played, Favorites).

@Chaphasilor what do you think?

lib/components/ArtistScreen/artist_screen_content.dart
lukaslindnermusic: It's still like that because it was like that in the previous version, that's the only reason 😊 
I will move the offline code for that to this provider as well and then watch it from the allTracks provider, similar to online mode.
It previously was just an effort to reduce initial loading, as the allTracks happened in the background but the UI depended on futures. But I guess the offline database querying happens so fast that this makes no difference (but improves consistency and clarity when moving the code) 👍🏻

lib/components/GenreScreen/genre_screen_content.dart
lukaslindnermusic: Ah yes, you explained when to use watch vs. read before, so I should've known. Sorry bout that, will edit it and all other mentioned ref.watchs 😊

lib/components/ArtistScreen/artist_screen_content.dart
Komodo5197: The syncing makes a lot of server requests and does a lot of reading and writing to the database, so we don't actually fully sync everything every time.  Specifically, we don't run the full sync step on tracks and we don't even fetch the children of albums.  Updating playcount would require requesting every single downloaded track from the server and going through every download, which I'd rather avoid.  This is all generally happening in the background while we try to load the music screen, and potentially are restoring and starting playback on a queue, so performance considerations are meaningful.  In a related performance note, I'm a bit hesitant about the idea of adding track-based downloading to genres, because those can easily get very large and often have a lot of overlap, and I'm less convinced of the utility of this than I am of the track-based artist downloads.

lukaslindnermusic: How about updating the counts of only the 5 tracks that get shown for an artist/genre once that screen is accessed in online mode?

lukaslindnermusic: And if you think track-based downloads for genres make the performance that much worse, then I think sticking with albums is okay. 
It's basically just a consideration for edge-cases, where there is a track with a genre, but its album does not have the genre assigned.

Chaphasilor: So yeah, keeping track of play counts in offline mode and pushing offline plays to the server is something I'd like to do at some point. But Jellyfin doesn't allow us to easily update the play count without side effects, so all would be going through the Playback Reporting plugin. That would make syncing much more efficient though, since we can query via SQL.  
But I don't see this happening anytime soon, especially not before I implement pushing offline plays to the plugin (without dealing with play counts).  
So if play counts are essentially unknown in offline mode, there is no sense in showing them. @lukaslindnermusic are you still using the chip selector to control which tracks shown in the section? If so, I think removing one of the chips in offline mode is absolutely fine, and offers a nice and natural fallback. You could even disable the chip and add a tooltip to make this more obvious.  
As for only updating the first 5 tracks, that seems odd and inconsistent.

@Komodo5197 am I correct to assume that the performance concerns you have about downloading genre-tracks are due to syncs having to sync each track individually, because they were explicitly downloaded? That might be an issue indeed.  
We probably could avoid downloading any tracks which are already incidental downloads, but the improvement would depend on the users' metadata I guess.  
I'm fine with just going by album to avoid overloading the syncing, although I would prefer downloads being as "thorough" as possible. What Lukas said is also how I feel, consistency between online and offline modes is definitely desirable. But tradeoffs are to be expected, *especially* if we want to avoid re-creating the entire Jellyfin database within Finamp.

lukaslindnermusic: > As for only updating the first 5 tracks, that seems odd and inconsistent.

why that? We essentially display only the 5 most played tracks. So once we get them in online mode, we already have the 5 most played tracks, so if we only update those 5 in the background, we have the must accurate data for offline.............. nevermind :P during writing this, I noticed that we might have downloaded albums only that are not containing top tracks and therefore won't update the correct information in that case :P 

I will disable the filter on the Genre Screen when offline and switch to Favorites in case Most Played was selected.

What about the Artist Screen? Should we fall back to Favorites there as well? or keep it as it currently is and just hide it?

lukaslindnermusic: Ok @Chaphasilor, I removed the Top Tracks form the ArtistScreen again. However, I've left in the code commented out (its just a few lines), so if we decide later to include it in some way, the current logic is already there as a starting point.

I also disabled the "Most Played" item curation type on the GenreScreen in Offline Mode, and added a fallback (that even is configurable in the Settings, because: Favorites would be the most obvious fallback choice, but if you're not using that feature a lot, you might rather want to see random picks instead of empty favorites lists).
In case "Most Played" was selected when switching to Offline Mode, the fallback will only be used until back online, or if you switch it manually.
Also added a tooltip to the disabled FilterChip.

lukaslindnermusic:
@Chaphasilor I don't know if we should link these issues:
- https://github.com/jmshrv/finamp/issues/254
We decided to do the play all / shuffle all buttons later when we're redesigning the queue system, so this might be done in a different PR?

- https://github.com/jmshrv/finamp/issues/182
As far as I can understand it, this refers to the Genres being present next to an album cover in the AlbumScreen.
That was not part of the GenreScreen redesign.
Should I try to add it, though? 
I think it technically would be easy, but one problem I guess is the available space, as we currently already have four rows for Artist, Track Count, Duration and ReleaseDate, and those 4 rows perfectly take the height of the album cover, so a fifth row for genres would destroy that.
Do you have any layout suggestion on how to do that?

Another thing, that I realized that really bugs me, is, that on the ArtistScreen, when you click on a genre in the top section next to the Artist Image, it always takes you only to the first genre in the list. 
Not quite sure why that is, probably because if the list grows too much, it gets shortened with an "..." at the end? 
Maybe we could do a horizontally scrollable list of all genres, similar to the ArtistChips on the PlayerScreen?

For those two things (GenreChips Redesign/Addition on both AlbumScreen and ArtistScreen), would you rather have that in this PR as well, or should we open a separate PR?

lib/components/ArtistScreen/artist_screen_content.dart
Chaphasilor: Okay that sounds good. For updating only the top tracks, I wasn't happy about having up-to-date info for some tracks, but not for all.  
On the artist screen things could work the same way as for genres. I'd try to go for consistency again, so show most played (otherwise @Tiefseetauchner will be sad/mad again), and then fall back to the same sort as on the genre screen (using the same setting?). You could include the same settings toggle widget on both settings screens.  
Or if you feel like that's better, create a separate setting. I don't think a setting is needed at all though, it's easy to switch between chips. But I also don't think it would hurt!

Chaphasilor:
Regarding the issue I linked, I knew we discussed it but I didn't remember if we decided to keep or ditch the play/shuffle buttons :)

I won't get around to a full review today, but I'll build and install this PR to test it tomorrow during the day :)

lib/models/finamp_models.dart
```diff
@@ -200,6 +200,12 @@ class DefaultSettings {
   static const preferHomeNetwork = false;
   static const homeNetworkAddress = "http://0.0.0.0:8096";
   static const autoReloadQueue = false;
+  static const genreCuratedItemSelectionType = GenreCuratedItemSelectionType.mostPlayed;
+  static const genreItemSectionsOrder = GenreItemSections.values;
+  static const genreFilterArtistScreens = true;
+  static const genreListsInheritSorting = false;
```
Chaphasilor, line 206: How important is this default value to you? As previously said, I expect the "See All" button to let me see all items within this section - naturally in the same order as the section.  
Seems like you disagree?

Chaphasilor:
I noticed the favorite filter (music screen tab view) doesn't seem to work for the genre tab. It also didn't work before, but this PR might be a good place to fix it?  
Unless Jellyfin doesn't support the filter when fetching genres?

lukaslindnermusic:
> I noticed the favorite filter (music screen tab view) doesn't seem to work for the genre tab. It also didn't work before, but this PR might be a good place to fix it? Unless Jellyfin doesn't support the filter when fetching genres?

Good catch. 
Actually, the filter is missing in our jellyfin_api.dart. I expect it does work, but we are just not using it in our backend for some reason (similar to the sortBy and sortOrder before).
I'm trying to fix that tomorrow.

lib/models/finamp_models.dart
lukaslindnermusic: Yes, I disagree :P 
I don't understand as "See All" as expansion of the short sections on the GenreScreen, but as a path to really see all items, in the order you would naturally browse on the MusicScreen.
But I can understand why you might see that differently. If this is an important point to you (or if other people see it differently as well), I can change the default to true here. :)

Chaphasilor: Hmm. I can see where you're coming from. Maybe we should have both? Especially for things like the "Favorites" section it really doesn't make sense to me that there would only be 5 "top" items. Favorites are unordered, and I can have more than 5. So especially there I would still want a button to see all favorite tracks by that artist.

But I definitely also makes sense to have a way to see all tracks of that genre!  
Maybe we can figure out a way to shift around the buttons to have one "global" see all, and a local one per section :)  
I can try to create some mockups with options for that.

lukaslindnermusic: You can already click the count sections on top, so if we change it so that the setting only affects the See All button, and then change the default value to true, would that be a good solution for now?

Doesn't make sense anyway that you can click on the total count of albums but then only get your favorites.

lukaslindnermusic: The counts always display the total counts, even if the favorite filter is selected, so I think this would be a good method to have both.

But maybe, we should add a visible box or something around those counts so that its more visible that you can click it?

lukaslindnermusic:
Some Last Fixes I'm working on:

- [ ]  Limit GenreScreen results to currently active library (check also other places in the app).
- [ ]  Change default setting for "See All" to inherit=true. Exclude the count-sections from that setting so that they always lead to the entire lists, and change the look so that it is more obvious that you can click them.
- [ ]  Changing FavoriteFilter, SortBy and SortOrder while those values are overwritten should not modify the MusicScreen but directly the overwrite-value, so that we have those settings completely separate from the main MusicScreen.
- [ ]  Inherited Random Sorting in Lists does not remove the Jump to letter thing yet

lib/services/jellyfin_api_helper.dart
```diff
@@ -272,14 +277,18 @@ class JellyfinApiHelper {
             isFavorite: isFavorite,
           );
         }
-      } else if (parentItem?.type == "MusicArtist") {
+      } else if (parentArtistItem?.type == "MusicArtist") {
         // For getting the children of artists, we need to use
         // artistIDs or albumArtistIds instead of parentId
+        // also, in order to only get the items from within one library
```
Komodo5197, line 283: This seems a bit confusing?  I'd make the additianl arg be called something like 'artistLibraryFilter' and then put the library in that and keep the artist in parentItem.  Then the actual api calls put them in the right slots.

lib/services/jellyfin_api_helper.dart
```diff
@@ -272,14 +277,18 @@ class JellyfinApiHelper {
             isFavorite: isFavorite,
           );
         }
-      } else if (parentItem?.type == "MusicArtist") {
+      } else if (parentArtistItem?.type == "MusicArtist") {
```
Komodo5197, line 280: Is this filtering artist specific, or does it also work for genres?

lib/services/jellyfin_api_helper.dart
lukaslindnermusic: I changed it because on the MusicScreen, the parentItem is used for the library.
And as in the api call for artist child items, we specifically reassign the parentItem to the artistIds instead of parentItem doing the actual api call, I decided to separate it for the artist. this seems more in-line with what the actual Jellyfin API gets.

lib/services/jellyfin_api_helper.dart
lukaslindnermusic: See my other comment :)

lukaslindnermusic: For genre childs, we don't need library, artist filter and genre filter, we only need library and genreFilter.
So parentItem is just used for the library, and the genreFilter is used to specify the genre.
We don't get the genre items by passing the genre as parent, but by passing it to the "genreIDs" field.

lukaslindnermusic: And the genre list in the musicscreen is a different endpoint and therefore just works already.

Chaphasilor (commented):
These are the things I've found until now.  
I'm not a fan of the section reload behavior, but I also couldn't think of a better solution :)

I'll respond to your comments in the morning

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -0,0 +1,457 @@
...
+          child: Row(
+            mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+            children: [
+              Expanded(
+                child: _buildCountColumn(
+                  count: albumCount,
+                  label: AppLocalizations.of(context)!.albums,
+                  onTap: () {
+                    openSeeAll(TabContentType.albums);
+                  },
```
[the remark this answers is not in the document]
lukaslindnermusic: Already changed during your review :)

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -0,0 +1,457 @@
...
+            case GenreItemSections.tracks:
+              return SliverPadding(
+                padding: const EdgeInsets.only(bottom: 12.0),
+                sliver: TracksSection(
+                  parent: widget.parent,
+                  tracks: tracks,
+                  childrenForQueue: Future.value(tracks),
+                  tracksText: genreCuratedItemSelectionType
+                      .toLocalisedSectionTitle(context, BaseItemDtoType.track),
+                  seeAllCallbackFunction: () => openSeeAll(TabContentType.tracks),
```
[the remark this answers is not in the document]
lukaslindnermusic: Already fixed during your review :) 
The last one was easily fixed by adding the overwrite values to the refresh hash :)

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -0,0 +1,457 @@
...
+        builder: (context) => MusicScreen(
+          genreFilter: widget.parent,
+          tabTypeFilter: tabContentType,
+          sortByOverrideInit: sortByOverride,
+          sortOrderOverrideInit: sortOrderOverride,
+          isFavoriteOverrideInit: isFavoriteOverride,
+        ),
+      ),
+    );
+  }
```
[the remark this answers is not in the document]
lukaslindnermusic: With SideDrawer, you mean the left menu with Logs, Settings, etc.?
Is it active on your device? I can't open it on both the iOS Simulator and my iPhone.

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -0,0 +1,457 @@
...
+                        ];
+                      }).toList(),
+                    ),
+                  ),
+                ),
+              ),
+            );
+          },
+        ),
+      ),
```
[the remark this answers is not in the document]
lukaslindnermusic: > So you still have this as a global selector. 
Yes because I thought it looked way cleaner with it.

> I'm not sure if that's a good idea, since not all curated categories apply to all item types. For example, we don't have a `playCount` for anything but tracks.
You are right, that doesn't make much sense. I thought maybe the server is counting also for the albums and artists, which theoretically would be possible. And honestly, the items that showed up for me seemed very reasonable, it always was music where I know I've listened to it a lot. But it definitely is 0, I checked it. Maybe I am listening to much to music so that I get that correct feeling for everything that might show up there 😅

> It also means that users have to scroll up, change the selection, and then scroll back down to i.e. see their favorite artists for that genre. The amount by which they need tl scroll might also vary between up and down, since if there are no favorite albums, that section will be much shorter, affecting the scroll distance.
That's a good point.

> I believe this would be better handled by moving the curated category selector within each section, below the section heading. This might even help with the "See all" discussion, because now we can call the tracks section just "Tracks", have the "See all" button open all genre tracks, and then use the chips to change the first 5 items, possibly with a way to "See more" at the end of the preview list. And we could customize the chips based on the item type which they are ordering/filtering.
> What do you think?

Yeah, you convinced me :) 
I'm just wondering: I already suggested somehwere else, that we could use the Counts at the top as way to open the not-inherited full list. If we keep the See All for the filtered, this might become unintuitive if the filterchips are actually below that, right? thats why you're suggesting to add something like a bottom row with "See more"?

Chaphasilor:
I'm really happy about d0d89ca5ef964de7a79b0056f0f300151b448933, that has been a long standing issue/inconsistency, many thanks for handling that!  
Now that you have some experience with it I just want to make sure I'm making correct assumptions: the Jellyfin API doesn't allows us to easily include results from multiple, *specific* libraries in one request, right? So we can't have the library picker (in the drawer) be a multi-select to compose content from multiple libraries into one?  

Because some people have requested that in the past, and I would be a fan of it myself (including Christmas tracks at the end of the year, and then excluding them again, for example). But afaik Jellyfin always expects a single parentID, and has no separate filter for libraries.

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
Chaphasilor: I wish Jellyfin would track plays of albums, artists, and playlists. But no luck. It doesn't mark them as played automatically, and it also doesn't support marking them as played manually (I tried).  

(But I just found out we can now mark items as played at an earlier date, which is nice. So we *could* properly register offline plays, although it requires one request per play. And it doesn't end up in Playback Reporting, so we'd also need to manually submit that. But still, the possibility is there!)

> I'm just wondering: I already suggested somehwere else, that we could use the Counts at the top as way to open the not-inherited full list. If we keep the See All for the filtered, this might become unintuitive if the filterchips are actually below that, right? thats why you're suggesting to add something like a bottom row with "See more"?
 
I agree, it could work well to use the top buttons for the unsorted lists, and See all for the sorted ones. And I don't think it would be too confusing, as long as we keep renaming the section title to reflect the current curation type. So don't just call it tracks, call it "Recently Added Tracks", just like you were doing before :)  
We *could* also move the see all in the row with the filter chips, but I don't think that's much better than having it above the chips :)

And no, you can't listen to too much music :D

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
Chaphasilor: Yeah, take a look:  
https://github.com/user-attachments/assets/cfa999e0-005b-4f35-8b7c-b3dc7575e4ee

It's a bit tricky to open without triggering the back gesture, but it's there.  
And yesterday I did notice some "glitching" while going back, which is why I noted it here.

This is not on the latest commit though, but still on fe5aba0

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
lukaslindnermusic: Alright, I will implement it that way :)

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
Chaphasilor: Yep, still happening, here's a video of the back glitch, where it opens and closes the drawer instead of going back:

https://github.com/user-attachments/assets/f7e9fb8d-f144-4464-8c81-9ae8ada0c3c4

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
lukaslindnermusic: Just to verify then. We would have:
- Tracks: Most Played (default), Favorites (offline fallback default), Random, Latest Release, Recently Added
- Albums: Favorites, Random, Latest Release (default), Recently Added
- Artists: Favorites, Random, Recently Added (default)

(I also thought that Artists might inherit the Latest Release from the album with the latest release date assigned to them.... but no 🙄)

Or would you use other defaults?

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
lukaslindnermusic: Okay... seems like thats an Android problem.
I will take a look into if you can conditionally disable the side drawer altogether.
But if not, would you say its okay that we live with it until we switch to your ShowAllScreen lists from the HomeScreen PR?

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -0,0 +1,530 @@
...
+                ),
+              ),
+              Expanded(
+                child: Padding(
+                  padding: const EdgeInsets.only(left: 4),
+                  child: _buildCountColumn(
+                    count: artistCount,
+                    label: (ref.read(finampSettingsProvider.artistListType) == ArtistType.albumartist)
+                        ? AppLocalizations.of(context)!.albumArtists
+                        : AppLocalizations.of(context)!.performingArtists,
```
Chaphasilor, line 364: This changes depending on which type is selected on the artist tab. But the count doesn't actually change.  
Maybe we should have two separate buttons, or combine both counts into one?

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
Chaphasilor: I'm not sure if Recently Added for artists makes all that much sense. I'd go with favorite artists instead.  
But it doesn't matter too much :)

But maybe change the chip order so the default is always first, and that chips for different types still have the same order (but missing a few options). I think that should work for the three cases here, if you go with most played, recently added, favorites, and then the rest.

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
Chaphasilor: Yeah, sure :)

lukaslindnermusic:
> I'm really happy about [d0d89ca](https://github.com/jmshrv/finamp/commit/d0d89ca5ef964de7a79b0056f0f300151b448933), that has been a long standing issue/inconsistency, many thanks for handling that! 

No problem. There still is a glitch that we are discussing in the discord thread. As the download system still fetches every library, in offline mode you still would get all items because the children inherit the library view without checking again and then it looks to the system as if the album is twice in one library. But we can fix that. However, there is a different discussion in general: If downloading a genre or an artist should download the genre/artist on every library or just the currently selected one. 
If all, we have to fix the getStatus to only include the current library, otherwise we get the yellow sync buttons again as the items dont match up if there are items in a different lib for a genre/artist.
If only the current one, Komodo wrote out a few steps that I did not remember yet, but I'm sure they know what to do :P 

> Now that you have some experience with it I just want to make sure I'm making correct assumptions: the Jellyfin API doesn't allows us to easily include results from multiple, _specific_ libraries in one request, right? So we can't have the library picker (in the drawer) be a multi-select to compose content from multiple libraries into one?
> 
> Because some people have requested that in the past, and I would be a fan of it myself (including Christmas tracks at the end of the year, and then excluding them again, for example). But afaik Jellyfin always expects a single parentID, and has no separate filter for libraries.

As far as I understand it, thats correct.
You filter by library by using the parentItem (in the API, not in our helper function). As far as I know, you can't pass multiple parentIDs at once, so its either one parent library or none at all (aka. include all libraries). I could not find a separate parameter where you could specifiy libraries separately from the parentId.
So if you want to only combine Library A and B, but not C, you would have to make double requests (Requests x number of selected libraries, unless all libraries selected). That sounds terrible lol :P 

Workaround for your christmas case, in case you didn't thought about that yet:
You can give Tags to items in Jellyfin Web. Just give every christmas album a specific tag, and then you can use the Parental Control to hide everything with a specific tag. Then, when christmas time is back, you can manually remove the filter again and so on. (I do that to hide work-in-progress film-versions (that I only want to watch occasionally but they should not clutter my list) from my film library.)

Chaphasilor (commented):
reviewing on the GitHub mobile app works terrible 😅  
This is just so you can actually see my comments.

lib/components/GenreScreen/genre_screen_content.dart
lukaslindnermusic: Really? For me, the count changes. Do you have a different number of album and performing artists on that genre? Often they are identical.


https://github.com/user-attachments/assets/10b95629-c984-47b0-8291-08e0847d4198

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
lukaslindnermusic: @Chaphasilor Regarding the order: I chose it intentionally because logically similar options are grouped together.
Most Played and Favorites both show stuff that you (probably) like a lot.
Latest Release and Recently Added are both time-based.
and Random... well... is random.
So the order 
Most Played - Latest Releases - Favorites - Random - Recently Added
or
Most Played - Latest Releases - Favorites - Recently Added - Random
will make it so that the default option for each section always will be the first one, but it will also destroy that logical grouping. 
With that context, would you still prefer one of the two new solutions (and which one) or would you trade the default first option for the logical grouping?

lib/components/GenreScreen/genre_screen_content.dart
Chaphasilor: Ah that might be it. The code also looked like it didn't handle the types differently, but apparently I didn't look close enough...

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
Chaphasilor: I'm aware that you but it into a logical order/grouping ^^  

If the chips are reset each time the user opens a genre, I still think the first chip should always be selected :)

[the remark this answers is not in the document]
lukaslindnermusic: > If the chips are reset each time the user opens a genre, I still think the first chip should always be selected :)

Oh, I think it definitely should stay remembered :) 
Alright then, then I would suggest we use 
`Most Played - Latest Releases - Favorites - Random - Recently Added`

The only thing I am not quite sure is if it is okay to store the current setting in an array. It would be nice if only the affected item type would reload after its filterchip changes, not all. 
But I'm not sure how providers are treating arrays when they change but you are watching a specific value in that array.
I'll give that a try, and if all providers reload after one value in the array changes, I will go with three separate variables.

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
lukaslindnermusic: That was fairly easy, I think :P It was literally right there haha 😅

lib/components/GenreScreen/genre_screen_content.dart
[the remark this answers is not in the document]
Komodo5197: Using select() on a provider filters rebuilds to only happen when oldValue !=newValue.  For Lists/Maps, equals() is identical(), which means if you update a child element instead of replacing the list with a different ones you might filter out a rebuild you want.  But you should be selecting specific entries any way in this case, and we generally don't modify Lists like that, only maps.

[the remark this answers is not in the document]
lukaslindnermusic: yeah, ChatGPT also suggested using .select(), but I think I will just go with separate variables. I mean... it doesn't cost money to have three variables instead of one :P

Chaphasilor:
Sorry about the conflicts, but I'm positive that you can handle them easily :)  
Is your ToDo list further up still up-to-date? Do you need help with anything?

lukaslindnermusic:
> Sorry about the conflicts, but I'm positive that you can handle them easily :)

No worries. I hope I merged it correctly.

> Is your ToDo list further up still up-to-date? Do you need help with anything?

I implemented the filterchips for every individual section (for now I kept the logical grouping as settings will be remembered). and I probably fixed the android sideDrawer thing. If you want to give it another go and the drawer bug is gone, I can also check this ToDo-point.

If you don't have anything else that needs change, then the only thing left before I think we are ready to merge is to at least fix the getStatus in the download service, so that we won't get the yellow sync icon for downloaded artists/genres that also have items in other libraries. 

If @Komodo5197 can guide me through the actual download changes (so that we actually only download from the current library), then I guess I can do that as well. But I think it will be faster if Komodo does it afterwards because they no best what to do.

Komodo5197:
The merge looks good to me.  The download changes should be reasonably straightforward:
1. Add a new type to the FinampCollectionType enum, ArtistWithLibrary or something.  Is this also being used for library-specific genre downloads?  Becuase I guess the name should be more generic.
2. Add a new field final BaseItemDto? item to FinampCollection.  Add the new type to the name and id functions.  Modify the asserts on the constructer to require both args on the new type, and enforce item=null othherwise.
3. Add a new case to _getFinampCollectionChildren which fetches the library-filtered albums and tracks.
4. Anywhere we produce a download stub from the artist item, produce it from a Finampcollection instead.  I think this includes the download button on the screen, all the context menus for the artists, and the stateProvider calls?  You might be able to just look at every call to DownloadStub.fromItem, this is probably like half of them.

Komodo5197:
Actually, the userDownloadedItemsProvider may need modification so that the new collection type shows up under artists/genres as desired, instead of in the special downloads category.  But we need to make sure we don't accidentally hide the cross-album downloads.  That might get a bit messy.

lukaslindnermusic:
@Chaphasilor you can also link https://github.com/jmshrv/finamp/issues/1187 to this PR. 
The issue will be fixed once this PR gets merged.

lib/services/downloads_service_backend.dart
```diff
@@ -1458,6 +1460,37 @@ class DownloadsSyncService {
               []);
           outputItems.removeWhere((element) => element.imageId == null);
           typeOverride = DownloadItemType.image;
+        case FinampCollectionType.collectionWithLibraryFilter:
+          outputItems = await _jellyfinApiData.getItems(
+                parentItem: (parent.baseItemType == BaseItemDtoType.genre)
```
Komodo5197, line 1465: parent.baseItemType is not set because the Downloadstub iteslf does not have a baseItemDto, it has a FinampCollection.  You need to use BaseItemDtoType.form(collection.item).

lib/services/downloads_service_backend.dart
```diff
@@ -1458,6 +1460,37 @@ class DownloadsSyncService {
...
+          // but they will get filtered out later
+          if (parent.baseItemType == BaseItemDtoType.artist) {
+                outputItems.addAll(await _jellyfinApiData.getItems(
+                  parentItem: item,
+                  libraryFilter: collection.library!,
+                  includeItemTypes: BaseItemDtoType.track.idString,
+                  filters: "Artist=${parent.name}",
+                  artistType: ArtistType.artist,
+                  fields:
+                      "${_jellyfinApiData.defaultFields},MediaSources,MediaStreams,SortName"
```
Komodo5197, line 1490: You can just use the fields variable here.

lib/services/downloads_service_backend.dart
```diff
@@ -1458,6 +1460,37 @@ class DownloadsSyncService {
...
           typeOverride = DownloadItemType.image;
+        case FinampCollectionType.collectionWithLibraryFilter:
+          outputItems = await _jellyfinApiData.getItems(
+                parentItem: (parent.baseItemType == BaseItemDtoType.genre)
+                    ? collection.library!
+                    : item,
+                libraryFilter: (parent.baseItemType == BaseItemDtoType.artist)
+                    ? collection.library!
+                    : null,
+                genreFilter: (parent.baseItemType == BaseItemDtoType.genre)
```
Komodo5197, line 1471: I believe using the genre as the parentItem and then the library filter should also work with how we have getItems set up now, and that slightly simplifies this code.

lukaslindnermusic: > I believe using the genre as the parentItem and then the library filter should also work with how we have getItems set up now, and that slightly simplifies this code.

No it does not work how it is currently set up, because the libraryFilter is missing in the `} else if (parentItem?.type == "MusicGenre") {` section of the api helper method, but we can add it.

However, on the genreScreen, we are not using this section of the helper function at all, because we have unified the query to avoid code deduplication and have a cleaner code, and we are also fetching artists with the same method and query, and for artists we have to use the different endpoint and to filter them by genre, I introduced the genreFilter argument... it can be used with tracks, artists and albums.

While it might work using libraryFilter: library and parentId: genreid in this case (because we are only fetching album-children and not artists), I think it's better to be consistent with how the genreScreen fetches.

I'm not quite sure, but I think, due to the introduction of the genreFilter the `} else if (parentItem?.type == "MusicGenre") {` part of the api helper is not necessary anymore?

lukaslindnermusic:
@Komodo5197 thanks for your help 😊 
Now individual downloads are separate and work. 
However, I noticed two things that are not yet optimal:

a) For me, it takes around 1-5 seconds for the download to visually start. Previously, that was way faster. Is there a way to improve that or is it the trade-off for the bugfix?

b) the SyncButton for the Artists is showing outdated again. the Genre SyncButton is working (Don't be confused, I've switched the DownloadButton for genre to use finampcollection directly after the last commit, so I did not test the wrong thing).
So it seems like there is some mismatch here again.

lukaslindnermusic:
> b) the SyncButton for the Artists is showing outdated again. the Genre SyncButton is working (Don't be confused, I've switched the DownloadButton for genre to use finampcollection directly after the last commit, so I did not test the wrong thing). So it seems like there is some mismatch here again.

Ahhhh my bad, I forgot to change one `parent.baseItemType` to use `BaseItemDtoType.fromItem(collection.item!);` instead, which caused performing tracks to not get downloaded.
Now it works!

Komodo5197:
What exactly do you mean by the download visually starting?  Random guess, does changing the viewId =null line for FinampCollections at the top of _syncDownload to not apply to the new collection type do anything?

lukaslindnermusic:
> What exactly do you mean by the download visually starting? 

https://github.com/user-attachments/assets/67b6189d-0e13-4370-bcc2-1a4176a8f524


> Random guess, does changing the viewId =null line for FinampCollections at the top of _syncDownload to not apply to the new collection type do anything?

You mean, doing something like that?
```      if (parent.finampCollection!.type != FinampCollectionType.collectionWithLibraryFilter) {
        viewId = null;
      }```

**Comment by lukaslindnermusic** [comment]
> Random guess, does changing the viewId =null line for FinampCollections at the top of _syncDownload to not apply to the new collection type do anything?

Uhh yes, that makes it much faster. Nice!

Now whats still left is to change every download button and then fix the DownloadsScreen to display Artists and Genres in their categories.

How would you suggest applying the DownloadScreen fix?
Also, would you add some indication to the legacy downloads for now (like i.e. appending "(Legacy Download)")?

**Comment by Komodo5197** [comment]
Maybe add an (All libraries) on the end?  It's probably fine to just leave it though.  For the downloads screen, the key thing is to modify userDownloadedItemsProvider but the way it's set up doesn't match that well with the new requirements.  First of all, we need to filter by finampCollection type which IO don't believe isar has access to, so you may need to add a .map() call to the output stream that filters the list by finampcollectiontype as needed.  And then there's the fact that we need to show both the artist items and the new collections under the same artist section.  So it might make sense to rip out the fancy logic going on where we filter on attributes in the enum, and just have a big switch case in the provider with all the filters baked into the individual cases.  I'm not really sure what method is best, feel free to do whatever.

**Comment by lukaslindnermusic** [comment]
With help from ChatGPT, I could figure it out... but please double-check it.

Two problems still exist now.
a) We don't have images for the new FinampCollection items.
b) Legacy Downloads keep showing "Syncing..." forever. 
Simulator Screenshot - iPhone 15 Pro - 2025-05-10 at 01 14 33

File: lib/services/downloads_service_backend.dart [hunk 26]
```diff
@@ -888,22 +888,35 @@ class DownloadsSyncService {
       Set<int> requireCompleted,
       Set<int> infoCompleted,
       BaseItemId? viewId) async {
-    if (parent.type == DownloadItemType.image ||
```
Komodo5197, line 891: Re-add this block setting asRequired.

lib/services/downloads_service.dart
```diff
@@ -154,21 +154,51 @@ class DownloadsService {
...
+              item.finampCollection?.item?.type == BaseItemDtoType.artist.idString);
+        }).toList();
+
+      case DownloadsScreenCategory.genres:
+        return allItems.where((item) {
+          return (item.type == category.type &&
+              (category.baseItemType == null ||
+              item.baseItemType == category.baseItemType)) ||
+              (item.finampCollection?.type == FinampCollectionType.collectionWithLibraryFilter &&
+              item.finampCollection?.item?.type == BaseItemDtoType.genre.idString);
```
Komodo5197, line 190: I'd prefer checking BaseItemDtoType.from(item.finampCollection!.item!)==BaseItemDtoType.genre here.

lukaslindnermusic:
- When you hit Repair, the sizes disappear and the repair seems to freeze.
- Probably unrelated to our current changes, but the Artists show all albums but also all tracks, because we are having them as separate children for the performing artist tracks. I guess we have to filter the tracks of each artist to exclude the ones that have an album that is also downloaded for the artist on the `downloaded_items_list.dart`?

lib/services/downloads_service_backend.dart
lukaslindnermusic: Ok, the hanging "Syncing..." is fixed, but there still are now images on the DownloadScreen for the new FinampCollections.

Komodo5197:
For the images, you can just modify DownloadedItemsList to pull the item from the finampCollection instead of stub.baseItem when showing these collections.  If you want to hide tracks that are inside albums on artists, you will have to do additional filtering, and it probably does make sense to do this on the UI side  The syncing is coming from the fact that the loss of the asRequired section prevents images from being downloaded.  Is the  repair still behaving oddly once that gets fixed?

lukaslindnermusic:
> Is the repair still behaving oddly once that gets fixed?

No, the Repair finishes. Near the end, the nodes syncing count seems stuck for a few times, then it goes up and down again, and eventually it finishes. (I'm not sure if this is normal?)

Komodo5197:
The nodes syncing count fluctuating a bit is normal, because some nodes are things like albums or artists that add a bunch more nodes to the sync when processed.  It's probably normal?  Anyway, looking at the downloads screen I noticed that the new downloads don't specify which library they are associated with.  We probably should do that in some way, because the fact that these are library specific is something the user should probably know, and having exact duplicates if you download across multiple libraries seems weird.

lukaslindnermusic:
> Anyway, looking at the downloads screen I noticed that the new downloads don't specify which library they are associated with. We probably should do that in some way, because the fact that these are library specific is something the user should probably know, and having exact duplicates if you download across multiple libraries seems weird.

Yeah thats a good point. I noticed that yesterday night as well while testing.
I guess, for the new artist and genre downloads, that's fairly easy because we have the info in the finampCollection?
Just load the library Id and then get the library name, and maybe add it right to the right of the size?
the "all Libraries" string I added could go there, too. and as the download is "orphaned" and cannot be deleted from other places in the app anymore, I think its a good idea to give that string some yellow color or something?

However, the same could be true for albums and tracks as well. The problem here is that - instead of the artist and genre -  they have different ids for each library. I don't know if the viewId gets stored alongside those downloads? I don't think so, so for identical albums and tracks in multiple libraries, we can't add the library name. 

By the way, I just took a look at jellyfin web, and they also have the bug where there are items from all libraries on artist pages. 😂 Honestly, Jellyfin... why not a simple "libraryId" datapoint for each api request? That would make things so much easier :P

lukaslindnermusic:
> However, the same could be true for albums and tracks as well. The problem here is that - instead of the artist and genre - they have different ids for each library. I don't know if the viewId gets stored alongside those downloads? I don't think so, so for identical albums and tracks in multiple libraries, we can't add the library name.

But of course, that is a very special edge case. Why would someone have the same album in two locations in the first place? Artist and Genre, yes, I can see use-cases. But the identical album or individual tracks downloaded twice is really really an edge case, so I think its fine that we don't have the library name for those.

Chaphasilor (commented):
1. Got a widget error + 401 error snackbar on login (password). Not sure if this PR is causing it.
2. Can we dynamically set the artist filter type to "recently added" or even random if there are no favorites? Showing an empty list by default seems not ideal
3. The sorting for the See All/Music screen isn't *exactly* the same as the preview section. For "Most played", the order between tracks with equal play count can be different between the two lists.
4. We're still only showing album artists or performing artists at the top, and change it depending on what was last active. That is odd imo. We should show both, either as two sections, or as two counts within one section. Or just count distinct artists from both types (either a or b or both) and just call it "artists"? Then the user can switch types after clicking on it. And when we download a genre, I'm guessing we also download both album and performing artists too?
5. Drawer issue seems fixed
6. Wasn't there a filter option for the artists top tracks section, that allowed choosing between different sortings? I don't see that...

Everything else works well. Haven't tested downloading yet, but I'll start a few over night.  
Performance is much better too, even genres with 2.5k+ albums load very quickly, great!

lukaslindnermusic:
> 1. Got a widget error + 401 error snackbar on login (password). Not sure if this PR is causing it.

I did not change anything for the login. 🤔

> 2. Can we dynamically set the artist filter type to "recently added" or even random if there are no favorites? Showing an empty list by default seems not ideal

That's a great idea. 😊  Would you say this should be a separate fallback setting, or would you just decide for a default.
The reason why I'm voting for a setting is that then the user knows why that is happening and can decide how. I think, it can feel buggy if the user decides to always show Favorites and then suddenly it shows something else.

> 3. The sorting for the See All/Music screen isn't _exactly_ the same as the preview section. For "Most played", the order between tracks with equal play count can be different between the two lists.

Yes, that's intentional because actually we don't have sorting for MostPlayed on the MusicScreen. It's just an alphabetic list in that case, so no sorting inheritance at all :P. I know that that's not ideal, but adding Most Played as sorting to the MusicScreen was something that I ruled out early in the process of this PR because it seems not related to the direct redesign of the genreScreen itself, and actually I wanted to keep this PR focused on the GenreScreen (which turned out to not be the case now as a lot of different stuff was handled as well already :P) 
Should I add MostPlayed as a sorting option to the MusicScreen in this PR or would you accept it if I do this in a later, separate PR?

> 4. We're still only showing album artists or performing artists at the top, and change it depending on what was last active. That is odd imo. We should show both, either as two sections, or as two counts within one section. Or just count distinct artists from both types (either a or b or both) and just call it "artists"? Then the user can switch types after clicking on it. And when we download a genre, I'm guessing we also download both album and performing artists too?

Okay, multiple things here.
First: We are not downloading artists at all when downloading a genre. An artist can get new albums that are NOT assigned to that genre at any time, and can also have multiple albums with completely different genres... so that is not an option. (But yes, the info for both album artists and performing artists are downloaded for the tracks and albums that have the genre assigned.)
You can change the type of artist by clicking on the count and then switching it in the list.
I think it is very intuitive that you can decide app-wide if you want to see album artists or performing artists, and that you can switch between them at the spots where you can see artist lists. I belive most people either default to one or the other and mostly ignore the other one, so having both counts besides each other I think is unnecessary - it clutters the available space and also we would need an additional request, but the way how we get that count also doesn't even allow us to get a combined distinct count for both types - Because we are getting that count via totalRecordCount from the api, while we are loading only the actual 5 items that get displayed - that's a reason why the performance is so greatly improved - we only load whats really really necessary. 
One thing I could maybe add is some sort of toggle functionality, somewhere near the artist count, where you can switch without having to enter the list. Or maybe an icon next to the favorite icon on the top right?

> 5. Drawer issue seems fixed

Nice 😊 

> 6. Wasn't there a filter option for the artists top tracks section, that allowed choosing between different sortings? I don't see that...
>    Related: There is no "See All" button on the artist screen. Perfect opportunity for a follow-up PR? :D

It's actually in the Artist Screen Settings. Have you checked them out? 😊 
I just thought this might be cleaner, as I see a difference between the section on the artistscreen and the sections on the GenreScreen. On the ArtistScreen, it is just a little "appetizer": "Here are 5 tracks you might like, and if you want to browse the artist, down there are all albums." 
We hide the section on the ArtistScreen completely if there are no items to show, which I think is really clean and nice and mirrors the behavior for the Albums and Appears On sections. Hiding the section if empty would not be possible if we would have another filter row directly there. 

In the Artist Page settings, there is also more functionality for filtering the artist page by genre tap instead of opening the genre.
Also, there are a few nice settings for the GenreScreen, like reordering the item sections.

Regarding the See All button on the ArtistScreen: that would require us to add an Artist filter to the MusicScreen list. and that is more painful than one might think at first glance, because it would involve introducing a second query for performing artists, and then filtering distinct tracks... and that all would basically make the pagination that we have impossible, because we can't predict how many artists will be the same in both queries.
All of this greatly exceeds the scope of the GenreScreen redesign, and there already are a lot of things I tackled in this PR besides the main goal, so I definitely think, this should be a separate PR (if at all, given the limitations).

> Everything else works well. Haven't tested downloading yet, but I'll start a few over night. Performance is much better too, even genres with 2.5k+ albums load very quickly, great!

Thanks 😊

lukaslindnermusic:
> 1. Got a widget error + 401 error snackbar on login (password). Not sure if this PR is causing it.

I am having login difficulties sometimes with the current redesign branch as well, so I think the cause has to be something else and it is not related to this PR.

lukaslindnermusic:
> Yes, that's intentional because actually we don't have sorting for MostPlayed on the MusicScreen. It's just an alphabetic list in that case, so no sorting inheritance at all :P. I know that that's not ideal, but adding Most Played as sorting to the MusicScreen was something that I ruled out early in the process of this PR because it seems not related to the direct redesign of the genreScreen itself, and actually I wanted to keep this PR focused on the GenreScreen (which turned out to not be the case now as a lot of different stuff was handled as well already :P)
> Should I add MostPlayed as a sorting option to the MusicScreen in this PR or would you accept it if I do this in a later, separate PR?

Looool sorry, I take it back. 
When I was looking for PlayCount, I was also under the impression that albums and artists have a PlayCount as well, but they don't. For tracks, there is a PlayCount, and it works great. Inheritance for PlayCount is back, sorry bout that :P 

Also, I have a great idea for switching between Album Artists and Performing Artists, directly on the Genre Screen.
Already onto it.

lukaslindnermusic:
> Also, I have a great idea for switching between Album Artists and Performing Artists, directly on the Genre Screen.
> Already onto it.

What I initially intended seems odd from a layout harmony perspective:

https://github.com/user-attachments/assets/5e644bab-f7a4-4dd1-a123-221c1bcac7a4

A simple Button at the top right would look like this:

https://github.com/user-attachments/assets/0da00b7c-556c-42f9-a5b3-ed52f12cdfaa

lukaslindnermusic:
> Related: There is no "See All" button on the artist screen. Perfect opportunity for a follow-up PR? :D

We are currently loading all tracks for the playback on the ArtistScreen, though. It is loaded delayed, but we are loading it. We *could* generate a list from ths api call once its finished, but the loading of allTracks currently is a performance killer for artists with a lot of tracks. Once we re-work the queue system, I think we should remove that api call to improve performance.
However, we can re-use it if we still want a See All and just load it on demand. However, as previously mentioned, pagination will be hard with that.

Chaphasilor:
7. The "fully downloaded" filter doesn't seem to work on the genre tab (or does it require all downloads to be completed?)

Chaphasilor:
Re 2: Maybe we could simply disable the favorites tab in that case? So it naturally falls back onto the next curation option, but will reset ti favorites if favorites are available. And a swipe-to-refresh on the genre screen should re-run the check, in case the user just added a favorite.

Re 3: There is a "Play Count" sort order on the music screen tracks tab, is that broken?

Chaphasilor:
The new switching button really doesn't fit, especially not without proper labels. But since it's a global setting it's fine, it's just my uncommon usage that's causing issues...

lukaslindnermusic:
First, regarding the See All on Artist Screens:

> I also don't understand the issue with artist see all? Aren't we talking about showing all tracks or all albums from a _specific_ artist? That doesn't require any fancy requests, does it?

Yes, we are talking about a specific artist. Albums are easy, you can re-use the MusicScreen (or later after the HomeScreen PR the SeeAll-Screen?) for that. just filter the albums by albumArtist. But for the Appears On section, you can't just filter the albums by performing artist, because then you would also include albums where the artist is albumartist AND performing artist, but those are supposed to be under "albums". The logic is to both fetch albumartist-albums and albums where the user is an performing artist on. We have to filter the list of albums-as-performing-artist to exclude the items on the album-artist-albums, so that we only show the albums where the artist is ONLY a performing artist but not an album artist.
This is what currently happens on the ArtistScreen. With albums, that's reasonably fast (unless you have 200+ albums I guess).

With tracks, you basically have to do the same if you want to include all
a) tracks where the artist is albumartist AND performingartist
b) tracks where the artist is albumartist and NOT performingartist
c) tracks where the artist is NOT albumartist BUT performingartist (these would be the tracks that are in the Appears On Albums).

This requires two queries and then the distinct results, and it is currently already happening on the ArtistScreen (only for the Playback when you hit the buttons and for the trackcount, which we actually could get much faster with the totalRecordCount column from toptracks instead 🙈), but in the Background. The trackcount is generated with that, and you can see that it shows up 1-3 seconds later, after everything else is already there.

Pagination with these requests CAN be done, by limiting each request to lets say 50, which results in 100 items fetched for each "page". But what you can't predict is how many matching items will be in both queries. So maybe, in the first 100 results, you will get 40 matching pairs between both queries, and therefore only get 60 items in total... then with the next 100, you get only 10 matching items that you filter distinct, so you get 90 more items with that. 
that's what I meant previously... you will get uneven pagination.

I agree, that it would be really cool, though... I will tackle this in a follow-up PR if you want 😊

lukaslindnermusic:
Now for the GenreScreen and what's still to do in this PR 😊 


> 7\. The "fully downloaded" filter doesn't seem to work on the genre tab (or does it require all downloads to be completed?)

Huh? yeah it also fails on the Artist Tab. Probably related to the download system changes, I will look into that.





> Re 2: Maybe we could simply disable the favorites tab in that case? So it naturally falls back onto the next curation option, but will reset ti favorites if favorites are available. 

Okay.

> And a swipe-to-refresh on the genre screen should re-run the check, in case the user just added a favorite.

👍




> Re 5: Okay, the setting being global makes sense I guess, I just haven't really noticed it because I almost always clear the storage when switching branches, so my default is always reset 😅

_and_

> The new switching button really doesn't fit, especially not without proper labels. But since it's a global setting it's fine, it's just my uncommon usage that's causing issues...

okay, then I will leave it as is 😊
But just to clarify: you mean **both** proposed options don't fit? 




> Re 6: I have to disagree. The tracks section doesn't have to be an appetizer, you made me excited about all the improvements to the artist screen that are possible! It should be possible to see all tracks and all albums on the artist screen. Especially when there is the appears on section below the albums, I don't think it's a good idea to show the full list of (potentially 100+) albums directly, but hide it behind the "see all" button instead. I'm aware it makes some things a tiny bit slower to find, but makes others faster to find or even possible to find (artist tracks, where you don't know the album). And I really don't think there's a big enough semantic difference between genre tracks and artist tracks to justify handling it completely different (one with filters, one with settigs default).

Alright, you convinced me. 
I will tackle this in a future PR 😊 

> The fallback settings also don't "vibe" with me. They shouldn't disappear when a different default is selected, at least not with the current settings design (disabled instead), but I think simply settings the user reorder the chips would be a much cleaner solution. Could you remove all related settings, and just go with a sensible default instead? Then we can add the reordering later on. You could also do it immediately, but I'm sure you want this getting merged too 😅

Okay, I think this PR still is suitable for adding the flter row to the ArtistScreen, removing the Fallback settings form Artist and Genre Screen Settings and adding the re-ordering of the Filters instead. I think, we should be able to define different orders for ArtistScreen and GenreScreen, though.

Chaphasilor:
Ah, the issue with Appears On makes sense, but I think the distinction between album artists and performing artists would be enough for finding the artist itself. But we can show all tracks/albums associated with that artist?  
It would fulfill my personal use case, but might not be ideal. Still, I think it's better than not being able to see the tracks at all.

Yes, both options don't fit :)

Yeah distinct orders sounds good. Just the concept should be the same.

Thanks for handling this! <3

Chaphasilor:
Still crashing for me when downloading a lot of tracks.  
Here are the app logs: [finamp-logs_genre-downloads-crashing.txt](https://github.com/user-attachments/files/20157143/finamp-logs_genre-downloads-crashing.txt)

And (part of) the system logs, couldn't figure out how to grab all of them: [asus-ASUS_AI2302-Android-15_2025-05-12_100802.txt](https://github.com/user-attachments/files/20157120/asus-ASUS_AI2302-Android-15_2025-05-12_100802.txt)

I'll try simply downloading my library via current redesign, maybe that also has issues?

lukaslindnermusic:
> Still crashing for me when downloading a lot of tracks. Here are the app logs: [finamp-logs_genre-downloads-crashing.txt](https://github.com/user-attachments/files/20157143/finamp-logs_genre-downloads-crashing.txt)
> 
> And (part of) the system logs, couldn't figure out how to grab all of them: [asus-ASUS_AI2302-Android-15_2025-05-12_100802.txt](https://github.com/user-attachments/files/20157120/asus-ASUS_AI2302-Android-15_2025-05-12_100802.txt)
> 
> I'll try simply downloading my library via current redesign, maybe that also has issues?

Yeah, please try it on the redesign as well.
In the logs, there are a few `Null check operator used on a null value` errors caused by the login flow, which I didn't change.
You are downloading loads of lyrics :P 

There are a few AssertionErrors like these:
[finamp-logs_genre-downloads-crashing_assertion_finamp-models_verifyEnums.txt](https://github.com/user-attachments/files/20162415/finamp-logs_genre-downloads-crashing_assertion_finamp-models_verifyEnums.txt)

It seems, that its _verifyEnums in class DownloadedImage in finamp_models.dart that's causing this. also, in the line above, it says `FinampOffline: true`, so maybe your device lost connection during syncing for a short time and the AutoOffline mode just switched to offline mode?
I have no idea if this happens because of some changes we made in this PR, or if it also happens in the current redesign branch.
@Komodo5197 do you have an idea?

@Chaphasilor Did you already check on the redesign branch by downloading the same items?

Chaphasilor:
I was on the downloads screen with the screen on while things were downloading when the crash happened, if that's any clue...

Chaphasilor:
Btw, no crash on redesign when downloading roughly the same genres and the favorite collection. I'll try a library now as a stress-test. I am downloading on my local network (using the local URL) now, instead of via the remote URL from a different network last night. It shouldn't have an impact aside from slower speeds, but it could be a factor.

Chaphasilor:
"Good" news! It also crashed during the library download on `redesign`: [finamp-logs_downloads-crash_redesign.zip](https://github.com/user-attachments/files/20171978/finamp-logs_downloads-crash_redesign.zip)  
So the issue seems to be somewhere else.  
I have logcat streaming logs to a file now, hopefully I can capture the crash this time :)  
But that means it's fine by me to merge this PR if the other issues are resolved. I'll wait about 30 minutes for a new crash, and then I'll switch to this PR again to test the remaining stuff...

Chaphasilor:
Wow, what a massive improvement! tysm for putting in the work @lukaslindnermusic <3

PR #1188, opened 2025-04-24
Fix issue templates

Tiefseetauchner:


Tiefseetauchner:
Ah forgot a description. Uhm

Well, \ is the way to do comments in markdown (on gh) so I changed that to fix the issue templates

PR #1191, opened 2025-04-24
Fix crash in MediaItemContentProvider when the pipe breaks

### LINKED ISSUE #1185: [FEATURE NAME]
<You can keep text in these brackets since it wont be visible>

<!!! Keep in mind that new features are currently only added to the redesign/beta version of the app !!!>

### Type:
- [ X ] New feature
- [ X ] Improve or extend an existing feature
- [ X ] Other: Half Bug / Half Feature. More of a feature request

### Explanation 
<Describe what your suggestion is>
Jellyfin allows you to upload a gif for an album cover, artist photo, etc. It has a really cool effect while the audio is playing. It can make it feel much more lively. However, when the album is played on the Finamp app, the app crashes. I think this has to do with the fact that android can't handle a gif as the album cover. 

### Reasoning
This is a just a cool feature that could make the experience feel more lively (or could even be similar feeling to Spotify's canvas feature)

### Implementation Notes
My suggestion would be that while the audio is playing and the primary picture is a gif, use a potential secondary cover picture for everything besides the screen that the audio is actually playing. In this scenario, the primary picture could be a gif, but the secondary picture would be a normal image file.

Maxr1998:
This can happen when loading especially large files like GIF cover arts. Unfortunately, the crash happens within (very old) framework code, so we have to copy the called function and write our own version that simply catches the exception.

Fixes #1185.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -25,7 +25,7 @@ if (flutterVersionName == null) {
 def keystoreProperties = new Properties()
 def keystorePropertiesFile = rootProject.file('key.properties')
 if (keystorePropertiesFile.exists()) {
-   keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
+    keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
 }
 
 android {
@@ -95,4 +95,5 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.3.1'
     implementation 'androidx.mediarouter:mediarouter:1.7.0'
     implementation "androidx.core:core-splashscreen:1.0.0"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.8.1"
 }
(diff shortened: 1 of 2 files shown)


Chaphasilor:
Does this happen with any gif? Or do I need something specific to test this?

Also, is there a good source for animated covers?  
CC @jar-113

Maxr1998:
> Does this happen with any gif? Or do I need something specific to test this?

I thought yes, but apparently not. I only had it happen on release builds too, interestingly. The same image was fine later on when on another branch with a debug build.

> Also, is there a good source for animated covers?  

I grabbed mine from Apple Music, you can extract them with [this website](https://covers.musichoarders.xyz/).
To convert it to a GIF, I just used [ezgif.com](https://ezgif.com/), somehow worked better than even `ffmpeg` for me in the past.

Chaphasilor:
Seems to work fine, got a crash without the PR and don't get a crash with this :)

I thought maybe this messed with loading some covers on Android Auto, but seems like that was broken before...

PR #1195, opened 2025-04-27
feat(android): add locale config for Android 13+

edeuss:
This will add a language picker to the info page for the app for Android 13+ devices.


Maxr1998 (approved):

Chaphasilor:
@edeuss where did you get the list of languages from? The iOS manifest? :)

edeuss:
@Chaphasilor the arb folder, just copied the codes from the files in there.

Chaphasilor:
Okay perfect, then I see no reason why this shouldn't be correct :D  
Thanks for the PR!

PR #1202, opened 2025-05-03
Set maximum line length to 120 characters

Maxr1998:
Now that the Hackathon is over and most PRs merged, we can finally apply this change.
Note that it requires at least Dart 3.7.0, so Flutter 3.29.0.

To batch-reformat all existing files, I used `dart format lib/`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,33 @@
+name: Test
+
+on:
+  pull_request:
+  push:
+
+# Due to a flutter bug, flutter build is not generating localizations
+# workaround by always running flutter pub get immediately before building,
+# and performing build for msix:create manually to allow implementing this workaround
+jobs:
+  check-formatting:
+    name: Ensure correct formatting
+    runs-on: ubuntu-latest
+    steps:
+    - name: Clone repository
+      uses: actions/checkout@v4
+    - uses: actions/setup-java@v4
+      with:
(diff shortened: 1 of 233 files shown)


mihawk90:
Since this is a batch reformatting that doesn't actually change how the code works, it might be a good idea to mark this commit (or the merge commit if one is done) to be ignored by git blame. GitHub has supported this with a [`.git-blame-ignore-revs` file in the project root](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view) for a while now.

Otherwise half the codebase will be attributed to this reformatting which is a PITA to track down the actual introduction of the code you're looking for.

Maxr1998:
Yup, already planned to do that after this is merged. I'll add that in a follow-up though, to make sure we use the correct commit hash. As of right now, we will wait until most open PRs are merged and then do a fresh reformat.

Chaphasilor:
I've marked this as a draft since we will need to re-run the formatting before merging :)

Maxr1998:
This somehow totally got lost in my notifications. PR is updated now @Chaphasilor, feel free to merge at any time!

Chaphasilor:
Which merge strategy is ideal here? Rebase or merge commit?

Maxr1998:
And thanks for adding the workflow already, looking good!

PR #1203, opened 2025-05-03
[Redesign] Miscellaneous fixes

Komodo5197:
Various fixes to a number of bugs I noticed while looking at PRs.  Built on top of #1162 to be more up-to-date with the autoOffline changes, so that needs to be merged first.

- Fix incorrect state tracking in playon handler, mostly revolving around changes happening while the socket is connecting.
- Fix hive_registrar.g.dart sometimes not being generated.
- Tweak log exports on desktop to still allow share method.
- Fix log censoring with empty local network address.
- Refactor background services expressed as providers to use ProviderContainer instead.
- Make getItems by itemId automatically split up requests with more than 200 items.
- Fixes to loadSaved queue on desktop when no items can be loaded.
- Fixes to loadSavedQueue seeking when current track fails to load.
- Fix android build not working in github actions.
- Fix track downloads not showing file sizes during transcode choice.
- Save and restore window size when running on desktop.


Chaphasilor:
Oh man, you really never disappoint with your fixes! A lot of good stuff in there, so thanks!  

FYI I just added another commit to also prompt/reload on transcoding settings change, so you might wanna rebase onto that (or just merge it in, not sure how well Git would handle that).  
Also, how did you end up fixing the build issues with the hive_registrar and GitHub? For GitHub it seems to have been the memory limit?

Komodo5197:
I'll just merge it in.  Git seems to handle this stuff well enough?  And I don't really like rebasing, personally.  The github build failures were indeed the memory limit.  The hive_registrar issue was an upstream issue I fixed, and then this upgrades our version.  https://github.com/IO-Design-Team/hive_ce/pull/130

Chaphasilor:
Hmm, so maybe I shouldn't have squash-merged #1162, but the commit history was a bit verbose, so you probably want to rebase your changes onto `redesign` now. Sorry about that; I tried reverting but that just creates a revert-commit.  
Only realized your PR would be affected a few seconds after hitting the merge button.

But I can also rebase it, since I'm the one who messed up.

Either way, I'll get to testing this now, since content-wise it is up-to-date with `redesign` already, just the commit histories are conflicting.

Komodo5197:
You can try rebasing if you want.  Alternatively, you can just revert and then do a normal merge, and it should be able to fold this in normally, even with the extra commits.  Or you could even force-push a normal merge if you really want.

Chaphasilor:
I'll take a look at the options tomorrow. I *think* rebasing should work well since this PR, by design, doesn't conflict with the changes in #1162, and there shouldn't be any major conflicts with pre-merge `redesign` either.

lib/services/censored_log.dart
```diff
@@ -27,9 +27,13 @@ extension CensoredMessage on LogRecord {
 
       if (user != null) {
         workingLogString = workingLogString.replaceAll(
-            CaseInsensitivePattern(user.baseUrl), "BASEURL");
+            CaseInsensitivePattern(user.baseURL), "ACTIVEURL").replaceAll(
```
Chaphasilor, line 30: I know you didn't change this, but what do you think about always resolving if it's the local or remote URL? That might help more with debugging than seeing `ACTIVEURL` and having to figure out which URL was switched to last.  
Alternatively we could put put an `active` qualifier in parentheses to confirm that this was used for the current request, and isn't some error from an old request that was made before the switch.

Chaphasilor:
Okay, I've rebased and force-pushed the changes now. I hope everything is still in order; the merges of the other PR confused git during the rebase but I think I caught the incorrect changes.  
Maybe keep your old branch around somewhere before resetting to this new state...  
I also think we can merge this tomorrow in any case :)

Chaphasilor:
Just tested the download location deletion, and it's not possible to delete if there is a download error. And since trying to download images to custom locations will fail on Android, that might be a problem.  
A repair also didn't fix this, possibly because it already deleted the matching track?  

Here are the logs:
[finamp-logs.txt](https://github.com/user-attachments/files/20067452/finamp-logs.txt)

Komodo5197:
I believe the core of the issue was that when you delete an item, the download settings are only synced for required items.  So you download an item into the location, and various images get pulled down with it.  Then you download another item into a different locaation that also needs those images.  Then you delete the item in the first location, but the syncTranscodeProfile of the images is never updated from the old location because its only using info links.  Then when the downloads repair marks all items with an outdated fileTranscodeProfile as needing redownloaded, these images get skipped because the sync profile still matches.  I've updated the code so that downloads repairs will update the sync profile around info links.  You'll still almost assuredly need a downloads repair after deleting the items and before the deletion works.  I also noticed that notDownloaded items can still have an unused fileTranscodeProfile, so we now ignore that as well, which hopefully should be all the issues.

lib/services/finamp_logs_helper.dart
```diff
@@ -89,42 +90,18 @@ class FinampLogsHelper {
...
-      // On desktop, we need to copy the file to the user-picked location
-      await tempFile.copy(filename);
-    }
-
-    await tempFile.delete();
+    await FilePicker.platform.saveFile(
+        fileName: "finamp-logs.txt",
+        // This is ignored on desktop
+        initialDirectory: (await getApplicationDocumentsDirectory()).path,
+        bytes: utf8.encode(await getFullLogs()));
```
Chaphasilor, line 105: If these are ignored on desktop, don't we still need to create a temp file and copy it manually?

Chaphasilor:
That indeed fixed it. After a repair the failed download was gone and I was able to delete the location. Thanks for the fix!

Now that #1176 is also close to being merged, do you have any preference which PR to merge first?

lib/services/finamp_logs_helper.dart
Komodo5197: The comment is incorrect.  I meant to say mobile in reference to initialDirectory.  I have verified that the bytes export does work on desktop.

Komodo5197:
I don't think the order matters too much, there doesn't seem to be any major conflicts.  I'd go with this just because I think its slightly further ahead.

Chaphasilor:
Alright, let's merge this then :)  
Thanks as always!

Issue #1090, opened 2025-03-05
Bug: Rearranging tabs doesn't seem to work


AllergicToGrass:
version: 0.9.14-beta
os: android

Steps to reproduce:
1. Open Finamp
2. Tap the hamburger menu in the top left
3. tap Settings
4. tap Layout & Theme
5. tap Tabs
6. Try moving "Tracks" from the end of the list to the beginning

Expected behavior:
Once I back out, The first tab to be displayed on the home screen is "Tracks"

Actual Behavior:
The tabs stay in the same location. One of the few times I tried rearranging tabs, Finamp locked up and I had to force close the app

e-v-o-l-v-e:
Hey,

Same here, it happens to me everytime I have to erase data or reinstall, this time I managed to 'fix' it by first disabling tracks tab, then forcing stop of the app . After that enabling tracks and moving tabs around worked as expected 🤷

I'll add that when it didn't work I had a weird flash of a grey screen when leaving tabs setting, which I had not when it worked as intended, I had a similar grey screen when using the reset button before doing what I mentioned before (even though on this one I managed to see the dialog asking confirmation)

lukaslindnermusic:
@Chaphasilor this was fixed with my GenreScreen PR (https://github.com/jmshrv/finamp/pull/1176) 😊

Chaphasilor:
This has now been resolved as part of [beta version 0.9.18](https://github.com/jmshrv/finamp/releases/tag/0.9.18-beta)!

PR #1218, opened 2025-05-16
Various Fixes and Refactoring

lukaslindnermusic:
Hi there 😊 

I've found a few oddities and minor bugs over the past few weeks and thought I could maybe just try to fix them and suggest my changes.

1. On the Artist and Genre Screens, the title jumped around during load because the downloadButton might take some time to load (and we also hide it because it displays as outdated until the providers fetched all data). This was because flutter automatically centers the title when there is enough space. I disabled that, which avoids the jumping aorund of the title and imo looks and feels just smoother.
2. the SortBy and SortOrder Buttons on the MusicScreen occasionally did not work. I inspected this and realized, that they do actually work, but the embedded widgets are not properly rebuild and therefore they edit the sort orders on other tabs as the current one. I edited some callback I found that was not doing anything anymore, and ensured that the widget is rebuild on every tabswitch. That seems to have fixed it.
3. I edited the available SortBy options to better represent the current tab. Previously, for everything except Tracks, the same options where shown. But there is no sense in having an "Name" and an "Album Artist" option on the Artists tab, or an "Album Artist" option on playlists and genres. Or "Premiere Date / Release Date" on Playlists, Genres and Artists, etc.. I also added the "Date Last Played" and "Runtime" options to fitting tabs. 
I ~~unommented~~ commented out CriticRating and CommunityRating, as I've not heard yet of any person that is using these to rank their music. Of course, this is just from my personal experience, so I can add it back if it actually does get used - Are there people doing that? Do you? 
4. the SortBys "Most Played" and "Date Last Played" now get disabled in Offline-Mode (as they are not compatible / don't show up-to-date data). This mirrors the behavior we have with PlayCount in the entire app already. If these SortBys were selected when changing to offline mode, it will temporarily switch to Name. If you change it, the changed will stay, otherwise it will jump back to the previously selected option when back online.
5. the Swipe Left/Right Actions on Track List Items display the action string and icon underneath the ListTile when you swipe it in a direction. I noticed that when you swipe it all the way to the other side, actually BOTH actions are shown, the left and the right, and you can see both for a split second. (Even "disabled" gets shown.) I fixed that by having a secondBackground that gets used when the other side is swiped, so we only display the action that is actually happening.
6. Some Date-based sort options in the sortBy function for some reason did not reliably work for me, so I added DateTime parsing for those, which seemed to have fixed it.
7. @Chaphasilor did it already, but before I realized that I did it too :P Renamed Premiere Date SortBy user-facing string to "Release Date", as this better reflects Music libraries. Jellyfin seems to prefer movie and series libraries, where "PremiereDate" totally makes sense. Not so much with music, though. However, I think its best to keep the internal enum value to premiereDate, as the other enum values also reflect the jellyfin side of things and we would destroy that integrity.
8. One thing that bugged me since I'm using the app: The Artist and Genre Lists do not have a subtitle, but they have an empty subtitle line underneath the Artist/Genre name. Of course, when you download an item, there is the downloadedIndicator, but otherwise there is nothing, and the title is out of place. ~~I changed this, so that if there is no subtitle, we don't use the "subtitle" in ListTile but add the downloadedIndicator via a Column in the title instead, styled exactly like if there would be a subtitle. This way, we get the title/subtitle layout we are used too when there is a downloadedIndicator, and if not and if there is no other subtitle, we only get the title, vertically centered.~~ 
--> https://github.com/jmshrv/finamp/pull/1218#issuecomment-2888277737
9. Since I've added "Date Last Played" to the SortBys, I had the idea of adding it as another CuratedItemSelectionType for the GenreScreen and ArtistScreen tracks. It also gets disabled on offline mode.
10. We're already disabling the DownloadButton when an artistScreen is genreFiltered. That's because we don't want to give the impression nor do we want to add the functionality to only download a genre from an artist but nothing else. Downloading an artist should always download the full artist, and it might be unexpected behavior if you see an artistprofile filtered for "Soundtrack" and there is one album, you hit download and then you've downloaded 20 albums with different genres. To make this clear to the user, I now added it back but greyed out, and clicking it shows "Please remove the Genre Filter to ~~download~~ enable the download actions for this artist." ~~(if you've already downloaded the artist, it shows the normal sync or delete icons. The greyed-out and message thing only applies to the actual "Download" button)~~ (This also applies to the other download action buttons)
11. I fixed the issue with the genreScreen on-pull refresh, where the genreFilter got resetted to its initial value. now it stays at the current value, while still invalidating all provider data and clearing the disabledFilters.
12. The Add/Shuffle to Queue/Next Up stuff in the long-press menu still fetched the wrong ArtistTracks (not including Performing Artist tracks). I know that @Chaphasilor is working on new menus, so this might be removed or integrated in their PR very soon, but for now, I changed it to use the allTracksProvider from the ArtistScreen instead. That one returns the correct items.
13. I renamed Favourite to Favorite in the entire app (user-facing as well as internal). We currently use a mix of both. A quick question on Discord as well as looking at Jellyfin Web and the API made me think that we should use the american english version.
14. I added better messages for empty ItemCurationSections. When switching automatically is enabled and you manually press on i.e. Favorites (so not if it was initially selected, only manually selecting it), but there are no items, you now also get a Snackbar Message in addition to the diabling filter, as in my recent days testing the feature, it seemed a little rapid and I not always noticed directly that I'm now on a different filter. This should improve that. It now also specifies that i.e. "You've no favorites yet for this artist in that genre.", so that it gets clearer to the user that it might be a genrefilter thats causing your favorites to not show up.
15. I fixed a bug where a queue started from a genre did say its from the genre, but ontap lead to the ArtistScreen instead of the new Genre Screen.
16. Last but not least, I refactored the AlbumSliverList, AlbumItemListTile and AlbumItemCard stuff a little bit. As it currently handles not just albums but also artists, playlists and genres, and as the downloadSystem already calls it that way, I thought, "Collection" is a better name of representing what it does. for the artistScreen only (which I then recently moved to the CuratedItemSections), there was a separate album_list_tile file, that did exactly the same as the other one... and there also was a comment about `TODO should this be unified with artist screen version?` so I did that now.

I hope, some of these bug fixes and minor improvements are welcome 😊

Edit:
17. Also fixed Favorite Icons not getting displayed for favorite collections (artists/albums) on other Screens such as Artist and Genre Screen when the Favorite Filter on the MusicScreen was active. They do get hid on the MusicScreen, because we obviously only get the Favorites then and can display more from the title etc., but this accidentally also affected other screens that used this widget.

Edit 2:
18. Added the current tabOrder to the refreshHash, so that the MusicScreen refreshes automatically when the tabOrder changes. This has to happen because otherwise tabs would still show the previous content from the tabs that previously were on the loaded indexes. This should fix https://github.com/jmshrv/finamp/issues/1090.


lukaslindnermusic:
Regarding the DownloadButton:
> * To make this clear to the user, I now added it back but greyed out, and clicking it shows "Please remove the Genre Filter to download this artist." (if you've already downloaded the artist, it shows the normal sync or delete icons. The greyed-out and message thing only applies to the actual "Download" button)

On second thought, though, it might be better to disable all downloadbuttons. deleting stuff that is not shown on screen is also bad :P Syncing I guess would be okay, but due to consistency I think its better to disable it as well.

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -125,6 +133,24 @@ void openSeeAll(
     if (newDisabledArtistFilters != null) {
       _disabledArtistFilters.addAll(newDisabledArtistFilters.whereType<CuratedItemSelectionType>());
```
Komodo5197, line 134: Is the whereType here just doing the equivilant of nonNulls?  Because if it is, I'd prefer using that for clarity.  Also, I'd love a comment explaining about the fallback logic above all the main provider watches to reasons this disabled logic is set up this way clearer.

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -125,6 +133,24 @@ void openSeeAll(
...
     }
+    _disabledTrackFilters.remove(genreCuratedItemSelectionTypeTracks);
+    _disabledAlbumFilters.remove(genreCuratedItemSelectionTypeAlbums);
+    _disabledArtistFilters.remove(genreCuratedItemSelectionTypeArtists);
+    newSelectedCuratedItemSelectionTypeTracks = sendEmptyItemSelectionTypeMessage(
```
Komodo5197, line 139: I don't like how the logic that makes this work is split up and hidden away.  I'd prefer a more explicit display of the logic, something like

```
if (newDisabledTrackFilters != null) {
      if(newDisabledTrackFilters.contains(newSelectedCuratedItemSelectionTypeTracks)){
         sendEmptyItemSelectionTypeMessage()
     }
     newSelectedCuratedItemSelectionTypeTracks=null;
    }
```

Also, I'd prefer to rename newSelectedCuratedItemSelectionTypeTracks to something like clickedCuratedItemSelectionTypeTracks to differentiate it more from newDisabledTrackFilters which has a different lifecycle.

lukaslindnermusic:
Thank you Komodo 😊 
I will implement the suggested changes later 😊 

Regarding this:
> * One thing that bugged me since I'm using the app: The Artist and Genre Lists do not have a subtitle, but they have an empty subtitle line underneath the Artist/Genre name. Of course, when you download an item, there is the downloadedIndicator, but otherwise there is nothing, and the title is out of place. I changed this, so that if there is no subtitle, we don't use the "subtitle" in ListTile but add the downloadedIndicator via a Column in the title instead, styled exactly like if there would be a subtitle. This way, we get the title/subtitle layout we are used too when there is a downloadedIndicator, and if not and if there is no other subtitle, we only get the title, vertically centered.

There actually was a tiny padding difference for items without subtitle and then with vs. without the downloadedindicator icon. It's very very minimal, so that I didn't notice it on static lists, but if you hit download on an item and pay close attention as soon as you selected your download quality, you can see it increase a little. No matter how I've set the contentPadding, as soon as I've hit download on an item, that tiny padding difference got added. I thought about using IntrinsicHeight or something, but then I decided that it would be best to instead get a bool if the downloadedIndicator shows something and then we can always use the default "subtitle" from ListTile when we have either a subtitle and/or downloadedindicator icon. I did this by providing the option to override the `ref.watch(downloadsService.stateProvider(item))` inside the downloadedIndicator, and added the same ref.watch in the `collection_item_list_tile.dart` and then passing it as argument to the downloadedIndicator, which uses this provider instead of watching again. Then I simply added a helper method that also takes a `AsyncValue<DownloadItemState?>? status` and then just returns true if there will be an icon shown.
This fixed the little padding jump and additionally removed the need for the "hacky" solution of using a Column inside of the title when there is no subtitle.

lib/components/curated_item_filter_row.dart
```diff
@@ -57,13 +58,14 @@ Widget buildCuratedItemFilterRow({
...
                             },
                           selected: isSelected,
-                          tooltip: (isOffline && type == CuratedItemSelectionType.mostPlayed)
-                              ? AppLocalizations.of(context)!.curatedItemsMostPlayedOfflineTooltip
-                              : ((disabledFiltersList.contains(type)) 
-                                  ? ((type == CuratedItemSelectionType.mostPlayed)
-                                      ? AppLocalizations.of(context)!.curatedItemsMostPlayedDisabledTooltip
-                                      : AppLocalizations.of(context)!.curatedItemsFavoritesDisabledTooltip)
-                                  : null),
+                          tooltip: (isOffline &&
```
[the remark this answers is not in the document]
lukaslindnermusic: You're right. Well, since I now have the Snackbar Message anyway, I don't think that we deliberately need that tooltip anymore.

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -40,6 +42,7 @@ class _ArtistScreenContentState extends ConsumerState<ArtistScreenContent> {
   final _downloadsService = GetIt.instance<DownloadsService>();
   final Set<CuratedItemSelectionType> _disabledTrackFilters = {};
   BaseItemDto? currentGenreFilter;
+  CuratedItemSelectionType? newSelectedCuratedItemSelectionType;
 
   StreamSubscription<void>? _refreshStream;
```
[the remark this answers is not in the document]
lukaslindnermusic: ah yes, because the state-refresh won't change the provider data so deleted offline tracks would not be immediately removed from the UI?

lib/components/GenreScreen/genre_screen_content.dart
lukaslindnermusic: Yes. But actually, as newDisabledArtistFilters is a `Set<CuratedItemSelectionType>?`, we don't need to check for null at all and can just addAll. 😅

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -366,6 +366,7 @@ class _MusicScreenTabViewState extends State<MusicScreenTabView>
             widget.view?.id,
             settings.isOffline,
             settings.tabOrder.indexOf(widget.tabContentType),
+            settings.tabOrder,
```
Komodo5197, line 369: settings.tabOrder.indexOf(widget.tabContentType) is supposed to handle rebuilding the moved tabs when tabOrder changes.  If that isn't working correctly, it should be removed.

lib/components/curated_item_filter_row.dart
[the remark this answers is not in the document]
Komodo5197: Yeah, but it would still be nice for the ones that are disabled on initial load or when offline.  I think you can just wrap the FilterChip in a Tooltip widget.

lib/components/AlbumScreen/downloaded_indicator.dart
```diff
@@ -62,3 +64,26 @@ class DownloadedIndicator extends ConsumerWidget {
     }
   }
 }
+
+bool downloadedIndicatorIsVisible(AsyncValue<DownloadItemState?>? status) {
```
Komodo5197, line 68: I don't like how much logic this is forcing into the parent.  What I'd do is move this function into the DownloadedIndicator class and have the only argument be a WidgetRef.  Then it can do ref.watch(downloadsService.stateProvider(item) itself, and all the parent needs to do is call isVisible ion the widget it created.   Also, you don't need the statusOverride stuff, its fine to just watch the provider in two places.

lukaslindnermusic: Like this?
https://github.com/jmshrv/finamp/pull/1218/commits/0070e82a278abd29263985a5581be4c7a2ad828a

Komodo5197: Yeah, that's pretty much what I was imagining.

lukaslindnermusic:
Okay, so we have different sort options, but when we select one that is not displayed per default in the list, we can't tell with what we are actually dealing with. For example, if I sort by Release Date, I will get the order but I can't actually tell at what Release Date I am starting at and how many albums are in one year etc. etc. 

So I added some additional optional info displays for Collection- and Track-Lists and linked them on the MusicScreen to the selected SortBy and on CuratedItemSelectionLists to the selected filters sortby.
So for example, if you select "Most Played" on a genre-tracks-section, the play count will get shown. If you select "Runtime" on playlist music screen sorting, the playlist total-duration will get shown before the default childCount. Release Date... you get the point 😊 

(The Play Count for tracks was already implemented (for the previous Top Tracks section), but I moved it to the left, because with multiple artists or artists with moderately long names you couldn't even see them.)

lukaslindnermusic:
for the Date Added and Date Last Played, maybe it would be nice to have something like "Played" or "Added " and then "just now", "5min ago", "2h ago", "yesterday", "2 days ago", "on 05/05/2025"...
I will work on implementing that tonight.

Also, there is still a bug that you can't remove an artist or a genre from a mix once added. I will look into that as well.
Maybe we should also disabling adding an entire genre to a mix for now?

Chaphasilor:
Awesome, thank you for the changes! Let me just comment on them:

1. Alright. The title needs to go somewhere else at some point, on my phone I can often only see a single word before the ... ellipsis.
2. I also noticed the sorting options not working just today. I'll see if it is fixed by this
3. Yeah that's nice. I vote to rename "Runtime" to "Duration", at least user-facing.
4. Can we disable and move them to the bottom of the list instead?
5. Nice! Thanks for figuring that out :D
6. Okay, didn't notice anything odd, but as long as it won't be worse, it's fine.
7. 😉
8. Okay. I've added a chip that shows the child count (e.g. number of tracks in an album/playlist, number of albums by an artist) within the new menus, I think we should unify those with the subtitle contents as best as possible (once that PR is merged)
9. Nice
10. I like it, disabling is often less confusing than hiding. Especially if such an explanatory message is shown. But we can show the sync button.
11. Alright, previous behavior was fine too, but I'm also happy with this solution :)
12. Guess I'll deal with that in the merge conflicts :P
13. Awesome, thanks!
14. That's nice. Maybe put the genre stuff in parentheses? And I prefer "You don't have any" over "You've no". We can probably also remove the "yet".
15. Ah yes, we finally have separate screens. Although I think there's still a bunch of booleans like "isGenre" on the artist screen, if I  remember that correctly? You could remove those if you notice them.
16. Collection is good, but maybe we should call it "FinampCollection" or "ItemCollection" to make is clear that this is not a Jellyfin "Collection". But collection is definitely the right word to refer to this.
17. Nice
18. I would've expected this to work already. Probably broke again at some point...

That's a lot of fixes, many thanks!

Regarding the "2 min ago" strings, please re-use the logic from the Playback History screen, as best as possible, or modify it in accordance with your changes. It should be consistent. Show minutes ago if it's less than 1h, show hours if it's less than 1 day, maybe show "Yesterday", and then just show the date. We *could* think about showing "x days ago" for the last week, but not beyond that :) 

If adding to a mix adds all genre albums, then just don't fix it. I wan't to keep the functionality, and removing it is not that important.

Oh and please use an ordered/numbered list next time xD

lukaslindnermusic:
Thanks for the feedback 😊 
I will address your mentioned points as soon as I have time. For now, I've edited my first comment to use the numbered list to not confuse people any further :P

Chaphasilor:
The sorting bug I encountered is fixed with this, as expected :)

lukaslindnermusic:
Before merging this, can someone running Windows can test if the icons and strings in the WidgetSpans on ListTiles for DateAdded and DateLastPlayed are misplaced as well?
Its kind of odd, on my iphone everything looks perfectly, but for the macOS build I did, Offsets were necessary. So maybe this is necessary for all desktop builds?
How does it look on Android?

lukaslindnermusic:
I now renamed release_date_helper.dart to datetime_helper.dart and added a unified class that from now on can be used to format datetimes in the app.
It either can format in the way currently done on the Playback History Screen, or it can display as „just now|x min ago|x h ago|yesterday|x days ago|Date" (which then also stays up-to-date/reformats every minute).

No worries, the ReleaseDateHelper stayed unchanged. But as DateTimeHelper spans a wider variety and ReleaseDateHelper is very specific, I thought it might be good to rename the file.

Komodo5197:
This is what it looks like on windows:
datetext
And this is android:
Screenshot_20250519-122613_Finamp Debug
Both seem incorrect in different ways?  We probably need to figure out a more principled way to do this.

lukaslindnermusic:
@Komodo5197 Okay, the offsets/positions are now perfect on macOS as well as iOS.
Hopefully, this also applies now to Android and Windows?

Chaphasilor (approved):
Icons and text size on Windows seems fine now (not "perfect", but let's leave it at that).  
Android is pretty much perfect from what I can tell.

This is ready to merge from my side.  
@Komodo5197 feel free to merge it in if you're also happy with it!

Chaphasilor:
Ah, that explains why it looks correct for me, I have the default size.

lukaslindnermusic:
@Komodo5197 I could reproduce on iOS Simulator by changing the default text size there, too.
For now, I've added a boolean parameter to disable it for the returned text widget.
For me, this improved it, but it did not resolve it perfectly. It's still not aligned as if you would use the default text size.

Is it better for you?

lukaslindnermusic:
In the future, is there no option to do this second row without a RichText but instead with something that is more cross-platform controllable? for example by using a simple Row and Text/Icon/SizedBox-children? 
Or is the use of RichText a necessity in this case?

Komodo5197:
This fixes the alignment on android, as far as I can tell.  Regarding non-richtext options, there probably is a way to do this with just a row widget, which might be more standard?  I also don't know why we have a column in the title slot of the listtile instead of just using the subtitile slot.  This wasn't really a cross-platform issue, it was a text scaling issue that could happen anywhere if the text scale was changed.  You reproduced it on ios, and I've reproduced it on windows as well.

lukaslindnermusic:
Yeah, true. With the cross-platform stuff, I was also refering to my fiddeling around with icon offsets. I had to use very specific offsets for desktop vs. mobile, and I'm not sure if this is standard for this kind of stuff?

Komodo5197:
I wasn't aware you still had macos specific offsets for the text.  I don't really know why the text would be aligned differently there than on windows, android, and ios.  As for the different offsets on the icons, looking at things closer there's a >1px discrepancy between the icon positions on android and windows, relative to the text, so I don't think this is actually a desktop vs mobile thing.  I think it might be due to the different fonts all the devices use.  Anyway, I think the widgetspan alignment to baseline instead of top may give a more consistent experience?

lukaslindnermusic:
> Anyway, I think the widgetspan alignment to baseline instead of top may give a more consistent experience?

Yeah, the Text widgets are baseline aligned. 
I just tried to align the icons with baseline as well, which caused a red assertion error screen.
.bottom worked, but then I would have to adjust all the offsets for the icons again.

I think, for now, it is okay how it is, and we can later move the entire thing to a Row if we want to improve it further?

Komodo5197:
Did you remember to also set baseline:TextBaseline.alphabetic when setting the alignment on the widgetspans?  But yeah, this doesn't really matter that much.  The icons don't really have a precise correct location like the text does.

lukaslindnermusic:
> Did you remember to also set baseline:TextBaseline.alphabetic when setting the alignment on the widgetspans?

Nope, of course I forgot :P 
I now use it and have adjusted the offsets. On macOS and iOS, it looks good again, and behaves a little bit better when resizing the text-size.

When it's good on Windows and Android, too, I think you can merge, @Komodo5197  😊 

I also wrapped the zoomed image from https://github.com/jmshrv/finamp/pull/1150 in  SafeArea (and adjusted the positioning of the x), as on iOS the x clashed with the status bar.

Komodo5197:
It looks good to me.  The zoom tweak has made me notice a fairly major issue where the album image zoom makes a new server request for every image size during the zoom transition.  But I'll fix that separately.

PR #1225, opened 2025-05-20
[Redesign] Album Image Zoom Fixes

Komodo5197:
Cleans up a couple of issues I noticed with the image zoom feature, primarily that a new resized image will be requested from the server for every frame in the transition animation.  Also:
- Use existing small image as placeholder for zoomed image during transition and while zoomed.
- Don't animate between images on unrelated routes.
- Linear transition animation.
- Always use maximum image resolution when zoomed.
- Removed dismiss gesture because it interfered with zooming and you can already both click outside the image or use the system back gesture.

### DIFF
--- a/lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
+++ b/lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
@@ -1,10 +1,10 @@
 import 'package:finamp/components/Buttons/cta_medium.dart';
 import 'package:finamp/components/global_snackbar.dart';
+import 'package:finamp/l10n/app_localizations.dart';
 import 'package:finamp/models/finamp_models.dart';
 import 'package:finamp/services/feedback_helper.dart';
 import 'package:finamp/services/queue_service.dart';
 import 'package:flutter/material.dart';
-import 'package:finamp/l10n/app_localizations.dart';
 import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
 import 'package:get_it/get_it.dart';
 
@@ -232,8 +232,7 @@ class AlbumScreenContentFlexibleSpaceBar extends StatelessWidget {
                   children: [
                     SizedBox(
                       height: 125,
-                      child: TapToZoomImage(
-                          albumImage: AlbumImage(item: parentItem)),
+                      child: AlbumImage(item: parentItem, tapToZoom: true),
                     ),
                     const Padding(
                       padding: EdgeInsets.symmetric(horizontal: 4),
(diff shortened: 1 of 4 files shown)


Chaphasilor:
Thanks for the fixes!

1. What's you reasoning for switching to a boolean option instead of a wrapping widget? I quite liked how explicit and descriptive the wrapper-approach was, and now every single album in the app is a stateful widget. Or was the stateful approach the only way to fix the continuous re-fetching?

2. I was debating getting rid of the dismiss gesture, but unltimately thought it worked well enough. Did you have issues with it?  
I did try to make it only dismissible when swiping outside (above/below) the image, but that didn't really work either.

Komodo5197:
The wrapper method was taking a finished AlbumImage widget, but to avoid the refetching I had to create a version of the widget with a bunch of non-default parameters.  And then to get the existing image to work as the placeholder, I needed direct access to the final imageListenable of the normal album image, which is only available once the albumimage has built.  So this made the wrapper method quite awkward.

Basically every time I tried to scale up the zoomed image on my phone the dismissable stole the gesture so nothing happened.  I had to explicitly orient my pinch to be exactly horizontal for it to work.  I thought about messing with the dismiss gesture more, but then decided that it didn't seem very worthwhile when its easy to close the image in other ways.

Chaphasilor:
Okay, fair enough.  
I see you managed to fix (most of) the build issues, thanks for that. Seems like semantic versioning doesn't really work all that well in the Flutter ecosystem 🙃  
If I can get this to build on Flutter 3.32 after upgrading, can I merge this already?

Chaphasilor:
Hmm, so one build (pull_request) worked, the other (push) didn't. Seems like a memory issue, I also had this problem on my machine where just (closing a few programs and) retrying made the build work. So it's probably not reliable yet, but good enough for now.

I'm guessing this is related to the [Kotlin rewrite of gradle tooling](https://medium.com/flutter/whats-new-in-flutter-3-32-40c1086bab6e#:~:text=Kotlin%20language%20rewrite%20of%20Gradle%20tooling) in Flutter 3.22, so I expect this to improve with upcoming Flutter patches

Chaphasilor:
I think we should merge this even with the remaining build issues since at least the build seems to work sometimes, whereas I can't build any other branches with Flutter 3.32 without applying these same fixes again...

Maxr1998 (approved):

Komodo5197:
The default arguments for a new project seem to be org.gradle.jvmargs=-Xmx8G -XX:MaxMetaspaceSize=4G -XX:ReservedCodeCacheSize=512m -XX:+HeapDumpOnOutOfMemoryError with a full 8 gigs of max memory allowed.  Should we just jump all the way to that?

Chaphasilor:
If that's the default I see no reason why not to do it. Artificially limiting the memory seems like an outdated approach anyway with virtual memory.  

My poor laptop might struggle even more, but its replacement is already ordered, so it's fine.

PR #1228, opened 2025-05-22
[Redesign] Android Backup Config

Komodo5197:
Configure android to back up settings while excluding downloaded files and other large replaceable assets.

### DIFF
--- a/android/app/src/main/res/xml/backup_rules.xml
+++ b/android/app/src/main/res/xml/backup_rules.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<data-extraction-rules>
+    <cloud-backup disableIfNoEncryptionCapabilities="false">
+        <include domain="sharedpref" path="./"/>
+        <include domain="root" path="app_flutter/"/>
+        <exclude domain="root" path="app_flutter/songs/"/>
+        <exclude domain="root" path="app_flutter/flutter_assets/"/>
+        <exclude domain="root" path="app_flutter/finamp-logs.txt"/>
+        <exclude domain="root" path="app_flutter/finamp-logs-old.txt"/>
+    </cloud-backup>
+</data-extraction-rules>
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -20,8 +20,9 @@
     android:usesCleartextTraffic="true"
     android:requestLegacyExternalStorage="true"
     android:supportsRtl="true"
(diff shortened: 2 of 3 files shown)


android/app/src/main/res/xml/backup_rules.xml
Maxr1998, line 1: - databases should be included too, I think, if we need the current download tasks
- not only songs should be excluded, but also images, for cloud backup
- we might want to specify device transfer rules too, and decide whether songs/images should be copied there. Generally, that seems appropriate in that mode, but for very large collections it could also make the download take a really long time
- likewise, we might want to include the files folder and external files folder, in that mode, if we decide to include songs/images

lib/services/downloads_service_backend.dart
```diff
@@ -679,6 +678,7 @@ class DownloadsDeleteService {
     if (item.file != null && item.file!.existsSync()) {
       try {
         await item.file!.delete();
+        _deleteLogger.finer("Deleted files for ${item.name}");
```
Maxr1998, line 681: Is it multiple files or just a single file? I'd prefer it to be specific.

android/app/src/main/res/xml/backup_rules.xml
Komodo5197: - As far as I can tell, we don't use the database path.  All information is stored either in isar, which is in the app_flutter documents path, or in sharedprefrences.
- The app_flutter location is the legacy documents folder, which generally shouldn't have an images directory, but I think it might be theoretically possible for one to get created so I'll exclude it.
- Without a device transfer block, we should theoretically be getting the default complete backup which includes the files and external paths.  It could take awhile, but otherwise we'd need a large download from the server instead, so neither seems much better.  Just transferring everything seemed a bit more sensible to me.

Komodo5197: - That file appears to be from flutter_downloader, which is what we used before the downloads rewrite.  We don't need it.
- Anything downloads inside app_flutter are using the legacy documents location that was the default in main but can't be selected in the redesign.  I think getting an image directory to match the song directory there would actually be almost guaranteed though, because we'd put the new artist and genre images there when migrating the old albums.  The files directory is the new default location for the songs and images directories.

Maxr1998: - Oh, I didn't realize that. Then we can definitely ignore it. Thanks for clearing my confusion!
- Alright, then the current configuration seems good. :+1:

Maxr1998 (approved):

PR #1239, opened 2025-05-26
[Redesign] Fix genrefiltered Playlist List

lukaslindnermusic:
Fix for https://github.com/jmshrv/finamp/pull/1232
Sorry for the oversight. 🙃🤦‍♂️
Now it should be working as expected.

Also some spacing fix for the ArtistScreen and some code formatting.

### DIFF
--- a/lib/components/GenreScreen/genre_screen_content.dart
+++ b/lib/components/GenreScreen/genre_screen_content.dart
@@ -31,8 +31,7 @@ class GenreScreenContent extends ConsumerStatefulWidget {
   final BaseItemDto? library;
 
   @override
-  ConsumerState<GenreScreenContent> createState() =>
-      _GenreScreenContentState();
+  ConsumerState<GenreScreenContent> createState() => _GenreScreenContentState();
 }
 
 class _GenreScreenContentState extends ConsumerState<GenreScreenContent> {
@@ -56,7 +55,7 @@ class _GenreScreenContentState extends ConsumerState<GenreScreenContent> {
     super.dispose();
   }
 
-void openSeeAll(
+  void openSeeAll(
     TabContentType tabContentType, {
     bool doOverride = true,
     CuratedItemSelectionType? itemSelectionType,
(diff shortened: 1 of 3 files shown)


lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -56,16 +55,18 @@ class _GenreScreenContentState extends ConsumerState<GenreScreenContent> {
...
   }) {
     bool isFavoriteOverride = false;
     SortBy? sortByOverride;
     SortOrder? sortOrderOverride;
-   
-    if (doOverride && ref.read(finampSettingsProvider.genreListsInheritSorting) && itemSelectionType != null) {
+
+    if (doOverride &&
+        ref.read(finampSettingsProvider.genreListsInheritSorting) &&
+        itemSelectionType != null) {
```
Chaphasilor, line 69: what happened here?

Chaphasilor: ahh nevermind, dart format strikes again. GitHub just displayed it weirdly...

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -112,36 +112,53 @@ class _MusicScreenTabViewState extends State<MusicScreenTabView>
...
+                        settings.onlyShowFavorites)))
+            ? true
+            : null,
         startIndex: pageKey,
         limit: _pageSize,
         artistType: settings.artistListType,
-        genreFilter: widget.genreFilter,
+        genreFilter: widget.tabContentType == TabContentType.playlists
+            ? null
+            : widget.genreFilter,
```
Chaphasilor, line 142: can't we include the filter for online mode? fetching all playlists is more overhead in online mode than in offline mode...

lukaslindnermusic: We *could* and I think it would work just fine, but we still get the wrong genreIds then, and I thought it might be more on the safer side to be consistent and filter by name as well than relying on the server to always show the correct results.
I thought, nobody has 100+ playlists, and also, they are paginated anyway.

Should I change it so that online uses the genreFilter again?

Chaphasilor: I mean if it doesn't works both ways, and not just due to a bug/inconsistency in Jellyfin, then I think we should leave it to the server. I'm also worried that pagination could break if we do manual filtering, although I guess we don't really filter but just find/map the wrong IDs to the right genres?

PR #1243, opened 2025-05-31
Add Centralized Device, App, and Server Metadata Logging

aman-boop:
"NOT FINISHED YET"

This PR introduces centralized collection and logging of device, app, and server metadata to enhance diagnostic capabilities. The primary changes include:

Added a new log.dart service to gather device, app, and server information across supported platforms.
Updated setup_logging.dart to initialize and register this metadata at startup using GetIt for dependency injection.
Modified finamp_logs_helper.dart to prepend collected metadata to full log exports, making support and debugging easier.
Updated pubspec.yaml to bump intl to 0.20.2 for compatibility.
Key enhancements:

Device info (model, OS version, etc.)
App info (name, version, build)
Server info (address, protocol, port)
Structured metadata is now included at the top of exported logs.
These improvements should make error reports and support requests more informative and actionable.

### DIFF
--- a/lib/services/finamp_logs_helper.dart
+++ b/lib/services/finamp_logs_helper.dart
@@ -5,12 +5,14 @@ import 'dart:math';
 import 'package:clipboard/clipboard.dart';
 import 'package:file_picker/file_picker.dart';
 import 'package:finamp/services/censored_log.dart';
+import 'package:finamp/services/log.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path_helper;
 import 'package:path_provider/path_provider.dart';
 import 'package:share_plus/share_plus.dart';
+import 'package:get_it/get_it.dart';
 
 class FinampLogsHelper {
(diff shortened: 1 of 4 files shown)


lukaslindnermusic:
I'm a little bit confused. Aren't the logs already saying stuff like:
```
[Startup/INFO] 2025-05-31 12:21:17.902565: This is Finamp version 0.9.16+117 (Signature ''), installed via com.apple.testflight.
[Startup/INFO] 2025-05-31 12:21:17.903949: Running on 18.4.1 on iPhone.
```
And I'm not quite sure about the Server info: As far as I know, we are censoring data in the logs that can get linked/dox a person. Aren't we censoring stuff like the Server address and port by design (i.e., it becomes `BASEURL` in the logged api responses)?

lukaslindnermusic:
Ahh okay, I see, you're just taking these loglines and reformatting them so they stand out more?

Still, I think the server address should stay censored and is not that relevant for debugging anyway?

I think, stuff like the server version is actually more of use here. That would be the api endpoint ‘System/Info‘.
maybe we are already calling that somewhere. (If not, I think we should only call it on-demand when exporting the logs and if possible.)

PR #1247, opened 2025-06-01
[Redesign] Use system title bar on wayland

Komodo5197:
This makes Finamp use the system title bar when on wayland.  It seems to work on KDE, someone should test what this looks like on GNOME because it sounds like they have some unusual behaviour related to window decorations on wayland.

### DIFF
--- a/linux/my_application.cc
+++ b/linux/my_application.cc
@@ -25,8 +25,7 @@ static void my_application_activate(GApplication* application) {
   // desktop).
   // If running on X and not using GNOME then just use a traditional title bar
   // in case the window manager does more exotic layout, e.g. tiling.
-  // If running on Wayland assume the header bar will work (may need changing
-  // if future cases occur).
+  // If running on Wayland always use a title bar because I don't know how to check the WM
   gboolean use_header_bar = TRUE;
 #ifdef GDK_WINDOWING_X11
   GdkScreen* screen = gtk_window_get_screen(window);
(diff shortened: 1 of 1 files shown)


Maxr1998:
I can also check on swaywm when I'm back home 👍

Chaphasilor (approved):
I can't really test this, but on Windows everything is unchanged as expected xD  '
Is gnome the only other window manager for Linux? Or do the other wms at least behave like it?
And did you manage to give it a try @Maxr1998?

linux/my_application.cc
```diff
@@ -25,15 +25,14 @@ static void my_application_activate(GApplication* application) {
   // desktop).
   // If running on X and not using GNOME then just use a traditional title bar
   // in case the window manager does more exotic layout, e.g. tiling.
-  // If running on Wayland assume the header bar will work (may need changing
-  // if future cases occur).
-  gboolean use_header_bar = TRUE;
+  // If running on Wayland always use a title bar because I don't know how to check the WM
```
Maxr1998:
```suggestion
  // If running on Wayland, never show the header bar, as this should generally work.
```

Maxr1998 (approved):
Sorry, had a busy week. Just tested it on Sway, it looks much nicer without the header bar, which was redundant on this window manager anyway.
I would reword that one comment though, since “title bar” seems to refer to the system title bar (if provided), which is a bit confusing imo.

Maxr1998:
> Is gnome the only other window manager for Linux?
There are quite a few window managers on Linux, probably more than I could count right now. However, since Gtk is closely connected to GNOME, it makes sense to use that header bar style on GNOME to fit in. However, this will only apply to X11, on Wayland we'll change from *header always enabled* to *header always disabled* with this PR, which should be fine.

Chaphasilor:
Alright, I guess it's best to let more users test this, thanks for the PR!

PR #1248, opened 2025-06-02
Enhance Logging with Metadata, Refactor Setup, Update intl, and add some c files to .gitignore

aman-boop:
### Logging Improvements

Introduced a new log metadata system (lib/services/log.dart) to collect device, app, and server info for logs, with privacy-safe redaction.
The log metadata is now prepended to exported logs in **FinampLogsHelper**.

### Refactored Logging Setup

**setup_logging.dart** now registers a singleton Log instance, simplifying device/app info handling.
Removed old, manual platform/device info logging.

### Dependency Update

Updated **intl** package to version 0.20.2 in **pubspec.yaml**.

### .gitignore

Ignores generated Linux Flutter plugin files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,3 +125,6 @@ app.*.map.json
 
 # generated localization files
 **/lib/l10n/*.dart
+linux/flutter/generated_plugin_registrant.cc
+linux/flutter/generated_plugin_registrant.h
+linux/flutter/generated_plugins.cmake
--- a/lib/services/finamp_logs_helper.dart
+++ b/lib/services/finamp_logs_helper.dart
@@ -66,9 +66,10 @@ class FinampLogsHelper {
     // Get the Log instance and add its metadata at the top
     final logMeta = GetIt.instance.isRegistered<Log>() ? GetIt.instance<Log>() : null;
     if (logMeta != null) {
-      fullLogsBuffer.writeln("=== Device/App/Server Info ===");
-      fullLogsBuffer.writeln(jsonEncode(logMeta.toJson()));
-      fullLogsBuffer.writeln("==============================");
+      // Use the on-demand censored metadata, which fetches server info/version as needed
+      final censoredMeta = await logMeta.toCensoredJsonOnDemand();
+      // Prepend this metadata to the logs
+      fullLogsBuffer.writeln(jsonEncode(censoredMeta));
     }
 
     if (_logFileWriter != null) {
--- a/lib/services/finamp_logs_helper.dart
+++ b/lib/services/finamp_logs_helper.dart
@@ -5,12 +5,14 @@ import 'dart:math';
 import 'package:clipboard/clipboard.dart';
 import 'package:file_picker/file_picker.dart';
 import 'package:finamp/services/censored_log.dart';
+import 'package:finamp/services/log.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path_helper;
 import 'package:path_provider/path_provider.dart';
 import 'package:share_plus/share_plus.dart';
+import 'package:get_it/get_it.dart';
 
 class FinampLogsHelper {
   final List<LogRecord> logs = [];
@@ -60,6 +62,15 @@ class FinampLogsHelper {
 
   Future<String> getFullLogs() async {
     final fullLogsBuffer = StringBuffer();
+
+    // Get the Log instance and add its metadata at the top
+    final logMeta = GetIt.instance.isRegistered<Log>() ? GetIt.instance<Log>() : null;
+    if (logMeta != null) {
+      fullLogsBuffer.writeln("=== Device/App/Server Info ===");
+      fullLogsBuffer.writeln(jsonEncode(logMeta.toJson()));
+      fullLogsBuffer.writeln("==============================");
+    }
+
     if (_logFileWriter != null) {
       final basePath = (Platform.isAndroid || Platform.isIOS)
           ? await getApplicationDocumentsDirectory()
--- a/lib/services/jellyfin_api_helper.dart
+++ b/lib/services/jellyfin_api_helper.dart
@@ -135,9 +135,8 @@ class JellyfinApiHelper {
       List<BaseItemDto> output = [];
       // Limit itemIds per request to 200.  Execute up to 10 requests in parallel.
       for (final slice in itemIds!.slices(2000)) {
-        final futures = slice
-            .slices(200)
-            .map((subSlice) => _fetchGetItemsResponse(itemIds: subSlice, fields: fields));
+        final futures = slice.slices(200).map((subSlice) =>
+            _fetchGetItemsResponse(itemIds: subSlice, fields: fields));
         final results = await Future.wait(futures);
         for (var subSliceResult in results) {
           output.addAll(subSliceResult.items ?? []);
@@ -155,7 +154,7 @@ class JellyfinApiHelper {
       searchTerm: searchTerm,
       itemIds: itemIds,
       filters: filters,
-      fields:fields,
+      fields: fields,
       recursive: recursive,
       artistType: artistType,
       genreFilter: genreFilter,
@@ -203,7 +202,7 @@ class JellyfinApiHelper {
     return response;
   }
 
-  Future<QueryResult_BaseItemDto> _fetchGetItemsResponse ({
+  Future<QueryResult_BaseItemDto> _fetchGetItemsResponse({
     BaseItemDto? parentItem,
     BaseItemDto? libraryFilter,
     String? includeItemTypes,
@@ -227,7 +226,8 @@ class JellyfinApiHelper {
       // since this error usually happens because the listeners on MusicScreenTabView
       // update milliseconds before the page is popped.
       // This shouldn't happen in normal use.
-      return QueryResult_BaseItemDto(totalRecordCount: 0, startIndex: 0, items: []);
+      return QueryResult_BaseItemDto(
+          totalRecordCount: 0, startIndex: 0, items: []);
     }
     assert(!FinampSettingsHelper.finampSettings.isOffline);
     assert(itemIds == null || parentItem == null);
@@ -239,6 +239,12 @@ class JellyfinApiHelper {
       _jellyfinApiHelperLogger.fine("Getting children of ${parentItem.name}");
     } else if (itemIds != null) {
       _jellyfinApiHelperLogger.fine("Getting items with ids $itemIds");
+      if (itemIds.isEmpty) {
+        // An empty itemIds list will not apply a filter to the results, but
+        // it should actually return no results
+        return QueryResult_BaseItemDto(
+            totalRecordCount: 0, startIndex: 0, items: []);
+      }
     } else {
       _jellyfinApiHelperLogger.fine("Getting items.");
     }
@@ -726,8 +732,7 @@ class JellyfinApiHelper {
 
   /// Gets a Playlist
   Future<dynamic> getPlaylist(BaseItemId playlistId) async {
-    final response =
-        await jellyfinApi.getPlaylist(playlistId: playlistId);
+    final response = await jellyfinApi.getPlaylist(playlistId: playlistId);
     return response;
   }
 
@@ -784,6 +789,7 @@ class JellyfinApiHelper {
   Future<void> updateItem({
     /// The item id.
     required BaseItemId itemId,
+
     /// the new Item.
     required BaseItemDto newItem,
   }) async {
@@ -795,14 +801,15 @@ class JellyfinApiHelper {
   }
 
   /// Updates playlist.
-  Future<void> updatePlaylist({
-    /// The item id.
-    required BaseItemId itemId,
-    /// The new Item.
-    required NewPlaylist newPlaylist    
-  }) async {
-    final response =
-        await jellyfinApi.updatePlaylist(playlistId: itemId, playlist: newPlaylist);
+  Future<void> updatePlaylist(
+      {
+      /// The item id.
+      required BaseItemId itemId,
+
+      /// The new Item.
+      required NewPlaylist newPlaylist}) async {
+    final response = await jellyfinApi.updatePlaylist(
+        playlistId: itemId, playlist: newPlaylist);
     if (response.toString().isNotEmpty) {
       throw response as Object;
     }
--- a/lib/services/log.dart
+++ b/lib/services/log.dart
@@ -1,6 +1,11 @@
 import "package:device_info_plus/device_info_plus.dart";
 import 'package:package_info_plus/package_info_plus.dart';
+import 'package:get_it/get_it.dart';
 
+import 'finamp_user_helper.dart';
+import 'jellyfin_api.dart';
+
+// Collects device information from the platform
 class DeviceInfo {
   String deviceName;
   String deviceModel;
@@ -12,6 +17,7 @@ class DeviceInfo {
     required this.osVersion,
   });
 
+  // Factory method to create DeviceInfo from the current platform
   static Future<DeviceInfo> fromPlatform() async {
     final deviceInfoPlugin = DeviceInfoPlugin();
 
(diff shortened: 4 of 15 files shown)


aman-boop:
waiting for feedback (.

lib/services/log.dart
```diff
@@ -0,0 +1,227 @@
...
+    final userHelper = GetIt.instance<FinampUserHelper>();
+    final user = userHelper.currentUser;
+    if (user == null) {
+      // Fallback if no user is logged in
+      return ServerInfo(
+        serverAddress: 'REDACTED',
+        serverPort: 443,
+        serverProtocol: 'https',
+      );
+    }
```
Chaphasilor: If no user is logged in, we should return `null`. So modify the method to return a `Future<ServerInfo?>` instead.

lib/services/log.dart
```diff
@@ -0,0 +1,227 @@
...
+    return Log(
+      deviceInfo: deviceInfo,
+      appInfo: appInfo,
+    );
+  }
+
+  // Serializes log info, fetching server info on demand
+  Future<Map<String, dynamic>> toJson() async{
+    // Fetch server info on demand, this can be slow so we do it here
+    final serverInfo = await ServerInfo.fromServer();
```
Chaphasilor: This is actually not true, it just parses the BaseURL. So please move it into `create()` :)

lib/services/log.dart
```diff
@@ -0,0 +1,227 @@
...
+
+  // Serializes server info to a map, optionally including the server version
+  Map<String, dynamic> toJson({String? serverVersion}) {
+    return {
+      'serverAddress': serverAddress,
+      'serverPort': serverPort,
+      'serverProtocol': serverProtocol,
+      if (serverVersion != null) 'serverVersion': serverVersion,
+    };
+  }
```
Chaphasilor: All of these `toJson()` methods should be generated via the `@JsonSerializable` annotation and code generation :)

lib/services/log.dart
```diff
@@ -0,0 +1,227 @@
...
+class DeviceInfo {
+  String deviceName;
+  String deviceModel;
+  String osVersion;
+
+  DeviceInfo({
+    required this.deviceName,
+    required this.deviceModel,
+    required this.osVersion,
+  });
```
Chaphasilor: can we also just include the `platform`? Might be obvious, but there's also no good reason to leave it out :)

lib/setup_logging.dart
```diff
@@ -1,21 +1,24 @@
...
+  // Create and store the Log instance for later use
+  final log = await Log.create();
+  GetIt.instance.registerSingleton<Log>(log);
+
   Logger.root.level = Level.ALL;
   Logger.root.onRecord.listen((event) {
     final finampLogsHelper = GetIt.instance<FinampLogsHelper>();
 
+    // Example: You can now use log.toJson() or log fields here if needed
+
```
Chaphasilor: please remove this, it's unnecessary.

lib/setup_logging.dart
```diff
@@ -28,42 +31,4 @@ Future<void> setupLogging() async {
...
-    } else {
-      final webInfo = await deviceInfo.webBrowserInfo;
-      deviceInfoString =
-          "Web browser ${webInfo.userAgent} on ${webInfo.platform}";
-    }
-    startupLogger.info("Running on $deviceInfoString.");
-  } catch (e) {
-    startupLogger.warning("Failed to get device info: $e");
-  }
 }
```
Chaphasilor, line 43: One issue with doing it like this is that we don't log the metadata on startup anymore, only when exporting.  
But having a trace of the startups and versions is definitely useful. So please log the metadata on startup

lib/services/log.dart
```diff
@@ -0,0 +1,227 @@
...
+  String packageName;
+  String version;
+  String buildNumber;
+
+  AppInfo({
+    required this.appName,
+    required this.packageName,
+    required this.version,
+    required this.buildNumber,
+  });
```
Chaphasilor: What's missing here is the version history that we discussed in the old PR. Where we check if the current version is different from the previous version, and if it is we push the timestamp and new version to a list, that we then also export.

Chaphasilor (dismissed):
Haven't tested it yet, but it looks pretty clean so far!  
There are a few things that I'd like to change before merging this though.  
Once everything else is working I can also implement getting some additional info from the server via HTTP requests :)

Let me know if you need help with any changes!

lib/services/log.dart
EpicDarkGalaxy: That horror, I'll add that soon.

lib/services/log.dart
aman-boop: It now returns NULL.

.gitignore
```diff
@@ -125,3 +125,6 @@ app.*.map.json
 
 # generated localization files
 **/lib/l10n/*.dart
+linux/flutter/generated_plugin_registrant.cc
```
Komodo5197: Aren't these already covered by lines 99-101?

lib/services/log.dart
```diff
@@ -0,0 +1,247 @@
...
+      final version = response.body?['Version'] ?? response.body?['version'];
+      return version?.toString();
+    } catch (_) {
+      return null;
+    }
+  }
+}
+
+/// Extension method to redact sensitive server address.
+extension ServerInfoCensor on ServerInfo {
```
Komodo5197: Why are we creating extensions for classes we created instead of just adding these methods directly to the class?

lib/services/finamp_logs_helper.dart
```diff
@@ -60,6 +62,16 @@ class FinampLogsHelper {
 
   Future<String> getFullLogs() async {
     final fullLogsBuffer = StringBuffer();
+
+    // Get the Log instance and add its metadata at the top
+    final logMeta = GetIt.instance.isRegistered<Log>() ? GetIt.instance<Log>() : null;
```
Komodo5197: I don't see a reason to store the log metadata in GetIt like this.  I think we should create a fresh Log instance on export to make sure that it is fully up to date with everything.

lib/setup_logging.dart
```diff
@@ -1,17 +1,23 @@
...
 import 'package:logging/logging.dart';
-import 'package:package_info_plus/package_info_plus.dart';
-import 'package:device_info_plus/device_info_plus.dart';
+import 'package:hive_ce/hive.dart';
 
 import 'services/finamp_logs_helper.dart';
+import 'services/log.dart';
 
 Future<void> setupLogging() async {
+  await Hive.openBox<dynamic>('app_info');
```
Komodo5197: The problem with this is that setupLogging is called too early in main to have access to hive.  It might be worth using something simpler like shared_preferences for this data?

EpicDarkGalaxy: Chaphasilor said that it's not cross-platform so use hive

.gitignore
```diff
@@ -121,10 +121,4 @@ app.*.map.json
 !**/ios/**/default.perspectivev3
 !/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
 !/dev/ci/**/Gemfile.lock
-!.vscode/settings.json
-
-# generated localization files
-**/lib/l10n/*.dart
```
Komodo5197: The l10n lines need to stay.  The localization dart files get autogenerated on build.

lib/setup_logging.dart
Komodo5197: I'm pretty sure it is cross platform, including desktop.
https://pub.dev/packages/shared_preferences
And it would be nice to get this separate from hive, because hive issues are the primary cause of version migration issues.

.gitignore
EpicDarkGalaxy: i am dead!

lib/setup_logging.dart
EpicDarkGalaxy: okay

Chaphasilor: Ah interesting, didn't know there was a Flutter plugin for that! I only knew the Android SharedPreferences, which are Android specific, unless I'm mistaken...

In that case we can try to use that. But the only real data I'd like to store is the version/upgrade history, so maybe adding a whole new store is overkill?

lib/services/log.dart
Chaphasilor: Vibe coding 😅

EpicDarkGalaxy: 🥲🥲🥲🥲🥲

lib/setup_logging.dart
Komodo5197: The actual SharedPrefrences are an android thing, but it looks like ios/mac have something similar, and the plugin just puts a json file in the applicationsupport directory on desktop.  The plugin usage seems trivial without all the initialization of hive, and storing version info in a simpler manner that doesn't need initialization makes sense when we want to log this info before the main hive initialization.  I don't think adding it is too crazy.

On a related intialization order note, I think the server info field will never be full on the startup run because we're running before the user helper is initialized.  I'm not sure how much I actually care about this when we will eventually get this logged during the export, though.

Chaphasilor:
@Komodo5197 I'm now using the shared preferences plugin, and made some other fixes. What do you think?  
I'd like to get this PR in before we release anything to stable, so that we can tell who upgraded from stable and who from beta :)

lib/setup_logging.dart
```diff
@@ -1,73 +1,38 @@
...
   Logger.root.onRecord.listen((event) {
-    final finampLogsHelper = GetIt.instance<FinampLogsHelper>();
 
     // We don't want to print log messages from the Flutter logger since Flutter prints logs by itself
     if (kDebugMode && event.loggerName != "Flutter") {
       debugPrint("[${event.loggerName}/${event.level.name}] ${event.time}: ${event.message}");
     }
-    if (kDebugMode && event.loggerName != "Flutter" && event.getStack != null) {
-      debugPrintStack(stackTrace: event.getStack);
+    if (kDebugMode && event.loggerName != "Flutter" && event.stackTrace != null) {
```
Komodo5197: These should stay getStack.  The native stackTrace does not cover all possible ways for a log to have a stack trace.

lib/services/log.dart
```diff
@@ -0,0 +1,223 @@
...
+  /// Extracts server info from the current user's base URL.
+  static Future<ServerInfo?> fromServer() async {
+    final userHelper = GetIt.instance.isRegistered<FinampUserHelper>() ? GetIt.instance<FinampUserHelper>() : null;
+    final jellyfinApiHelper = GetIt.instance.isRegistered<JellyfinApiHelper>()
+        ? GetIt.instance<JellyfinApiHelper>()
+        : null;
+    final user = userHelper?.currentUser;
+    if (user == null) return null;
+
+    final serverInfo = await jellyfinApiHelper?.loadServerPublicInfo();
```
Komodo5197: We should have this in a try/catch and a timeout so that being offline or having server issues doesn't prevent you from exporting logs.  Maybe we should have more try/catch in general throughout this, because we really don't want log sharing to fail

lib/services/log.dart
```diff
@@ -0,0 +1,223 @@
...
+  Map<String, dynamic> toJson() => _$ServerInfoToJson(this);
+
+  /// Extracts server info from the current user's base URL.
+  static Future<ServerInfo?> fromServer() async {
+    final userHelper = GetIt.instance.isRegistered<FinampUserHelper>() ? GetIt.instance<FinampUserHelper>() : null;
+    final jellyfinApiHelper = GetIt.instance.isRegistered<JellyfinApiHelper>()
+        ? GetIt.instance<JellyfinApiHelper>()
+        : null;
+    final user = userHelper?.currentUser;
+    if (user == null) return null;
```
Komodo5197: So I believe this check prevents this data from every being logged during the startup, because we call this too early.  We should probably be more explicit about this, with at least a comment or something.  Having the app wait on a server query in main() before the app becomes interactive seems bad, so we should be explicit that we aren't and shouldn't do that.  Maybe pass an explicit flag to not collect this info during the startup log creation?

lib/setup_logging.dart
```diff
@@ -1,73 +1,38 @@
...
-      final webInfo = await deviceInfo.webBrowserInfo;
-      deviceInfoString = "Web browser ${webInfo.userAgent} on ${webInfo.platform}";
-    }
-    startupLogger.info("Running on $deviceInfoString.");
-  } catch (e) {
-    startupLogger.warning("Failed to get device info: $e");
-  }
+  // Create and store the Log instance for later use
+  final log = await Log.create();
+  startupLogger.info(jsonEncode(await log.toJson())); // Log metadata on startup
```
Komodo5197: Does json produce a reasonable looking result in the log, with good formatting?

lib/services/log.dart
```diff
@@ -0,0 +1,223 @@
...
+
+  /// Detects app metadata using package_info_plus and updates stored version history.
+  static Future<AppInfo> fromPlatform() async {
+    final packageInfo = await PackageInfo.fromPlatform();
+    final currentVersion = "${packageInfo.version} (${packageInfo.buildNumber})";
+
+    final SharedPreferencesAsync prefs = SharedPreferencesAsync();
+
+    final history = List<String>.from(
+      (await prefs.getStringList(_SharedPreferencesVersionHistoryKey) ?? <String>[]),
```
Komodo5197: I'd prefer a try/catch around the shared preferences to be extra sure exporting logs can never fail.

lib/setup_logging.dart
Chaphasilor: I'll have to try again, but I changed it because flutter complained that `.getStack` doesn't exist and couldn't find an import.

lib/services/log.dart
Chaphasilor: Fair point. Is there some kind of inline try-catch in Dart, that will log an error and return null if an exception is thrown? That would make things much less verbose...  
Maybe something similar to the `unawaited()` helper

lib/services/log.dart
Chaphasilor: I guess we could do that. We could also log server data separately after the user helper is available, although I don't really think we need it at all, if it's shared with logs anyway.

lib/setup_logging.dart
Chaphasilor: No, not with good formatting. It is easier to parse if it's just a single line though, in case we ever want to do that.  
I could also log both formats, first line in JSON for potential parsing, and second line in properly-formatted plain text?

lib/setup_logging.dart
Chaphasilor: okay yeah the `censored_log` wasn't imported, and dart wasn't smart enough to find it

lib/services/jellyfin_api_helper.dart
```diff
@@ -465,13 +465,21 @@ class JellyfinApiHelper {
...
-  Future<PublicSystemInfoResult?> loadServerPublicInfo() async {
+  Future<PublicSystemInfoResult?> loadServerPublicInfo({Duration? timeout}) async {
     // Some users won't have a password.
-    if (baseUrlTemp == null) {
+    if (_finampUserHelper.currentUser?.baseURL == null && baseUrlTemp == null) {
       return null;
     }
 
     var response = await jellyfinApi.getPublicServerInfo();
+    if (timeout != null) {
```
Komodo5197, line 475: This seems wrong.  You're awaiting the future before attaching the timeout.  I'm not sure the typing even makes sense?

lib/services/finamp_logs_helper.dart
```diff
@@ -59,6 +60,22 @@ class FinampLogsHelper {
 
   Future<String> getFullLogs() async {
     final fullLogsBuffer = StringBuffer();
+
+    // Get the Log instance and add its metadata at the top
+    final logMeta = await EnvironmentMetadata.create();
+
+    // Prepend this metadata to the logs
+    fullLogsBuffer.writeln(jsonEncode(logMeta.toJson()));
```
Komodo5197: If we have the readable version, I'd skip the json one.

lib/services/finamp_logs_helper.dart
```diff
@@ -59,6 +60,22 @@ class FinampLogsHelper {
...
+    fullLogsBuffer.writeln("=== METADATA ===");
+    fullLogsBuffer.writeln("Device Info: ${logMeta.deviceInfo.pretty}");
```
Komodo5197: We have to print this twice, so maybe we should do it in a pretty method on the main metadata class.

PR #1249, opened 2025-06-06
Some OfflineMode Tweaks

flloschy:
Added Info label to display the current state (translations to be added soon:tm:)
image
image

Naturally the label also shows this
image

Made that the override can override both offline mode to be enabled and disabled instead of just enabled as before

Changed the icon of Offline Mode to show when the automation is reevaluating (since the 7 second delay exists)
image

Removed the LifeCycle from AutoOffline.startWatching since we are unsure why I added this in the first place haha

And Tweaked the AutoOffline code a bit in therms of readability (shouldnt introduce functional differences)


Chaphasilor:
Nice!  
I think we can find some less technical wording, so instead of "automation active", maybe "Will switch automatically" or something like that? Open for suggestions. Same for "Override active", how about "Automatic switching paused"?

I'm also wondering, how do I re-enable the automation? If offline mode was turned on automatically and I turn it off, do I then have to turn it back on for the automation to kick in again? Do I have to have no connectivity when doing this, so that it counts as a valid state?  
Maybe there should just be a button when the override is active, which allows disabling the override?

Oh and unrelated, but it looks like you have some kind of gradient effect going on? Which platform is that screenshot from, Arch?

lib/services/network_manager.dart
```diff
@@ -90,35 +90,40 @@ Future<void> _onConnectivityChange(
...
   // this prevents an issue on ios (and mac?) where the
-  // listener gets called even though it shouldnt.
+  // listener gets called even though it shouldn't.
   // The wait also acts as an timeout so offline mode is less
-  // likely to engage when it doesnt need to and this helps
+  // likely to engage when it doesn't need to and this helps
   // with queue reloading
+  autoOfflineStatusStream.add(++activeDelayCounter);
   await Future.delayed(Duration(seconds: 7), () => {});
+  autoOfflineStatusStream.add(--activeDelayCounter);
```
Chaphasilor, line 111: why exactly is this a counter? due to parallel executions?

flloschy: exactly
a boolean could make the icon flicker

lib/components/MusicScreen/offline_mode_status_label.dart
```diff
@@ -0,0 +1,35 @@
...
+class OfflineModeStatusLabel extends ConsumerWidget {
+  const OfflineModeStatusLabel({
+    super.key,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {    
+    AutoOfflineOption automation = ref.watch(finampSettingsProvider.autoOffline);
+
+    String displayText = "Automation disabled";
```
Komodo5197: I'd prefer to just not show anything at all if the setting is disabled.

lib/components/MusicScreen/offline_mode_switch_list_tile.dart
```diff
@@ -11,20 +12,23 @@ class OfflineModeSwitchListTile extends ConsumerWidget {
...
       onChanged: (value) {
-        // when offline mode is enabled manually,
-        // prevent the auto offline feature from turning off
-        // offline mode automatically. When offline mode
-        // is manually turned off, reenable the listener.
-        FinampSetters.setAutoOfflineListenerActive(!value);
+        // when using the switch it toggles the automation on and off again
+        // this is so the user can overwrite the automation at any time with
+        // any value.
+        bool currentValue = FinampSettingsHelper.finampSettings.autoOfflineListenerActive;
```
Komodo5197: Now that we have a line dedicated to the automation, we might be able to make the behavior clearer by always disabling on manual switch flips, and then adding a reenable button to the automation disabled line.

lib/services/network_manager.dart
```diff
@@ -24,36 +24,36 @@ Logger _networkAutomationLogger = Logger("Network Automation");
 Logger _autoOfflineLogger = Logger("Auto Offline");
 Logger _networKSwitcherLogger = Logger("Network Switcher");
 
+int activeDelayCounter = 0;
+
+/// This stream receives update when autoOffline enters/exists the 7 second confirmation/validation timeout
+final autoOfflineStatusStream = StreamController<int>.broadcast();
```
Komodo5197, line 29: This might be slightly simpler if you just make a provider class with an increment/decrement method.

flloschy: I felt like this is a lot of code to achieve the same thing in a file which is already all over the place, and this isnt too unreadble i think

lib/components/MusicScreen/offline_mode_switch_list_tile.dart
flloschy: a button or a button which looks like a link?

lib/components/MusicScreen/offline_mode_status_label.dart
flloschy: that was my original plan but there is no Empty widget which could have been returned so I did this instead

lib/services/network_manager.dart
Komodo5197: If prefer this solution that's fine.  I think you can at least directly return the stream instead of yielding its element when building the provider, though.

lib/components/MusicScreen/offline_mode_status_label.dart
Komodo5197: Can't you just use SizedBox.shrink()?

lib/components/MusicScreen/offline_mode_switch_list_tile.dart
Komodo5197: I think a button-looking thing would probably look better?  But UI design isn't really my expertise.

flloschy: Mmm Thinking about this both options arent optimal...

using the same switch for offline mode and automation overrides is confusing

and having an additional button which is only visible when using the switch could be a lot to take in and make the UI complicated and probably also confusing and overwhelming 

on the other hand Chaphasilors new home page comes ... at some point ... so both solutions (current function / extra button) would be temporary to a certain extend anyway

Chaphasilor: I don't think adding a clearly labeled button to get out of a non-default state is overly complicated or confusing :)  
It's also fine if the button is wrapped into the next line. I think a `SimpleButton` with an icon and text should be sufficient, there's probably not really enough space to add one of the CTAs...

Chaphasilor:
I think we should rename "Prefer Home Network" to "Prefer Local Address", and "Home Network Address" to "Local Address". Would you mind throwing that into this PR as well?

flloschy:
my testing concluded that everything works as it should 
translations still need to be added though

Chaphasilor:
Just tried it out and can confirm that it's working. The logic for showing the different icons was slightly broken, but I've fixed that and added more granular options.  
I really am a fan of the "button" that's just a title and description, that's really clear. I just highlighted it a bit more to emphasize that it's clickable, and updated the wording to make it (hopefully) clearer and shorter.

Take a look!

Chaphasilor:
Thanks for the PR <3

PR #1252, opened 2025-06-08
Minor changes for Finamp desktop

Maxr1998:


lib/components/LayoutSettingsScreen/TabsSettingsScreen/hide_tab_toggle.dart
```diff
@@ -16,15 +16,19 @@ class HideTabToggle extends ConsumerWidget {
...
-    return SwitchListTile.adaptive(
-      title: Text(tabContentType.toLocalisedString(context)),
-      secondary: ReorderableDragStartListener(
-        index: index,
-        child: const Icon(Icons.drag_handle),
+    return ReorderableDelayedDragStartListener(
+      index: index,
+      child: SwitchListTile.adaptive(
+        title: Text(tabContentType.toLocalisedString(context)),
+        secondary: ReorderableDragStartListener(
```
Komodo5197, line 23: What's with this nested drag listener?

Chaphasilor: I think the delayed handler is for long press drag on the tab item, and the non-delayed one is for dragging on the dedicated drag handle?  
Just to have both options

Maxr1998: Pretty much that. If you don't disable `buildDefaultDragHandles`, `ReorderableListView` also adds both of those. We have to do this manually, since it is unfortunately not possible to control just the desktop behavior here..

Chaphasilor (approved):
Seems fine, can't test the linux stuff though...

Maxr1998:
I tested it on my machine, still builds and runs fine.

PR #1253, opened 2025-06-08
Support word-by-word synced lyrics

Maxr1998:
Still far from finished, but this branch already contains a good base to work on, as well as some cleanup that I'll probably push to redesign earlier than the rest.

### DIFF
--- a/lib/screens/lyrics_screen.dart
+++ b/lib/screens/lyrics_screen.dart
@@ -555,25 +555,24 @@ class LyricsLinePainter extends ChangeNotifier implements CustomPainter {
       _paintWithWordHighlighting(canvas, size);
     } else {
       // Default painting for non-current lines or lines without cues
-      // Calculate the offset based on text alignment for consistent alignment
-      double offsetX = 0.0;
-      switch (textAlign) {
-        case TextAlign.start:
-        case TextAlign.left:
-          offsetX = 0.0;
-          break;
-        case TextAlign.center:
-          offsetX = (size.width - textPainter.width) / 2.0;
-          break;
-        case TextAlign.end:
-        case TextAlign.right:
-          offsetX = size.width - textPainter.width;
-          break;
-        case TextAlign.justify:
-          offsetX = 0.0; // Justify behaves like start for single lines
-          break;
-      }
+      final offsetX = _calculateTextAlignmentOffset(size.width, textPainter.width);
       textPainter.paint(canvas, Offset(offsetX, 0.0));
+    }
+  }
+
+  /// Calculates the horizontal offset for text alignment
+  double _calculateTextAlignmentOffset(double containerWidth, double textWidth) {
+    switch (textAlign) {
+      case TextAlign.start:
+      case TextAlign.left:
+        return 0.0;
+      case TextAlign.center:
+        return (containerWidth - textWidth) / 2.0;
+      case TextAlign.end:
+      case TextAlign.right:
+        return containerWidth - textWidth;
+      case TextAlign.justify:
+        return 0.0; // Justify behaves like start for single lines
     }
   }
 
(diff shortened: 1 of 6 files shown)


lukaslindnermusic:
https://github.com/jmshrv/finamp/pull/1253/commits/0a074af699a8285fe7ddb604841d7216b6df845e thanks <3 that annoyed me as well :P 
Would you mind also fixing the Sort Order Settings on the Artist and Genre Screen Settings for the Item Sections and the Filter Type Order? Thanks <3

lib/screens/lyrics_screen.dart
```diff
@@ -465,15 +467,13 @@ class _LyricLine extends ConsumerWidget {
...
-                (line.text?.trim().isNotEmpty ?? false) &&
-                (finampSettings?.showLyricsTimestamps ?? true))
+            if (showTimestamp)
               WidgetSpan(
                 alignment: PlaceholderAlignment.bottom,
                 child: Padding(
                   padding: const EdgeInsets.only(right: 8.0),
                   child: Text(
-                    "${Duration(microseconds: (line.start ?? 0) ~/ 10).inMinutes}:${(Duration(microseconds: (line.start ?? 0) ~/ 10).inSeconds % 60).toString().padLeft(2, '0')}",
+                    "${Duration(microseconds: line.startMicros).inMinutes}:${(Duration(microseconds: line.startMicros).inSeconds % 60).toString().padLeft(2, '0')}",
```
Chaphasilor: I believe this won't work as the server sends timestamps in "ticks"?

Maxr1998:
> [0a074af](https://github.com/jmshrv/finamp/commit/0a074af699a8285fe7ddb604841d7216b6df845e) thanks <3 that annoyed me as well :P

I just merged that as part of #1252 :grin:

> Would you mind also fixing the Sort Order Settings on the Artist and Genre Screen Settings for the Item Sections and the Filter Type Order? Thanks <3

I can look into those, sure!

lib/screens/lyrics_screen.dart
Maxr1998: Yes, it still sends ticks, the `startMicros` is simply a getter that extracts this common conversion.

lukaslindnermusic:
> I just merged that as part of #1252 😁
> > Would you mind also fixing the Sort Order Settings on the Artist and Genre Screen Settings for the Item Sections and the Filter Type Order? Thanks <3
> 
> I can look into those, sure!

Thanks <3 It's basically the same as the Tab List 😊

Maxr1998:
> Thanks <3 It's basically the same as the Tab List 😊

Fixed in 09caf159901e606c0129742d8001e90df8c8d9cc.

lib/screens/lyrics_screen.dart
Chaphasilor: Ah, I see. I had it backwards in my mind 🙈

Chaphasilor:
what's the status of this PR? 0.9.19 will probably be the last release before 10.11 drops, and I'm targeting this weekend for release atm

Maxr1998:
I think I'll only be able to finish it at it after 10.11 is released, realistically. But with the current RC3 I could at least work on it in the meantime, the server API should be good.

Maxr1998:
If you can finish it before this weekend, sure. I somehow feel like the whole thing might be a bit more complex, but maybe also not. Though, is it really set that 10.11 will drop soon enough that there won't be another Finamp release till then?

Chaphasilor:
Given that our release cadence is ~2 months, I'm pretty sure. I just would like Finamp to support all newly-released features right from the start

Chaphasilor:
@Komodo5197 how mad would you be if I merge this in its current state? 😅

Komodo5197:
I haven't even fully read it yet... It looks like no meaningful changes outside the lyrics screen, though?

Maxr1998:
> Yes, it's just the lyrics screen. Just not the best code, but all I could manage until now. I just want to make sure this feature is in Finamp before Jellyfin adds support for it, and since we don't know when exactly 10.11 will be released, I'd like to get this in with 0.9.19 :)  
> Since @Maxr1998's and my testing was positive, even on 10.10, I'll just merge this in now :) 

Sounds good to me. I glanced at the code earlier and it seemed fine. There are a few optimizations I'd like to make but that can wait until after the next release.

PR #1255, opened 2025-06-13
[Redesign] Improve error handling

Komodo5197:
Stop riverpod from eating errors inside providers.  Prevent error during AutoOffline.startWatching from the settings provider not being fully initialized.  Properly log errors that occur outside of flutter stack frames.  Always create snackbar for assert failures to reduce the chance that they're missed while debugging.  Throw assert error if jellyfin api calls occur during widget build().

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -113,7 +113,7 @@ void main() async {
     _mainLog.info("Setup offline listen tracking");
     await _setupDownloadsHelper();
     _mainLog.info("Setup downloads service");
-    _setupProviders();
+    await _setupProviders();
     _mainLog.info("Setup providers");
     await _setupOSIntegration();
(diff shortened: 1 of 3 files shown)


Chaphasilor:
didn't notice any issues, only thing I did was include [`didChangeDependencies`](https://api.flutter.dev/flutter/widgets/State/didChangeDependencies.html) as afaik network request are also fine in there.  
The rest is just formatting...

Komodo5197:
I think the scenarios where didChangeDependancies would be used for network requests are generally filled by our use of riverpod.  None of this is really hard and fast - you can technically get away with doing all the requests in build() if you use the state correctly, and you can accidentally spam network requests from anywhere if you really want.  I mostly just wanted to stop people from directly putting api requests in widget build().

Issue #1224, opened 2025-05-20
Queue / Now Playing often not restored at startup


Terrance:
### Type:

- [x] Bug
- [ ] Major Bug (makes the app unusable)
- [ ] Other

### What version of Finamp are you using?
- [x] Beta/Redesign: 0.9.16
- [ ] Stable

### Affected Device/Platform

- [x] Android
- [-] Android Auto
- [-] iOS
- [-] Linux
- [-] Windows
- [-] Mac

### Description & Steps to Reproduce

My queue only seems to get auto-restored about 50% of the time; the other half, I have to retrieve it myself via _Restore Now Playing_.

Observations from some experimentation:
* restoring manually without starting playback, then closing and reopening the app loses it again
* restoring and playing seems to persist for some time, but not across a force-close (which may also include the OS stopping the app in the background at some later date)

### Logs

📎 [2025-05-20-finamp-logs.txt](https://github.com/user-attachments/files/20357888/2025-05-20-finamp-logs.txt)

### Further information

https://github.com/user-attachments/assets/6f7eced0-7287-47be-bfec-77cae08a2868

 _Originally posted in https://github.com/jmshrv/finamp/issues/1215#issuecomment-2887798294_

Komodo5197:
The loss of the manual restore if you don't play is intentional.  We don't start saving the queue after a restore until it is 'accepted' by the user actually playing it for a bit.  This helps avoid saving queue that only partially loaded due to offline mode or connection issues.  For other cases where it doesn't restore, I'm not really sure what's going on.  I've added a bit more logging to the restore process so if this keeps up in the next beta release, the logs should hopefully be more enlightening.

Terrance:
Is there a reason not to just always restore the queue?  If your last session is restored and you don't want to continue, you can just start playing something else instead.

Terrance:
> swiping down on the now playing bar

I didn't know that was even a gesture until now 🙃

> something unusual with the OS that would send a full stop command instead of just a pause

This device is sometimes a bit memory- or process-constrained, so it would not surprise me if the OS decides to stop the app after a few days without use.  If I stop playback, leave the app and come right back, the queue's still there.  It tends to be when I come back the next day or two later that it's disappeared.

>  You load up the app and the auto restore just doesn't happen?

Correct, there's no sign of the bottom bar at all; no temporary loading state that I can see.

Komodo5197:
[...]

PR #1265, opened 2025-06-18
[Redesign] Add setting to ignore OS stop commands

Komodo5197:
#1224 seems to be related to the OS sending unexpected 'stop' commands, killing the queue.  I've added a setting to just ignore these commands, requiring you to use the in-app gestures if you want to close the player.  Off by default.  Also:
- Prevent non-playback commands like UserDataChanged from triggering aggressive playback reporting.
- Add a throttle to aggressive playback reporting when controlled - desktop player seems to fire this ~14 times per second.
- Capture stacktraces that were being missed when directly logging Errors without manually passing the stacktrace.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -2498,6 +2498,14 @@
   "@autoReloadQueueSubtitle": {
     "description": "Subtitle for the setting that controls if the queue is automatically reloaded when the source changes"
   },
+  "ignoreExternalStopTitle": "Ignore External Stop",
+  "@ignoreExternalStopTitle": {
+    "description": "Title for the setting that controls if external stop commands are ignored."
+  },
+  "ignoreExternalStopSubtitle": "Ignore audio 'stop' commands sent by the OS.  Only in-app gestures will stop playback and clear the queue.",
+  "@ignoreExternalStopSubtitle": {
+    "description": "Subtitle for the setting that controls if external stop commands are ignored."
+  },
   "autoReloadPrompt": "Reload queue to apply new {type, select, network{network settings} transcoding{transcoding settings} other{settings}}",
   "@autoReloadPrompt": {
     "description": "Prompt shown when the queue needs to be reloaded after changing network settings",
--- a/lib/models/finamp_models.dart
+++ b/lib/models/finamp_models.dart
@@ -212,6 +212,7 @@ class DefaultSettings {
   static const playlistTracksSortBy = SortBy.defaultOrder;
   static const playlistTracksSortOrder = SortOrder.ascending;
   static const genreFilterPlaylists = false;
+  static const ignoreExternalStopCommands = false;
 }
 
 @HiveType(typeId: 28)
@@ -327,6 +328,7 @@ class FinampSettings {
     this.playlistTracksSortBy = DefaultSettings.playlistTracksSortBy,
     this.playlistTracksSortOrder = DefaultSettings.playlistTracksSortOrder,
     this.genreFilterPlaylists = DefaultSettings.genreFilterPlaylists,
+    this.ignoreExternalStopCommands = DefaultSettings.ignoreExternalStopCommands,
   });
 
   @HiveField(0, defaultValue: DefaultSettings.isOffline)
@@ -688,6 +690,9 @@ class FinampSettings {
 
   @HiveField(116)
   SleepTimer? sleepTimer;
+
+  @HiveField(117, defaultValue: DefaultSettings.ignoreExternalStopCommands)
+  bool ignoreExternalStopCommands;
 
   static Future<FinampSettings> create() async {
     final downloadLocation = await DownloadLocation.create(
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -550,19 +550,12 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
   @override
   Future<void> stop() async {
     try {
-      _audioServiceBackgroundTaskLogger.info("Stopping audio service");
+      _audioServiceBackgroundTaskLogger.info("Audio service received stop command");
+
+      if (FinampSettingsHelper.finampSettings.ignoreExternalStopCommands) return;
 
       final queueService = GetIt.instance<QueueService>();
       await queueService.stopPlayback();
-      // await _player.seek(_player.duration);
-
-      // await handleEndOfQueue();
-
-      // TODO: Do we want to actually cancel the sleep timer if we stop the music?
-      _timer.value?.cancel();
-      _timer.value = null;
-
-      await super.stop();
     } catch (e) {
       _audioServiceBackgroundTaskLogger.severe(e);
       return Future.error(e);
@@ -571,6 +564,10 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
 
   Future<void> stopPlayback() async {
     try {
+      // TODO: Do we want to actually cancel the sleep timer if we stop the music?
+      _timer.value?.cancel();
+      _timer.value = null;
+
       await _player.stop();
     } catch (e) {
       _audioServiceBackgroundTaskLogger.severe(e);
--- a/lib/components/global_snackbar.dart
+++ b/lib/components/global_snackbar.dart
@@ -1,10 +1,10 @@
 import 'dart:async';
 
 import 'package:chopper/chopper.dart';
+import 'package:finamp/l10n/app_localizations.dart';
 import 'package:finamp/services/feedback_helper.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
-import 'package:finamp/l10n/app_localizations.dart';
 import 'package:logging/logging.dart';
 
 @Deprecated("Use GlobalSnackbar.error(dynamic error) instead")
@@ -83,9 +83,6 @@ class GlobalSnackbar {
   static void error(dynamic event) => _enqueue(() => _error(event));
   static void _error(dynamic event) {
     _logger.warning("Displaying error: $event", event);
-    if (event is Error && event.stackTrace != null) {
-      _logger.warning(event.stackTrace);
-    }
     BuildContext context = materialAppNavigatorKey.currentContext!;
     String errorText;
     if (event is Response) {
--- a/lib/models/finamp_models.g.dart
+++ b/lib/models/finamp_models.g.dart
@@ -219,6 +219,7 @@ class FinampSettingsAdapter extends TypeAdapter<FinampSettings> {
       playlistTracksSortBy: fields[113] == null ? SortBy.defaultOrder : fields[113] as SortBy,
       playlistTracksSortOrder: fields[114] == null ? SortOrder.ascending : fields[114] as SortOrder,
       genreFilterPlaylists: fields[115] == null ? false : fields[115] as bool,
+      ignoreExternalStopCommands: fields[117] == null ? false : fields[117] as bool,
     )
       ..disableGesture = fields[19] == null ? false : fields[19] as bool
       ..showFastScroller = fields[25] == null ? true : fields[25] as bool
@@ -230,7 +231,7 @@ class FinampSettingsAdapter extends TypeAdapter<FinampSettings> {
   @override
   void write(BinaryWriter writer, FinampSettings obj) {
     writer
-      ..writeByte(110)
+      ..writeByte(111)
       ..writeByte(0)
       ..write(obj.isOffline)
       ..writeByte(1)
@@ -450,7 +451,9 @@ class FinampSettingsAdapter extends TypeAdapter<FinampSettings> {
       ..writeByte(115)
       ..write(obj.genreFilterPlaylists)
       ..writeByte(116)
-      ..write(obj.sleepTimer);
+      ..write(obj.sleepTimer)
+      ..writeByte(117)
+      ..write(obj.ignoreExternalStopCommands);
   }
 
   @override
--- a/lib/screens/audio_service_settings_screen.dart
+++ b/lib/screens/audio_service_settings_screen.dart
@@ -47,6 +47,7 @@ class _AudioServiceSettingsScreenState extends State<AudioServiceSettingsScreen>
           BufferDisableSizeConstraintsSelector(key: _updateChildren),
           const LoadQueueOnStartupSelector(),
           const AutoReloadQueueToggle(),
+          const IgnoreExternalStopToggle(),
         ],
       ),
     );
@@ -186,6 +187,20 @@ class AutoReloadQueueToggle extends ConsumerWidget {
       subtitle: Text(AppLocalizations.of(context)!.autoReloadQueueSubtitle),
       value: ref.watch(finampSettingsProvider.autoReloadQueue),
       onChanged: FinampSetters.setAutoReloadQueue,
+    );
+  }
+}
+
+class IgnoreExternalStopToggle extends ConsumerWidget {
+  const IgnoreExternalStopToggle({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return SwitchListTile.adaptive(
+      title: Text(AppLocalizations.of(context)!.ignoreExternalStopTitle),
+      subtitle: Text(AppLocalizations.of(context)!.ignoreExternalStopSubtitle),
+      value: ref.watch(finampSettingsProvider.ignoreExternalStopCommands),
+      onChanged: FinampSetters.setIgnoreExternalStopCommands,
     );
   }
 }
--- a/lib/services/censored_log.dart
+++ b/lib/services/censored_log.dart
@@ -9,7 +9,18 @@ import 'finamp_user_helper.dart';
 extension CensoredMessage on LogRecord {
   /// The uncensored log string to be shown to the user
   String get logString =>
-      "[$loggerName/${level.name}] $time: $message${stackTrace == null ? "" : "\n${stackTrace.toString()}"}";
+      "[$loggerName/${level.name}] $time: $message${getStack == null ? "" : "\n${getStack.toString()}"}";
+
+  StackTrace? get getStack {
+    var stack = stackTrace;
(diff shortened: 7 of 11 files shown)


lib/services/music_player_background_task.dart
```diff
@@ -571,6 +567,10 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
 
   Future<void> stopPlayback() async {
     try {
+      // TODO: Do we want to actually cancel the sleep timer if we stop the music?
```
flloschy: I can see a world where both ways can be annoying to a lot of users
Stopping makes more sense to me though

Chaphasilor: We could keep it running as long as it's properly shown in the UI, although I don't recall seeing that anywhere else.  
But the old behavior was that the sleep timer was restarted when playback was resumed (after the timer expired and playback was stopped), which is definitely not what we want :P

lib/l10n/app_en.arb
```diff
@@ -2498,6 +2498,14 @@
   "@autoReloadQueueSubtitle": {
     "description": "Subtitle for the setting that controls if the queue is automatically reloaded when the source changes"
   },
+  "clearQueueOnStopTitle": "Clear Queue On External Stop",
+  "@clearQueueOnStopTitle": {
+    "description": "Title for the setting that controls if OS stop commands clear the queue."
+  },
+  "clearQueueOnStopSubtitle": "Clear the queue and close the player if a 'stop' command is received from the OS.  In-app stop gestures will always close the player and clear the queue.",
```
Maxr1998:
```suggestion
  "clearQueueOnStopSubtitle": "Clear the queue and close the player if a 'stop' command is received from the OS. In-app stop gestures will always close the player and clear the queue.",
```

lib/models/finamp_models.dart
```diff
@@ -689,6 +691,9 @@ class FinampSettings {
   @HiveField(116)
   SleepTimer? sleepTimer;
 
+  @HiveField(117, defaultValue: DefaultSettings.clearQueueOnStop)
+  bool clearQueueOnStop;
```
Maxr1998:
```suggestion
  bool clearQueueOnStopEvent;
```

lib/l10n/app_en.arb
Chaphasilor: Just commit formatting directly :)

Komodo5197:
@Chaphasilor I don't see any way to set up consistent logging of media commands either.  With the newest version of the code, I've made it so that the default behavior of the media stop command does still call stop on the player, releasing the buffers and native resources.  But the player screen and queue don't get disposed, so you can still re-enter the app and press play.  I feel like this is probably what the OS/user actually wants if calling the media stop command?  And then you can toggle the setting if you want the old behavior back.

Chaphasilor:
I guess that makes sense, something like the difference between sleep and hibernate on Windows PCs?  
I don't think anyone would complain that the queue is not cleared, and even if, we now have the "Clear queue and stop playback" option in the player menu.

So this is ready to merge from your side? I'll give it a try in a bit...

lib/services/music_player_background_task.dart
Komodo5197: I think clearing the timer if you call stop makes sense.  With the old behavior especially questioning this seems odd - would you really want to clear the queue, close the player screen, and still have the sleep timer running?  Of course, the old behavior also forgot to do that if the queue service stopPlayback was called directly, which is what we usually do, so it was just not thought through.  With the new behavior maybe not clearing it seems a bit more plausible, but I'm still thinking that clearing it seems correct.  Also, I've noticed that we're saving off the current sleep timer and then never doing anything with it - were there plans for this?  My guess is that we wanted to pre-fill the old values in the sleep timer menu but forgot to implement that.

Komodo5197:
I think everything's pretty much how I wanted it now.  If you're looking it over, you might want to also check out the sleep timer.  I ended up messing with that a bunch while trying to make the menu pre-fill the values from the last used timer.

lib/l10n/app_en.arb
Maxr1998: Did a quick review on my phone so I couldn't 😁

lib/l10n/app_en.arb
```diff
@@ -755,7 +755,16 @@
       },
       "suffix": {
         "type": "String",
-        "example": "minutes or tracks"
+        "example": "minutes"
+      }
```
Chaphasilor: You removed the `suffix` parameter, the example should be moved to the `time` parameter

lib/l10n/app_en.arb
```diff
@@ -755,7 +755,16 @@
...
+    }
+  },
+  "sleepTimerRemainingTracks": "Sleeping {time,plural,=1{at end of track} other{in {time} tracks}}",
```
Chaphasilor:
```suggestion
  "sleepTimerRemainingTracks": "Sleeping {time,plural,=1{after this track} other{in {time} tracks}}",
´´´

File: lib/models/finamp_models.dart [hunk 6]
```diff
@@ -3174,101 +3173,112 @@ enum ArtistItemSections {
...
+    remainingNotifier.value = secondsLength + tracksLength;
 
-    if (type == SleepTimerType.duration) {
-      timer = Timer.periodic(const Duration(seconds: 1), (t) async {
+    if (secondsLength > 0) {
+      _timer = Timer.periodic(const Duration(seconds: 1), (t) async {
         final secondsLeft = remainingDuration.inSeconds;
 
-        remainingNotifier.value = secondsLeft;
+        remainingNotifier.value = secondsLeft + tracksLength;
```
Chaphasilor, line 3178: why are you adding the `tracksLength` (in track count, according to the comment) to the seconds?  
Is that just to not need to differentiate between the two modes, since you ensure that only one of them is ever non-zero?

lib/models/finamp_models.dart
```diff
@@ -3174,101 +3173,112 @@ enum ArtistItemSections {
...
 
         if (secondsLeft <= 0) {
           t.cancel();
-          sleepTimerLogger.info("Sleep timer finished");
-          await this.callback();
+          _timer = null;
+          if (tracksLength > 0) {
+            sleepTimerLogger.info("Sleep timer switching to track count");
```
Chaphasilor, line 3184: I can't think of a scenario where we would ever switch from seconds to tracks for an active timer?

lib/models/finamp_models.dart
```diff
@@ -3174,101 +3173,112 @@ enum ArtistItemSections {
...
+    } else {
+      _tracksRemaining = tracksLength;
     }
 
-    sleepTimerLogger.info(
-        "Sleep timer started for ${type == SleepTimerType.duration ? Duration(seconds: length) : "$length tracks"}, finishTrack: $finishTrack");
+    sleepTimerLogger.info("Sleep timer started for ${Duration(seconds: secondsLength)}, $tracksLength tracks");
+  }
+
+  void trackCompleted() {
```
Chaphasilor:
```suggestion
  void onTrackCompleted() {
```

Maxr1998:
Not sure how much work that would be, but would it make sense to move those formatting changes to a separate PR?

Komodo5197:
It's probably doable, but I've never been particularly concerned with keeping PRs focused.  I don't foresee too much discussion/objections regarding the formatter fixes, so I'd rather just push the whole thing.

lib/services/music_player_background_task.dart
Chaphasilor: I'm not sure what the plan was, but remembering the last value (like you did now) could be useful, but probably isn't a perfect solution either. I'd be happy with either one, but remembering the last value might have less potential to be frustrating.  
And let's stick with clearing the sleep timer on stop, I think we all agree that it makes more sense than the other option, and I don't want to make this customizable unless someone actively requests it.

Chaphasilor (commented):
This review applies to changes until 788b7f189738ba804b071a8a072802b78bac63fe.  
I had no issues with the sleep timer functionality during testing, so that seems to work.  
I think it would've been nice to remember the last used tab/type, and remember the last used value per type, but it's probably not that important, and would add unneeded complexity. We can always add that if someone actually requests that.

Aside from that I'm having issues with the `verifyCallable` assertion, that is getting triggered when trying to reload the music screen tabs. Restoring a queue and downloading works, same as navigating to other screens via the player screen or menus. But the tab content only loads the first time (also not reliably), and otherwise triggers the assertion error.  
Only after a few minutes will the content load for the active tab, but the other tabs remain in a loading state.  
@flloschy also experienced this issue I think: [Discord link](https://discord.com/channels/1064474779043770408/1240083343652028486/1385914124436902040).  
This is also the case on the latest commit.

I don't recall experiencing that when you originally added the `_verifyCallable` check, but I also don't think you changed the check...

lib/models/finamp_models.dart
Komodo5197: This is how the finish track button works now.  We run the seconds timer, then switch to a tracks timer with length 1.

lib/models/finamp_models.dart
Komodo5197: We can end up with both modes in conjunction during duration timers with finish track on, and need to prevent it from ever hitting 0 while the timer is still active.  The value doesn't need to mean anything, we just need the notifier to tick every time we might have a UI update.

lib/l10n/app_en.arb
```diff
@@ -2498,6 +2507,14 @@
   "@autoReloadQueueSubtitle": {
     "description": "Subtitle for the setting that controls if the queue is automatically reloaded when the source changes"
   },
+  "clearQueueOnStopEventTitle": "Clear Queue On External Stop",
+  "@clearQueueOnStopEventTitle": {
+    "description": "Title for the setting that controls if OS stop commands clear the queue."
+  },
+  "clearQueueOnStopEventSubtitle": "Clear the queue and close the player if a 'stop' command is received from the OS. In-app stop gestures will always close the player and clear the queue.",
```
Chaphasilor:
```suggestion
  "clearQueueOnStopEventSubtitle": "Clear the queue and close the player if a 'stop' command is received from the OS. In-app stop actions will always close the player and clear the queue.",
```

lib/l10n/app_en.arb
Komodo5197: The parameter still exists so that the existing translations still work, even though english doesn't use it anymore.

Chaphasilor: That's an interesting way to handle this. So we'll just remove this in "a bit"?

lib/models/finamp_models.dart
Chaphasilor: Ah so the duration simply has priority, and if it runs out but `tracksLength` is not yet 0, it will finish the last track? And in theory we could also have a 5 minute timer which will play three more track after it ends?  
That's so strange xD  
But I guess this is fine then :)

Komodo5197:
It looks like the paging controller uses postFrameCallbacks for background page loads, so I've added that as allowed.  Most other things fire during the build phase and needed stack reading exceptions, or were fully in the background during the idle phase.  Hopefully that clears up the issue.

lib/models/finamp_models.dart
Chaphasilor: Right yeah, I had it backwards once again...

lib/l10n/app_en.arb
Komodo5197: Yeah.  I think this stays a bit messy until we make a change that requires this to be fully re-translated, if we ever do.

lib/models/finamp_models.dart
Komodo5197: Yes, we could in theory play multiple tracks after after the end of the duration, but it would be strange for anyone to want that.  This just seemed like the cleanest way to represent everything because it means all public fields are always valid and used.

lib/l10n/app_en.arb
Chaphasilor: I guess that's fine. We could also remove any existing translations and have people re-translate it before the next release; most language haven't yet translated the new sleep timer strings anyway.

Komodo5197: Would you prefer I do it that way?

Chaphasilor: I know that I'll forget about removing the parameter. If you don't, we can leave it in.  
But it's probably the same amount of work for translators if we remove it, since they'll need to update the translations either way because the source string was changed.

Chaphasilor: But we could simply check again before the next release, and if any translations are still using the suffix parameter, we leave it in, otherwise we remove it?

Chaphasilor:
Even with your fix I'm still getting the assertion errors, sadly.

lib/l10n/app_en.arb
Komodo5197: If touching the source string already triggers retranslation, I'll just delete them all so I can remove the parameters.

Chaphasilor:
There you go :)
```log
I/flutter (27779): [GlobalSnackbar/WARNING] 2025-06-23 00:07:16.913018: Displaying error: 'package:finamp/services/jellyfin_api_helper.dart': Failed assertion: line 224 pos 12: '_verifyCallable()': is not true.
I/flutter (27779): #2      JellyfinApiHelper._fetchGetItemsResponse (package:finamp/services/jellyfin_api_helper.dart:224:12)
jellyfin_api_helper.dart:224
I/flutter (27779): #3      JellyfinApiHelper.getItems (package:finamp/services/jellyfin_api_helper.dart:141:28)
jellyfin_api_helper.dart:141
I/flutter (27779): #4      _MusicScreenTabViewState._getPage (package:finamp/components/MusicScreen/music_screen_tab_view.dart:105:49)
music_screen_tab_view.dart:105
I/flutter (27779): #5      _MusicScreenTabViewState.initState.<anonymous closure> (package:finamp/components/MusicScreen/music_screen_tab_view.dart:229:7)
music_screen_tab_view.dart:229
I/flutter (27779): #6      PagingController.notifyPageRequestListeners (package:infinite_scroll_pagination/src/core/paging_controller.dart:214:17)
paging_controller.dart:214
I/flutter (27779): #7      _PagedLayoutBuilderState.build.<anonymous closure> (package:infinite_scroll_pagination/src/widgets/helpers/paged_layout_builder.dart:160:31)
paged_layout_builder.dart:160
I/flutter (27779): #8      _ListenableListenerState._handleChange (package:infinite_scroll_pagination/src/utils/listenable_listener.dart:53:22)
listenable_listener.dart:53
I/flutter (27779): #9      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:435:24)
change_notifier.dart:435
I/flutter (27779): #10     ValueNotifier.value= (package:flutter/src/foundation/change_notifier.dart:560:5)
change_notifier.dart:560
I/flutter (27779): #11     PagingController.value= (package:infinite_scroll_pagination/src/core/paging_controller.dart:97:11)
paging_controller.dart:97
I/flutter (27779): #12     PagingController.refresh (package:infinite_scroll_pagination/src/core/paging_controller.dart:123:5)
paging_controller.dart:123
I/flutter (27779): #13     _MusicScreenTabViewState._refresh (package:finamp/components/MusicScreen/music_screen_tab_view.dart:340:23)
music_screen_tab_view.dart:340
I/flutter (27779): #14     _MusicScreenTabViewState.build.<anonymous closure>.<anonymous closure> (package:finamp/components/MusicScreen/music_screen_tab_view.dart:504:34)
music_screen_tab_view.dart:504
I/flutter (27779): #15     RefreshIndicatorState._show.<anonymous closure> (package:flutter/src/material/refresh_indicator.dart:578:64)
refresh_indicator.dart:578
I/flutter (27779): (elided 8 frames from class _AssertionError and dart:async)

```

Komodo5197:
Okay, I just allowed anything in the page listener request this time, so that should clear that up.  I discovered a correct failure where the add to playlist menu was doing requests during build and fixed that, as well as making a couple more methods check _verifyCallable.  And I decided to just delete all the translations that were actually using the prefix/suffix so that I could remove the arguments.

Chaphasilor:
@Komodo5197 got another Stack Trace for you:

```log
[PlaybackHistoryService/WARNING] 2025-06-26 12:11:52.864641: 'package:finamp/services/jellyfin_api_helper.dart': Failed assertion: line 658 pos 12: '_verifyCallable()': is not true.
		#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:67:4)
		#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:49:5)
		#2      JellyfinApiHelper.stopPlaybackProgress (package:finamp/services/jellyfin_api_helper.dart:658:12)
		#3      PlaybackHistoryService.onTrackChanged (package:finamp/services/playback_history_service.dart:353:34)
		#4      new PlaybackHistoryService.<anonymous closure> (package:finamp/services/playback_history_service.dart:104:15)
		#5      _RootZone.runUnaryGuarded (dart:async/zone.dart:1778:10)
		#6      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:381:11)
		#7      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:312:7)
		#8      _MultiStreamController.addSync (dart:async/stream_impl.dart:1197:36)
		#9      _MultiControllerSink.add (package:rxdart/src/utils/forwarding_stream.dart:135:35)
		#10     _StartWithStreamSink.onData (package:rxdart/src/transformers/start_with.dart:12:31)
		#11     _RootZone.runUnaryGuarded (dart:async/zone.dart:1778:10)
		#12     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:381:11)
		#13     _DelayedData.perform (dart:async/stream_impl.dart:573:14)
		#14     _PendingEvents.handleNext (dart:async/stream_impl.dart:678:11)
		#15     _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:649:7)
		#16     _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
		#17     _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)	
```

I'm guessing it's related to when/where the stream waa registered?

Issue #1006, opened 2025-01-10
When trying to download on mobile data with WiFi only download, no warning message is shown


felix920506:
When trying to download with mobile data and the require WiFi for downloads option enabled, download doesn't start but no warning or hint is shown, which is confusing.

Chaphasilor:
So you just want a note that the setting is enabled and the download won't start right-away, yes?

felix920506:
@Chaphasilor yes.

PR #1269, opened 2025-06-21
Added another AutoOffline Option

### LINKED ISSUE #1180: Do not turn off offline mode in some Wi-Fi Networks
### Type:
- [ ] New feature
- [X] Improve or extend an existing feature
- [ ] Other: <describe here>

### Explanation 
Give ability to set Wi-Fis, in which offline mode will be turned off. In other Wi-Fis offline mode will be turned on Or allow to turn on offline mode in limited Wi-Fi networks.

### Reasoning
Every day my phone gets connected to some Wi-Fi networks, but I can connect to my server only from my home network. I have set other Wi-Fi as limited, but offline mode still turns off, so I get connection errors.

### Implementation Notes
Maybe something like list of allowed networks? And then check, if WiFi SSID = SSID from list, so turn the offline mode off.

flloschy:
Not quite sure why 8b0bc2c and f51bee8 are included in the PR
~~Also some `.g.dart` files are trying to sneak into commits. Should I push them? Looks like mostly formatting changes.~~ That was bc I didnt format... Also shouldn't generated files be in the gitignore?

Anyway, this should close #1180 since enabling offline Mode when the server cant be reached is kinda like filtering network names


lib/services/network_manager.dart
```diff
@@ -175,6 +175,8 @@ int _getDownloads() {
 }
 
 void _notifyOfPausedDownloads(List<ConnectivityResult> connections) async {
+  if (!FinampSettingsHelper.finampSettings.isOffline) return;
```
Komodo5197, line 178: This seems incorrect.  This prevents the notification from ever firing when online, which would prevent it from showing if we switch to cellular and pause downloads due to that?  And why would we ever want to be showing this while offline anyway?

flloschy: Order of operatioms thing (line 73)

- We are online
- We switch networks
- We go offline
- We notifiy

And

- We are offline
- We switch networks
- We go online
- We dont notify

Komodo5197: I don't think we want to notify in any situation where the offline mode has just toggled, because that has its own notification, so a download specific one isn't needed.  The primary concern is generating a notification when we go online wifi -> online cellular, and maybe also if we loose connection while downloading but autooffline is disabled.  And I'm pretty sure both those scenarios are broken by this.

flloschy: This whole thing isnt meant to notify when going into cellular in the first place (or is it?) since it checks for no connection at all and not just wifi, though maybe it should

Komodo5197: I'm pretty sure that notifying when you go online wifi to online cellular, pausing downloads when require wifi is on, is the primary purpose of this function.  I think the no connection thing is a bonus?  Honestly, I'm not sure why we feel the need to notify you if your downloads pause due to lack of connection - that seems pretty obvious and also affects the rest of online mode.

flloschy: @Chaphasilor suggested add it

Chaphasilor: Don't blame *me* :P

This was requested in https://github.com/jmshrv/finamp/issues/1006, although the main issue was having a notification for newly-added downloads.

In general, I agree that there is no need to show two notifications at once. We could either combine them into one, or get rid of one. I think changing the auto offline notification to add "Any downloads will be paused/postponed" would be best.

PR #1275, opened 2025-07-05
Server Discovery Emulation & Quick Connect Authorization

Chaphasilor:
## Changes

Screenshot showing the sharing menu panel
Screenshot showing the quick connect authorization menu panel

This adds two new menus, one for sharing the Jellyfin server currently configured in Finamp with other clients, and one for approving pending Quick Connect requests.  

For the former it uses the Server (Client?) Discovery over UDP.
Essentially, when this is active, we listen for incoming UDP messages that match `who is Jellyfin?` and reply to those with the same piece of JSON that the server would normally send.  
I'm not entirely sure if this isn't a terrible idea to do, but so far I couldn't find a convincing reason why not. Still, the plan is to automatically disable this after a very short timeout, and on app startup.

My motivation for this was to help setting up Jellyfin clients (like Finamp? 👀) on things like TVs in a hotel or airbnb, where typing out the URL sucks and the client on the TV can't talk to the server via the local network. 

The Quick Connect authorization menu was added as the logical next step after sharing the server URL: if we don't want to type in the URL, we probably also don't want to type in a password.

## Todo before merging

- [x] Translations

- [x] ~~Reset Settings~~

- [x] ~~Extended CONTRIBUTING.md~~

## Related Issues

Nothing really, I just had the idea a while ago and it seems to work well for me. So now I'm looking for feedback.  
So if anyone wants to give this a try, that would be much appreciated!


lib/services/server_discovery_emulation_service.dart
```diff
@@ -0,0 +1,80 @@
...
+        if (datagram != null) {
+          _serverDiscoveryEmulationLogger.finest("Received datagram: ${utf8.decode(datagram.data)}");
+          final requestMessage = utf8.decode(datagram.data);
+          if (requestMessage.toLowerCase().contains(discoveryMessage.toLowerCase())) {
+            _serverDiscoveryEmulationLogger.fine(
+              "Received discovery message from ${datagram.address}:${datagram.port}",
+            );
+            // Respond with the server's information
+            final response = ClientDiscoveryResponse(
+              address: _finampUserHelper.currentUser?.baseURL,
```
flloschy: Do we want to send the baseURL? Since it could be the local or public address
I mean it makes sense to share the *active* url though... Maybe show in the messages which url will be shared, just to clarify what the user is doing?
Edit: Sharing both would work too i guess? Dont know how useful that would actually be

Chaphasilor: Yeah, so if two URLs are configured, and we need server sharing, the remote URL is probably the active one. So that should be fine.

Sharing both is an interesting idea, but not sure how well that will work for other clients.

Komodo5197:
Creating a setting and background service for this seems counterproductive to me.  I don't see why you would ever want the broadcast to start on app, launch without intervention, so saving the state is undesirable, and without a setting we don't need a background service and can just rely on the widget lifecycle.  I'd make it so that if you hit 'start sharing' we call the start function, and then if you hit stop we call dispose, but then if the widget gets disposed due to the sheet closing we also call dispose on the broadcaster, and also send a notification that we've stopped.  And then the broadcast stays live for as long as the sheet is open without a timer.

Chaphasilor:
@Komodo5197 I don't remember why I opted to create a setting *and* service (it's been a few months since the original PoC), I agree that having both is not great.  
Your proposed solution does seem sensible to me, but I'll have to check that this also works when the app is placed in the background.  
I'd like this to be usable for logging into another client on the same device, if feasible.

If not, that's also not a dealbreaker, I just want to try. But I need to check anyway to see if we need a lifecycle hook / timer in addition or not.

Chaphasilor:
Okay, made some changes and decided to keep things simple. No timer, just turning it off when the menu is dismissed as you suggested, and when the app is hidden on mobile.  
It now replies with both public and local URL if both are configured
Sharing to another client on the same device still works when opening them in split view, if really needed.  

I also tweaked the receiving end (login flow) to better handle the same server being shared with multiple URLs.

Sadly the Jellyfin Android app doesn't seem to pick up all the shared servers, it probably deduplicates by server ID. But basic sharing when no server is available definitely works with that as well.

I decided to keep it as a service simply because registering it via GetIt and having it as a singleton made sense to me and was consistent. And I wanted to keep it as a separate class and not integrate it too tightly with the menu.

From my side this is done now, let me know if there's still things that seem off.  
I might improve code sharing between the discovery emulation service and the login flow for the message strings, addresses, and ports before merging.

Komodo5197:
- The primary use of the empty theme is during queue restore, which I'd prefer to stay grey, so double-check that you haven't broken that.  If you have, maybe just explicitly use the blue theme if the themed sheet doesn't receive an item.
- Having this as a singleton service makes no sense.  It should never be accessed from anywhere else, and does not have the capacity to handle more than one widget interacting with it at once.  It has no persistent background state, and clutters up main for no reason.  We have plenty of non-singleton service classes.
- Doesn't disabling the broadcast when the app is in the background explicitly contradict your desire to configure other apps on the phone?  Have you decided that isn't actually important?
- I think that the AppLifecycleListener is the only place updating the toggleable list key is needed?  But I don't like reaching into the playlist menu file for this, or the weird interface it presents us.  If we want the theming, it might make sense to factor that out into a stateless widget which takes the current loading/state values instead of initialValue+forceLoading, and then the playlist menu uses a widget that just wraps that with state tracking.  Or maybe the loading behavior stays with the core widget, and we only factor out the currentState tracking?
- I'd prefer to use an explicit setServerSharing instead of  toggleServerSharing.  We should know our target value in every scenario.
- We should probably have some error handling somewhere for all the socket work?

Chaphasilor:
- I remember that we discussed this at some point. I'll make sure it says gray while loading.
- I believe the only issue with multiple widgets accessing the service was that an old socket wouldn't be closed if `advertiseServer()` was called multiple times. I've now added a check for avoiding multiple calls to `advertiseServer()`. Either way, I unregistered it :) I also consolidated it with the other class from the login flow, since they are essentially doing the same thing.
- As I wrote earlier, it's still possible to set up other clients by using a split view to open both apps at the same time. However I don't consider this to be an essential feature, and prefer the lower complexity.
- I also thought about this, since I would like to use this kind of list tile more often in the future. I've done that now.
- What kind of error handling are you thinking of? Handling the `send` and `receive` methods? For the JSON encoding? Or just a big try/catch block?

Another thought I had: maybe this is a good time to add Quick Connect authentication support, i.e. entering a Quick Connect code to authenticate another device....

lib/services/server_client_discovery_service.dart
```diff
@@ -0,0 +1,142 @@
...
+  static final _clientDiscoveryLogger = Logger("JellyfinServerClientDiscovery");
+
+  final _finampUserHelper = GetIt.instance<FinampUserHelper>();
+
+  late RawDatagramSocket _discoverySocket;
+  late RawDatagramSocket _advertisingSocket;
+  bool isDiscovering = false;
+  bool isAdvertising = false;
+
+  bool _isDisposed = false;
```
Komodo5197: This variable should be removed, alll uses should be using isDiscovering instead.

lib/components/toggleable_list_tile.dart
```diff
@@ -0,0 +1,103 @@
...
+    this.tapFeedback = true,
+  });
+
+  final String title;
+  final String? subtitle;
+  final Widget leading;
+  final IconData? icon;
+  final Widget? trailing;
+  final bool state;
+  final Future<bool> Function(bool state) onToggle;
```
Komodo5197: There's not really a reason to have this value return a bool, we don't use it anymore.

lib/menus/output_menu.dart
```diff
@@ -187,10 +189,17 @@ class _OutputTargetListState extends State<OutputTargetList> {
...
               final route = snapshot.data![index];
+              if (route.isSelected) {
+                switchingToRoute = null; // Reset switching state if route is selected
+              }
               return OutputSelectorTile(
                 routeInfo: route,
-                onSelect: () {
-                  setState(() {});
+                isLoading: switchingToRoute == route.name,
+                onSelect: ({bool loading = false, bool value = false}) {
```
Komodo5197, line 198: This loading argument seems to be pointless, it's exclusively called with true.

lib/menus/output_menu.dart
```diff
@@ -240,12 +249,13 @@ class OutputSelectorTile extends StatelessWidget {
...
   final FinampOutputRoute routeInfo;
   final bool isLoading;
-  final VoidCallback? onSelect;
+  final void Function({bool loading, bool value})? onSelect;
 
   @override
   Widget build(BuildContext context) {
     return ToggleableListTile(
-      forceLoading: isLoading,
+      key: ValueKey(routeInfo.name),
```
Komodo5197: Is this key really needed?

lib/menus/output_menu.dart
```diff
@@ -264,17 +274,12 @@ class OutputSelectorTile extends StatelessWidget {
...
         }),
       ),
-      positiveIcon: TablerIcons.device_speaker_filled,
-      negativeIcon: TablerIcons.device_speaker,
-      initialState: routeInfo.isSelected,
+      icon: routeInfo.isSelected ? TablerIcons.device_speaker_filled : TablerIcons.device_speaker,
+      state: routeInfo.isSelected,
       onToggle: (bool currentState) async {
         final audioHandler = GetIt.instance<MusicPlayerBackgroundTask>();
+        onSelect?.call(loading: true, value: currentState);
```
Komodo5197, line 280: Doesn't this cause a race between where if getRoutes completes faster than setOutputToRoute we never update the display?

lib/menus/quick_connect_authorization_menu.dart
```diff
@@ -0,0 +1,346 @@
...
+    minDraggableHeight: 0.60,
+    buildSlivers: (context) {
+      final menuEntries = [QuickConnectInput()];
+
+      var menu = [
+        SliverStickyHeader(
+          header: ServerSharingPanelHeader(),
+          sliver: MenuMask(
+            height: ServerSharingPanelHeader.defaultHeight,
+            child: SliverList(delegate: SliverChildListDelegate.fixed(menuEntries)),
```
Komodo5197: We only have one item, so it might make sense to just use a sliverToBoxAdapter here?

lib/menus/quick_connect_authorization_menu.dart
```diff
@@ -0,0 +1,346 @@
...
+                    ),
+                  ),
+                );
+              }),
+            ),
+          ),
+        ),
+        // Hidden text field for actual input handling
+        Opacity(
+          opacity: 0.01,
```
Komodo5197: Do we really not have a better way to do this?

lib/menus/quick_connect_authorization_menu.dart
```diff
@@ -0,0 +1,346 @@
...
+    );
+  }
+
+  @override
+  void initState() {
+    super.initState();
+    _controller.addListener(() {
+      if (_controller.text.trim().length >= 6) {
+        // Automatically authorize when the code is long enough
+        if (_state != QuickConnectAuthorizationState.processing &&
```
Komodo5197: This guard should probably be inside _authorizeQuickConnect so that it applies to the onSubmitted call?

lib/menus/output_menu.dart
Chaphasilor: is it? if `switchingToRoute` is `null` (i.e. we're currently not switching), `isLoading` will be `false`

lib/menus/output_menu.dart
Chaphasilor: probably not, I added it while diagnosing a different bug...

lib/menus/output_menu.dart
Chaphasilor: hmm that could be. I originally had two calls, one with isLoading true, the other with isLoading false and the new state.  
I should probably add that back

lib/menus/quick_connect_authorization_menu.dart
Chaphasilor: I mean the MenuMask is only needed when the sheet content becomes vertically scrollable. That could probably happen on small phones or foldables. Or is the comment about the `SliverList` with delegate?

lib/menus/quick_connect_authorization_menu.dart
Chaphasilor: I'm sure there is. I'll take another look tomorrow.

lib/menus/quick_connect_authorization_menu.dart
Chaphasilor: good point.

lib/menus/output_menu.dart
Komodo5197: I meant the loading given to onSelect.  But this is moot if you're re-adding a call with false.

lib/menus/quick_connect_authorization_menu.dart
Komodo5197: I mean the sliverList.  The header setup + menu mask makes sense.

lib/menus/music_screen_drawer.dart
```diff
@@ -85,6 +87,16 @@ class MusicScreenDrawer extends StatelessWidget {
                         title: Text(AppLocalizations.of(context)!.logs),
                         onTap: () => Navigator.of(context).pushNamed(LogsScreen.routeName),
                       ),
+                      ListTile(
+                        leading: const Padding(padding: EdgeInsets.only(right: 16), child: Icon(TablerIcons.access_point)),
+                        title: Text(AppLocalizations.of(context)!.serverSharingMenuButtonTitle),
+                        onTap: () => showServerSharingPanel(context: context),
+                      ),
+                      ListTile(
```
Komodo5197: Having two buttons here seems excessive.  Honestly even one seems a bit much, I personally would have probably put this inside settings, alongside the logoff button, because I'm not sure how frequently this would be accessed?

Chaphasilor: I'm not happy with the placement either. In Jellyfin web the quick connect option is in the user menu, which in some way is the equivalent to our drawer.  
Now that I added a button to the sharing menu for opening the quick connect menu we can probably move the quick connect button somewhere else, although I really don't think it counts as a setting. But having it at the same level as the logout button could work.  
That being said the rest of the menu probably also needs an overhaul; the restore now playing option should be next to the playback history. An overhaul for the drawer was planned anyway.

I think I'll move the options above the logout button, and move the restore now playing into the upper section, does that work for you?

lib/menus/quick_connect_authorization_menu.dart
Chaphasilor: simple `Visibility()` with fully maintained state and interactivity worked as well 👍

lib/menus/music_screen_drawer.dart
Komodo5197: Looks good to me.  I did notice that when I clicked 'enter quick connect code' in the sharing menu I had to close the menu and reopen the sharing menu to get back.  It might be better to leave the first menu open in the background?  Or would that be unexpected.  Also, I noticed that themed bottom sheet is using a partially transparent background.  Do we actually want that?  The menus are completely opaque when a background image is used.

Chaphasilor: I mean if you are at the stage where you can see a quick connect code I guess you don't need the server sharing anymore? I consciously did it like this. Leaving the other menu open probably wouldn't hurt, but I also don't think it would be super useful.  
I also noticed the transparency, but I think it's definitely subtle enough to not cause accessibility issues. I'm impartial to making it transparent.

Komodo5197: If you think closing it is better, that's fine.  Regarding the transparency, I'd prefer to remove it just to make things a bit more consistent.  While messing with this, I noticed that I'm not sure about how the all-black bottom sheet actually looks.  Maybe we should avoid setting the modal sheet background color to use the same background color as the rest of the app?  Also, I noticed that I create a new theme from scratch instead of using copyWith on the existing theme in PlayerScreenTheme, so if you disable the blurred background the player just becomes grey.  I'd appreciate you fixing this.

Chaphasilor: alright, should be fixed now. the other option for maintaining consistency would've been to introduce transparency to the themed bottom sheets as well, which might've looked nice but definitely wouldn't have improved performance.  
I'm not sure if you wanted me to adjust both instances where the `ThemeData()` was newly created instead of copied, but I changed them both and think it's still behaving correctly.  
I'll go ahead and merge this now, thanks for your feedback :)

PR #1276, opened 2025-07-08
Add support for Android user-installed certificates

### LINKED ISSUE #106: Failed to connect to server - HandshakeException

jfly:
This fixes https://github.com/jmshrv/finamp/issues/106

Frustratingly, Dart/Flutter ignores user-installed certificates. Working around this requires rooting your Android device to install certificates as "system" certs, which isn't an option for all people.

This is a known issue with Dart, see
https://github.com/dart-lang/sdk/issues/50435 and
https://github.com/flutter/flutter/issues/56607 for details.

This depends on my fork of
`johnstef99/flutter_user_certificates_android`, which I've sent a PR for here
<https://github.com/johnstef99/flutter_user_certificates_android/pull/2>. If y'all don't like the supply chain implications of that, I'm happy to inline the implementation here instead.

I opted not to make this configurable as I don't see the value. People using finamp are sophisticated self-hosters, and should understand the risks of configuring their OS to trust certificates.

## Related Issues

fixes #106

### DIFF
--- a/android/app/src/main/res/xml/network_security_config.xml
+++ b/android/app/src/main/res/xml/network_security_config.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+  <base-config>
+    <trust-anchors>
+      <certificates src="system"/>
+      <certificates src="user"/>
+    </trust-anchors>
+  </base-config>
+</network-security-config>
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -46,6 +46,7 @@ import 'package:flutter_localizations/flutter_localizations.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_svg/svg.dart';
 import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
+import 'package:flutter_user_certificates_android/flutter_user_certificates_android.dart';
 import 'package:get_it/get_it.dart';
 import 'package:hive_ce_flutter/hive_flutter.dart';
 import 'package:intl/date_symbol_data_local.dart';
@@ -105,6 +106,8 @@ void main() async {
     _migrateDownloadLocations();
     _migrateSortOptions();
     _mainLog.info("Completed applicable migrations");
+    await _trustAndroidUserCerts();
+    _mainLog.info("Trusted Android user certs");
     await _setupFinampUserHelper();
     _mainLog.info("Setup user helper");
     await _setupJellyfinApiData();
@@ -418,6 +421,13 @@ void _migrateSortOptions() {
   if (changed) {
     FinampSettingsHelper.overwriteFinampSettings(finampSettings);
   }
+}
+
+Future<void> _trustAndroidUserCerts() async {
+  // Extend the default security context to trust Android user certificates.
+  // This is a workaround for <https://github.com/dart-lang/sdk/issues/50435>.
+  WidgetsFlutterBinding.ensureInitialized();
+  await FlutterUserCertificatesAndroid().trustAndroidUserCertificates(SecurityContext.defaultContext);
 }
 
 Future<void> _setupFinampUserHelper() async {
--- a/lib/services/jellyfin_api_helper.dart
+++ b/lib/services/jellyfin_api_helper.dart
@@ -10,6 +10,7 @@ import 'package:finamp/services/http_aggregate_logging_interceptor.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_user_certificates_android/flutter_user_certificates_android.dart';
 import 'package:get_it/get_it.dart';
 import 'package:http/io_client.dart' as http;
 import 'package:isar/isar.dart';
@@ -59,6 +60,11 @@ class JellyfinApiHelper {
   static Future<void> _processRequestsBackground((SendPort, RootIsolateToken) input) async {
     BackgroundIsolateBinaryMessenger.ensureInitialized(input.$2);
     ReceivePort requestPort = ReceivePort();
+
+    // Extend the default security context to trust Android user certificates.
+    // This is a workaround for <https://github.com/dart-lang/sdk/issues/50435>.
+    await FlutterUserCertificatesAndroid().trustAndroidUserCertificates(SecurityContext.defaultContext);
+
     input.$1.send(requestPort.sendPort);
     final dir = (Platform.isAndroid || Platform.isIOS)
         ? await getApplicationDocumentsDirectory()
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -23,6 +23,7 @@
     android:allowBackup="true"
     android:fullBackupContent="@xml/backup_rules_legacy"
     android:dataExtractionRules="@xml/backup_rules"
+    android:networkSecurityConfig="@xml/network_security_config"
   >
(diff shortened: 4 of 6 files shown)


lib/services/jellyfin_api_helper.dart
```diff
@@ -59,6 +60,12 @@ class JellyfinApiHelper {
   static Future<void> _processRequestsBackground((SendPort, RootIsolateToken) input) async {
     BackgroundIsolateBinaryMessenger.ensureInitialized(input.$2);
     ReceivePort requestPort = ReceivePort();
+
+    // Extend the default security context to trust Android user certificates.
+    // This is a workaround for <https://github.com/dart-lang/sdk/issues/50435>.
+    final _flutterUserCertificatesAndroidPlugin = FlutterUserCertificatesAndroid();
+    await _flutterUserCertificatesAndroidPlugin.trustAndroidUserCertificates(SecurityContext.defaultContext);
```
jfly: Unfortunately, Dart isolates don't seem to inherit the `SecurityContext.defaultContext` we configured at app startup. I don't know if there's a cleaner way of handling this.

Chaphasilor: While this is fine, have you tested if this also causes issues with `just_audio`/ExoPlayer? That uses a separate http client, and I'm not sure if that will trust the certificates. It *probably* does since it's Android-native, but if not that would mean that users could browser and download, but not stream music.

android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+  <base-config>
+    <trust-anchors>
+      <certificates src="system"/>
+      <certificates src="user"/>
```
Chaphasilor, line 6: ahh I'm guessing this will fix ExoPlayer?

lib/main.dart
```diff
@@ -420,6 +423,14 @@ void _migrateSortOptions() {
   }
 }
 
+Future<void> _trustAndroidUserCerts() async {
+  // Extend the default security context to trust Android user certificates.
+  // This is a workaround for <https://github.com/dart-lang/sdk/issues/50435>.
+  WidgetsFlutterBinding.ensureInitialized();
+  final flutterUserCertificatesAndroidPlugin = FlutterUserCertificatesAndroid();
+  await flutterUserCertificatesAndroidPlugin.trustAndroidUserCertificates(SecurityContext.defaultContext);
```
Chaphasilor: I think we don't need the variable at all.
```suggestion
  await FlutterUserCertificatesAndroid().trustAndroidUserCertificates(SecurityContext.defaultContext);
```

lib/services/jellyfin_api_helper.dart
```diff
@@ -59,6 +60,12 @@ class JellyfinApiHelper {
   static Future<void> _processRequestsBackground((SendPort, RootIsolateToken) input) async {
     BackgroundIsolateBinaryMessenger.ensureInitialized(input.$2);
     ReceivePort requestPort = ReceivePort();
+
+    // Extend the default security context to trust Android user certificates.
+    // This is a workaround for <https://github.com/dart-lang/sdk/issues/50435>.
+    final flutterUserCertificatesAndroidPlugin = FlutterUserCertificatesAndroid();
+    await flutterUserCertificatesAndroidPlugin.trustAndroidUserCertificates(SecurityContext.defaultContext);
```
Chaphasilor: Same here
```suggestion
    await FlutterUserCertificatesAndroid().trustAndroidUserCertificates(SecurityContext.defaultContext);
```

lib/services/jellyfin_api_helper.dart
jfly: > While this is fine, have you tested if this also causes issues with just_audio/ExoPlayer?

As you say, if it's Android-native, then it probably works.

How would I test this? I just deleted my downloaded library and confirmed I can stream music from my server with a user-installed certificate. Is that good enough?

Chaphasilor: Yes, that's good enough :)

android/app/src/main/res/xml/network_security_config.xml
jfly: I don't know what ExoPlayer is, sorry.

These `<trust-anchor>`s affect "native" Java/Kotlin network traffic, for example `javax.net.ssl.HttpsURLConnection`. Documented here: <https://developer.android.com/privacy-and-security/security-config#network-security-config>

lib/services/jellyfin_api_helper.dart
jfly: Done!

lib/main.dart
jfly: Done!

android/app/src/main/res/xml/network_security_config.xml
Chaphasilor: ExoPlayer is Android's build-in media player  
https://developer.android.com/media/media3/exoplayer

jfly:
> Hey, thanks for the PR!
I originally though it was another "+1" comment on the issue, and was pleasantly surprised when I saw that it's actually a PR last night.

Thanks for the review =) Finamp is awesome, was so happy when I found it.

> There's also no issue with adding a dependency. We can either keep this PR open until the upstream PR is merged, or merge it now and migrate later. I'd prefer the former, but if you want to be done with this we can also merge now.

Awesome! Let's wait a week and see? I haven't heard from the maintainer yet... :crossed_fingers:

jfly:
> If you don't want to keep that fork active, I can also create a clone of your fork.

I intend to keep my fork active!

Chaphasilor:
Huh, I think I meant to comment something, guess I forgot to. Either way, let's merge this then!

Thanks for the PR! <3

Chaphasilor:
@jfly FYI it seems like the plugin (or your new method?) isn't checking if the current platform is actually Android or not. And since there's no native implementation for any platform *other than Android*, it will throw an error when running e.g. on Linux.  
So you might wanna add such a check, also in your inlined implementation for Immich :)  

Let me know if you update your fork, then I can remove our little bandaid again :)

jfly:
Ah, sorry for the breakage! The change you made in https://github.com/jmshrv/finamp/commit/e158083353dca12fee816f5d5b995adc70d2ceb9 LGTM.

I've updated my fork accordingly: https://github.com/johnstef99/flutter_user_certificates_android/pull/2/files#r2232202219

Chaphasilor:
@jfly another heads-up: your newly-introduced `network_security_config.xml` seems to overwrite the value of `android:usesCleartextTraffic` (in `AndroidManifest.xml`), setting it to `false`. That means all native-code features like downloading

jfly:
> If you want to create a flutter package/plugin on pub.dev for this anyway, feel free to open a PR for this.

Sorry, busy week, and then I've got some travel. I should have some time to do this late next week.

PR #1282, opened 2025-07-15
[Redesign] Fix outdated album sync

Komodo5197:
The sync download button that appears on items with outdated downloads doesn't actually fix outdated albums because the quicksync setting causes them to be skipped.  This has been fixed by forcing a full sync when explicitly resyncing these types.  Also:
- Fix null exceptions in album/artist chips when switching to offline mode when a non-downloaded item is playing.
- Fix restore queue failing to present the retry button when errors are thrown.
- Show played seek bar color behind non-square albums on now-playing bar to match adjacent section of seek bar.

### DIFF
--- a/lib/components/PlayerScreen/album_chip.dart
+++ b/lib/components/PlayerScreen/album_chip.dart
@@ -73,7 +73,7 @@ class _EmptyAlbumChip extends StatelessWidget {
 }
 
 class ReleaseDate extends StatelessWidget {
-  const ReleaseDate({this.baseItem, this.backgroundColor, this.color});
+  const ReleaseDate({super.key, this.baseItem, this.backgroundColor, this.color});
 
   final BaseItemDto? baseItem;
   final Color? backgroundColor;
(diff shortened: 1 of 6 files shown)


Chaphasilor:
Thanks! Added some fixes, feel free to merge this if these look fine to you :)

Komodo5197:
There was some sort of overlay that the now playing bar didn't have, so the background in the queue didn't actually match the scrubber.  I've fixed that.  Also, queue_list.dart wasn't formatted correctly, so I fixed that.  Do you have some sort of setting wrong, or are you formatting manually and just forgot to run it?

I've also noticed that we have a lot of cases where we use transparencies to do color mixing.  I feel like that's probably a bad practice?  My guess is that the 'correct' method would probably be to more thoroughly fill in all those different entries in the ColorScheme and actually use them.  But I don't have experience with this sort of thing, so I'm not actually sure how much of this is true or meaningful.

Chaphasilor:
Hmm the overlay was there on purpose for contrast reasons (I was aware of the overlay when adding the color, it still looked better), but now that you mention it, it probably should've been on the now playing bar too in the first place. Well, but since no-one really complained about contrast on the now playing bar yet, I guess we can also remove in the queue panel... 

We probably could, although Flutter tends to use these named colors in unexpected places, so changing the default can lead to unexpected visual changes (like for the disc header). If there is a way to create a separate namespace for Finamp-specific accent colors that could work?  
In general it's sometimes really handy to use slightly different opacities depending on what is shown where. Probably nothing that couldn't be done through a color scheme, but there are more than just 3 shades of the accent color.  
Speaking of colors, there's still an outstanding accessibility issue for adding a high-contrast color scheme and/or manually picking a global accent color. For those it would probably also help to go through a color scheme, because at least in the case of the high-contrast theme we might need to tweak the opacities.

PR #1291, opened 2025-07-20
Some Kotlin cleanup

Maxr1998:
Reformatted and cleaned up the Android MainActivity Kotlin code, neatly split into individual commits.


Chaphasilor:
Thanks! Change looks good to me (although I haven't tested it yet).  
But I'm wondering if there's a better way to handle communication channels, i.e. sharing constants (instead of copying channel name strings). This is not really required to merge this PR, but I'd like to know if you have any experience with that @edeuss

Maxr1998:
I doubt you can directly share constants between Dart and Kotlin. It might be possible with codegen, but I haven't looked into it. Adding a try/catch should probably be another PR, since I'd like to keep the changes to a minimum (mostly refactoring) here.

Chaphasilor (approved):

PR #1293, opened 2025-07-20
Fix downloaded files being marked as cache on Android 

Maxr1998:
An earlier version of the `background_downloader` library contained a bug that caused all downloaded files to be marked as cache files on Android.
This was due to an incorrect group id being set on those files.
The newest version fixes this issue, but we have to update all previously downloaded files to the correct group id.


Chaphasilor:
some more info / links about what this fixes would be nice for future reference :)

Maxr1998:
That's why it was still in draft :grin:

lib/main.dart
```diff
@@ -217,8 +217,9 @@ Future<void> _setupDownloadsHelper() async {
     }
   }
 
-  await FileDownloader().configure(globalConfig: (Config.checkAvailableSpace, 1024));
-  await FileDownloader().resumeFromBackground();
+  await fileDownloader.configure(globalConfig: (Config.checkAvailableSpace, 1024));
+  await fileDownloader.resumeFromBackground();
+  await MethodChannel("com.unicornsonlsd.finamp/output_switcher").invokeMethod('showId');
```
Komodo5197: Why is this line here?

lib/main.dart
```diff
@@ -230,6 +231,11 @@ Future<void> _setupDownloadsHelper() async {
       FinampSetters.setHasDownloadedPlaylistInfo(true);
     });
   }
+
+  if (!FinampSettingsHelper.finampSettings.hasCompletedDownloadsFileOwnerMigration) {
+    await MethodChannel("com.unicornsonlsd.finamp/downloads_service").invokeMethod("fixDownloadsFileOwner");
```
Komodo5197: Shouldn't we be providing a list of download locations to the method call?

lib/main.dart
Maxr1998: That's an unintentional remainder from my tests, but I haven't even self-reviewed the code yet 😅

lib/main.dart
Maxr1998: Definitely should 😬
Right now, it's still broken.

Maxr1998:
Ok, I rebased my code, fixed a few issues, added some docs and tested it locally. Should be ready for review now!

Chaphasilor:
Would you mind handling the conflicts? :)

Chaphasilor:
Awesome work!

PR #1294, opened 2025-07-21
Add support for custom `finamp://` URL scheme registration & handling

### LINKED ISSUE #1281: Define a Finamp custom URL scheme

### Type of Feature

- [x] Something new
- [ ] Improvement of a feature
- [ ] None of the above

### Description

Opening a `finamp://` link should open the Finamp application.

We can pick another schema but as far as I can tell `finamp://` [is available](https://github.com/bhagyas/app-urls).

### Reasoning

There is currently as far as I can see no custom URL scheme defined to open the app in iOS. This means people who use alternate launchers for iPhone like [Dumbify](https://apps.apple.com/us/app/dumbify/id6480082872) can't add Finamp to their home screen, as there's no link the launcher can use to launch Finamp.

### Implementation Notes

Finamp should define "finamp://" as a [custom url scheme](https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app) which simply opens the application. No further parsing of the URL is necessary for the feature to work.

### Additional Information

_No response_

Chaphasilor:
## Changes

- Finamp will open / move to foreground when a URL with the `finamp://` scheme is opened
- URLs starting with `finamp://internal/` can be used to open registered screens based on the remaining path
  - i.e. `finamp://internal/settings/layout` will open the "Layout & Theme" settings screen (directly, going back will return to the previous screen and *not* to the top-level settings screen)

## Todo before merging

- [x] ~~Translations~~

- [x] ~~Reset Settings~~

- [x] ~~Extended CONTRIBUTING.md~~
- [ ] Configure scheme support for iOS

## Related Issues

- closes #1281

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:io';
 import 'dart:ui';
 
+import 'package:app_links/app_links.dart';
 import 'package:audio_service/audio_service.dart';
 import 'package:audio_session/audio_session.dart';
 import 'package:background_downloader/background_downloader.dart';
@@ -443,10 +444,27 @@ class Finamp extends StatefulWidget {
 
 class _FinampState extends State<Finamp> with WindowListener {
   static final Logger windowManagerLogger = Logger("WindowManager");
+  static final Logger linkHandlingLogger = Logger("LinkHandling");
+
+  StreamSubscription<Uri>? _uriLinkSubscription;
 
   @override
   void initState() {
     super.initState();
+    // Subscribe to all events (initial link and further)
+    _uriLinkSubscription = AppLinks().uriLinkStream.listen((uri) {
+      linkHandlingLogger.info("Received link: $uri");
+      final parsedUrl = Uri.parse(uri.toString());
+      if (parsedUrl.host == "internal") {
+        Navigator.of(GlobalSnackbar.materialAppNavigatorKey.currentContext!).pushNamed(
+          parsedUrl.path,
+        );
+      } else {
+        //TODO eventually we could do something here, but for now we just log the link
+      }
+    });
+
+    // If the app is running on desktop, we add a listener to the window manager
     if (Platform.isWindows || Platform.isLinux || Platform.isMacOS) {
       WindowManager.instance.addListener(this);
       // windowManager.setPreventClose(true); //!!! destroying the window manager instance doesn't seem to work on Windows release builds, the app just freezes instead
@@ -455,6 +473,8 @@ class _FinampState extends State<Finamp> with WindowListener {
 
   @override
   void dispose() {
+    _uriLinkSubscription?.cancel();
+
     if (Platform.isWindows || Platform.isLinux || Platform.isMacOS) {
       WindowManager.instance.removeListener(this);
     }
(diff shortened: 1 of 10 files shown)


Chaphasilor:
... I should probably re-enable format on save -.-

linux/flutter/generated_plugins.cmake
```diff
@@ -3,6 +3,7 @@
 #
 
 list(APPEND FLUTTER_PLUGIN_LIST
+  gtk
```
Komodo5197: This file  is in .gitignore.  You should delete it from git.

linux/flutter/generated_plugin_registrant.cc
```diff
@@ -6,13 +6,17 @@
...
 
+#include <gtk/gtk_plugin.h>
 #include <isar_flutter_libs/isar_flutter_libs_plugin.h>
 #include <media_kit_libs_linux/media_kit_libs_linux_plugin.h>
 #include <screen_retriever_linux/screen_retriever_linux_plugin.h>
 #include <url_launcher_linux/url_launcher_plugin.h>
 #include <window_manager/window_manager_plugin.h>
 
 void fl_register_plugins(FlPluginRegistry* registry) {
+  g_autoptr(FlPluginRegistrar) gtk_registrar =
```
Komodo5197: This file  is in .gitignore.  You should delete it from git.

lib/main.dart
```diff
@@ -443,10 +444,27 @@ class Finamp extends StatefulWidget {
...
+
+  StreamSubscription<Uri>? _uriLinkSubscription;
 
   @override
   void initState() {
     super.initState();
+    // Subscribe to all events (initial link and further)
+    _uriLinkSubscription = AppLinks().uriLinkStream.listen((uri) {
+      linkHandlingLogger.info("Received link: $uri");
+      final parsedUrl = Uri.parse(uri.toString());
```
Komodo5197: What's the purpose of reparsing the uri?

Chaphasilor: That is a good point. I somehow assumed it would be a plain string, but that's a different stream...

Chaphasilor:
Also, can you confirm that the changes I made for macOS also work? I couldn't test that :)

jmshrv:
> thanks @jmshrv! Did you test this already?  
> And what's the "editor" type in the .plist files? 

I believe "Editor" means that it's our own URL standard, and not one we just happen to adopt

> Choose a role for your app: either an editor role for URL schemes you define, or a viewer role for schemes your app adopts but doesn’t define.

I tested the change on iOS and it worked as intended, will test on macOS today :)

jmshrv:
Can confirm that it works on macOS :)

Chaphasilor:
@flloschy yes it needs to be installed I think, probably integrates with the DE via the .desktop file or something like that  
@Maxr1998 did you manage to test this with the aur package yet?

Chaphasilor:
> main.dart:458 is the link handler, so there's something wrong still.

@Maxr1998 it's actually the Navigator, and the issue seems to be that the global scaffold key is still undefined (that's the failing null check).  
I've added a (crude) fix, please try again now

Chaphasilor:
Thanks for helping out with this everyone! <3

PR #1295, opened 2025-07-22
[Redesign] Logout fixes.

Komodo5197:
Various fixes to better handle logoffs.


Chaphasilor:
argh, enabled the wrong format on save option...

Komodo5197:
image

It seems like there's some sort of memory leak during the sync process.  I'm not sure where, the dart dev tools only list ~300 MB of objects in the heap.  I'll keep looking into it a bit.

Chaphasilor:
Okay. Is this a regression on Isar's part? Because in your original PR I did some pretty extensive testinf but I don't think I ran into this issue. So maybe we could just go back a few versions?

Komodo5197:
Okay, I experimented with running a whole bunch of past versions and eventually narrowed it down.  The problem was the changes to the download screen in 9.18.  To accommodate the new categories, the query watch in userDownloadedItemsProvider was changed to be less efficient, and we run more copies of it.  I knew the query watches were relatively heavy, but I guess nobody was looking out for anything like this.  I've refactored it to avoid a query watch entirely in 029e185cfb4ae040abb62aeb0848b0a194163c16, and the performance and memory usage seem greatly improved.  It seems like syncs were also leaking memory in all the previous versions I tested, but in much more manageable amounts, and that continues to be the case now.  Also, because the escalation seems to have been caused by userDownloadedItemsProvider, any sync where you weren't continually watching the downloads screen would be fine.  Anyway, I only tested on windows, so you should probably verify that this improves things for you with your database on android.

Chaphasilor:
I meant to reply but forgot. Thanks for investigating, that sounds like an acceptable solution. Is that memory leak a known isar issue?  
I haven't properly tested this on Android yet, but I might do that some time this week.

Chaphasilor:
Just to confirm, yes, this appears to be fixed. Synced 10k downloads last night on the downloads screen with the screen on, not a single crash :)

Issue #1278, opened 2025-07-11
Genre Shuffle


Ichijoe2112:
### Are you using the newest Version?

- [x] Yes, the feature is still missing on the newest version
- [x] No, there aren't any open request about this. I checked!

### Type of Feature

- [x] Something new
- [ ] Improvement of a feature
- [ ] None of the above

### Description

Could the Devloper's please consider adding a Genre Shuffle, to kinda sorta create a faux Radio Station. in this instance Country would be my drug of choice. 

Oddly enough, Jellyfin already has that ability, but Jellyfin, cant run offline, and Data burns to eaisly round these parts.

Thank you kindly for your time!

lukaslindnermusic:
Are you using the redesign beta?

The Play and Shuffle Buttons were temporarily removed (and kept removed with the new genre screen) because currently, adding a certain amount of tracks simultaneously to the queue crashes the app, and genres are the most likely to be above that number.
We are considering a better large-queue-handling, though.

However, there is a "workaround". You can shuffle a genre by tapping on the tracks count at the top and then on the blue "Shuffle All" button on the bottom right. This applies a limit, though. Its the same as with the regular track tab, and defaults to 250.

PR #1296, opened 2025-07-22
[Redesign] Added Customization Settings for Additional Item Infos + Temporary Shuffle Genre Button

lukaslindnermusic:
## Todo before merging

- [x] Translations

- [x] Reset Settings

## Related Issues

- fixes a discussion we had on Discord about the Additional Item Info
- sort of fixes https://github.com/jmshrv/finamp/issues/1278, though it's temporary

<br>

I've added Customization Settings where you can change the "additional info" that is shown in the subtitles of item list tiles. Previously with my SortBy-overhaul, the additional info was hardcoded to the selected SortBy or Item Filter (aka. show Release Date when sorted by Release Date or "Latest Release" was selected on Genre/Artist-Screen and so on).

In the `Layout & Theme -> Customization` Settings, you can now decide (for each itemType separately) what info you want to see:
The current approach stays the default for every item type, and is now called "Adaptive".
But you can now for example explicitly choose to always show the duration for playlists and the release date for Albums, no matter the selected SortBys.
And of course, you can also disable the additional info completely.

<br>

Unrelated to the above, I've also added a temporary "Shuffle" button to the Genre Screen, because it felt to me like that was asked by multiple people recently.

The Shuffle functionality was already there via a workaround with the floating action button in the genre-filtered track list, but that's a little bit hidden away and you have to know about that. 

And while it's still planned to improve the large queue handling at some point, I thought it would be better to temporarily add a button to the genre screen that does the same as the floating action button.
It does not shuffle the entire genre but adds the `shuffle all tracks limit` from the settings.....
but it's better than nothing, and its only temporary, so....


lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -170,10 +170,10 @@ class _AlbumScreenContentState extends ConsumerState<AlbumScreenContent> {
...
                 parent: widget.parent,
                 onRemoveFromList: onDelete,
-                showDateAdded: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.dateCreated),
-                showPlayCount: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.playCount),
-                showDateLastPlayed: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.datePlayed),
-                showReleaseDate: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.premiereDate),
+                showDateAddedAdaptive: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.dateCreated),
+                showPlayCountAdaptive: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.playCount),
+                showDateLastPlayedAdaptive: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.datePlayed),
+                showReleaseDateAdaptive: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.premiereDate),
```
Chaphasilor: how about passing a single enum here instead, that decides which info is shown? I guess we'll only ever show one extra field.  

```suggestion
                showExtraInfo: widget.parent.type == "Playlist" ? playlistSortBy : null,
```

Chaphasilor: should've caught that last time, but I think it would simplify the code quite a bit

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -626,6 +626,59 @@ class TrackListItemTile extends ConsumerWidget {
...
+        ? AdditionalBaseItemInfoTypes.none
+        : additionalBaseItemInfos[BaseItemDtoType.track];
+
+    final additionalInfo = switch (additionalBaseItemInfo) {
+      AdditionalBaseItemInfoTypes.playCount => (
+        playCount: true,
+        releaseDate: false,
+        dateAdded: false,
+        dateLastPlayed: false,
+      ),
```
Chaphasilor: this also seems quite verbose, especially since you're dissecting it into separate variables at the end anyway...

lib/models/finamp_models.dart
```diff
@@ -710,6 +718,9 @@ class FinampSettings {
   @HiveField(119, defaultValue: DefaultSettings.syncPlaybackSpeedAndPitch)
   bool syncPlaybackSpeedAndPitch;
 
+  @HiveField(120, defaultValue: DefaultSettings.additionalBaseItemInfo)
+  Map<BaseItemDtoType, AdditionalBaseItemInfoTypes> additionalBaseItemInfo;
```
Chaphasilor: make sure to handle missing entries in that map elsewhere in the code

Chaphasilor (commented):
just a quick code review, now onto functionality

Chaphasilor:
Thanks for the PR!

Hmm, you really like adding customization settings 😅  
Has anyone actually requested this much customizability? This is slowly getting into a territory where it becomes hard to maintain. All of these features need to be supported when changing the design, and should theoretically also work in grid layout.

Disabling the adaptiveness seems fine. Setting a static option also *seems* sensible, but doing this for each item type might be a bit overkill...  
If we keep this, do you think you could make this more dynamic, so that if we add new tab types it will automatically render a setting for that? Instead of using a bunch of hard-coded settings widgets. Otherwise I guarantee that this will eventually be out-of-sync...

Only bug I found is that setting albums to show the release date will actually show the date added :)

lib/components/AlbumScreen/album_screen_content.dart
lukaslindnermusic: Yes, good idea... I left it open so that we maybe later can decide to show multiple, but I guess then we could just undo it, and it definitely reduces code lines :P

lib/components/AlbumScreen/track_list_tile.dart
lukaslindnermusic: Yes, that's true. It got less more verbose by simplifying the passing of the showPlayCount etc. to just passing down one single SortBy. 👍

Maxr1998:
> Hmm, you really like adding customization settings 😅 Has anyone actually requested this much customizability?

I might be to blame for this specific setting, actually 🫣
I love the idea of always seeing the release date, but still ordering new albums by the date they were imported into Jellfyin.

Chaphasilor:
Oh wait, I even agreed to that on Discord 🙃  
But that was over a month ago 😅

I think you could still use TabContentType, there should be a `TabContentType.fromBaseItemDtoType()` or something similar?  
We could eventually rename this to be ListContentType or something similar.

I'm actually not sure regarding Hive, I don't think we used a map before. Maybe for download locations?  
Either way, does the adapters file get properly regenerated? Is there a Hive annotation for BaseItemDtoType? I'm on mobile right now...

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -633,51 +609,24 @@ class TrackListItemTile extends ConsumerWidget {
...
+        (additionalBaseItemInfo == AdditionalBaseItemInfoTypes.adaptive &&
+            adaptiveAdditionalInfoSortBy == SortBy.premiereDate));
+    bool showDateAdded =
+        (additionalBaseItemInfo == AdditionalBaseItemInfoTypes.dateAdded ||
+        (additionalBaseItemInfo == AdditionalBaseItemInfoTypes.adaptive &&
+            adaptiveAdditionalInfoSortBy == SortBy.dateCreated));
+    bool showDateLastPlayed =
+        (additionalBaseItemInfo == AdditionalBaseItemInfoTypes.dateLastPlayed ||
+        (additionalBaseItemInfo == AdditionalBaseItemInfoTypes.adaptive &&
+            adaptiveAdditionalInfoSortBy == SortBy.datePlayed));
```
Chaphasilor: I think this is a bit easier to read:
```suggestion
        : (additionalBaseItemInfos[BaseItemDtoType.track] ?? AdditionalBaseItemInfoTypes.adaptive);

    bool showPlayCount = additionalBaseItemInfo == AdditionalBaseItemInfoTypes.playCount:
    bool showReleaseDate = additionalBaseItemInfo == AdditionalBaseItemInfoTypes.dateReleased;
    bool showDateAdded = additionalBaseItemInfo == AdditionalBaseItemInfoTypes.dateAdded;
    bool showDateLastPlayed = additionalBaseItemInfo == AdditionalBaseItemInfoTypes.dateLastPlayed;;
    if (additionalBaseItemInfo == AddititionalBaseItemInfoType.adaptive) {
      showPlayCount = showPlayCount || adaptiveAdditionalInfoSortBy == SortBy.playCount;
      showReleaseDate = showReleaseDate || adaptiveAdditionalInfoSortBy == SortBy.premiereDate;
      showDateAdded = showDateAdded || adaptiveAdditionalInfoSortBy == SortBy.dateCreated;
      showDateLastPlayed = showDateLastPlayed || adaptiveAdditionalInfoSortBy == SortBy.datePlayed;
    }
```

lib/components/GenreScreen/genre_screen_content.dart
```diff
@@ -272,9 +277,35 @@ class _GenreScreenContentState extends ConsumerState<GenreScreenContent> {
...
+        /// and once we redesigned the play/shuffle buttons, they should get added here
+        /// for now as a temporary solution, I've added a single Shuffle Button, as users kept asking about it
+        /// it basically is the same as the Shuffle All button on the tracks tab + genrefilter
+        /// so the limit defined in the settings applies here as well to prevent crashing
+        SliverToBoxAdapter(
+          child: Padding(
+            padding: const EdgeInsets.only(left: 8, right: 8, top: 16),
+            child: Center(
+              child: CTAMedium(
+                text: AppLocalizations.of(context)!.shuffleButtonLabel,
```
Komodo5197: Could we use a different name like 'random' or something to clarify that we're not necessarily using all tracks?

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -626,6 +602,27 @@ class TrackListItemTile extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final highlightTrack = isCurrentTrack && highlightCurrentTrack;
     final isOnDesktop = Platform.isMacOS || Platform.isWindows || Platform.isLinux;
+    final isOffline = ref.watch(finampSettingsProvider.isOffline);
+    final additionalBaseItemInfos = ref.watch(finampSettingsProvider.additionalBaseItemInfo);
+    final additionalBaseItemInfo =
```
Komodo5197: Is it possible to accidentally show playcount/lastPlayed offline via adaptiveAdditionalInfoSortBy?  Also, I'd like to avoid watching isOffline and triggering rebuilds in scenarios where we never would have shown playcount/lastPlayed and thus do not actually need the information.

lib/models/finamp_models.dart
```diff
@@ -715,6 +723,9 @@ class FinampSettings {
   @HiveField(120, defaultValue: DefaultSettings.useHighContrastColors)
   bool useHighContrastColors;
 
+  @HiveField(121, defaultValue: DefaultSettings.additionalBaseItemInfo)
+  Map<TabContentType, AdditionalBaseItemInfoTypes> additionalBaseItemInfo;
```
Komodo5197: You need the SettingsHelperMap annotation here for this to work properly.

lib/components/LayoutSettingsScreen/CustomizationSettingsScreen/additional_base_item_info_dropdown_list_tile.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  const AdditionalBaseItemInfoDropdownListTile({required this.tabContentType, super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final additionalBaseItemInfos = ref.watch(finampSettingsProvider.additionalBaseItemInfo);
+    final currentType = additionalBaseItemInfos[tabContentType] ?? AdditionalBaseItemInfoTypes.adaptive;
+
+    // Filter dropdown items based on tabContentType
+    final dropdownItems = AdditionalBaseItemInfoTypes.values.where((type) {
```
Komodo5197: I'd prefer a list with conditional entries here.

lib/models/finamp_models.dart
Komodo5197: I don't like the name additionalBaseItemInfo.  Maybe something like itemListSubtitleType?

lib/components/AlbumScreen/track_list_tile.dart
lukaslindnermusic: > Is it possible to accidentally show playcount/lastPlayed offline via adaptiveAdditionalInfoSortBy?

No, I don't think so, because the adaptive context changes as well. It's only relevant if for tracks the additional info is explicitly set to playCount or lastPlayed.

Should I make the ref.watch and fallback to 'none' conditional for these scenarios and/or use ref.read instead?

lib/components/GenreScreen/genre_screen_content.dart
lukaslindnermusic: Hmm... I don't know if I think "Random" is a better name. I'm more expecting it to play everything with "Random", to be honest.
But I definitely get your point and have also thought about using a different name.
However, since this is only a temporary solution, I came to the conclusion that it is unnecessary to introduce a new string that has to be translated etc.

@Chaphasilor what do you think regarding this?

lib/components/AlbumScreen/track_list_tile.dart
Komodo5197: You still want ref.watch.  I basically just wanted the watch statement in an if block.  Maybe something like this:

```
    final isOffline = ref.watch(finampSettingsProvider.isOffline);
    final additionalBaseItemInfos = ref.watch(finampSettingsProvider.additionalBaseItemInfo);
    final additionalBaseItemInfo =additionalBaseItemInfos[TabContentType.tracks] ? AdditionalBaseItemInfoTypes.adaptive;

    bool showPlayCount = additionalBaseItemInfo == AdditionalBaseItemInfoTypes.playCount;
    bool showReleaseDate = additionalBaseItemInfo == AdditionalBaseItemInfoTypes.dateReleased;
    bool showDateAdded = additionalBaseItemInfo == AdditionalBaseItemInfoTypes.dateAdded;
    bool showDateLastPlayed = additionalBaseItemInfo == AdditionalBaseItemInfoTypes.dateLastPlayed;

    if (additionalBaseItemInfo == AdditionalBaseItemInfoTypes.adaptive) {
      showPlayCount = showPlayCount || adaptiveAdditionalInfoSortBy == SortBy.playCount;
      showReleaseDate = showReleaseDate || adaptiveAdditionalInfoSortBy == SortBy.premiereDate;
      showDateAdded = showDateAdded || adaptiveAdditionalInfoSortBy == SortBy.dateCreated;
      showDateLastPlayed = showDateLastPlayed || adaptiveAdditionalInfoSortBy == SortBy.datePlayed;
    }

if(showPlayCount||showDateLastPlayed){
  if(ref.watch(finampSettingsProvider.isOffline)){
    showPlayCount=false;
    showDateLastPlayed=false;
}
}
```

lib/components/GenreScreen/genre_screen_content.dart
Komodo5197: My intuition would be that shuffle contains everything, like shuffling a deck of cards, while random just gets a random selection of arbitrary size.  But I agree the name isn't the best.  We're already using this algorithm for the shuffle all songs feature, so having a more descriptive name would help there as well.  Regarding the temporary nature of this, I've been thinking about a way to implement a shuffle all for genres or all songs, but I haven't come up with anything reasonable.  Even if we update the queue to allow generated instead of fixed track lists, that only gets us clean playback of all items in order.  Trying to get a shuffle requires sending all already played tracks to the server, or requesting all tracks from the server to shuffle ourselves, neither of which seems like a scalable option.  We might need to continue using random instead of shuffle, but we keep drawing new random lists as we play without guaranteeing everything plays exactly once.

lib/models/finamp_models.dart
lukaslindnermusic: Oh, why is that? For me it worked, but I've not tested any edge cases, so theres probably something that could go wrong in certain circumstances?

SettingsHelperMap then just stores strings, correct?

lib/components/LayoutSettingsScreen/CustomizationSettingsScreen/additional_base_item_info_dropdown_list_tile.dart
lukaslindnermusic: What exactly do you mean? 
Something like that?
```
final Map<TabContentType, Set<AdditionalBaseItemInfoTypes>> allowedTypesMap = {
  TabContentType.tracks: {
    AdditionalBaseItemInfoTypes.playCount,
    AdditionalBaseItemInfoTypes.dateLastPlayed,
    AdditionalBaseItemInfoTypes.dateReleased,
  },
  TabContentType.albums: {
    AdditionalBaseItemInfoTypes.duration,
    AdditionalBaseItemInfoTypes.dateReleased,
  },
  TabContentType.playlists: {
    AdditionalBaseItemInfoTypes.duration,
  },
  TabContentType.artists: {
    AdditionalBaseItemInfoTypes.duration,
  },
};

// Types that are always shown
const commonTypes = {
  AdditionalBaseItemInfoTypes.adaptive,
  AdditionalBaseItemInfoTypes.dateAdded,
  AdditionalBaseItemInfoTypes.none,
};

// Build the list
final dropdownItems = AdditionalBaseItemInfoTypes.values.where((type) {
  return commonTypes.contains(type) ||
      allowedTypesMap[tabContentType]?.contains(type) == true;
}).toList();
```

lib/models/finamp_models.dart
lukaslindnermusic: Hmm, but its not the only thing influencing the subtitle as a whole, so I wouldn't make it so deterministic with calling it SubtitleType.
Maybe `additionalSubtitleInfo`?

lib/components/GenreScreen/genre_screen_content.dart
lukaslindnermusic: We could maybe call it similar to what I think Apple Music either does or did: Radio?
"Start Radio"

or is that too confusing or metaphorical?

lib/models/finamp_models.dart
Komodo5197: I'm not sure it's really important for the setting to specify additional, but I'll defer to you.  I do feel that we want to end with Type, because its storing an enum and not the actual info.

lib/components/LayoutSettingsScreen/CustomizationSettingsScreen/additional_base_item_info_dropdown_list_tile.dart
Komodo5197: I was thinking like

var dropdownItem = [
AdditionalBaseItemInfoTypes.adaptive,
AdditionalBaseItemInfoTypes.dateAdded,
if([TabContentType.artists,TabContentType.playlists].contains(tabContentType))
AdditionalBaseItemInfoTypes.duration,];

lib/models/finamp_models.dart
Komodo5197: The generated provider doesn't work properly when watching the whole map, because the object id never changes.  You need the annotation to make add the key argument so you can set/watch individual values.

lib/components/GenreScreen/genre_screen_content.dart
Chaphasilor: So, Jellyfin also handles shuffle that way, by having a fixed-length queue and stopping playback after the last track has played.  
I think as long as it's a true shuffle based on all tracks, and *then* we slice it to a certain length, it should be find and close enough to the user's intention.  
Toggling shuffle after the queue has been created won't un-shuffle the tracks naturally, but I think that's fine for now and consistent with other shuffle buttons in Finamp.

Only thing I'd consider is showing the amount, e.g. "Shuffle 250 tracks", but that might be more confusing than helpful.  
Other languages might also not need this distinction at all.

I agree that there is a semantic difference between shuffling and random selection, especially when it comes to requesting additional "chunks"; a true shuffle implies that every track is in the selection exactly once.  
I also haven't been able to think of a good solution involving the server here, the main issue is that it's too slow. Even if we just want all track IDs of the entire library (to shuffle locally), that will take many seconds for larger libraries, so it's not feasible. For the generative queue, we could keep track of all tracks played in the current queue and use that to filter out duplicates, and *eventually* that should give us all tracks. But the amount of requests required to get to this point is undetermined.  
But at the same time, the issues appear mainly for larger libraries, and for those it isn't really feasible to play the entire library in one session anyway. So there probably *are* ample opportunities to fetch all tracks little-by-little.

Chaphasilor: Oh and I wouldn't call this a radio either. A radio lets you discover new tracks that you enjoy, this is not really the same.

lib/models/finamp_models.dart
Chaphasilor: How about `TileExtraInfoType` or `TileAdditionalInfoType`? Since this refers to additional info shown on item tiles, and is not really related to the BaseItemDto. Sure, that influences what can be shown, but I think this type is more related to the UI than to the underlying data.

lib/components/GenreScreen/genre_screen_content.dart
lukaslindnermusic: How about calling it "Shuffle (250 tracks)", And then offer the option to "Shuffle 250 to end of queue" if the user wants to expand the queue? 
We could also think about removing duplicates locally... let's be real: when someone plays a queue of more than 250 items, they probably don't care if there are a few duplicates in the next (second, third,...) 250 tracks. But if they do care, we could filter it locally (to avoid sending all the excludeItemIds to the server), but then we should be transparent about the fact that it may be less then 250. @Chaphasilor, please don't shout at me, but... this could be a setting 😅

Chaphasilor: I was talking about local filtering all this time. We could simply have a margin for requesting more tracks than we need, and then filtering. Or making another request. I really don't think local vs. remote filtering needs to be user-configurable?  
The "Shuffle 250 to end of queue" button seems out-of-scope for this PR. I'd be *fine* with the "Shuffle (250 tracks)" label, although not thrilled. You decided if you want to include this here :)

lib/components/LayoutSettingsScreen/CustomizationSettingsScreen/additional_base_item_info_dropdown_list_tile.dart
```diff
@@ -24,48 +24,28 @@ class AdditionalBaseItemInfoDropdownListTile extends ConsumerWidget {
...
-        case TabContentType.albums:
-          return type == AdditionalBaseItemInfoTypes.duration || type == AdditionalBaseItemInfoTypes.dateReleased;
-        case TabContentType.playlists:
-        case TabContentType.artists:
-          return type == AdditionalBaseItemInfoTypes.duration;
-        default:
-          return false;
-      }
-    }).toList();
+    final dropdownItems = [
```
Komodo5197: This was more of an example than a full suggestion.  This list needs more types.

lib/components/MusicScreen/item_collection_list_tile.dart
```diff
@@ -101,17 +101,19 @@ class ItemCollectionListTile extends ConsumerWidget {
       associatedTabContentType = null;
     }
 
-    final additionalBaseItemInfos = ref.watch(finampSettingsProvider.additionalBaseItemInfo);
-    final additionalBaseItemInfo =
+    final tileAdditionalInfos = associatedTabContentType != null
```
Komodo5197: Call this tileAdditionalInfoSetting or something.  tileAdditionalInfos doesn't mean anything.

lib/models/finamp_models.dart
```diff
@@ -723,8 +723,9 @@ class FinampSettings {
   @HiveField(120, defaultValue: DefaultSettings.useHighContrastColors)
   bool useHighContrastColors;
 
-  @HiveField(121, defaultValue: DefaultSettings.additionalBaseItemInfo)
-  Map<TabContentType, AdditionalBaseItemInfoTypes> additionalBaseItemInfo;
+  @HiveField(121, defaultValue: DefaultSettings.tileAdditionalInfo)
+  @SettingsHelperMap("tabContentType", "tileAdditionalInfo")
+  Map<TabContentType, TileAdditionalInfoType> tileAdditionalInfo;
```
Komodo5197: I'd prefer tileAdditionalInfoType or something.  It's not the actual info, just an enum of which type of info to show.  tileAdditionalInfo is still decently readable if you disagree, though.

lukaslindnermusic:
Thanks @Komodo5197 , that did it 🥳
@Chaphasilor I think we can merge now 😊

Chaphasilor:
Awesome, thanks for the PR! <3

Issue #1056, opened 2025-02-18
Discord rich presence intergration


bensonchow123:
It would be nice for finamp to have discord rich presence, and it is the only feature from spotify that i actively uses that is not in finamp.

Issue #738, opened 2024-05-21
[Feature Request] Discord Rich Presence for desktop


akitakedits:
Title says it all, Discord Rich Presence would be nice for desktop, so you can let people can know what you're listening to.
Low priority obviously.

Titaniumtown:
Why though?

jmshrv:
This would be fun to have, afaik there aren't really any mature Dart libraries for Discord Rich Presence so it would be quite a bit of work (there is https://pub.dev/packages/dart_discord_rpc, but it doesn't support all platforms)

erictbar:
Since Finamp uses Jellyfin, I just use this: [https://github.com/Radiicall/jellyfin-rpc](https://github.com/Radiicall/jellyfin-rpc) and it displays whatever is playing on the server (as long as Discord is running on my PC), whether it be music or video

aliak74:
I would really love to have this feature as well and could try and test this. I use finamp several times a day usually. Just a user though.

flloschy:
Will be included in the next release :)

PR #1298, opened 2025-07-22
Feature: Discord RPC

flloschy:
## Changes
Adding Discord RPC functionality to show what song Finamp is playing inside Discord (:O)
- RPC stops when offline mode is active
- RPC updates every 15 seconds
- Displays songName, artist, album.
- Also sends image urls (cover+artist) to discord (customizable fallback finamp logo if no public address is in use)
- Clears RPC when playback is stopped or paused

## Todo before merging
- [x] Translations
- [x] Reset Settings
- [x] ~~Extended CONTRIBUTING.md~~ nothing worth adding I believe
- [x] Adding Settings
- [x] Initializing the Object (and then testing this) 
- [x] Handle network connectivity (the library doesnt seam to do this. Maybe not even required since discord is locally accessible?)

## Related Issues
closes https://github.com/jmshrv/finamp/issues/738
closes https://github.com/jmshrv/finamp/issues/1056


lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,84 @@
...
+    ProviderContainer container = GetIt.instance<ProviderContainer>();
+    container.listen(discordRpcProvider, (_, enabled) {
+      enabled ? DiscordRpc.start() : DiscordRpc.stop(); 
+    });
+  }
+
+  static Future<void> start() async {
+    if (!running) {
+      running = true;
+      await FlutterDiscordRPC.initialize("1363567299948314906");
```
Chaphasilor: you could use the unique device ID from the auth header for this?

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:get_it/get_it.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+import 'package:flutter_discord_rpc/flutter_discord_rpc.dart';
+part 'discord_rpc.g.dart';
+
+var running = false;
+StreamSubscription<MediaState>? listener;
+
+@riverpod
```
Chaphasilor: Is there really any reason for making this a provider? We never actually read any data *from* it, do we?  

And you're already using the new `ProviderContainer` too :)

lib/services/discord_rpc.dart
Chaphasilor: ah nevermind, this ID is coming from Discord.  
Is it a secret or can we just expose it like that?

Chaphasilor (commented):
Nice work! One thing I'd definitely like to see is a setting to toggle this on or off. I guess the default should be off for privacy reasons :)

lib/services/discord_rpc.dart
flloschy: I was just copying what ever was done to AutoOffline

lib/services/discord_rpc.dart
flloschy: This is public, yes
When you go to discord.dev and create an application with the name "Jellyfin" (thats what I did) and then simply use the appID, the RPC will show "listening to Jellyfin"

lib/services/discord_rpc.dart
Chaphasilor: I believe that is either a) unneeded or b) was done to apply new settings immediately. I think the solution from the playon_service works just as well, without code generation...

lib/services/discord_rpc.dart
Chaphasilor: but we only need to do this once, right? Or does every user need to do this?  
Also, why not Finamp? Or "Jellyfin via Finamp"?

flloschy: yes, only once
because Im currently reusing the id from my rpc xd

Chaphasilor: did you need to configure the app in the Discord portal? or is the "listening" bit done just through the code?

flloschy: Yes its code. It can be listening, streaming, watching and playing

flloschy:
For some reason the RPC wont bet updated 100% of the time and may display outdated information :/

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,138 @@
...
+      startListener();
+    } else {
+      _rpcLogger.info("Attempted to Start RPC even though its already running");
+    }
+  }
+
+  static Future<void> stop() async {
+    if (_running) {
+      _rpcLogger.info("Stopping RPC");
+      _running = false;
```
Komodo5197: I think this should be changed after the awaits, not before.

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,175 @@
...
+
+  static Future<void> _start() async {
+    if (!_running) {
+      _rpcLogger.info("Starting RPC");
+      _running = true;
+      _createTimer(); 
+      await FlutterDiscordRPC.instance.connect(autoRetry: true);
+      _startListener();
+    } else {
+      _rpcLogger.info("Attempted to Start RPC even though its already running");
```
Komodo5197: _start and _stop get called on every settings change. Don't log their redundant calls.

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,175 @@
...
+    final enabled = FinampSettingsHelper.finampSettings.rpcEnabled;
+    final offline = FinampSettingsHelper.finampSettings.isOffline;
+    (enabled && !offline) ? _start() : _stop();
+  }
+
+  static Future<void> _start() async {
+    if (!_running) {
+      _rpcLogger.info("Starting RPC");
+      _running = true;
+      _createTimer();
```
Komodo5197: Start the timer after we've connected.

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,175 @@
...
+  //   return details && end && start && largeImage && smallImage && largeText && smallText;
+  // }
+
+
+  static void _startListener() {
+    
+
+    _listener = mediaStateStream.listen((state) async {
+      if (!state.playbackState.playing) {
+        if (lastState != null) {
```
Komodo5197: Why are we still doing things with lastState instead of just setting curtrentState for the timer to update?

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,175 @@
...
+          currentState = null;
+          await _updateRPC();
+        };
+        return;
+      }
+
+      final baseItem = BaseItemDto.fromJson(state.mediaItem!.extras!["itemJson"] as Map<String, dynamic>);
+      final artistItem = await _jellyfinApiHelper.getItemById(baseItem.artistItems!.first.id);
```
Komodo5197: I'm concerned about this potentially spamming requests.

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,175 @@
...
+    // if (_isDuplicate(currentState)) return;
+
+    await FlutterDiscordRPC.instance.setActivity(activity: currentState!);
+    lastState = currentState;
+
+    _rpcLogger.finer("Updated");
+  }
+
+  static Future<void> _stopListener() async {
+    await _listener?.cancel();
```
Komodo5197: I'd prefer to inline this and _createTimer.

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  static void _createTimer() {
+    // updates the rpc regularly to fix potential desyncs, keeps conection alive and also to prevent ratelimits
+    _timer = Timer.periodic(Duration(seconds: 5), (Timer time) {_updateRPC();});
+  }
+
+  static void _reEvaluate() {
+    final enabled = FinampSettingsHelper.finampSettings.rpcEnabled;
+    final offline = FinampSettingsHelper.finampSettings.isOffline;
```
Komodo5197, line 37: This probably makes sense, but not having a jellyfin connection does not necessarily imply not having a discord connection.

flloschy: While true I figured this makes sence since offline Mode implies no active api calls etc. But running the RPC means discord will have web traffic due to finamp which circumvents offline Mode

lib/services/discord_rpc.dart
flloschy: I think I fixed it by caching, but it would be nice to get the data locally, which surely is possible I just dont know how to get this data from a MediaItem's id

lib/services/discord_rpc.dart
flloschy:
```
>>> playback progress
[lastState=something currentState=something_else]
*rpc updated*

>>> stopped playback
[lastState=something currentState=null]
*rpc disabled*

>>> song switched
[lastState=null currentState=null]
*no rpc call*
```
Does that make sense?

lib/services/discord_rpc.dart
flloschy: My thought was to prevent a second call which might happen ASAP so the awaited functions dont get called for no reason

Komodo5197: Yeah, but that means we could get early runs of start().  We might need something more like the playonservice with a 3 state enum that we can set to a processing state at the beginning of start/stop.  Then we can use a try/finally to make sure it always gets set to a non-processing value at the end.

lib/services/discord_rpc.dart
Komodo5197: Yeah, I see how we still need lastState to make sure that we only fire once on null state.  I'd prefer to have all the laststate logic inside _updateRPC and just set currentstate=null here, though.

lib/services/discord_rpc.dart
flloschy: like so? https://github.com/jmshrv/finamp/pull/1298/commits/54d6cf515b8ce262908414b6e63006abec7841cf + https://github.com/jmshrv/finamp/pull/1298/commits/5b283156aa9310a1fd5b5fef6b252bd428352860

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,145 @@
...
+    final state = GetIt.instance<MusicPlayerBackgroundTask>();
+    final playbackState = state.playbackState.valueOrNull;
+    final mediaItem = state.mediaItem.valueOrNull;
+
+    if (playbackState?.playing == null || !playbackState!.playing) {
+      return null;
+    }
+
+    final baseItem = BaseItemDto.fromJson(mediaItem!.extras!["itemJson"] as Map<String, dynamic>);
+    if (artistItem == null || !baseItem.artists!.any((v) => v == artistItem?.name)) {
```
Komodo5197: Why are we checking by name instead of id?

lib/services/discord_rpc.dart
Komodo5197: Looks good to me.  I'm pretty sure we can't find the artist image without the full item - we already do this for artist_chips.dart.

lib/services/discord_rpc.dart
Komodo5197: Yeah.  Although as I said, I'd like a try/finally so that we don't get stuck inn starting/stopping if the rpc functions throw an error.  Also, if you want calling start while stopping or stop while starting to work, it might make sense to create a targetState bool that's set immediately on function call.  Then at the end of start you have if(!targetState) stop(), and vice-versa instead of having to use a retry timer.

lib/services/discord_rpc.dart
Komodo5197: It's a bit nebulous, because we didn't anticipate calls to outside servers. I'd imagine the average user would want this to track internet connectivity, not jellyfin connectivity. But that feels overcomplicated.

lib/services/discord_rpc.dart
flloschy: > Although as I said, I'd like a try/finally so that we don't get stuck inn starting/stopping if the rpc functions throw an error

Im unsure why a start/stop loop should happen when a rpc error happens?

> if(!targetState) stop(), and vice-versa instead of having to use a retry timer

good call

lib/services/discord_rpc.dart
flloschy: Does this still apply now?

lib/main.dart
```diff
@@ -455,10 +462,11 @@ class _FinampState extends State<Finamp> with WindowListener {
   }
 
   @override
-  void dispose() {
+  Future<void> dispose() async {
     if (Platform.isWindows || Platform.isLinux || Platform.isMacOS) {
       WindowManager.instance.removeListener(this);
     }
+    await DiscordRpc.stop().timeout(Duration(milliseconds: 500));
```
Komodo5197: Shouldn't we check if the setting is enabled, or at least if we are on desktop?

lib/services/discord_rpc.dart
Komodo5197: I'm not 100% sure, but we can probably do better than getting stuck in transitioning and never doing anything again until an app restart?

lib/main.dart
flloschy: We could, though the stop method already only executes code when the status is running, which is only possible when we are on desktop and the setting is enabled

lib/services/discord_rpc.dart
Komodo5197: Haven't you put all the last state management inside updateRPC now, fulfilling the request?

lib/services/discord_rpc.dart
Chaphasilor: I'd say we isolate Finamp as much as possible when offline mode is enabled; it might also be used as a form of privacy protection. If people request that this should be changed, we can make it configurable. But I'd rather err on the side of caution and privacy here :)

lib/services/discord_rpc.dart
flloschy: yeah, though a lot changed so I wasnt sure if this is still valid

lib/services/discord_rpc.dart
flloschy: I highly doubt this will happen since the user needs to enable/disable this feature repeatedly for this to happen and at that point its expected behavior

flloschy:
The checks lie
I can build locally with no problem :)

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,197 @@
...
+  /// - `172.16.X.X` - `172.19.X.X` - `172.20.X.X` - `172.29.X.X` - `172.30.X.X` - `172.31.X.X`
+  /// - `192.168.X.X`
+  static bool isAddressInLocalAddressRange(String address) {
+    // regex from https://stackoverflow.com/a/2814102
+    final regex = RegExp(r"(^127\.)|(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)");
+    final host = Uri.parse(address).host;
+    return regex.hasMatch(host);
+  }
+
+  static List<String?> _fetchImageUrls(BaseItemDto baseItem) {
```
Komodo5197: It would make sense for this to return a tupple instead of a list

Chaphasilor: I wouldn't even mind creating a proper class/type for this, for self-explanatory typing. It's not like we're paying per line.  
It *is* probably overkill, but I don't think it has any other drawback?

Chaphasilor:
Is discussion for the unresolved conversations still ongoing? Or should they be marked as resolved? @Komodo5197

lib/services/discord_rpc.dart
Komodo5197: It starts/stops every time offline mode toggles.  If I were using it and saw an error pop up, I'd probably try toggling offline mode or the discord setting to try to make it retry, or expect it to retry itself.  I don't know the full details of how fallible this connection is though, or what sort of error handling/retrying is already built in.

flloschy: The build-in error handling only handles lost connection and then attempts to reconnect, all other errors will be handled by us
But i dont know either what errors could realistically happen...

But back to the original suggestion, what does a try/final block do different than the system currently is?

Chaphasilor: I guess setting the state tracking variables to a consistent state?

flloschy: I dont know what that means

Komodo5197: If any of the discordrpc functions throws an exception, we never change the state away from processing, so both start and stop will refuse to run.  A try/finally would allow us to always set the state to something before the function exits and allow one of those methods to be attempted again.

flloschy: ohhhhhhhhhhhhhhhhhhhhhhhh duh yeah of course, my bad 
For some reason I was thinking you want to throw a custom error no matter what or something like that which - logically - doesn't really make sense

flloschy: [c4a82d9](https://github.com/jmshrv/finamp/pull/1298/commits/c4a82d94b618e35dad6d8bfd53b8e64db5775540) :+1:

lib/services/discord_rpc.dart
```diff
@@ -0,0 +1,217 @@
...
+      _status = _RpcStatus.transition;
+      _rpcLogger.info("Starting RPC");
+
+      try {
+        await FlutterDiscordRPC.instance.connect(autoRetry: true, retryDelay: Duration(seconds: 3));
+        await FlutterDiscordRPC.instance.clearActivity();
+      } catch (e) {
+        _rpcLogger.severe("Failed to Start", e);
+        _targetStatus = _RpcStatus.stopped;
+        await stop();
```
Komodo5197: This doesn't do anything because our _status is still transition.  Did you mean to set _status instead of _targetStatus?

flloschy: Good catch yeah. Shouldnt commit things with a 3 am brain ':D

Chaphasilor:
I've decided to squash-merge since the changes were actually pretty narrowly focused and there wasn't some important detail in the commit messages :)  
Thanks for the PR!

Issue #1102, opened 2025-03-10
[Feature request] (Beta) Support sequental album track within playlists in hybrid ReplayGain mode


HeroBrine1st:
I'm on 0.9.14. Currently, if I place tracks with gapless playback in playlist with hybrid replaygain, there may be sudden jump of volume on next track, and due to tracks being "unite" it is easily seen.

# Current behavior

Hybrid ReplayGain uses track gain even if both tracks are from the same album and list [previousTrack, nextTrack] is a sublist of all tracks in their album.

# Expected behavior

Hybrid ReplayGain uses album gain in such cases, preventing volume jump on gapless albums.

# Proposed solution

The most optimal solution is to find "sequental tracks" in queue and use album gain in those ranges, using this algorithm:

<details>

<summary>I wrote that for other project so that it is python, anyway I want to show the algorithm</summary>

```python
    for playlist in playlists:
        sequental_ranges = []
        i = 0
        while i < len(playlist["tracks"]):
            start = i
            # Find the end of the current sequential range
            while (
                (i + 1) < len(playlist["tracks"])
                and playlist["tracks"][i + 1]["album_uri"] == playlist["tracks"][i]["album_uri"]  # same album
                and (  # same disk
                    (playlist["tracks"][i]["disc_number"] == playlist["tracks"][i + 1]["disc_number"] and
                     playlist["tracks"][i]["track_number"] + 1 == playlist["tracks"][i + 1]["track_number"])
                    or  # next disk
                    (playlist["tracks"][i]["disc_number"] + 1 == playlist["tracks"][i + 1]["disc_number"] and
                     playlist["tracks"][i + 1]["track_number"] == 1)
                )
            ):
                i += 1
            end = i
            # Add the range to the list if it contains at least two tracks
            if end > start:
                sequental_ranges.append(range(start, end + 1))
            i += 1
        playlist["sequental_ranges"] = sequental_ranges
```
"sequental_ranges" then contains every range where you need to use album gain
</details>

There's also other solution, I think it is the one Spotify uses. It involves adding fourth option to volume normalization - "always use album gain". Looking at source code, I see that you already considered that:

https://github.com/jmshrv/finamp/blob/1f3e37bb78839a46b13cd5fb27a01ffb0ea0795e/lib/services/music_player_background_task.dart#L897-L919

[...]

PR #1302, opened 2025-07-27
Use album gain outside of album contexts

HeroBrine1st:
## Changes

This PR uses album gain separately fetched from jellyfin server via `currentTrackMetadataProvider`.

<details>
<summary>Previous text</summary>

This fixes https://github.com/jmshrv/finamp/issues/1102.. and adds a bunch of new bugs (don't know any but it is inevitable, I know)

I had a bunch of ideas how to fix that, however I found [this issue in jellyfin](https://github.com/jellyfin/jellyfin/issues/14346) (can be linked as upstream to #1102 btw), where [it is told](https://github.com/jellyfin/jellyfin/issues/14346#issuecomment-2988016426) that jellyfin averages track gains to get album gain. Knowing that, my implementation averages known part of album, removing the need to do any network requests or changing jellyfin API.

<sup>(however the latter would be very good but I know dotnet (0 hours) even less than Dart (3-4 hours before I abandoned it))</sup>

The algorithm: take tracks from the same album before and after currently playing track (so that resulting list is subset of tracks of their album), then average their gain to get "album" gain.

Due to my lack of experience in Dart language, I think this PR is more a PoC than a reliable fix. Currently it has production-working asserts (I *probably* tested it on production using `flutter build linux`).

Tested on album Myth by Two Steps from Hell (inside a playlist ofc), first two tracks in my case have like 10db difference and those tracks are the cause of my issue ;-)

And the first commit is to enhance code clarity and improve compiler assurance on soundness (primarily via extracting pointer to queue list for DRY).
</details>

## Todo before merging

- [x] Make a new `Album-based gain` setting
- [x] Handle sequental album tracks in hybrid gain (so that `contextNormalizationGain` is obsolete)
  - [ ] Remove contextNormalizationGain?
- [x] Handle offline mode
- ~~Add "Album-based with fallback to track-based" gain without handling sequental tracks (my own testing feedback :P for some reason singles don't have album gain).~~  
  ~~I think it is worth making current "Album-based gain" the "Album-only gain" (because it is, well, album-only) and add fallback to track gain for "Album-based"~~ We decided to add fallback for album-based gain. The remaining is related to "remove contextNormalizationGain" item
- [x] Create an "album provider" so that albums are cached in memory and not requested for every track in queue

## Related Issues

Fixes https://github.com/jmshrv/finamp/issues/1102

P.s. additional thanks to the author of flake.nix - won't link PR 1299 not to spam but you helped me so much. Before I thought I better base my changes on release tag for usage of `nixpkgs`'s derivation.
P.s.s. those averages can also be calculated at queue modification time for performance at the energy cost (as most of the time those averages won't be used on large playlists)

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -1474,13 +1474,13 @@
   "@volumeNormalizationModeSelectorSubtitle": {
     "description": "Subtitle for the dropdown that selects the replay gain mode"
   },
-  "volumeNormalizationModeSelectorDescription": "Hybrid (Track + Album):\nTrack gain is used for regular playback, but if an album is playing (either because it''s the main playback queue source, or because it was added to the queue at some point), the album gain is used instead.\n\nTrack-based:\nTrack gain is always used, regardless of whether an album is playing or not.\n\nAlbums Only:\nVolume Normalization is only applied while playing albums (using the album gain), but not for individual tracks.",
+  "volumeNormalizationModeSelectorDescription": "Dynamic:\nWhen playing multiple tracks from the same album in a row, album gain is used (i.e. when playing an album, or tracks from the same album in a playlist). Otherwise, track gain is used.\n\nTrack-based:\nTrack gain is always used, regardless of whether an album is playing or not.\n\nAlbum-based:\nAlbum gain is always used, but if it's not available (i.e. for singles), track gain is used instead.\n\nAlbums Only:\nVolume normalization is only applied while playing albums (using the album gain), but not for individual tracks.",
   "@volumeNormalizationModeSelectorDescription": {
     "description": "Description for the dropdown that selects the replay gain mode, shown in a dialog that opens when the user presses the info icon next to the dropdown"
   },
-  "volumeNormalizationModeHybrid": "Hybrid (Track + Album)",
+  "volumeNormalizationModeHybrid": "Dynamic",
   "@volumeNormalizationModeHybrid": {
-    "description": "'Hybrid' option for the replay gain mode dropdown. In hybrid mode, the track gain is used for regular playback, but if an album is playing (either because it's the main playback queue source, or because it was added to the queue at some point), the album gain is used instead."
+    "description": "'Dynamic' option for the replay gain mode dropdown. In dynamic mode, the track gain is used for regular playback, but if an album is playing (either because it's the main playback queue source, or because it was added to the queue at some point), or successive tracks from a same album are in the queue, the album gain is used instead."
   },
   "volumeNormalizationModeTrackBased": "Track-based",
   "@volumeNormalizationModeTrackBased": {
--- a/lib/services/metadata_provider.dart
+++ b/lib/services/metadata_provider.dart
@@ -43,12 +43,14 @@ class MetadataProvider {
   LyricDto? lyrics;
   bool isDownloaded;
   bool qualifiesForPlaybackSpeedControl;
+  double? parentNormalizationGain;
 
   MetadataProvider({
     required this.mediaSourceInfo,
     this.lyrics,
     this.isDownloaded = false,
     this.qualifiesForPlaybackSpeedControl = false,
+    this.parentNormalizationGain,
   });
 
   bool get hasLyrics => mediaSourceInfo.mediaStreams.any((e) => e.type == "Lyric");
@@ -167,8 +169,7 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
   }
 
   BaseItemDto? parent;
-  if (request.item.parentId != null &&
-      (request.checkIfSpeedControlNeeded && !metadata.qualifiesForPlaybackSpeedControl)) {
+  if (request.item.parentId != null) {
     if (!FinampSettingsHelper.finampSettings.isOffline) {
       try {
         parent = await jellyfinApiHelper.getItemById(request.item.parentId!);
@@ -200,6 +201,10 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
       parent != null &&
       (parent.runTimeTicks ?? 0) > MetadataProvider.speedControlLongAlbumDuration.inMicroseconds * 10) {
     metadata.qualifiesForPlaybackSpeedControl = true;
+  }
+
+  if (parent != null) {
+    metadata.parentNormalizationGain = parent.normalizationGain;
   }
 
   if (request.includeLyrics && metadata.hasLyrics) {
--- a/lib/services/metadata_provider.dart
+++ b/lib/services/metadata_provider.dart
@@ -157,7 +157,6 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
 
   final metadata = MetadataProvider(mediaSourceInfo: playbackInfo, isDownloaded: localPlaybackInfo != null);
 
-  // check if item qualifies for having playback speed control available
   if (request.checkIfSpeedControlNeeded) {
     for (final genre in request.item.genres ?? []) {
       if (MetadataProvider.speedControlGenres.contains(genre.toLowerCase())) {
@@ -170,30 +169,43 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
             MetadataProvider.speedControlLongTrackDuration.inMicroseconds * 10) {
       // we might want playback speed control for long tracks (like podcasts or audiobook chapters)
       metadata.qualifiesForPlaybackSpeedControl = true;
+    }
+  }
+
+  BaseItemDto? parent;
+  if (request.item.parentId != null &&
+      (request.checkIfSpeedControlNeeded && !metadata.qualifiesForPlaybackSpeedControl)) {
+    if (!FinampSettingsHelper.finampSettings.isOffline) {
+      try {
+        parent = await jellyfinApiHelper.getItemById(request.item.parentId!);
+      } catch (e) {
+        metadataProviderLogger.warning(
+          "Failed to get parent item ('${request.item.parentId}') for '${request.item.name}' (${request.item.id})",
+          e,
+        );
+      }
     } else {
-      // check if "album" is long enough to qualify for playback speed control
-      if (request.item.parentId != null) {
-        if (FinampSettingsHelper.finampSettings.isOffline) {
-          final parent = await downloadsService.getCollectionInfo(id: request.item.parentId);
-          if ((parent?.baseItem?.runTimeTicks ?? 0) >
-              MetadataProvider.speedControlLongAlbumDuration.inMicroseconds * 10) {
-            metadata.qualifiesForPlaybackSpeedControl = true;
-          }
-        } else {
-          try {
-            final parent = await jellyfinApiHelper.getItemById(request.item.parentId!);
-            if ((parent.runTimeTicks ?? 0) > MetadataProvider.speedControlLongAlbumDuration.inMicroseconds * 10) {
-              metadata.qualifiesForPlaybackSpeedControl = true;
-            }
-          } catch (e) {
-            metadataProviderLogger.warning(
-              "Failed to check if '${request.item.name}' (${request.item.id}) qualifies for playback speed controls",
-              e,
-            );
-          }
-        }
+      final parentInfo = await downloadsService.getCollectionInfo(id: request.item.parentId!);
+      if (parentInfo == null) {
+        metadataProviderLogger.warning(
+          "Couldn't find parent collection '${request.item.parentId}' for track '${request.item.id}' in offline mode",
+        );
+      } else if (parentInfo.baseItem == null) {
+        metadataProviderLogger.warning(
+          "Offline metadata for '${request.item.parentId}' (parent of '${request.item.id}') does not include jellyfin BaseItemDto",
+        );
+      } else {
+        parent = parentInfo.baseItem;
       }
     }
+  }
+
+  // check if item qualifies for having playback speed control available
+  if (request.checkIfSpeedControlNeeded &&
+      !metadata.qualifiesForPlaybackSpeedControl &&
+      parent != null &&
+      (parent.runTimeTicks ?? 0) > MetadataProvider.speedControlLongAlbumDuration.inMicroseconds * 10) {
+    metadata.qualifiesForPlaybackSpeedControl = true;
   }
 
   if (request.includeLyrics && metadata.hasLyrics) {
--- a/lib/services/metadata_provider.dart
+++ b/lib/services/metadata_provider.dart
@@ -11,24 +11,19 @@ import 'jellyfin_api_helper.dart';
 final metadataProviderLogger = Logger("MetadataProvider");
 
 class MetadataRequest {
-  const MetadataRequest({required this.item, this.includeLyrics = false, this.checkIfSpeedControlNeeded = false})
-    : super();
+  const MetadataRequest({required this.item, this.includeLyrics = false}) : super();
 
   final BaseItemDto item;
 
   final bool includeLyrics;
-  final bool checkIfSpeedControlNeeded;
 
   @override
   bool operator ==(Object other) {
-    return other is MetadataRequest &&
-        other.includeLyrics == includeLyrics &&
-        other.checkIfSpeedControlNeeded == checkIfSpeedControlNeeded &&
-        other.item.id == item.id;
+    return other is MetadataRequest && other.includeLyrics == includeLyrics && other.item.id == item.id;
   }
 
   @override
-  int get hashCode => Object.hash(item.id, includeLyrics, checkIfSpeedControlNeeded);
+  int get hashCode => Object.hash(item.id, includeLyrics);
 }
 
 /// A storage container for metadata about a track.  The codec information will reflect
@@ -153,19 +148,17 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
 
   final metadata = MetadataProvider(mediaSourceInfo: playbackInfo, isDownloaded: localPlaybackInfo != null);
 
-  if (request.checkIfSpeedControlNeeded) {
-    for (final genre in request.item.genres ?? []) {
-      if (MetadataProvider.speedControlGenres.contains(genre.toLowerCase())) {
-        metadata.qualifiesForPlaybackSpeedControl = true;
-        break;
-      }
-    }
-    if (!metadata.qualifiesForPlaybackSpeedControl &&
-        (metadata.mediaSourceInfo.runTimeTicks ?? 0) >
-            MetadataProvider.speedControlLongTrackDuration.inMicroseconds * 10) {
-      // we might want playback speed control for long tracks (like podcasts or audiobook chapters)
+  for (final genre in request.item.genres ?? []) {
+    if (MetadataProvider.speedControlGenres.contains(genre.toLowerCase())) {
       metadata.qualifiesForPlaybackSpeedControl = true;
+      break;
     }
+  }
+  if (!metadata.qualifiesForPlaybackSpeedControl &&
+      (metadata.mediaSourceInfo.runTimeTicks ?? 0) >
+          MetadataProvider.speedControlLongTrackDuration.inMicroseconds * 10) {
+    // we might want playback speed control for long tracks (like podcasts or audiobook chapters)
+    metadata.qualifiesForPlaybackSpeedControl = true;
   }
 
   BaseItemDto? parent;
@@ -196,8 +189,7 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
   }
 
   // check if item qualifies for having playback speed control available
-  if (request.checkIfSpeedControlNeeded &&
-      !metadata.qualifiesForPlaybackSpeedControl &&
+  if (!metadata.qualifiesForPlaybackSpeedControl &&
       parent != null &&
       (parent.runTimeTicks ?? 0) > MetadataProvider.speedControlLongAlbumDuration.inMicroseconds * 10) {
     metadata.qualifiesForPlaybackSpeedControl = true;
--- a/lib/services/metadata_provider.dart
+++ b/lib/services/metadata_provider.dart
@@ -10,22 +10,6 @@ import 'jellyfin_api_helper.dart';
 
 final metadataProviderLogger = Logger("MetadataProvider");
 
-class MetadataRequest {
-  const MetadataRequest({required this.item, this.includeLyrics = false}) : super();
-
-  final BaseItemDto item;
-
-  final bool includeLyrics;
-
-  @override
-  bool operator ==(Object other) {
-    return other is MetadataRequest && other.includeLyrics == includeLyrics && other.item.id == item.id;
-  }
-
-  @override
-  int get hashCode => Object.hash(item.id, includeLyrics);
-}
-
 /// A storage container for metadata about a track.  The codec information will reflect
 /// the downloaded file if appropriate, even for transcoded downloads.  Online
 /// transcoding will not be reflected.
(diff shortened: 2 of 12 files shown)


Chaphasilor:
Hey @HeroBrine1st, sorry for never replying to your issue! I read the notification and wanted to read it in detail, but it simply drowned in other notifications...

Thanks for the PR, that's definitely appreciated!  
I agree that averaging the gain is an acceptable, if not ideal, solution. But I think we can actually make the needed request:  
In the metadata provider, we pre-fetch information for upcoming tracks anyway. Maybe it would be possible to fetch the needed information there?  

I honestly don't think performance is much of a concern, but if you want to improve the performance I definitely won't stop you!  
I haven't looked at the code yet but will do that in the coming days :)

HeroBrine1st:
> In the metadata provider, we pre-fetch information for upcoming tracks anyway. Maybe it would be possible to fetch the needed information there?  

That's something I scratched today right before this implementation :-) but I should still have it shelved in my IDEA. It was untested due to the issues in the issue (sic!), otherwise should be good, iirc it added only one big request on each playlist open. And it will definitely wipe out race condition issues if any.

Chaphasilor:
@HeroBrine1st for every downloaded track, Finamp's download system should fetch the parent album's `BaseItemDto` as metadata. So you should be able to simply fetch the album gain from there when you need it, instead of injecting it somewhere during the download process!  
So you only need to handle the offline case in the metadata provider, where if Finamp is offline you fetch the album from the download system as opposed to the server.

Regarding race conditions and such, so, there are no guarantees that the metadata will be resolved before the playback starts. I've seen part of the metadata missing on slower connections myself.  
But I don't think this is really an issue, you could simply call `PlayerVolumeController.setReplayGainVolume()` from the metadata provider if the metadata fetching in done while the track is already playing, and you determine that the current gain is incorrect.  
I believe that approach is nicer than stalling playback until all metadata has arrived.

Overall really good job so far, thank you so much!

PS: Could you elaborate on the bug you've found and fixed? Maybe share a screen recording of how to reproduce it on `redesign`?

lib/services/metadata_provider.dart
```diff
@@ -170,32 +171,48 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
...
       // we might want playback speed control for long tracks (like podcasts or audiobook chapters)
       metadata.qualifiesForPlaybackSpeedControl = true;
+    }
+  }
+
+  BaseItemDto? parent;
+  if (request.item.parentId != null) {
+    if (!FinampSettingsHelper.finampSettings.isOffline) {
+      try {
+        parent = await jellyfinApiHelper.getItemById(request.item.parentId!);
```
Komodo5197: It might make sense to create a Future<BaseItemDto?> parentFuture with this offline logic above, somewhere before jellyfinApiHelper.getPlaybackInfo, and then just do parent=await parentFuture here.  This would allow both requests to run at once, minimizing potential latency.  Also, it might make sense to add some sort of caching to the parent requests.  Maybe something like adding 

`final itemByIdProvider = Provider.family.autoDispose((ref, BaseItemId id)=>GetIt.instance<JellyfinApiHelper>().getItemById(id));`

outside this function and then replacing the parent api call with 

`Future<BaseItemDto?> parentFuture = ref.watch(itemByIdProvide(request.item.parentId!));`

lib/services/metadata_provider.dart
```diff
@@ -170,32 +171,48 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
             MetadataProvider.speedControlLongTrackDuration.inMicroseconds * 10) {
       // we might want playback speed control for long tracks (like podcasts or audiobook chapters)
       metadata.qualifiesForPlaybackSpeedControl = true;
+    }
+  }
+
+  BaseItemDto? parent;
+  if (request.item.parentId != null) {
+    if (!FinampSettingsHelper.finampSettings.isOffline) {
```
Komodo5197: All of the uses of FinampSettingsHelper.finampSettings.isOffline in this function should be replaced with ref.watch(finampSettingsProvider.isOffline), and then the unused ref.watch(finampSettingsProvider.isOffline) at the head of the function can be removed.  This is a pre-existing style issue.

lib/services/music_player_background_task.dart
```diff
@@ -940,7 +951,12 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
     if (FinampSettingsHelper.finampSettings.volumeNormalizationActive && currentTrack != null) {
       final baseItem = jellyfin_models.BaseItemDto.fromJson(currentTrack.extras?["itemJson"] as Map<String, dynamic>);
 
-      double? effectiveGainChange = getEffectiveGainChange(currentTrack, baseItem);
+      double? effectiveGainChange = getEffectiveGainChange(
+        currentTrack,
+        baseItem,
+        _queueService.getQueueStream().valueOrNull?.isSequental() ?? false,
```
Komodo5197: This should be _queueService.getQueue().isSequental()

lib/services/music_player_background_task.dart
```diff
@@ -1082,20 +1098,51 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
...
       item ?? jellyfin_models.BaseItemDto.fromJson(currentTrack.extras?["itemJson"] as Map<String, dynamic>);
   double? effectiveGainChange;
   switch (FinampSettingsHelper.finampSettings.volumeNormalizationMode) {
-    case VolumeNormalizationMode.hybrid:
-      // case VolumeNormalizationMode.albumBased: // we use the context normalization gain for album-based because we don't have the album item here
-      // use context normalization gain if available, otherwise use track normalization gain
-      effectiveGainChange = currentTrack.extras?["contextNormalizationGain"] as double? ?? baseItem.normalizationGain;
+    case VolumeNormalizationMode.hybrid when isSequental:
+    case VolumeNormalizationMode.albumBased:
+      effectiveGainChange = parentNormalizationGainProvider() ?? baseItem.normalizationGain;
```
Komodo5197: We should already have the album normalization for some items, so you should be able to do parentNormalizationGainProvider() ?? (currentTrack.extras?["contextNormalizationGain"] as double?) ??baseItem.normalizationGain

lib/services/music_player_background_task.dart
```diff
@@ -940,7 +951,12 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
     if (FinampSettingsHelper.finampSettings.volumeNormalizationActive && currentTrack != null) {
       final baseItem = jellyfin_models.BaseItemDto.fromJson(currentTrack.extras?["itemJson"] as Map<String, dynamic>);
 
-      double? effectiveGainChange = getEffectiveGainChange(currentTrack, baseItem);
+      double? effectiveGainChange = getEffectiveGainChange(
+        currentTrack,
+        baseItem,
+        _queueService.getQueueStream().valueOrNull?.isSequental() ?? false,
+        () => _providerContainer.read(currentTrackMetadataProvider).valueOrNull?.parentNormalizationGain,
```
Komodo5197: Maybe we should verify that the iotemin the currentTrackMetadataProvider matches the one in currentTrack before using, to guard against possible ordering issues.  Also, why are we passing a function instead of just getting the value?

lib/services/music_player_background_task.dart
```diff
@@ -1082,20 +1098,51 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
...
+
+    return previousTrackAlbum == currentTrackAlbum || currentTrackAlbum == nextTrackAlbum;
+  }
 }
 
-double? getEffectiveGainChange(MediaItem currentTrack, jellyfin_models.BaseItemDto? item) {
+double? getEffectiveGainChange(
+  MediaItem currentTrack,
+  jellyfin_models.BaseItemDto? item,
+  bool isSequental,
```
Komodo5197: I don't see much reason to pass isSequental or parentNormalizationGainProvider as function arguments instead of calculating them internally.

lib/services/music_player_background_task.dart
HeroBrine1st: > Maybe we should verify that the iotemin the currentTrackMetadataProvider matches the one in currentTrack

Yes, I missed that. But what should it do when it doesn't match? Falling back to track gain will have the same effect of applying wrong volume, and my experience only suggests placing a logger or an assert here.

> Also, why are we passing a function instead of just getting the value?

It is not always needed, so for performance at least when it isn't needed. The normalization gain is delayed to the start of next track so I decided not to make it worse. But now I think everyone will still use hybrid mode so this will lead to nothing.

UPD: I have a patch that replaces function with plain value on my PC, however I'm not sure my initial concern (don't read what you don't use) is not valid anymore.

lib/services/music_player_background_task.dart
HeroBrine1st: In earlier versions of this function it used GetIt internally, which, I think, uses reflection, which is bad for performance, so I moved GetIt usage to constructor (and bindQueueService method due to circular dependency). Anyway this allows reusing instances already available somewhere, which should on itself give a slight performance boost.

lib/services/music_player_background_task.dart
HeroBrine1st: Actually contextNormalizationGain is not for the album played only. If you place next album or some unrelated tracks, it will use normalization gain from first album for them too.

This, I think, is internal reasoning. Other than that, I agree with you, and the behavior I described should be fixed in separate PR.

HeroBrine1st:
I marked as resolved the things I fixed. 

Also `dart format .` is mad at metadata_provider.dart and wants to reformat the whole file changing indent for some reason, but it doesn't do that on redesign branch of this repository. I don't think I made so much changes, so I constantly revert indent changes (or I will be the one to git blame).

lib/services/music_player_background_task.dart
Komodo5197: Are you sure?  I did find a bug where loadSavedQueue will use the initial source with normalization for everyone, but everything else looks good reading through the code.  Where you hitting that?  I think you could fix it by just setting the source contextnormalization to null before calling _replaceWholeQueue inside loadSavedQueue.

lib/services/music_player_background_task.dart
HeroBrine1st: After re-checking, I have found no way to decrease number of arguments without using GetIt. queueService (which is used for isSequental) should be then passed via parameter, and metadata cannot be passed without parameter due to it being read differently in UI and music player task.

Making a mere calculator function dependent on the whole QueueService is meh, and passing `FinampQueueInfo` is strange to my opinion. But passing either of them will remove the extension. Idk what is best.

UPD: Sorry I misunderstood you ha-ha. You meant encapsulation, I will do it in a minute.

lib/services/music_player_background_task.dart
Komodo5197: As far as I'm aware, the performance impact of GetIt or reading a provider is minimal, and there will not be a noticeable impact of reading them both here instead of passing down a function and pre-registering the _queueService variable.  I think both of these shouldn't be doing much more than looking up values in a map with a couple wrapper functions.  I've generally found the biggest performance/latency impacts to all come from usage of async.

Regarding the incorrect gain, I don't have a great solution.  I feel that using the track gain from the current track is probably better than using the album gain from a potentially unrelated album?  In practice, this should hopefully not come up due to the metadataProvider precaching.  If volume normalization regularly fires before the currentTrackMetadataProvider has switched over, we may need to start reading the relevant metadataProvider directly.

HeroBrine1st:
I'm sorry for hitting the "request review" button too early, probably should have waited at least one hour ha-ha.

I believe you can see the aggregated changes since your review [here](https://github.com/jmshrv/finamp/compare/6ce38c87488599348e4bb59215615946cf3c04e1..98f584cba53b8dff8b8deef9e836f8b125211018)

lib/services/music_player_background_task.dart
Komodo5197: The effective gain function is already essentially dependent on the queue service, because isSequential is always calculated based off the current track.  Pushing the getQueue calls out to the caller doesn't really help much there.  I'd prefer to rely on the queue and provider states directly in the function and just make this fact more explicit by renaming it or something.

lib/services/metadata_provider.dart
```diff
@@ -170,36 +173,30 @@ metadataProvider = FutureProvider.autoDispose.family<MetadataProvider?, Metadata
...
     }
   }
 
+  BaseItemDto? parent;
+  if (parentFuture != null) {
+    parent = await parentFuture;
+  }
+
+  // check if item qualifies for having playback speed control available
+  if (request.checkIfSpeedControlNeeded &&
```
Komodo5197: This flag was primarily to prevent fetching the parent unnecessarily.  Now that we always fetch the parent, we might as well remove field from metadatarequest and always run the full logic.

lib/services/music_player_background_task.dart
```diff
@@ -1082,20 +1099,54 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
...
-double? getEffectiveGainChange(MediaItem currentTrack, jellyfin_models.BaseItemDto? item) {
+double? getEffectiveGainChange(
+  MediaItem currentTrack,
+  jellyfin_models.BaseItemDto? item,
+  FinampQueueInfo queueInfo,
+  MetadataProvider? Function() metadataProvider,
+) {
   final baseItem =
       item ?? jellyfin_models.BaseItemDto.fromJson(currentTrack.extras?["itemJson"] as Map<String, dynamic>);
+  bool isSequental;
```
Komodo5197: Actually, I was fine with having isSequential as an extension method, if you prefer that.

lib/services/music_player_background_task.dart
Komodo5197: I don't think dart's true reflection capabilities are actually available in flutter applications.  You can access the Type of an object or parameterized function, which should be basically free, but you can't do anything but check equality with it.  So GetIt is basically just maintaining a Map<Type, dynamic> you can insert into or read.

lib/services/music_player_background_task.dart
HeroBrine1st: > we may need to start reading the relevant metadataProvider directly

Actually that's not impossible. Removing queueItem is trivial (it is unused) along with checkIfSpeedControlNeeded, leaving only `BaseItemDto item` and `boolean includeLyrics` as values of request. Then the problem is guessing the `includeLyrics` value, as it (supposedly) created not to prefetch lyrics for upcoming tracks. However, it is always true! So it is trivial to always fetch the required track, and if it is not available yet, the listener on `currentTrackMetadataProvider` will catch up.

I also think it calls to split the metadataProvider into metadataProvider and lyricsProvider.

HeroBrine1st:
Now getEffectiveGainChange matches metadata provider with provided track. It looks like `BaseItemDto` is a data class (with overriden `==` and `hashCode` because otherwise normalization would fall back to trackGain and I don't see that on the aforementioned "Myth" album (in playlist context of course).

Currently not intend to merge this, so I didn't rebase it.

lib/services/music_player_background_task.dart
```diff
@@ -1082,20 +1093,59 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
...
+    final currentTrackAlbum = currentTrack?.baseItem?.parentId;
+    if (currentTrackAlbum == null) return false;
+    final previousTrackAlbum = previousTracks.lastOrNull?.baseItem?.parentId;
+    final nextTrackAlbum = (nextUp.firstOrNull ?? queue.firstOrNull)?.baseItem?.parentId;
+
+    return previousTrackAlbum == currentTrackAlbum || currentTrackAlbum == nextTrackAlbum;
+  }
 }
 
 double? getEffectiveGainChange(MediaItem currentTrack, jellyfin_models.BaseItemDto? item) {
```
HeroBrine1st: What about `getGainForCurrentPlayback`? This name implies that this function depends on something other than `currentTrack` and `item` and so `currentTrack` must match the currently played item.

lib/services/music_player_background_task.dart
HeroBrine1st: > Are you sure?

I rechecked and no, I am not :P. It is apparently fixed. Pretty strange, I faced it not too long ago for an app to update..

HeroBrine1st:
This should be fully ready to merge. Remaining concerns are:

- At 1a2bb60a, it is tempting to move `await parentFuture` somewhere so that code for `qualifiesForPlaybackSpeedControl` is not split, but then the commit history would not be all clean (e.g. it is <sup>probably</sup> independent with 6fb225e4 due to 1a2bb60a being clean change)
- It is strange that metadataProvider fetches lyrics for all tracks, not only current one, bringing up the question about `bool includeLyrics` necessity
- And something else, I forgot

lib/services/music_player_background_task.dart
```diff
@@ -940,7 +946,7 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
     if (FinampSettingsHelper.finampSettings.volumeNormalizationActive && currentTrack != null) {
       final baseItem = jellyfin_models.BaseItemDto.fromJson(currentTrack.extras?["itemJson"] as Map<String, dynamic>);
 
-      double? effectiveGainChange = getEffectiveGainChange(currentTrack, baseItem);
+      double? effectiveGainChange = getGainForCurrentPlayback(currentTrack, baseItem);
 
       _volumeNormalizationLogger.info(
```
Komodo5197, line 957: _applyVolumeNormalization could now be called up to three times per track change, so I'd like to reduce the logging a bit.  Logging only on change would be nice, but may be complex as we don't store the raw effectiveGainChange anywhere.  At the very least, we can remove _volumeNormalizationLogger.info("Gain change: $effectiveGainChange"); as it is completely redundant with the previous log.

lib/services/music_player_background_task.dart
```diff
@@ -297,16 +300,19 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
       _volumeNormalizationLogger.info("non-Android base volume gain factor: $iosBaseVolumeGainFactor");
     }
 
+    _providerContainer = GetIt.instance<ProviderContainer>();
```
Komodo5197: I think this variable is unused?

HeroBrine1st: IIRC It is used in onQueueServiceAvailable, to listen on currenrTrackMetatataProvider

Komodo5197: Ah, yeah.  Might make sense to inline it if it's only used once, but that's up to you.

Chaphasilor:
@Komodo5197 thanks for reviewing and @HeroBrine1st thanks for the great implementation and discussion!  
I'll test this out properly tomorrow, and if it works I see no blocking issue with this either...  

I think I originally intended the `includeLyrics` parameter to avoid fetching lyrics when using the metadata provider for menus or when no lyrics are needed, but that doesn't really make sense anymore. And if we end up needing conditional logic for that we could always use a checking function or a provider instead of passing this as an argument.

Now that you' dealt with the code a bit more closely, do you have any new ideas how we could mitigate #769?  
I'm guessing the code is too unrelated to actual playback though :)

lib/services/music_player_background_task.dart
```diff
@@ -1082,20 +1088,59 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
...
+extension SequentalTracks on FinampQueueInfo {
+  bool isSequental() {
+    final currentTrackAlbum = currentTrack?.baseItem?.parentId;
+    if (currentTrackAlbum == null) return false;
+    final previousTrackAlbum = previousTracks.lastOrNull?.baseItem?.parentId;
+    final nextTrackAlbum = (nextUp.firstOrNull ?? queue.firstOrNull)?.baseItem?.parentId;
+
+    return previousTrackAlbum == currentTrackAlbum || currentTrackAlbum == nextTrackAlbum;
+  }
+}
```
Chaphasilor: why is this here instead of in `queue_service`? 🤔

lib/services/metadata_provider.dart
```diff
@@ -49,182 +22,200 @@ class MetadataProvider {
...
+        }
+      } else {
+        try {
+          return await jellyfinApiHelper.getItemById(parentId);
+        } catch (e) {
+          metadataProviderLogger.warning("Failed to get parent item '$parentId'", e);
+        }
+      }
+      return null;
+    });
```
Chaphasilor, line 221: I think it might make sense to eventually group such providers together into one file/directory. Having them embedded in a "random" file is a bit strange, and I don't think this provider is actually specific to this screen (even if it's currently only used here).  

We could probably reduce the code/logic by quite a lot if we used providers for fetching all base items, and handling online/offline mode, even if the providers are non-caching.  
That's out-of-scope for this specific PR though :)

lib/services/music_player_background_task.dart
HeroBrine1st: The same as albumProvider (which should be `itemProvider`) - tunnel vision :-)

I actually want it to be private. The only reason I returned this extension back (I inlined it once) is to neatly use it after `when` and so avoid extra computations when not needed. Do you spot other use cases for this function?

Chaphasilor: not currently, but it's always nice having some helper function available when it's needed. I'll leave it up to you.  

either way, there's a typo, it should be `isSequential` and `SequentialTracks` with an `i`, and I'd suggest naming this more clearly, like `isSequentialTrackFromSameAlbum` or something similar.

lib/l10n/app_en.arb
```diff
@@ -1474,7 +1474,7 @@
...
     "description": "Subtitle for the dropdown that selects the replay gain mode"
   },
-  "volumeNormalizationModeSelectorDescription": "Hybrid (Track + Album):\nTrack gain is used for regular playback, but if an album is playing (either because it''s the main playback queue source, or because it was added to the queue at some point), the album gain is used instead.\n\nTrack-based:\nTrack gain is always used, regardless of whether an album is playing or not.\n\nAlbums Only:\nVolume Normalization is only applied while playing albums (using the album gain), but not for individual tracks.",
+  "volumeNormalizationModeSelectorDescription": "Hybrid (Track + Album):\nWorks like track-based, but when playing multiple tracks from the same album in a row, works like album-based for them instead.\n\nTrack-based:\nTrack gain is always used, regardless of whether an album is playing or not.\n\nAlbums Only:\nVolume Normalization is only applied while playing albums (using the album gain), but not for individual tracks.\n\nAlbum-based:\nAlbum gain is always used, but if it is not available (e.g. Jellyfin does not provide it for singles), track gain is used.",
```
Chaphasilor: This is incorrect. "Hybrid" will not always use the tracks gain. When playing an album, the album gain will be used, that's why it's called hybrid. For albums album gain, for anything else track gain.  
And now with your improvement it's album gain for albums *and* successive tracks from the same album, and track gain for anything else.

And let's rename this to `Dynamic` instead of `Hybrid (Track + Album)`

Edit: I see that "multiple tracks from an album in a row" does technically include simply playing an album. But I'd like to emphasize all three scenarios, playing random tracks, playing an album, and playing any queue with consecutive tracks from the same album.

Chaphasilor (changes requested):
just some use-facing changes.  
functionality works great, gain is correctly applied when playing albums or consecutive album tracks in a playlist, but not when e.g. shuffling that playlist or playing the tracks individually.  
It even works when playing one track independently, and then adding another track from the same album to the queue. This will actually re-apply the new gain, which I'm not super sure is the right thing to do, but is probably okay :)

lib/services/metadata_provider.dart
```diff
@@ -49,182 +22,200 @@ class MetadataProvider {
   LyricDto? lyrics;
   bool isDownloaded;
   bool qualifiesForPlaybackSpeedControl;
+  double? parentNormalizationGain;
```
HeroBrine1st, line 25: This can be "BaseItemDto? parentItem` so that this change is generic to support other cases. Or is it better to leave as is?

Chaphasilor: not really. a parent item can also be a playlist, if playback was started via that playlist...

Chaphasilor:
Yeah that seems fine. I'll change the description then.  
I don't really think it necessary for you to rebase things, but feel free...

Chaphasilor:
From my side this is ready to merge. @HeroBrine1st ?

Chaphasilor:
Hmm, if you think that this really is because of the `mediaItem` listener firing when another track (not the current track) is changed, that would then also cause a bug with the "tracks" option of the sleep timer, so we should look into that in another PR.

But I'll merge this for now, thanks a ton for all your work!

Issue #1189, opened 2025-04-24
[FR] Add option to automatically expand player screen when clicking on a song


Ge082:
Actually, clicking on a music to start it just bring the small rectangular thing. Which is good, but some prefer to brîg the full player control when starting any music like all streaming app does

There should be a togglable option where, if activated, when you click on a song, it brings you the full player screen

### Type:
- [x] Improve or extend an existing feature

PR #1304, opened 2025-07-27
#1189 Implementation for toggle player screen full

Y0ngg4n:
Implements: https://github.com/jmshrv/finamp/issues/1189

## Changes

This adds a toggle under interactions and opens the player after loading the song directly when clicking on it in track list view.

## Todo before merging

- [ x] Translations

- [ x] Reset Settings


Komodo5197:
You should add a check to avoid doing this if split screen mode is currently active, because it will just close immediately.  Also, I think the most appropriate place to put this is probably _replaceWholeQueue?

Chaphasilor:
I agree that right now this is the best place to put it, and the easiest change.  
I'm not super happy with performing app navigation from within the queue service though (hence the listener idea), but that can be changed later on :)

lib/services/queue_service.dart
```diff
@@ -486,6 +488,10 @@ class QueueService {
       if (initialIndex >= itemList.length) {
         return Future.error("initialIndex is bigger than the itemList! ($initialIndex > ${itemList.length})");
       }
+
+      if (FinampSettingsHelper.finampSettings.autoExpandPlayerScreen) {
+        await NowPlayingBar.openPlayerScreen(GlobalSnackbar.materialAppNavigatorKey.currentContext!);
```
Komodo5197: I think this should be unawaited.

lib/screens/add_to_playlist_screen.dart
```diff
@@ -0,0 +1 @@
+
```
Komodo5197: You seem to have accidentally committed a deleted file.

Y0ngg4n: Oh i am very sorry

Chaphasilor:
Thanks for the changes. I've slightly changed the wording and added a subtitle.  
I also delayed the opening of the player screen slightly by moving it to a later position, that seemed a bit nicer for when selecting a new track, without showing the old track first or being too slow. And I made sure the screen only opens when playback actually starts right away, not when restoring the previous queue after startup :)

Y0ngg4n:
@Chaphasilor thank you so much. Sorry that i did not react faster.

PR #1305, opened 2025-07-28
Changed Offline Mode Icons

flloschy:
Changes: https://github.com/jmshrv/finamp/pull/1305#issuecomment-3139050144

### DIFF
--- a/lib/components/MusicScreen/offline_mode_switch_list_tile.dart
+++ b/lib/components/MusicScreen/offline_mode_switch_list_tile.dart
@@ -13,17 +13,17 @@ class OfflineModeSwitchListTile extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     AutoOfflineOption automation = ref.watch(finampSettingsProvider.autoOffline);
     bool overrideActive = !ref.watch(finampSettingsProvider.autoOfflineListenerActive);
-    int activeDelays = ref.watch(autoOfflineStatusProvider);
+    bool reevaluating = ref.watch(autoOfflineStatusProvider) > 0;
 
     IconData getCurrentIcon() {
-      if (overrideActive) {
-        return TablerIcons.cloud_pause;
-      } else if (automation == AutoOfflineOption.disabled) {
+      if (automation == AutoOfflineOption.disabled) {
         return TablerIcons.cloud_off;
-      } else if (activeDelays > 0) {
-        return Icons.cloud_sync_outlined;
+      } else if (overrideActive) {
+        return TablerIcons.robot_off;
+      } else if (reevaluating) {
+        return Icons.sync_sharp;
       } else {
-        return TablerIcons.cloud_search;
+        return TablerIcons.robot;
       }
     }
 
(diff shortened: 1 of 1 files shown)


Chaphasilor:
The `else` clause refers to active automation, right? I'm not sure if a magnifying glass makes a lot of sense for that case.  
I'm fine with using the sync icon while we're trying to figure things out, but I think simply swapping isn't a good idea.  
Can you try to find a better-fitting icon at <https://tabler.com/icons> please?

flloschy:
image

As a matter of fact, I cant find a better icon at tabler :p
But I *can* brute force search a fitting icon

flloschy:
image

I guess simply just `TablerIcons.cloud` is fine?

flloschy:
I think 4 is actually a sweet spot
What about

Automation: 
image

Automation Paused: 
image

Reevaluating: 
image

Automation Disabled:
image

Komodo5197:
I prefer the current cloud_off for when automation is disabled, myself.

flloschy:
Sure
When i did it, should i push or shall we not risk another RPC commit situation?

Komodo5197:
Using separate icons for online/offline when automation is disabled seems pointless to me, as the switch state already tells you that.  Note that none of the icons when automation is enabled in the latest proposal tell you about the offline state, just the automation state.  cloud_off is just supposed to illustrate the concept of offline mode, not indicate anything.  This doesn't 100% align with the icon's use as an indicator when automation is enabled, but you need to flip a setting to change between these, so it seems fine to me.

flloschy:
Im confused and not quite sure what you mean? What is fine and what is pointless?

Komodo5197:
Having the icon change in response to offline/online mode while automation is disabled seems pointless to me, because that info is already represented by the switch itself.  The thing that seems fine to me is the minor discrepancy between when automation is enabled and the icon tells you the automation status, and when automation is disabled and the icon is just illustrating the text 'offline mode'.

Chaphasilor:
@Komodo5197 I agree with you here. The reason I originally suggested it was simply because there was no good icon for indicating "you've disabled the automation and are in manual mode", that's why I opted to just show if it's online or offline.  
But you point it valid, we can just go back to having a regular icon that describes what the setting is about, and not its current state (as it was before the automation was introduced). So let's do that.

@flloschy that means we'll go with the following:

Automation: 
image

Automation Paused: 
image

Reevaluating: 
image

Automation Disabled Completely:
`cloud-off`

lib/components/MusicScreen/offline_mode_switch_list_tile.dart
```diff
@@ -13,17 +13,19 @@ class OfflineModeSwitchListTile extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     AutoOfflineOption automation = ref.watch(finampSettingsProvider.autoOffline);
     bool overrideActive = !ref.watch(finampSettingsProvider.autoOfflineListenerActive);
+
+    /// delays>0 auto offline is reevaluating
```
Chaphasilor: this could've been a bool variable ^^

Chaphasilor:
awesome, thanks!

PR #1315, opened 2025-08-03
Redesign

HeroBrine1st:
## Changes

Fixes https://github.com/issues/created?issue=jmshrv%7Cfinamp%7C1308
Fixes sleep timer reacting to queue change as track change (can't find issue)
Adds rustup to flake.nix so that developers on NixOS are happy (It is if I am of the `flake.nix` and `flake.lock` files ha-ha)
Reformats file from my previous PR

## Todo before merging

None, ready to merge

## Related Issues

None

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -65,6 +65,7 @@
               jdk17
               androidComposition.platform-tools
               cmake
+              rustup
             ];
           };
       }
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -353,9 +353,12 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
     // audible volume change).
     // This is possible because this callback is called on each queue change
     mediaItem.listen((currentTrack) {
-      sleepTimer?.onTrackCompleted();
-
       _applyVolumeNormalization(currentTrack);
+    });
+
+    // But sleepTimer doesn't want to listen on queue changes
+    mediaItem.distinct().listen((currentTrack) {
+      sleepTimer?.onTrackCompleted();
     });
 
     // trigger sleep timer early if we're almost at the end of the final track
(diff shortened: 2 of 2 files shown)


HeroBrine1st:
Not only I forgot to type "the maintainer" (see the first revision of first message, that was funny), but forgot to set the title so the default one meant I redesigned the whole app xD

HeroBrine1st:
> Nice, that `distinct()` looks perfect.  

It is a normal thing to do in idiomatic Kotlin so it transferred here :-) I think we can map it to itemId before distinct so that less fields are compared, also we can map settings to a tuple of three (baseGain, normGainActive and normGainMode) and use distinct there too.

> Any idea how we could get events only when the queue advances forwards? Probably by listening to changes in the current index, and checking if it's positive, since we can't prepend tracks before the current track?  
> The only problem I see with that is that enabling shuffle mode will move the current track to index 0, and all other tracks after it. So the index potentially decreases, without the actual track changing...

Shuffle is not a problem if we require change of both index and itemId (so if only one is changed just swallow the event and continue). There is negligible probability that user will change to shuffle at the same time as track ends but I estimate that probability as 0.

Another approach is to listen on itemId (saving previous value) and on change look to queue and see if previous itemId is in nextUp/queue or previousTracks. This will mitigate the shuffle problem leaving another probability: user clicks on shuffle at the same time track is changed, then new track is played but if the previous track is now next (due to shuffle) it will trigger as skipping to previous track. So it is probability of user timing a click to track change times probability of precious track being the next in queue (it is not 100% because that's the case of two tracks in queue, meaning shuffling it is.. strange). 

To make user mad, it will have to repeat, and repeat reliably, so it is a power of that probability and that approaches to zero very quickly even with first approach.

The most reliable approach, however, is to directly listen on just_audio player. Idk if there is an event for next/previous track but for me it is reasonable to have it, so I expect it to be there.

Chaphasilor:
Okay, so maybe it's easiest to compare the indices, and the MediaItem.id, and if the conditions match invoke a callback function, like `onNextTrack`?  
Alternatively we could create a Stream that pushes out a new even each time and can be listened to from multiple files, instead of having a local-only callback.

Chaphasilor:
@Komodo5197 or @jmshrv feel free to merge if you approve :)

Komodo5197:
Looks good to me.

PR #1316, opened 2025-08-04
[WIP]: Change all ListViews to SuperListViews

Y0ngg4n:
## Changes

This package provides a drop in replacement for ListViews and SliverLists with faster performance:
https://pub.dev/packages/super_sliver_list


Chaphasilor:
Just read through the readme for those packages again. While it looks great from a functionality standpoint, it also seems to be essentially abandoned for a few months now, and there are a few open issues and PRs that haven't been addressed yet. So I'm not sure if it really is a great idea to commit ourselves to this library...

Would you mind looking for instances where we scroll or animate to a specific index and see if we could use the new list capabilities instead? This might not work for the music screen tab view because it requires "checking" if the needed item is there, but it *could* work. And in the queue list where we know all items it should definitely work.

I also think there are some instances of `FixedExtentSliverList` being used, maybe we could get rid of those as well?

Y0ngg4n:
@Chaphasilor hm yeah valid points. I will check if we can squeeze more Performance out. Maybe i find something.

Y0ngg4n:
@Chaphasilor the docs for SliverFixedExtentList says its more efficient:
https://api.flutter.dev/flutter/widgets/SliverFixedExtentList-class.html
`[SliverFixedExtentList](https://api.flutter.dev/flutter/widgets/SliverFixedExtentList-class.html) is more efficient than [SliverList](https://api.flutter.dev/flutter/widgets/SliverList-class.html) because [SliverFixedExtentList](https://api.flutter.dev/flutter/widgets/SliverFixedExtentList-class.html) does not need to perform layout on its children to obtain their extent in the main axis.`

PR #1317, opened 2025-08-04
Add missing ValueKeys to some ListViews

Y0ngg4n:
## Changes

This just adds some ValueKeys, to the ListViews. Maybe this helps with performance. I have checked most of our ListViews and compared them with suggestions for more performance and in the most parts we are already using the best option.

### DIFF
--- a/lib/components/AlbumScreen/album_screen_content.dart
+++ b/lib/components/AlbumScreen/album_screen_content.dart
@@ -262,7 +262,7 @@ class _TracksSliverListState extends ConsumerState<TracksSliverList> {
         }
 
         return TrackListTile(
-          key: ValueKey(item),
+          key: ValueKey(item.id),
           item: item,
           children: widget.childrenForQueue,
           index: indexOffset,
--- a/lib/components/AlbumScreen/album_screen_content.dart
+++ b/lib/components/AlbumScreen/album_screen_content.dart
@@ -262,6 +262,7 @@ class _TracksSliverListState extends ConsumerState<TracksSliverList> {
         }
 
         return TrackListTile(
+          key: ValueKey(item),
           item: item,
           children: widget.childrenForQueue,
           index: indexOffset,
--- a/lib/components/LogsScreen/logs_view.dart
+++ b/lib/components/LogsScreen/logs_view.dart
@@ -16,7 +16,7 @@ class LogsView extends StatelessWidget {
       reverse: true,
       itemBuilder: (context, index) {
         return LogTile(
-          key: ValueKey(finampLogsHelper.logs.elementAt(index)),
+          key: ValueKey(finampLogsHelper.logs.reversed.elementAt(index).sequenceNumber),
           logRecord: finampLogsHelper.logs.reversed.elementAt(index),
         );
       },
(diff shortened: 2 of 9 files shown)


lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -262,6 +262,7 @@ class _TracksSliverListState extends ConsumerState<TracksSliverList> {
         }
 
         return TrackListTile(
+          key: ValueKey(item),
```
Chaphasilor: maybe we should use `item.id` here?

lib/components/LogsScreen/logs_view.dart
```diff
@@ -15,7 +15,10 @@ class LogsView extends StatelessWidget {
       itemCount: finampLogsHelper.logs.length,
       reverse: true,
       itemBuilder: (context, index) {
-        return LogTile(logRecord: finampLogsHelper.logs.reversed.elementAt(index));
+        return LogTile(
+          key: ValueKey(finampLogsHelper.logs.elementAt(index)),
```
Chaphasilor: I'm not sure if using the entire content as a key is such a good idea 😅

lib/components/PlaybackHistoryScreen/playback_history_list.dart
```diff
@@ -31,6 +31,7 @@ class PlaybackHistoryList extends StatelessWidget {
                   final actualIndex = group.value.length - index - 1;
 
                   final historyItem = TrackListTile(
+                    key: ValueKey(group.value[actualIndex].item),
```
Chaphasilor: use `item.id` here

lib/screens/lyrics_screen.dart
```diff
@@ -326,6 +326,7 @@ class _LyricsViewState extends ConsumerState<LyricsView> with WidgetsBindingObse
                           (nextLine == null || currentMicros < nextLine.startMicros);
 
                       return Column(
+                        key: ValueKey(line),
```
Chaphasilor: we could use `line.startMicros` here? the lyric text itself can repeat, but the timestamps should be unique?

lib/screens/queue_restore_screen.dart
```diff
@@ -24,7 +24,7 @@ class QueueRestoreScreen extends StatelessWidget {
         padding: const EdgeInsets.only(left: 0.0, right: 0.0, top: 30.0, bottom: 45.0),
         itemCount: queueList.length,
         itemBuilder: (context, index) {
-          return QueueRestoreTile(info: queueList.elementAt(index));
+          return QueueRestoreTile(key: ValueKey(queueList.elementAt(index)), info: queueList.elementAt(index));
```
Chaphasilor: `queueList.elementAt(index).creation` should work

lib/screens/view_selector.dart
```diff
@@ -93,6 +93,7 @@ class _ViewSelectorState extends State<ViewSelector> {
                 final view = _views.keys.elementAt(index);
 
                 return CheckboxListTile(
+                  key: ValueKey(view),
```
Chaphasilor: `view.id`

Issue #586, opened 2024-01-28
[Feature Request] Add queue to Playlist


kuflierl:
I think a "add queue to playlist" button would be quite helpful. It would be especially helpful when I want to save a mix that I like to a playlist.

flloschy:
This can be tagged as upcoming @Chaphasilor 

<img width="280" height="49" alt="Image" src="https://github.com/user-attachments/assets/9b029e2f-a730-4513-998d-620f2316b2ad" />

flloschy:
There you go

PR #1320, opened 2025-08-05
Add Queue to Playlist instead of needing a new playlist

flloschy:
## Changes
Instead of just creating a new playlist and adding all the items in the queue, it is now possible to add to an existing playlist

### Note
Since the whole adding to playlist system is based around adding one item at a time, I needed to change quite a bit and this may not be the optimal solution. My testing concludes that everything still works as expected at least, though some extra testing cant hurt :)

## Todo before merging
- [ ] Translations

## Related Issues
closes https://github.com/jmshrv/finamp/issues/586

### DIFF
--- a/lib/components/AddToPlaylistScreen/add_to_playlist_list.dart
+++ b/lib/components/AddToPlaylistScreen/add_to_playlist_list.dart
@@ -21,9 +21,9 @@ import 'new_playlist_dialog.dart';
 import '../../menus/playlist_actions_menu.dart';
 
 class AddToPlaylistList extends StatefulWidget {
-  const AddToPlaylistList({super.key, required this.itemToAdd, required this.playlistsFuture});
+  const AddToPlaylistList({super.key, required this.itemsToAdd, required this.playlistsFuture});
 
-  final BaseItemDto itemToAdd;
+  final List<BaseItemDto> itemsToAdd;
   final Future<List<BaseItemDto>> playlistsFuture;
 
   @override
@@ -54,7 +54,7 @@ class _AddToPlaylistListState extends State<AddToPlaylistList> {
               final (playlist, isLoading, playListItemId) = snapshot.data![index];
               return AddToPlaylistTile(
                 playlist: playlist,
-                track: widget.itemToAdd,
+                tracks: widget.itemsToAdd,
                 playlistItemId: playListItemId,
                 isLoading: isLoading,
               );
@@ -91,7 +91,7 @@ class _AddToPlaylistListState extends State<AddToPlaylistList> {
             onPressed: () async {
               var dialogResult = await showDialog<(Future<BaseItemId>, String?)?>(
                 context: context,
-                builder: (context) => NewPlaylistDialog(itemsToAdd: [widget.itemToAdd.id]),
+                builder: (context) => NewPlaylistDialog(itemsToAdd: widget.itemsToAdd.map((item) {return item.id;}).toList()),
               );
               if (dialogResult != null) {
                 var oldFuture = playlistsFuture;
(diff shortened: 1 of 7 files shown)


lib/components/AddToPlaylistScreen/add_to_playlist_list.dart
```diff
@@ -101,29 +105,31 @@ class _AddToPlaylistListState extends State<AddToPlaylistList> {
                   ];
                   playlistsFuture = oldFuture.then((value) => value + loadingItem);
                 });
-                try {
-                  var newId = await dialogResult.$1;
-                  // Give the server time to calculate an initial playlist image
-                  await Future.delayed(const Duration(seconds: 1));
-                  final jellyfinApiHelper = GetIt.instance<JellyfinApiHelper>();
-                  var playlist = await jellyfinApiHelper.getItemById(newId);
+                for (var track in widget.itemsToAdd) {
```
Komodo5197: What is the purpose of this whole loop?  Revert these changes, and add a check so that the non-track path also fires if we have more than one item.

lib/components/AddToPlaylistScreen/add_to_playlist_list.dart
```diff
@@ -177,7 +183,9 @@ class _AddToPlaylistTileState extends ConsumerState<AddToPlaylistTile> {
         itemIsIncluded = true;
       } else {
         final downloadsService = GetIt.instance<DownloadsService>();
-        itemIsIncluded = downloadsService.checkIfInCollection(widget.playlist, widget.track);
+        for (var track in widget.tracks) {
```
Komodo5197: This is just overwriting the same bool a bunch.  Just set it to false if we make it here and there are more than one item.

lib/components/AddToPlaylistScreen/add_to_playlist_list.dart
```diff
@@ -196,18 +204,18 @@ class _AddToPlaylistTileState extends ConsumerState<AddToPlaylistTile> {
           ? TablerIcons.circle_dashed_plus
           : TablerIcons.circle_plus,
       initialState: itemIsIncluded ?? false,
-      onToggle: (bool currentState) async {
-        if (currentState) {
+      onToggle: (bool alreadyInPlaylist) async {
+        if (alreadyInPlaylist) {
           // Only tracks can be removed from playlists
-          if (BaseItemDtoType.fromItem(widget.track) != BaseItemDtoType.track) {
+          if (BaseItemDtoType.fromItem(widget.tracks.first) != BaseItemDtoType.track) {
```
Komodo5197: Always return true here if we have more than one track, to match album behavior.

lib/components/AddToPlaylistScreen/add_to_playlist_list.dart
```diff
@@ -236,42 +244,70 @@ class _AddToPlaylistTileState extends ConsumerState<AddToPlaylistTile> {
           }
           return !removed;
         } else {
-          // add to playlist
-          if (BaseItemDtoType.fromItem(widget.track) != BaseItemDtoType.track) {
+          if (widget.tracks.length == 1) {
```
Komodo5197: Why have you duplicated so much code?  It looks like the only divergence from the album path is the prompt text.  Just send us down that and change the prompt calculation logic.

lib/components/PlayerScreen/queue_source_helper.dart
```diff
@@ -140,6 +145,10 @@ Future<bool> addItemToPlaylist(BuildContext context, BaseItemDto item, BaseItemD
   return true;
 }
 
+Future<bool> addItemToPlaylist(BuildContext context, BaseItemDto item, BaseItemDto parent) async {
```
Komodo5197: This helper method seems pointless.  We only call it in a method that's already handling a multi-item playlist addition.  Remove it.

Chaphasilor (approved):
I've addressed the review.  
For reference, while we *could* theoretically keep track of which tracks were added, and then remove them later on, this will run into issues if some tracks already are in the playlist before because a) we can't properly show that some tracks are already in there (we'd need a new icon, which I think is a bit overkill), and b) Jellyfin doesn't tell us which items were added, and since Jellyfin will never add duplicate items that would mean if we remove the tracks, the playlist wouldn't be in the original state but would have all matching tracks that we wanted to add removed (that's only a semantic difference of course, but is probably still unexpected). 

I also noticed the `addItemsToPlaylists()` helper is in `queue_source_helper.dart`, which doesn't make add that much sense to me. Would it make sense to move this into `jellyfin_api_helper`? I see no reason why we'd ever not want to update the downloaded playlists, if we actively modify a playlist...

flloschy:
Oh wow, thanks <3
Was gonna tackle those today xd

lib/components/AddToPlaylistScreen/add_to_playlist_list.dart
```diff
@@ -244,70 +248,46 @@ class _AddToPlaylistTileState extends ConsumerState<AddToPlaylistTile> {
...
+              BaseItemDtoType.genre => "genre",
+              BaseItemDtoType.playlist => "playlist",
+              BaseItemDtoType.track => "tracks",
+              _ => "unknown",
+            };
             await showDialog(
               context: context,
               builder: (context) => ConfirmationPromptDialog(
-                promptText: "Add X items to playlist?",
+                promptText: AppLocalizations.of(
```
Komodo5197: It's probably worth breaking this up into a tracks localization and an album/artist/ect. localization.

Chaphasilor: I wanted to avoid breaking existing translations, but sure, let's do it. Then I'll also rename the current translation key, since we're not just dealing with albums anymore

Chaphasilor:
Thanks for the PR! 😁

PR #1322, opened 2025-08-06
WIP: Add Search for category settings

Y0ngg4n:
## Changes

This adds fuzzy search to the settings page.
- [x] Search trough categories
- [ ] Search trough settings widgets


.direnv/bin/nix-direnv-reload
Chaphasilor, line 1: this file probably shouldn't be committed?

Y0ngg4n:
> Are you using `dart format .`? There are a lot of formatting changes
> 
> Also just throwing it in there: Instead of a Hardcoded list, maybe there is a way to use code generation or dynamically scan all `/settings` routes for strings? Maybe a custom `@settings(icon: x)` decorator to mark a class as Setting? Idk if that works how I think it would.
> 
> If there is a code-gen way or decorator way it would be neat to also do that to revert to default settings like `@resetToDefault(aValue: DefaultSettings.aValue)`. But still no clue if that works at all :)
> 
> Also whats up with `.direnv`? Are these your custom/personal files (which i think they are given the hardcoded username and ID file names?), then you should probably gitignore them and delete from your fork

I dont know how i would to that. Could you may provide a link to an example?

Komodo5197:
I've been thinking about how this could be done.  The biggest issue is that all the localization strings and whatnot aren't created until the individual settings widgets are built, but we need them statically somehow.  I've been thinking about how we could implement this with subclassing, as suggested above, and I think I have something vaguely reasonable.  I could probably get codegen to gather up all the title/subtitle strings from the subclasses into a static field somewhere and build page/global reset methods if you made all the individual settings classes extend SettingsBase, SettingsBase with BoolSetting, or StatefulSettingsBase as defined below.

```dart
abstract class SettingsBase extends StatefulSettingsBase {
  SettingsBase({super.key});

  Widget build(BuildContext context, WidgetRef ref);

  @override
  // ignore: library_private_types_in_public_api
  _SettingsBaseState createState() => _SettingsBaseState();
}

abstract class StatefulSettingsBase extends ConsumerStatefulWidget {
  String title(BuildContext context);
  String subtitle(BuildContext context);

  /// Will reset to DefaultSettings.<variable name> by default.  Override if
  /// different resetting is needed.
  void reset(FinampSettings settings) {
    FinampSetters.resetByName(routeName);
  }

  abstract final String routeName;
  abstract final String settingName;

  const StatefulSettingsBase({super.key});

  @override
  ConsumerStatefulElement createElement() {
    assert(FinampSetters.isGenerated(routeName));
    return ConsumerStatefulElement(this);
  }
}

class _SettingsBaseState<T extends SettingsBase> extends ConsumerState<T> {
  @override
  Widget build(BuildContext context) => widget.build(context, ref);
}

mixin BoolSetting on SettingsBase {
  @override
    return SwitchListTile.adaptive(
      title: Text(title(context)),
      subtitle: Text(subtitle(context)),
      value: ref.watch(FinampSetters.providerByName(routeName)),
      onChanged: FinampSetters.setterByName(routeName),
    );
  }
}
```

The only methods I could come up with for extracting all the text without having to modify all the individual settings widgets with something like this would be either building a copy of every settings page offscreen on the main page, which seems like a bad idea, or changing the localization string names to be extremely consistently <variableName>Title and <variable name>Subtitle and then collection them via codegen, but we still need to maintain a separate mapping of variables to pages with that method.  I'm not sure any of these strategies are that great.

By the way, if you want to look at what codegen is already doing, the relevant files are build.yml and lib/builders/*.  It's fairly esoteric, though.

Y0ngg4n:
@Komodo5197 oh nice thank you very much. This week i dont have time for this. But next week i will try to finish this.

Chaphasilor:
@Y0ngg4n any updates on this? anything you need help with? :)

Y0ngg4n:
@Chaphasilor i will definitely finish this. I am just currently a bit busy. Sorry for the delay.

Chaphasilor:
@Y0ngg4n friendly ping, but no worries if you're still busy. Just let me know if I should resolve the conflicts :)

Y0ngg4n:
@Chaphasilor i am really sorry for the long waiting time. i will write you a pm. feel free to resolve the conflicts.

Y0ngg4n:
Hi there. I have to admit it was not going as planned with this PR and i am sorry for the long waiting times. i thought i can handle this but i have to admit i can't currently. I don't have the capacity. My honest apologize.

Apart from resolving the conflicts, this PR lacks one thing i haven't found a solution for:
I don't get the texts of the subcomponents as i dont know how to extract the texts recursively.
So if you test the PR you will see that the category settings search works fine but the search for the individual components doesn't work.
@Komodo5197 said that it may be better to move to a 2 stage approach. i thought long about this but i don't not know how to implement this as my knowledge regarding code generators in dart is quite limited to be honest. The problem with a 2 stage approach is that you have to somehow get the instance of the widget to render it. So the things you can do is some reflection magic i really don't want to do or you could make it like i did now with the existing component references but that has the problem that the string extraction runs at runtime and not before with the codegen. 
Maybe you know a way but currently i am kinda out of ideas and time and don't know exactly why the existing solution does not work, but don't have the time to debug it. If somebody, maybe @Komodo5197 has the capacity to take this over or finish it, it would be greatly appreciated. If you have any questions, feel free to contact me.

PR #1325, opened 2025-08-08
[Redesign] Various themeing fixes

Komodo5197:
Various fixes, mostly to the currently playing album highlighting.
- Fix theme transitions not being properly skipped while widgets are in background.
- Switch some widgets with unneeded states to stateless.
- Give album list items rounded corners.
- Fix musicscreentabview rebuilding on unrelated settings changes.
- Fix original queue source album being highlighted after queue restore when the currently playing item is not necessarily from that album.


Chaphasilor:
Those rounded corners on the album tile look much nicer btw :D

Komodo5197:
Yeah.  Queue restores were taking the original source and applying it to every individual new item's source, not just the original source of the new queue, which confused the highlighting.

PR #1326, opened 2025-08-08
Add new edit screen for playlists

### LINKED ISSUE #256: allow changing order of tracks on playlist / add sorting option
Currently there is no possibility to (re-)sort tracks on a playlist.

Would be nice to implement an option to move tracks up and down to allow changing the track order of an existing playlist in Finamp.

### LINKED ISSUE #1250: upload playlist thumbnails in app
### Type:
- [x] New feature
- [ ] Improve or extend an existing feature
- [ ] Other: <describe here>

### Explanation 

Add the ability to change the playlist thumbnail to an image from the device finamp is running on
### Reasoning

This would be easier than having to switch to the jellyfin web client whenever users want to change the playlist thumbnail

F-4Dev:
## Changes
This new screen will replace the original edit dialog with a new screen which will have all the original capabilities and will add new functions like changing the cover and reordering tracks inside the playlist. 

## Todo before merging
- [x] old capabilities
- [x] change cover
- [x] reorder tracks

## Related Issues
- closes #1250
- closes #256


Chaphasilor:
Okay, I've now added the track list to the edit screen, which can already be used to reorder tracks within the playlist and remove tracks from the playlist.  
It's already working quite nicely, but still needs some visual polish.
Also still todo:

- [x] Only show edit button if user has permission to edit (i.e., is the owner or an admin), we can reuse the check for the delete button
- [x] Implement updating the cover
- [x] Implement "restore" functionality to add removed tracks back into list
- [x] ~~If possible, implement dragging between playlist content and remove tracks lists~~
- [ ] Visual improvements
  - [x] Removed tracks header
  - [x] Remove X button
  - [x] Restore button
  - [ ] Cover & Name layout/design
- [x] Add translation strings
- [ ] Once all that is working, use the screen for creating new playlists (not sure how feasible that is)

lib/screens/playlist_edit_screen.dart
```diff
@@ -230,231 +231,203 @@ class _PlaylistEditScreenState extends ConsumerState<PlaylistEditScreen> {
...
+                    child: Builder(
+                      builder: (context) {
+                        return FlexibleSpaceBar(
+                          background: Align(
+                            alignment: AlignmentGeometry.bottomCenter,
+                            child: Container(
+                              margin: const EdgeInsets.symmetric(horizontal: 8, vertical: 8),
+                              padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 16),
+                              decoration: BoxDecoration(
+                                color: ColorScheme.of(context).primary.withOpacity(0.15),
```
Chaphasilor: @Komodo5197 I've been struggling with creating a theme based on the playlist item. I tried to adapt the logic from the `ThemedBottomSheet`, but something's still missing. Any idea what the issue is?

Chaphasilor:
Functionally this is now complete. Using this to create new playlists can be done later on, especially since that's a completely new (if requested) feature.  
I'd like to properly use the accent color for the flexible space, and I'd appreciate any ideas for improving/simplifying the code.

lib/screens/playlist_edit_screen.dart
Komodo5197: I think you're using the out of scope ref when setting the theme data.  Anyway, you should probably just use ItemTheme.

Chaphasilor: Ah, that makes sense. Is there something like a `ProviderScopeBuilder` that directly exposes the new ref?  
I've switched to `ItemTheme`, that is exactly what I was looking for, I just forgot that it existed.

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,497 @@
...
+    final resultPlaylist = await _jellyfinApiHelper.getPlaylist(_id!);
+    setState(() {
+      _publicVisibility = resultPlaylist['OpenAccess'] as bool;
+    });
+  }
+
+  Future<void> _saveOrUpdatePlaylist() async {
+    if (_formKey.currentState != null && _formKey.currentState!.validate()) {
+      setState(() {
+        _isUpdating = true;
```
Komodo5197: This never seems to get set false when errors do not occur.  EDIT - I guess it doesn't matter sine the widget is disposed immediately after?

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,497 @@
...
+        }
+        return;
+      }
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final bool isOffline = ref.watch(finampSettingsProvider.isOffline);
+    SortBy playlistSortBySetting = ref.read(finampSettingsProvider.playlistTracksSortBy);
```
Komodo5197: Why do we read this instead of watching?

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,497 @@
...
+    } else {
+      return null;
+    }
+  }
+
+  Future<void> _fetchPublicVisibility() async {
+    if (_publicVisibility != null) return;
+    final resultPlaylist = await _jellyfinApiHelper.getPlaylist(_id!);
+    setState(() {
+      _publicVisibility = resultPlaylist['OpenAccess'] as bool;
```
Komodo5197: This needs to also update the dirty baseline.

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,497 @@
...
+        return;
+      }
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final bool isOffline = ref.watch(finampSettingsProvider.isOffline);
+    SortBy playlistSortBySetting = ref.read(finampSettingsProvider.playlistTracksSortBy);
+    final playlistSortBy =
```
Komodo5197: This seems to be unused.

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,497 @@
...
+            ),
+            itemId: widget.playlist.id,
+          );
+        }
+
+        if (_hasCoverChanged) {
+          await _jellyfinApiHelper.setItemPrimaryImage(itemId: widget.playlist.id!, imageFile: newAlbumImage!);
+        }
+
+        musicScreenRefreshStream.add(null); // refresh playlist content
```
Komodo5197, line 187: You should invalidate getSortedPlaylistTracksProvider here as well.

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,497 @@
...
+  @override
+  void initState() {
+    super.initState();
+    _name = widget.playlist.name;
+    _id = widget.playlist.id;
+    _albumImage = widget.playlist;
+    _fetchPublicVisibility();
+    final tracksAsync = ref.read(getSortedPlaylistTracksProvider(widget.playlist));
+    final (allTracks, playableTracks) = tracksAsync.valueOrNull ?? (null, null);
+    playlistTracks = allTracks ?? [];
```
Komodo5197: You should copy allTracks before using it.  playlistTracks is being modified, so any later user of getSortedPlaylistTracksProvider is getting the modified list, even if you discard the changes.

Komodo5197 (commented):
- The edit button in the playlist menu still uses the legacy edit menu.  That should be fixed, and the old code removed.  You'll need to account for getSortedPlaylistTracksProvide possibly being uninitialized in init.
- The message on image edit failure was confusing.  The publicly visible toggle and and playlist name are also tpyes of metadata from a user perspective, but are unrelated to that permission.
- Having the reorder drag only possible via the drag handle felt unnatural.  I really wanted to be able to use any part of the track to drag, including dragging into the removed section instead of swiping or using the button.

lib/screens/playlist_edit_screen.dart
Komodo5197: That's what Consumer does.

lib/screens/playlist_edit_screen.dart
Chaphasilor: Yeah, it doesn't matter at the moment. But it should be reset anyway, in case we should ever decide not to pop.

lib/screens/playlist_edit_screen.dart
Chaphasilor: Not sure, copied it from the playlist screen. I'll probably just remove it since it's not used. I was debating including the sorting settings on the edit screen to allow persistently reordering the playlist based on track names, durations, or albums, but I decided against it, since after all this is why we have the sorting options on the playlist screen to begin with.

Chaphasilor:
I've fixed your issues, except for dragging between the lists. I might take a look at that tomorrow, but feel free to tackle that if you have an idea how to do it in a nice way. I haven't found a good way to drag items between two lists yet, it's either dropping onto the full list and appending at the end (acceptable for removing, but not really for restoring) or mixing in drag targets with the regular list items.

lib/services/permission_providers.dart
```diff
@@ -0,0 +1,93 @@
...
+      .getPlaylistUser(id)
+      .then((response) {
+        return response.canEdit;
+      })
+      .catchError((_) {
+        return false;
+      });
+});
+
+final Provider<bool> canEditMetadataProvider = Provider((ref) {
```
Komodo5197: It might be sensible to autodispose these so that they can actually update.  I cahnged the permissions on my test user and had to restart finamp to pick it up.

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,659 @@
...
+              userId: GetIt.instance<FinampUserHelper>().currentUserId,
+            ),
+            itemId: widget.playlist.id,
+          );
+        }
+        if (_hasCoverChanged) {
+          await _jellyfinApiHelper.setItemPrimaryImage(itemId: widget.playlist.id!, imageFile: newAlbumImage!);
+        }
+        musicScreenRefreshStream.add(null); // refresh playlist content
+        ref.invalidate(getSortedPlaylistTracksProvider(widget.playlist));
```
Komodo5197: It looks like we actually need to invalidate  the getAlbumOrPlaylistTracks provider instead.

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,659 @@
...
+    }
+  }
+
+  // Controller to communicate save state transitions to FAB
+  final _playlistSaveFeedbackController = _PlaylistSaveFeedbackController();
+
+  @override
+  Widget build(BuildContext context) {
+    // wait for playlist tracks, then mark loading as false
+    ref.listen(getSortedPlaylistTracksProvider(widget.playlist), (_, tracksAsyncLoaded) {
```
Komodo5197: It might be a bit more readable for this to be a listenManual placed alongside the read in initState.

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,659 @@
...
+    final result = await FilePicker.platform.pickFiles(type: FileType.image);
+    if (result == null) return null;
+    return File(result.files.single.path!);
+  }
+
+  Future<void> _fetchPublicVisibility() async {
+    if (_publicVisibility != null) return;
+    final resultPlaylist = await _jellyfinApiHelper.getPlaylist(widget.playlist.id!);
+    setState(() => _publicVisibility = resultPlaylist.openAccess);
+    _resetDirtyBaseline();
```
Komodo5197: We should probably limit the reset to just publicVisibility baseline incase the response is delayed and the user has already started making changes.

lib/components/AlbumScreen/playlist_edit_button.dart
```diff
@@ -16,9 +15,11 @@ class PlaylistNameEditButton extends StatelessWidget {
     return IconButton(
       icon: const Icon(TablerIcons.edit),
       tooltip: AppLocalizations.of(context)!.editItemTitle(BaseItemDtoType.fromItem(playlist).name),
-      onPressed: () => showDialog(
-        context: context,
-        builder: (context) => PlaylistEditDialog(playlist: playlist),
+      onPressed: () => Navigator.push(
+        context,
+        PageRouteBuilder<PlaylistEditScreen>(
```
Komodo5197: The lack of an animation moving into the edit screen is fine, but it was a bit disorienting not having one for going back because it can happen automatically after the edit completes.

lib/components/AlbumScreen/playlist_edit_button.dart
```diff
@@ -16,9 +15,11 @@ class PlaylistNameEditButton extends StatelessWidget {
...
+          pageBuilder: (context, animation, secondaryAnimation) => PlaylistEditScreen(playlist: playlist),
```
Komodo5197: This isn't actually caused by these changes, but item_collection_wrapper uses a ref in its page route builder and it breaks hot reloads because its no longer mounted.  It would be nice if you moved the ref.watch call to outside the lambda.

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -589,42 +633,55 @@ class TrackListItem extends ConsumerWidget {
             child: unthemedItem,
           )
         : unthemedItem;
+    return features.contains(TrackListItemFeatures.fullyDraggable)
```
Komodo5197, line 636: This doesn't seem to actually work, at least on desktop.

lib/screens/playlist_edit_screen.dart
```diff
@@ -0,0 +1,659 @@
...
+        ref.invalidate(getSortedPlaylistTracksProvider(widget.playlist));
+        if (!mounted) return;
+        GlobalSnackbar.message((context) => AppLocalizations.of(context)!.playlistUpdated, isConfirmation: true);
+        setState(() {
+          _isUpdating = false;
+          _resetDirtyBaseline();
+        });
+        // Trigger success flash before pop
+        _playlistSaveFeedbackController.success();
+        await Future<void>.delayed(const Duration(milliseconds: 500));
```
Komodo5197: This doesn't work properly.  The button gets hidden due to the dirty baseline being reset.  We also already have the snackbar, so it just feels like the screen closing is out of sync.

Komodo5197:
I don't really have much experience with the nuance of the drag slivers, so I think you'd know more than me.  My guess would be that placing the header into the reordable list and just not giving it a drag handle would be how you do it, but that's just a guess.

lib/services/permission_providers.dart
Chaphasilor: I've noticed that too. I'll need to lookup how to do that, for the annotation it's the difference between `@riverpod` and `@Riverpod(keepAlive: true)`.

Chaphasilor: Okay, so it seems to be `isAutoDispose: true` as an argument to the provider before the `ref` callback in Riverpod 3 (https://riverpod.dev/docs/concepts2/auto_dispose), and the `AutoDispose...` variant for RP2.

lib/components/AlbumScreen/track_list_tile.dart
Chaphasilor: It's a delayed handler, maybe that's why? I'll probably have to use the non-delayed version on desktop?

Chaphasilor:
Yes, placing the header in the list probably works, but requires careful index management. Maybe that's worth it though.

lib/components/AlbumScreen/playlist_edit_button.dart
Chaphasilor: So *that's* where the error came from. How did you debug this? I wasn't able to tell what's going on from the stack trace...

lib/screens/playlist_edit_screen.dart
Chaphasilor: I've removed the snackbar now, I think the animated confirmation is nicer and more cohesive. The changes are also reflected on the playlist screen (except for visibility, at least for now), so there's no real need for a clear confirmation anyway.

lib/screens/playlist_edit_screen.dart
Chaphasilor: That's what I wanted to do initially, but I didn't know I had to use `listenManual` instead of `listen`...  
Thanks!

lib/screens/playlist_edit_screen.dart
Chaphasilor: I also update the local `BaseItemDto` of the playlist to get the name to change, since the `BaseItemDto` is directly passed to the screen.  
I think eventually we'd want to avoid that by using globally cached item providers that we can use to read the value reactively instead of passing it to the screens statically.

Komodo5197: Yeah, this is pretty messy but I don't see a way around it without large scale refactoring.  By the way, we should also call downloadsService.resync with forceSync=true here to update that.

lib/components/AlbumScreen/track_list_tile.dart
Komodo5197: I have no clue why, but that does seem to fix it.  I've actually been wondering if it makes sense to add more directly relevant platform flags to an extension so we don't have random scatterings of isIOS || isAndroid scattered around.  So we could make isDesktop, isMobile, isMediaKit, isTouchscreen, and so on.

lib/components/AlbumScreen/playlist_edit_button.dart
Komodo5197: I just walked down the stacktrace until I got to the first code that was actually written by us.  And then it was fairly obvious why that lambda was incorrect, because WidgetRefs are basically just BuildContext, so it's the pretty much same BuildContext across an async gap issue the linter's always warning about, except apparently a WidgetRef version of that isn't implemented or isn't working with our project settings.

lib/screens/playlist_edit_screen.dart
```diff
@@ -105,20 +107,43 @@ class _PlaylistEditScreenState extends ConsumerState<PlaylistEditScreen> {
...
+      );
     }
     _initialName = _name ?? '';
     _initialVisibility = _publicVisibility ?? false;
     _initialTrackIdsOrder = playlistTracks.map((t) => t.id.raw).toList();
   }
 
+  @override
+  void dispose() {
+    _playlistTracksSubscription?.close();
```
Komodo5197: You don't actually need to save the subscription and manually dispose it.  listenManual automatically tracks the life cycle of the WidgetRef, so it disposes on its own here.  We would need to manually dispose and recreate it in didUpdateWidge if we had that, but we don't expect that to occur for this widget.

Komodo5197:
I can't imagine that an edit screen is the most high traffic area of the app, so while the cross drag would be nice, I wouldn't put too much effort into it if presents a lot of difficulties.

lib/services/permission_providers.dart
```diff
@@ -0,0 +1,99 @@
...
+          });
+    });
+
+final AutoDisposeProviderFamily<bool, BaseItemDto> canEditItemProvider = AutoDisposeProviderFamily((
+  ref,
+  BaseItemDto item,
+) {
+  var itemType = BaseItemDtoType.fromItem(item);
+
+  // do not bother checking server for item types known to not be deletable
```
Maxr1998:
```suggestion
  // do not bother checking server for item types known to not be editable
```

lib/services/permission_providers.dart
```diff
@@ -0,0 +1,99 @@
...
+          .catchError((_) {
+            return false;
+          });
+    });
+
+final AutoDisposeProviderFamily<bool, BaseItemDto> canEditItemProvider = AutoDisposeProviderFamily((
+  ref,
+  BaseItemDto item,
+) {
+  var itemType = BaseItemDtoType.fromItem(item);
```
Maxr1998, line 56: Why does this have no offline check but delete has? And don't the call sites check for offline already?

Maxr1998 (commented):
Haven't tested it on device and can't comment on the provider stuff, but it generally seems fine to me.

lib/services/permission_providers.dart
Chaphasilor: I was thinking of offline edits to playlists, which eventually would be nice to have. For now we have checks in place at other positions though, as you noted.  
I didn't change the canDelete provider, but I think it's good to check the offline state in the provider, that should be less error-prone.

lib/screens/playlist_edit_screen.dart
Chaphasilor: I actually read that and deleted it, but I think that ended up getting revered when I tried something else (or maybe it ended up in another commit)

Chaphasilor:
If this is fine for you to merge as-is then I'd probably go ahead with this for now, if there is demand by other (or if someone wants to volunteer) we can always add the improved dragging.

lib/services/permission_providers.dart
Maxr1998: Understood, that sounds reasonable. Could you maybe add a comment to the code explaining that?

Chaphasilor: yeah I'll do that tomorrow :)

Komodo5197:
Feel free to merge it without the cross drag, and we'll see if anyone else cares enough to complain.  Maybe if we ever get it working for the queue we can port it over.  Also, it looks like you messed up the formatting again - you should probably get an autoformatter set up or reconfigure it or something.

Chaphasilor:
I actually have auto-format enabled, but that only applies to files I change, so it often gets messed up during merges. Not sure what happened here.  
I was thinking about just adding a pre-commit hook that formats things, how does that sound?

Chaphasilor:
@F-4Dev 🙌

lukaslindnermusic:
nice one 😍😍😍
thanks for implementing this functionality ❤️

one tiny thing I asked myself after checking it out: the track numbers in the reordering-list start with 0, which of course is the correct int when you are looking at it from a programming standpoint, but I think for the user-facing stuff, it makes more sense to start the list with id "1"?

F-4Dev:
I love how it turned out. Thank you for making it happen so fast. I really which I could have done more 😅
But anyways.. Thank you ❤️

Chaphasilor:
You're absolutely right about the index xD

PR #1331, opened 2025-08-12
Fix dynamic/album-based gain modes in background

HeroBrine1st:
## Changes

Fixes the following bug caused by Riverpod (see related issues):

<details>
<summary>The bug</summary>

To reproduce, you need 5 albums and a playlist:

- 4 unrelated tracks (from different albums)
- 2 tracks from the same album (called "reproducer album" later) with different track normalization gain (6 dB in my case) and track normalization gain of first of them should be more (i.e. track is louder) than album gain. Examples: Myth by Two Steps from Hell (~10 dB difference), ATTENTION ATTENTION by Shinedown (~6 dB difference, that's what I faced this bug on)

I recommend creating dedicated playlist instead of using a sublist of an existing playlist for tests in debug builds (they are very laggy to the point they are unusable!).

Then:

- You can start the queue from the first of the four unrelated tracks.
- Then stop the app ("Force stop" in android settings) and open it.
- Restore the queue or start again. Start playback, lay the phone down and shut the screen off (you can move it).
- Then don't do anything on phone and wait for reproducer album to play. The first track is very loud
  You can also skip instead of waiting, but skip three times (so that 4th track in a playlist is playing) and then seek it almost to the end, leaving some time to fetch the reproducer album metadata. Do not open the Finamp app!

Then either:

- Open the app and hear that volume changed to album-based ([this code](https://github.com/jmshrv/finamp/blob/1efd315d9a3d665a94db31912ad6eb5cbac3d314/lib/services/music_player_background_task.dart#L1115-L1124) triggers)
- Wait for second track and see that volume is suddenly changed ([#769](https://github.com/jmshrv/finamp/issues/769) helps to hear that)

This is caused by riverpod listeners being frozen when app is in background. This can be seen in Finamp logs - metadata is fetched only for currently playing track (which confirms `read` forces the fetching), but if you bring finamp to foreground after queue is shifted enough, other tracks are fetched simultaneously.

</details>

There are two ways to fix this (ignoring Jellyfin changes):

- Use `read` to force fetching of metadata even in background
- Split `ProviderContaier` as mentioned in https://github.com/rrousselGit/riverpod/issues/2671#issuecomment-2973693074

This PR goes the former way. This is dirtiest as it is possible as it uses (possibly) undocumented behavior however the latter way is only worse (it will double the requests). I will fully understand if you don't want a such change, then the only reasonable way is to expose album normalization gain with track metadata on Jellyfin (which I hope I'll soon dedicate time to!)

I moved `GetIt.instance<ProviderContainer>()` out of switch-case block because we need to consider `dynamic` mode in case tracks are not sequential.

Bug is reproduced on PR base. Used reproduction steps using both "album-based" and "dynamic" modes. Tested with debug apk build!

## Todo before merging

I recommend thoroughly reviewing this PR and merge only if you fully understand the parent issue, this issue and possible consequences of the change as I am not experienced with Flutter enough.

- Reproduce the bug on Android (so that I am not the only one affected)
- Reproduce the bug and test this PR on iOS too
- Consider the crash explained [here](https://github.com/rrousselGit/riverpod/issues/3354) because `dependencies.flutter_riverpod: ^2.6.1` is not 3.0.0 where it is fixed.

## Related Issues

https://github.com/rrousselGit/riverpod/issues/2671

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -44,6 +44,7 @@ import 'package:finamp/services/ui_overlay_setter_observer.dart';
 import 'package:finamp/services/widget_bindings_observer_provider.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/scheduler.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_localizations/flutter_localizations.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
@@ -96,6 +97,8 @@ import 'services/theme_mode_helper.dart';
 import 'setup_logging.dart';
 
 final _mainLog = Logger("Main()");
+
+final providerScopeKey = GlobalKey();
 
 void main() async {
   // If the app has failed, this is set to true. If true, we don't attempt to run the main app since the error app has started.
@@ -298,6 +301,14 @@ Future<void> _setupProviders() async {
 
   DataSourceService.create();
   AutoOffline.startWatching();
+
+  unawaited(
+    Stream.periodic(Duration(seconds: 1)).forEach((_) {
+      if (!SchedulerBinding.instance.framesEnabled) {
+        (providerScopeKey.currentContext as InheritedElement?)?.build();
+      }
+    }),
+  );
 }
 
 Future<void> _setupOSIntegration() async {
@@ -523,6 +534,7 @@ class _FinampState extends State<Finamp> with WindowListener {
   @override
   Widget build(BuildContext context) {
     return UncontrolledProviderScope(
+      key: providerScopeKey,
       container: GetIt.instance<ProviderContainer>(),
       child: GestureDetector(
         onTap: () {
--- a/lib/components/AlbumScreen/track_list_tile.dart
+++ b/lib/components/AlbumScreen/track_list_tile.dart
@@ -642,7 +642,7 @@ class TrackListItemTile extends ConsumerWidget {
     );
     final addSpaceAfterSpecialIcons =
         (downloadedIndicator.isVisible(ref) || (baseItem.hasLyrics ?? false)) && (showDateAdded || showDateLastPlayed);
-      
+
     final showPlaybackProgress = !highlightCurrentTrack && playbackProgress != null && playbackProgress! < 0.99;
 
     return ListTileTheme(
--- a/lib/components/PlaybackHistoryScreen/playback_history_list.dart
+++ b/lib/components/PlaybackHistoryScreen/playback_history_list.dart
@@ -36,7 +36,7 @@ class PlaybackHistoryList extends StatelessWidget {
                     item: group.value[actualIndex].item.baseItem!,
                     isShownInSearchOrHistory: true,
                     highlightCurrentTrack: groupIndex == 0 && index == 0, // only highlight first track
-                    playbackProgress: group.value[actualIndex].playPercentage
+                    playbackProgress: group.value[actualIndex].playPercentage,
                   );
 
                   return index == 0
--- a/lib/services/current_track_metadata_provider.dart
+++ b/lib/services/current_track_metadata_provider.dart
@@ -13,6 +13,9 @@ final currentTrackMetadataProvider = AutoDisposeProvider<AsyncValue<MetadataProv
     BaseItemDto? base = itemToPrecache.baseItem;
     if (base != null) {
       ref.listen(metadataProvider(base), (_, __) {});
+      ref.read(
+        metadataProvider(base),
+      ); // forces it even in background https://github.com/rrousselGit/riverpod/issues/2671
     }
   }
 
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -1129,11 +1129,15 @@ double? getGainForCurrentPlayback(MediaItem currentTrack, jellyfin_models.BaseIt
       item ?? jellyfin_models.BaseItemDto.fromJson(currentTrack.extras?["itemJson"] as Map<String, dynamic>);
 
   double? effectiveGainChange;
+  final providerContainer = GetIt.instance<ProviderContainer>();
+  providerContainer.read(
+    currentTrackMetadataProvider,
+  ); // forces it even in background https://github.com/rrousselGit/riverpod/issues/2671
+
   switch (FinampSettingsHelper.finampSettings.volumeNormalizationMode) {
     case VolumeNormalizationMode.hybrid
         when GetIt.instance<QueueService>().getQueue().isCurrentlyPlayingTracksFromSameAlbum():
     case VolumeNormalizationMode.albumBased:
-      final providerContainer = GetIt.instance<ProviderContainer>();
       // final parentNormalizationGain = providerContainer.read(currentTrackMetadataProvider).valueOrNull?.parentNormalizationGain;
       // includeLyrics is always true - fetch the metadataRequest directly.
       // Requires that provided arguments are the only fields of request,
--- a/lib/services/playback_history_service.dart
+++ b/lib/services/playback_history_service.dart
@@ -258,8 +258,7 @@ class PlaybackHistoryService {
           // played less than 30 seconds
           (element.playDuration ?? Duration.zero) >= const Duration(seconds: 30) ||
               // played less than 20% of the track duration
-              (element.playPercentage ?? 0.0) >=
-                  0.2)) {
+              (element.playPercentage ?? 0.0) >= 0.2)) {
         continue;
       }
 


Chaphasilor:
maybe we should think about upgrading to Riverpod 3?  
I think having two separate provider containers, one for actual background stuff like metadata fetching, which is independent from the UI, and another one that only is used in UI shown to the user, seems reasonable to me.  
Anything that is a widget would always use the second container, and all services would always use the first one.  
Only problem might be that the containers don't "talk" to each other, i.e. updating a value/notifier in one container wouldn't cause an update in the other? Haven't looked into it deeply enough yet.

Riverpod 3 looks nice in general, also for persisting providers across restarts and automatic retries :)

HeroBrine1st:
> It also seems a bit more optimized than a periodic re-build in the background.

Except when metadataProvider is slow to fetch album gain, then this code won't trigger:

https://github.com/jmshrv/finamp/blob/1efd315d9a3d665a94db31912ad6eb5cbac3d314/lib/services/music_player_background_task.dart#L1112-L1125

Periodic rebuild should handle that. It is pretty rare case though. I'm currently busy with implementing playing specific discs (it is somewhat 80% complete if my approach works), but I will test the proposed hack after that.

HeroBrine1st:
So, this hack works. I have working tree at 5c01bb98. My patch is hacky in its own way (passing value from main down the tree instead of using GetIt) but it is okay for testing it.

So, this rebuilds the UI. Probably not every widget, but I understood why it is questionable.

Also, we'll need some container for Globalkey (to avoid removing `const` from root widget) - just imagine a new contributor seeing `GetIt.instance<GlobalKey>()`.

So, which one?

- Current implementation - uses a glitch (not a bug-not a feature), requires fast metadata fetcher, ~~additional code in normalization gain calculation function but~~ that additional code can be extracted into another listener
- 5c01bb98-like - probably won't break easily in Riverpod if Riverpod stays to reusing Flutter scheduler but if Flutter changes something, this will fail. 

I also propose hybrid approach taking the best of both hacks, meaning it has twice the chance to fail :joy::

- New track plays, listener gets metadata for current track (reading `currentTrackMetadataProvider`, forcing it to fetch additional tracks)
- If metadata is not available yet, the Komodo5197's hack is used to periodically rebuild until metadata is available
- After that, periodic rebuild stops

Although it might break autoReloadQueue but idk how it can be triggered in background

UPD: Forgot to mention that I have tested slow metadata fetcher with this:

```diff
diff --git a/lib/services/metadata_provider.dart b/lib/services/metadata_provider.dart
--- a/lib/services/metadata_provider.dart	(revision 5c01bb9834c330e92672b700a540706faedde15b)
+++ b/lib/services/metadata_provider.dart	(date 1755353917303)
@@ -193,6 +193,8 @@
         "Fetched metadata for '${item.name}' (${item.id}): ${metadata.lyrics} ${metadata.hasLyrics}",
       );
 
+      await Future.delayed(Duration(seconds: 15));
+
       return metadata;
     });
```

It works.

UPD2: Hybrid approach can be replaced by polling though, so only two options remain

Komodo5197:
So, as an alternative to making the FinampApp widget not const, you can just make the GlobalKey a global constant.  This is basically what GlobalSnackBar does.  I don't think FinampApp being constant actually matters either way, though.

Regarding what rebuilds, widgets don't recursively build their children themselves, so calling build() on the provider scope does not rebuild the whole widget tree, as that would also be a problem during normal operation.  All it does is cause pending provider builds to activate.  So I think the worst thing it might be doing is keeping the theming color provider up to date?  But I haven't tested that very much.  And none of the actual widgets should be rebuilding, only providers.

Chaphasilor:
Regarding autoReloadQueue, that can be triggered by autoOffline kicking in, or in the future when transcoding settings are changed dynamically. So it would be good if that works.  
You said that a hybrid solution could be replaced by polling, but is that as "reactive"? If we have a fast metadata provider and need the album gain right away (i.e. user skips to ahead a few tracks and the to almost the end of the track, or the track is just very short), then a poling approach might also be too slow to trigger the rebuild. So to me it seems like the reactive approach of the `.read()` combined with periodic polling should be a good way to get around this? And if either approach breaks due to upstream changes, we would still have the other approach as a backup too?

HeroBrine1st:
> You said that a hybrid solution could be replaced by polling, but is that as "reactive"?

Well, yes, but actually they both aren't. To user, they are the same.

> then a poling approach might also be too slow to trigger the rebuild

Under polling approach I mean doing something like that:

```dart
mediaItem.listen((currentTrack) {
  while(true) {
    final a = providerContaier.read(currentTrackMetadataProvider).valueOrNull;
    if(a == null || a is not for currentTrack) {
      await Future.delayed(seconds: 10);
    } else { 
      applyNormalizationGain(); 
      break;
    }
  }
})
```

As you see, there can be at max 10 seconds between playing new track and applying gain. This is the same with hybrid solution (where read is accompanied by `(providerScopeKey.currentContext as InheritedElement).build();` and there's no explicit call to `applyNormalizationGain` - this is "reactive" but under the hood it is the same polling).

The only way to fix that is to increase polling frequency.

Btw, see the related issue - it is fixed in Riverpod 3.0 but is not released yet (there's no tag pointing to history with PR commit at least): https://github.com/rrousselGit/riverpod/pull/4238. I don't understand how it works, looks like it does the same `(providerScopeKey.currentContext as InheritedElement).build();` but from inside of an element and faster.

HeroBrine1st:
Didn't even test that, it is the same code as I tested on https://github.com/jmshrv/finamp/commit/5c01bb9834c330e92672b700a540706faedde15b but rearranged.

Comparison: [5c01bb98..96e86b10](https://github.com/jmshrv/finamp/compare/5c01bb9834c330e92672b700a540706faedde15b..96e86b105dd3074e23ef66ac1241b62a8e68b1b8)

HeroBrine1st:
> And if either approach breaks due to upstream changes, we would still have the other approach as a backup too

Valid point. I cherry-picked the commit. Didn't test too :-)

Chaphasilor:
Hmm. If it's fixed upstream, should we just wait and make use of the proper fix, instead of jerry-rigging this?

Komodo5197:
Riverpod 3.0 isn't released yet, and I don't know how soon it can really be expected.  I'd just move forward with this for now.  If this is similar to the official solution, it might make sense to make to speed up our timer.  Nothing should happen during the scope build if no providers have invalidated, and I don't think we have any providers that are expected to change more than once or twice per track, especially in the background, so we might be fine to pull the frequency up to once per second, or maybe more.  That sort of speed might also make reduced rebuild latency of the added read methods less necessary.

Chaphasilor:
Oh I see, that wasn't too obvious since the documentation already has a migration guide...  
Okay, let's try a 1s periodic interval then, and if there aren't any obvious drawbacks, we'll merge it for now.

HeroBrine1st:
Retested with:

```diff
diff --git a/lib/services/metadata_provider.dart b/lib/services/metadata_provider.dart
--- a/lib/services/metadata_provider.dart	(revision ee837028038bfc7cbdb7c43edbc412ca25b91ab4)
+++ b/lib/services/metadata_provider.dart	(date 1755520323785)
@@ -193,6 +193,8 @@
         "Fetched metadata for '${item.name}' (${item.id}): ${metadata.lyrics} ${metadata.hasLyrics}",
       );
 
+      await Future.delayed(Duration(seconds: 10));
+
       return metadata;
     });
 
```

Would be very good to test on iOS too (well, we can't anymore - ios job is broken ha-ha)

Chaphasilor:
@edeuss would you mind giving this a test on iOS? (bonus points if you can figure out why the build is failing there 😅)

Edit for reference:

> To reproduce, you need 5 albums and a playlist:
> 
> - 4 unrelated tracks (from different albums)
> - 2 tracks from the same album (called "reproducer album" later) with different track normalization gain (6 dB in my case) and track normalization gain of first of them should be more (i.e. track is louder) than album gain. Examples: Myth by Two Steps from Hell (~10 dB difference), ATTENTION ATTENTION by Shinedown (~6 dB difference, that's what I faced this bug on)
> 
> I recommend creating dedicated playlist instead of using a sublist of an existing playlist for tests in debug builds (they are very laggy to the point they are unusable!).
> 
> Then:
> 
> - You can start the queue from the first of the four unrelated tracks.
> - Then stop the app ("Force stop" in android settings) and open it.
> - Restore the queue or start again. Start playback, lay the phone down and shut the screen off (you can move it).
> - Then don't do anything on phone and wait for reproducer album to play. The first track is very loud
>   You can also skip instead of waiting, but skip three times (so that 4th track in a playlist is playing) and then seek it almost to the end, leaving some time to fetch the reproducer album metadata. Do not open the Finamp app!
> 
> Then either:
> 
> - Open the app and hear that volume changed to album-based ([this code](https://github.com/jmshrv/finamp/blob/1efd315d9a3d665a94db31912ad6eb5cbac3d314/lib/services/music_player_background_task.dart#L1115-L1124) triggers)
> - Wait for second track and see that volume is suddenly changed ([#769](https://github.com/jmshrv/finamp/issues/769) helps to hear that)
> 
> This is caused by riverpod listeners being frozen when app is in background. This can be seen in Finamp logs - metadata is fetched only for currently playing track (which confirms `read` forces the fetching), but if you bring finamp to foreground after queue is shifted enough, other tracks are fetched simultaneously.

HeroBrine1st:
Just a friently reminder that this PR slowly but surely becomes stale.

Btw, I have just opened a pull request in jellyfin that will definitely fix this issue: https://github.com/jellyfin/jellyfin/pull/14745

lib/main.dart
```diff
@@ -298,6 +301,14 @@ Future<void> _setupProviders() async {
 
   DataSourceService.create();
   AutoOffline.startWatching();
+
+  unawaited(
+    Stream.periodic(Duration(seconds: 1)).forEach((_) {
+      if (!SchedulerBinding.instance.framesEnabled) {
+        (providerScopeKey.currentContext as InheritedElement).build();
```
Komodo5197: I think the key context is nullable, so this should probably be as InheritedElement?.

HeroBrine1st: Yeah, you're right, `currentContext` returns `BuildContext?`. I thought linter will at least warn about that

Chaphasilor:
I guess we can just merge this into `redesign` and then people on iOS will be more likely to test it than testing this PR?  
I sadly can't test on iOS, it's working for me on Android...  
Any objections?

This might also be related to the favorite button state in the media notification not updating properly.

HeroBrine1st:
This should break nothing and if anything it can only add performance issues so I agree it is clear to merge

Chaphasilor:
Awesome. Thanks for the PR!

PR #1332, opened 2025-08-12
Fix Offline Automation not resuming when PreferLocalNetwork is enabled

flloschy:
## Changes
When PreferLocalNetwork is enabled, automation wouldnt resume after overwriting. This is now fixed (reason stated in code comment)

Also some formatting sneaked into the commit :)

### DIFF
--- a/lib/services/network_manager.dart
+++ b/lib/services/network_manager.dart
@@ -40,7 +40,8 @@ class AutoOffline extends _$AutoOffline {
   static void startWatching() {
     ProviderContainer container = GetIt.instance<ProviderContainer>();
 
-    container.listen(autoOfflineProvider, (_, automationEnabled) {
+    container.listen(autoOfflineProvider, (_, automationState) {
+      bool automationEnabled = automationState > 0;
       _networkAutomationLogger.info("${automationEnabled ? "Enabled" : "Paused"} Automation");
 
       if (automationEnabled) {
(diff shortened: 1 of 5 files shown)


lib/services/network_manager.dart
```diff
@@ -64,7 +65,14 @@ class AutoOffline extends _$AutoOffline {
         ref.watch(FinampUserHelper.finampCurrentUserProvider).valueOrNull?.preferLocalNetwork ??
         DefaultSettings.preferLocalNetwork;
 
-    return (autoOfflineEnabled && autoOfflineActive) || autoServerSwitch;
+    // instead of `(autoOfflineEnabled && autoOfflineActive) || autoServerSwitch` this function
+    // needs to return a not-bool. This is because when autoServerSwitch is true, the function always
+    // returns true, and since the value doesnt change the listener wont trigger. Thus this number workaround
+    int state = 0;
```
Komodo5197, line 77: We might want to return an enum here.  We could have

```
enum AutoOfflineListenerType {
disabled,
offlineOnly,
serverSwitchOnly,
offlineAndServerSwitch
}
```
or something like that.  Or maybe just massage the comment a bit to make it extra clear that we need to make sure the provider updates and _onConnectivityChange runs for several relevant setting changes, even if the connection listener is already active.

flloschy: I feel like making this an enum makes it more complicated and unreadable than just leaving it like this

Komodo5197: Okay, that's fine.  But I would like the comment altered, because it took me a minute or two to work out why triggering the listener on true/false wasn't enough.

Chaphasilor:
thanks for fixing this :)

Issue #632, opened 2024-03-12
[Redesign] [Feature Request] Play specific Disc of an album


mihawk90:
Currently it is possible to play an entire album or selected tracks, however (for multi-disc albums) there seems to be no way to play only a specific disc. The "Disc X" header has no function (other then the visual separation of course).

Ideally the header would behave the same as selecting a track would, i.e. it would play the disc on tap, while long-tap opens the established context menu.

Chaphasilor:
Sounds like a plan, thanks. And let's do the features on-by-one. I'm really not convinced about any kind of hidden tracks that are remembered client side...

Chaphasilor:
Forgot to mark this as closed. Think it dropped in 0.9.19?

mihawk90:
Oh damn thanks for the notice, I totally missed that

PR #1333, opened 2025-08-13
Allow playing specific disc of album

HeroBrine1st:
## Changes

- (NixOS-related) Adds second devShell which stubs rustup with [fenix](https://github.com/nix-community/fenix), decreasing build tool state persistence
- Adds disc album menu available by long pressing on disc header or pressing "dots_vertical" icon button. All playback actions use tracks of selected disc only
- Adds ability to start playback by tapping on disc header, which replaces queue with disc contents

I have fought with application architecture, so the way I've implemented it is up for debate.

Numbers here are also up to debate (that's disc header):

```dart
header: Padding(
  padding: const EdgeInsets.symmetric(horizontal: 8.0, vertical: 8.0),
  child: Material(
    color: Theme.of(context).colorScheme.surface,
    borderRadius: BorderRadius.circular(8),
    clipBehavior: Clip.antiAlias,
    child: InkWell(
      // ...
      child: Padding(
        padding: const EdgeInsets.symmetric(horizontal: 8.0, vertical: 8.0),
        child: Row(
```

I'm not a designer so I picked what doesn't stick out a mile and looks nice :-)

<details>
<summary>Screenshots</summary>

image

Hovering over disc header:

image

Disc menu:

image

There are no entries under "Add to playlist" currently (changed after screenshots were taken)

</details>

## Todo before merging

- [x] Translations

## Related Issues

https://github.com/jmshrv/finamp/issues/632


HeroBrine1st:
I completely forgot about "Add to playlist" and "Add to mix".. Give me a minute

UPD: And it looks like another widget which is too generic, so I'll do it after review as the way I changed playback actions is up for debate

https://github.com/jmshrv/finamp/blob/e2851074b658152939a4b03473aba1b681956e92/lib/menus/components/menuEntries/instant_mix_menu_entry.dart#L19-L34

I'll clone both `AddToPlaylistMenuEntry` and `InstantMixMenuEntry` and replace BaseItemDto with List<BaseItemDto>, as in this case there's no need to copy too much code. And remove `AdaptiveDownloadLockDeleteMenuEntry` (I doubt download system supports this exotic case), `ToggleFavoriteMenuEntry` and `DeleteFromServerMenuEntry` if `discTrackList` is not null of course.

HeroBrine1st:
As to generic playback actions, I see three ways:

- Copy all playback actions and `getPlaybackActionPages`. Not very good
- Add this `trackList` argument like I did. As Chaphasilor said, it adds inconsistency, and for some reason I still have done it that way!
- Exploit the fact `baseItem` is used only for queue:

https://github.com/jmshrv/finamp/blob/571cd985fb2817a21989e455a9b6738619f69800/lib/menus/components/playbackActions/playback_actions.dart#L88-L102

From there I initally proposed replacing `await loadChildTracks(baseItem: baseItem) ?? []` with a lambda parameter which is responsible for that. If source is also passed as a parameter (with factory for default lambda), `PlayPlaybackAction` loses dependency on baseItem, making it both consistent and generic. And it would be ideal if `AbstractPlaybackAction` is extracted accepting icon and label from child classes as constants and with overridable `onPressed` method, which is responsible for current onPressed except probably `Navigator.pop(context);`.

Idk why I haven't done it that way but went second way.

Komodo5197:
Another way to go about this would be with something like

```
sealed class PlayableItem {}

class AlbumDisc implements PlayableItem {
  AlbumDisc(this.parent, this.tracks);
  List<BaseItemDto> tracks;
  BaseItemDto parent;
}

class BaseItemDto with RunTimeTickDuration implements PlayableItem {
```
and then we could just upgrade QueueItemSource.formBaseItem and loadChildTracks to take a PlayableItem instead, and they could switch over the subclasses.

lib/l10n/app_en.arb
```diff
@@ -2743,5 +2743,22 @@
...
+        "example": "1"
+      },
+      "albumName": {
+        "type": "String"
+      }
+    }
+  },
+  "moreActionsOnAlbumDisc": "More actions on album disc",
+  "@moreActionsOnAlbumDisc": {
+    "description": "Text for more actions button that appears in the header of an album disc, and opens menu allowing to play specific disc"
```
Chaphasilor:
```suggestion
    "description": "Label for the disc header menu button."
```

lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -10,74 +10,67 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
 }
 
 class PlayPlaybackAction extends ConsumerWidget {
-  const PlayPlaybackAction({super.key, required this.baseItem});
+  const PlayPlaybackAction({super.key, required this.item});
 
-  final BaseItemDto baseItem;
+  final PlayableItem item;
 
   @override
```
Chaphasilor, line 71: copy-paste error? shouldn't this be `BaseItemDtoType.album`?

lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -10,74 +10,67 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
 import 'package:get_it/get_it.dart';
 
-Map<String, Widget> getPlaybackActionPages({required BuildContext context, required BaseItemDto baseItem}) {
+Map<String, Widget> getCommonPlaybackActionPages({required BuildContext context, required PlayableItem item}) {
```
Chaphasilor, line 15: I would like for `getPlaybackActionPages()` to still exist, and handle the request based on the provided arguments.  
If it's a playlists (or genre?), handle it with `getCommonPlaybackActionPages()` (maybe rename to `getDefaultPlaybackActionPages()` ?). If it's an artist, handle it with `getArtistPlaybackActionPages()`. And if it's an album, handle it with `getAlbumPlaybackActionPages()`.

This way I don't need to think about or even *know* which item type is getting handled, and things stay mostly consistent.

lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -10,74 +10,67 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
-            ShufflePlaybackAction(baseItem: baseItem),
-            if (queueService.getQueue().nextUp.isNotEmpty) ShuffleNextPlaybackAction(baseItem: baseItem),
-            ShuffleToNextUpPlaybackAction(baseItem: baseItem),
-            ShuffleToQueuePlaybackAction(baseItem: baseItem),
-          ],
-        ),
-      };
-  }
+
+  return {
```
Chaphasilor, line 18: Can we call this `overwriteTrackList`/`customTrackList`/`alternativeTrackList` instead? I want to discourage devs from using this option unless really needed. Maybe also add a doc comment (`///`) to it?

lib/menus/components/menu_item_info_header.dart
```diff
@@ -149,21 +179,37 @@ class TrackInfo extends ConsumerWidget {
...
   const AlbumInfo({super.key, required this.item, required this.condensed, required this.features});
 
-  final BaseItemDto item;
+  final PlayableItem item;
   final bool condensed;
   final List<MenuItemInfoHeaderFeatures> features;
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+    final item = this.item;
```
Chaphasilor, line 161: what's the point of this line?

HeroBrine1st:
This implements that and, honestly, it is much more clean now. I have implemented only "AddToPlaylistMenuEntry" though, other items lack flexibility.

I have one suggestion about new PlayableItem which I can't implement in SOLID manner (pun intended). Current `PlayableItem` doesn't support Liskov substitution and I identified that before I started doing it that way. Instead, I tried to go strategy pattern:

```dart
abstract interface class PlayableItem {
  QueueItemSource get source;
  Future<List<jellyfin_models.BaseItemDto>> getItems();
}

class RealPlayableItem implements PlayableItem {
  RealPlayableItem({required this.baseItem});

  final jellyfin_models.BaseItemDto baseItem;

  @override
  Future<List<jellyfin_models.BaseItemDto>> getItems() {
    return loadChildTracks(baseItem: baseItem);
  }

  @override
  QueueItemSource get source => QueueItemSource.fromBaseItem(baseItem);
}

class PrefetchedPlayableItem implements PlayableItem {
  PrefetchedPlayableItem({required this.items, required this.source});

  final List<jellyfin_models.BaseItemDto> items;

  @override
  final QueueItemSource source;

  @override
  Future<List<jellyfin_models.BaseItemDto>> getItems() {
    return Future.value(items);
  }
}
```

Generic, polymorfic, SOLID! And `PrefetchedPlayableItem` is very generic on its own.

Except the fact that QueueItemInfo has two additional items (name and type):

https://github.com/jmshrv/finamp/blob/4746ac5da3a651421bd21dd70b1badbc893b08bb/lib/models/finamp_models.dart#L1845-L1849
This is too much for wrapper class so I fall back to the proposed way because I couldn't do that cleanly. I can make `source` a `getSource` method and add those two `type` and `nameType` parameters to that method if you don't see the issue I see, but it will be probably too much responsibility and I didn't see Liskov substitution in the project as opposed to switch-case at the usage site.

I.e. make it like this:

```dart
abstract interface class PlayableItem {
  QueueItemSource getSource({QueueItemSourceType? type, QueueItemSourceNameType? nameType});
  Future<List<jellyfin_models.BaseItemDto>> getItems();
}
```

Or expose `parent` because MenuItemInfoSliverHeader needs it, idk.

Also I need to revert extracting MenuAlbumInfoSliverHeader, just spotted that

Chaphasilor (commented):
functionality looks good, I extended it to also support the swipe gestures that already exist for tracks.  
also left a few comments, but that was before you force-pushed your rebase, so I don't know how relevant they still are...

lib/menus/components/menu_item_info_header.dart
HeroBrine1st: Dart type inference engine doesn't understand that a final field can't suddenly change type in switch-case block, so that's a workaround to make it know it is definitely constant

Chaphasilor: alright

lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -109,6 +116,95 @@ class _AlbumScreenContentState extends ConsumerState<AlbumScreenContent> {
...
+    final swipeRightEnabled = ref.watch(finampSettingsProvider.itemSwipeActionRightToLeft) != ItemSwipeActions.nothing;
+    final allowedDismissDirection = (swipeLeftEnabled && swipeRightEnabled)
+        ? DismissDirection.horizontal
+        : swipeLeftEnabled
+        ? DismissDirection.startToEnd
+        : swipeRightEnabled
+        ? DismissDirection.endToStart
+        : DismissDirection.none;
+
+    Future<bool> discHeaderConfirmDismiss({
```
Komodo5197: I'd prefer not to declare such a long method inside of build().  Also, is it possible to consolidate any of this with the track copy?

lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -158,11 +254,66 @@ class _AlbumScreenContentState extends ConsumerState<AlbumScreenContent> {
...
+                    context: context,
+                    direction: DismissDirection.startToEnd,
+                    action: ref.watch(finampSettingsProvider.itemSwipeActionLeftToRight),
+                  ),
+                  secondaryBackground: buildSwipeActionBackground(
+                    context: context,
+                    direction: DismissDirection.endToStart,
+                    action: ref.watch(finampSettingsProvider.itemSwipeActionRightToLeft),
+                  ),
+                  child: Material(
```
Komodo5197: I don't really like how this turned out, visually.  There's not enough padding between the disc label and the track underneath once you scroll up.  You might need to adjust where you apply the padding to change how scrolling songs get clipped.  Also, the icons in the background for the swipes don't fit into the vertical space.  I'm also not sure about whether I like the rounded corners.  This is positioned directly below the appbar header, which is completely rectangular, and it feels like this should match.

lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -158,11 +254,66 @@ class _AlbumScreenContentState extends ConsumerState<AlbumScreenContent> {
...
+                    borderRadius: BorderRadius.circular(8),
+                    color: Theme.of(context).colorScheme.surface,
+                    clipBehavior: Clip.antiAlias,
+                    child: InkWell(
+                      onTap: () async => await GetIt.instance<QueueService>().startPlayback(
```
Komodo5197: I'm not sure I actually like treating the album like the songs in terms of interactions.  I feel like it might make more sense to treat this more like the album itself, where we have an icon only play/shuffle button on the right in front of the menu button, and then you can't click the actual header.  Having shuffle quickly available does seem like something worth doing, in contrast to songs which of course can't shuffle at all.  Maybe we keep the swipe gestures, but you can't rapidly add album discs to a queue in the same way you might want to with songs, so it might not be as needed.

lib/models/finamp_models.dart
```diff
@@ -1819,6 +1829,19 @@ class QueueItemSource {
...
   });
 
+  factory QueueItemSource.fromPlayableItem(
+    PlayableItem playableItem, {
+    QueueItemSourceType? type,
+    QueueItemSourceNameType? nameType,
+  }) {
+    switch (playableItem) {
+      case AlbumDisc():
+        return QueueItemSource.fromBaseItem(playableItem.parent, type: type, nameType: nameType);
```
Komodo5197, line 1844: Do we want to add a new queue source type for album discs?  Or would it be better to not get that granular?

lib/menus/components/playbackActions/playback_actions.dart
HeroBrine1st: We have changed that to PlayableItem since then, so it has clear semantics

lib/components/AlbumScreen/album_screen_content.dart
Chaphasilor: It probably is, I'll try to extract as much as possible later tonight

lib/components/AlbumScreen/album_screen_content.dart
HeroBrine1st: Yeah, this is somewhat non-symmetric. I would understand if background was transparent but it isn't 

image

Rounded corners were initially copied from tracks so that visually they are the same, though I can add nice little play/shuffle buttons there and remove InkWell. I imagined it again at that's not that bad as I initially thought. That's also response to the next comment.

I'm not sure we need swipes here at all. Consider the case of album with singular discs:

- Go to first album, press hamburger, "Add to next up"
- Go to second album, press hamburger, "Add to next up"
- Repeat until satisfied

So swipes for discs look out of place, there's no other place you swipe to add multiple tracks at once. I also thought about adding swipes but in my head it looked like whole disc including all its tracks is swiped, so user understands what is being added, but it seems to be very cumbersome (how to share state unidirectionally? only by manually rewiring it).

Chaphasilor: Hmm, I specifically adjusted to paddings with the clipping in mind. I would like the header to be closer to the tracks below than the tracks above, and didn't want a lot of padding to begin with.  
Also, for me the swipe background icon on the disc header is already smaller than on the track tile, do you think it's still too big?

I don't really see the point about the appbar header. The header is the old design, so it's not a big deal anyway, and the disc header should blend in with the background unless tapped or swiped, so that also shouldn't matter?

lib/models/finamp_models.dart
HeroBrine1st: I thought about that too and I decided that it is like user adds whole album to queue then removes extra tracks from queue. Also, if you know which album a track is added from, you can easily determine its disc.

lib/components/AlbumScreen/album_screen_content.dart
HeroBrine1st: It is smaller but not enough

<details>
image

image
</details>

lib/components/AlbumScreen/album_screen_content.dart
Chaphasilor: I'd like to keep the swipe gestures, since that allows quickly queuing up multiple albums using only specific discs. So instead of queuing up the entire album with a "normal" and an "acoustic" disc, you can choose to only queue up the normal disc, without going through another menu.

As for being able to shuffle, that's a fair point. Since the disc header label is quite short, we could most likely fit a play, shuffle, and menu icon on the right side, just like we have it for the album (just without labels). Then we could remove the tap interaction on the header in favor of the specific icons.

lib/models/finamp_models.dart
Chaphasilor: I thought about it, but since Jellyfin has no real concept of discs and we can't navigate to a "disc" screen, I decided against it for now

HeroBrine1st:
<details>
<summary>Screenshots</summary>
image

image
</details>

Komodo5197:
Looking at the screenshots, it might also make sense for us to recolor the disc header or something.  The disc headers are way harder to pick out from the songs when scrolling due to the fact that there's no longer a big open space on the right of the header which visually stood out from all the song icons.

lib/components/AlbumScreen/album_screen_content.dart
Chaphasilor: Interesting, it's not clipping on my phone. I guess it's related to font size / display scaling...

Chaphasilor:
Thanks @HeroBrine1st, those buttons look nice!  
Re: header color, I played around with using `primaryContainer` last night, but that was a bit too bold. But I'm also not convinced that visibility of the header is an issue, it looks quite distinct to me.

lib/components/AlbumScreen/album_screen_content.dart
Chaphasilor: Also, we're now literally declaring discs as a "PlayableItem" in the code, so I see no problem with adding the swipe interaction here.  
I'd add it to all items, but the main issue is that on the library tabs we can't have swipe gestures since it clashes with swiping between tabs. But I do think adding this universally would be a good idea.

lib/menus/components/menu_button.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class MenuButton extends ConsumerWidget {
```
Chaphasilor: I'm not happy with calling this `MenuButton`. It has nothing to do with menus. I'd rather call it `IconButton` or `SimpleUnlabledButton`, akin to the existing `SimpleButton`.  
In general there are very few cases where we should only have an icon without a visible text label.

HeroBrine1st: Oops, my undersight. I routinely removed "Overflow" from overflow menu without thinking much as I was sure it is specialised class :-) I'll rename that tomorrow (I also stopped rebasing due to branch not being isolated to me alone so you can safely do the swipe code extraction)

Chaphasilor:
Anything related to disc header swipe gestures should be resolved now. I'm also fine with the going back to the square header without any margins.  
So from my side the new class name and the merge conflicts are the only remaining blocks to getting this merged :)

Komodo5197:
It looks like the bottom of the text in the dismissable is still being cut off  Also, I still think we need more padding on the bottom of the header.  When in the list the top and bottom padding are the same, so the header is basically track sized, but when scrolling the bottom padding gets way smaller even though the top padding maintains its size.

image
image
image

HeroBrine1st:
I agree it looks better symmetric. For example, reverting everything back leads to this:

<details>
<summary>Screenshot and patch</summary>
image

```diff
diff --git a/lib/components/AlbumScreen/album_screen_content.dart b/lib/components/AlbumScreen/album_screen_content.dart
--- a/lib/components/AlbumScreen/album_screen_content.dart	(revision b5eddd17467dd5587852c55033360fdee418c133)
+++ b/lib/components/AlbumScreen/album_screen_content.dart	(date 1755595624765)
@@ -166,7 +166,6 @@
           for (var childrenOfThisDisc in childrenPerDisc)
             SliverStickyHeader(
               header: Container(
-                padding: const EdgeInsets.only(left: 8.0, right: 8.0, top: 8.0, bottom: 0.0),
                 color: Theme.of(context).colorScheme.surface,
                 child: Dismissible(
                   key: Key("${childrenOfThisDisc[0].id}-${childrenOfThisDisc[0].parentIndexNumber}"),
@@ -190,14 +189,16 @@
                     context: context,
                     direction: DismissDirection.startToEnd,
                     action: ref.watch(finampSettingsProvider.itemSwipeActionLeftToRight),
+                    iconSize: 40.0,
                   ),
                   secondaryBackground: buildSwipeActionBackground(
                     context: context,
                     direction: DismissDirection.endToStart,
                     action: ref.watch(finampSettingsProvider.itemSwipeActionRightToLeft),
+                    iconSize: 40.0,
                   ),
                   child: Padding(
-                    padding: const EdgeInsets.symmetric(horizontal: 10.0, vertical: 0.0),
+                    padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 16.0),
                     child: Row(
                       crossAxisAlignment: CrossAxisAlignment.center,
                       children: [
```

</details>

Which is now too big (at least compared to what we currently have) however we can do something like that:

<details>
<summary>Screenshot and patch</summary>
image


```diff
diff --git a/lib/components/AlbumScreen/album_screen_content.dart b/lib/components/AlbumScreen/album_screen_content.dart
--- a/lib/components/AlbumScreen/album_screen_content.dart	(revision 5f0d01d7f28182fb9d812f3395d53a0b3b507910)
+++ b/lib/components/AlbumScreen/album_screen_content.dart	(date 1755597058648)
@@ -166,7 +166,6 @@
           for (var childrenOfThisDisc in childrenPerDisc)
             SliverStickyHeader(
               header: Container(
-                padding: const EdgeInsets.only(left: 8.0, right: 8.0, top: 8.0, bottom: 0.0),
                 color: Theme.of(context).colorScheme.surface,
                 child: Dismissible(
                   key: Key("${childrenOfThisDisc[0].id}-${childrenOfThisDisc[0].parentIndexNumber}"),
@@ -197,7 +196,7 @@
                     action: ref.watch(finampSettingsProvider.itemSwipeActionRightToLeft),
                   ),
                   child: Padding(
-                    padding: const EdgeInsets.symmetric(horizontal: 10.0, vertical: 0.0),
+                    padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
                     child: Row(
                       crossAxisAlignment: CrossAxisAlignment.center,
                       children: [

```

I also took the opportunity here to revert the change from 16 dp padding to 8+10=18 dp (I think someone with good sight memory will get crazy so better not to do "small" changes).

</details>

Also I though pushing git merge will push all the commits from base branch but github filtered them out? That's probably my first time I resolve conflicts using `git merge` not `git rebase` (which changes the PR base so that commits from there are naturally filtered out)

Chaphasilor:
Yes GitHub only shows only commits that aren't in the target branch.  
I can't really reproduce the issue with the text being cut off, is this happening on Windows?

Either way, I've added the paddings as suggested, and now use a spacer element to create visual grouping between the header and the disc tracks without weirdly offsetting the header when scrolled. That's what I originally intended to do anyway...  
I hope this is acceptable now, sorry for delaying this...

HeroBrine1st:
Yeah, this is better

<details>
<summary>Screenshot as of 041f09d</summary>
image
</details>

I'm not sure if I correctly resolved conflicts. It looks like this branch conflicts only with d65b19750956ac31c1e5d3377ba9bd821f5f53aa, which I manually reapplied to conflicting file on conflict resolution. It at least builds and works. Also updated flake.lock file for new Dart SDK.

I also got my settings reset after merge, probably due to d65b19750956ac31c1e5d3377ba9bd821f5f53aa (?)

Chaphasilor:
@HeroBrine1st your settings should definitely not reset from https://github.com/jmshrv/finamp/commit/d65b19750956ac31c1e5d3377ba9bd821f5f53aa, since there were no changes to Hive.  
But I also noticed the settings resetting, and I suspect https://github.com/jmshrv/finamp/commit/0220d6f20a672b4480970ca96a4aa45418e4f19e to be the cause. @Komodo5197 in what way was the old location incorrect? I'm fine with updating the path, but then we should probably migrate the old database to the new location instead of creating a new database.

I guess the hive/settings issue can be addressed separately though. This looks good to merge, I just added a long-press gesture to open the menu for consistency with the track list tiles :)  
Thanks for adding this and being patient!

Komodo5197:
This change was made in response to an issue where some Linux users were having the app fail to start when the documents directory could not be located, even though we don't actually use that directory.  This was due to hive using it to set the default box location, even though we always supply the directory ourselves when opening a box.  This change shouldn't have affected the database locations at all due to our manually supplied paths.  I also can't reproduce the settings reset, so more data would be appreciated.

PR #1358, opened 2025-08-29
Report Transcoding Status to Server

### LINKED ISSUE #1048: Transcode not displayed in dashboard
In Finamp settings, when transcoding is enabled, it is not reported in the dashboard, and Jellyfin indicates that the media is played without modifications, looking at the logs the file is indeed transcoded

I had opened an issue at Jellyfin but they said it was client side

Chaphasilor:
## Changes

This change makes the server finally show that we're transcoding. The only real change needed was choosing a random ID and supplying that for the stream URL and to the playback reporting endpoint. All other changes are related to automatic server-side transcoding, which isn't feasible for Finamp since it requires *1 pre-emptive request per track* before the playback URL can be created.

Also bumped the Kotlin version due to a Flutter warning during build.  
And there's a bunch of house-keeping related to Jellyfin API types.

## Todo before merging

- [x] Extended CONTRIBUTING.md with information about how transcoding status works on the server, and why we do it this way
- [x] Test this on 10.11 RC

## Related Issues

- closes #1048


Chaphasilor:
I've tested this on both 10.10 and 10.11-RC5 and it seems to work without issues. The reasoning for these changes is also laid out in `CONTRIBUTING.md`.  
I haven't tested this on other platforms but don't expect and differences there either.  
If you are fine with these changes, feel free to merge @Komodo5197 :)  
(merge commit would be ideal to preserve the original attempts for reference)

Chaphasilor:
@Komodo5197 any objections to merging this? :)

lib/services/queue_service.dart
```diff
@@ -1181,11 +1186,14 @@ class QueueService {
       if (queueItem.item.extras!["isOffline"] as bool) {
         return Future.error("Offline mode enabled but downloaded track not found.");
       } else {
-        if (queueItem.item.extras!["shouldTranscode"] == true) {
-          return HlsAudioSource(await _trackUri(queueItem.item), tag: queueItem);
-        } else {
-          return AudioSource.uri(await _trackUri(queueItem.item), tag: queueItem);
-        }
+        final trackUri = await _trackUri(queueItem.item);
+        print("trackUri: $trackUri");
```
Komodo5197: You left a print statement in.

lib/services/queue_service.dart
```diff
@@ -1156,6 +1159,8 @@ class QueueService {
       artUri: artUri,
       title: item.name ?? "unknown",
       extras: {
+        "playSessionId": uuid
```
Komodo5197: Move the comment so that we don't have a line wrap between uuid and .v4().

Chaphasilor:
thanks for the review :)

Issue #1336, opened 2025-08-15
Artist Page Tracks Section only starts queue with visible


Tiefseetauchner:
### Are you using the newest Version?

- [x] Yes, the issue happens on the newest version
- [x] No, there aren't any open issues about this. I checked!

### Type of Report

- [ ] The app is unusable
- [x] Something isn't quite right
- [ ] Some functionality isn't accessible to me
- [ ] Accessibility
- [ ] None of the above

### The exact version of the app

0.9.19 (beta)

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [x] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Description & Steps to Reproduce

- Open artist page
- Click on Track in displayed Tracks
- Only 5 tracks are added to queue

### Logs

<details>

  <summary>Logs</summary>
  

[finamp-logs.txt](https://github.com/user-attachments/files/21790054/finamp-logs.txt)

  
</details> 

### Additional Information

I think it should play all (or at least a larger amount of) songs instead of just 5?

According to lukas this was on purpose, I think if we implement it that only on starting the playback does it load all tracks, it should be fine performance wise and I can listen to my top tracks and not just 5 of them :3

lukaslindnermusic:
Yes, that was on purpose to improve performance.

But it aligns well with my upcoming PR, where I will 
- bring Chaps PlaybackActionRow to the screens, replacing the current play/shuffle buttons.
- add a "See All" to the Top Tracks that opens a list with all the tracks in the selected sorting (Most Played, Favorites, etc.), similar to what I did on the Genre Screen.

The ActionRow already only loads the tracks once a user wants to play them, so I will port this to the top tracks.
The only consideration is the following:

Currently, top tracks distinctly combines two limited queries: One where the artist is album artist, one where it is performing artist. This is to catch all tracks where the artist is either one or both.

For the list, this combination is more tricky because we are reusing the MusicScreen.
Chap said it would be okay to only show the performing artist tracks, because for album artist tracks you would browse by album anyway.
I would've implemented a switch similar to the artist list.

[...]

Issue #982, opened 2024-12-22
Suggestion: All Tracks Listed under Artist


ryanvanmass:
It would be awesome when you are viewing a artist to to see a list of all tracks in addition to Top Tracks and Albums

ravtakhar:
is there a way to pull the most popular tracks as well? A view like Plexamp would be awesome!

lukaslindnermusic:
In [redesign beta 0.9.18](https://github.com/jmshrv/finamp/releases/tag/0.9.18-beta) we already added more "curation types" for the top 5 tracks (like favorites, last played, random, date added..), but I'm planning to adding a "See All" button to that section (similar to how its done on the new genre screen) soon, to browse a list of all tracks by a certain artist / album artist.

Chaphasilor:
@lukaslindnermusic any timeline for adding that "see all" button? should be a straight-forward integration of your existing solution for the genre screen?

Issue #818, opened 2024-07-22
[Feature Request] Save loop mode: sequential/shuffle


yunyuyuan:
I'm using the beta version `0.9.8`.

Currently, the default loop mode is set to sequential play. Users can manually switch to shuffle play, but this setting is not saved and must be set each time. This is inconvenient for users who prefer a shuffle loop mode.

I would like finamp to save the loop mode (sequential, shuffle) for each playlist. This way, users don't have to manually switch to their preferred loop mode every time they start play a list. 

Alternatively, having a global setting to set the default loop mode would also be helpful.

Thank you for considering this feature request!

HeroBrine1st:
Spotify has even better approach. Currently, Finamp has two buttons that do *almost* the same function: start playing playlist. Spotify, however, has toggle for shuffle and dedicated button to play (see screenshots below). This way there is only one "primary" button.

<details>
  <summary>Images</summary>
On mobile:

![image](https://github.com/user-attachments/assets/2e63ee5b-be53-4061-85bb-de9c73115693)

On PC:

![image](https://github.com/user-attachments/assets/849fee56-c0ef-4efb-b9a7-0d307d37a592)

![image](https://github.com/user-attachments/assets/cc084ada-e6d0-4f31-8f4b-c9bcfe37f9d7)

</details>

Chaphasilor:
Yeah these are both valid suggestions. It should be fairly easy to implement too.  
Just to be sure, you want regular shuffle, but with the first track being the selected one? As opposed to starting at a certain track and then only shuffling the *remaining* tracks.

yunyuyuan:
@Chaphasilor Yes,my current approach is to click on a track in the playlist to start playing, and then switch to shuffle mode.

BloodyIron:
Isn't this a duplicate of #797 ?

lukaslindnermusic:
Hmm...
This is fancy and all, but I'm honestly not sure if this makes it that much easier compared to if you just have all Playback Options directly available with just one tap or a swipe+tap, with the most recently used option presented to you by default.
And you would also not just have one play or shuffle button, but you can quickly add to queue or next up as well (which for example is my personal preference most of the time).
I think it is way more logical if each button has a clear and consistent function, while with this approach, the switch changes what the Play button does.

[...]

Issue #609, opened 2024-02-28
0.9.3 BETA: Playing Artist Plays Albums Alphabetically instead of Chronologically


9p4:
When pressing the "play" button on an artist, the expected behaviour is to play the albums chronologically. Instead the albums are queued up alphabetically.

Chaphasilor:
Any ideas how this should be handled? Do you want to change the default behavior? Have an option in settings to change between the two behaviors? Have an additional playback option? Have a way to sort artist albums and always play in that order?

9p4:
I think that the previous behaviour ought to be kept: play the albums chronologically. That is what I expected (and what happens in other music-playing apps as well)

hoggatt:
I agree queuing in order should be the default. It would also be nice to be able to customize the artist page a bit. For example, maybe I just want to see the albums and not the top songs. Also, I usually want the albums ordered from newest to oldest, so allowing this to be a setting would be nice too.

Tiefseetauchner:
The issue is the way Jellyfin sorts the songs grouped by album. The real question here is whether we want to locally sort the already sorted songs, or whether we should investigate Jellyfins options further. I don't know what Jellyfin can do, but I know that performance is an issue for me on my Fairphone 3 (given, an old and even back then slow phone)

Chaphasilor:
I'd like to rely on Jellyfin for sorting as much as possible. Sadly, the web version doesn't sort chronologically either (probably due to similar problems), and Jellyfin-Vue only queues a single track.

I guess the only way to properly achieve this is by loading all albums, sorted chronologically, and then loading the tracks for each album afterwards

Tiefseetauchner:
Good luck! If I may provide a bit of ideas:

Load the albums chronologically and then grab each albums songs until you reached a threshold (~~last I remember, which tbf it's been a while, this is also used on Genres~~ scrap that it isn't anymore) so to not load a thousand songs into the queue at once.

I'm literally walking through a forest, so my memory may be off, but I believe I put in a limit in the query, I'd use that. Maybe even grab the song count of the album in the initial query and fire the queries in a future.all or whatever it was called, if it's not possible or easy to append to the queue as the album lists come in. If it is, do just add them sequentially, probably better than a hundred simultaneous queries.

Anyways, just wanted to distract myself from this weird thing called *nature* and *trees* (ew), probably do not have to tell ya all that.

PR #1360, opened 2025-08-30
[Redesign] PlaybackActionRow instead of Buttons

lukaslindnermusic:
## Changes

This PR will bring @Chaphasilor s `PlaybackActionRow` to all dedicated screens (Album, Playlist, Artist, Genre), replacing the old buttons. The initial page of the `PlaybackActionRow` will depend on the ~~current Queue Order~~ last used queue destination.
I will also add a "See All" to the Artist Top Tracks.
I will also fix some bugs.

## Todo before merging

- [x] Add `PlaybackActionRow` to Album, Playlist and Artist Screens
- [x] Add option to remember last used page of `PlaybackActionRow`
- [x] Add `PlaybackActionRow` to GenreScreen
- [x] Fix Playback Order of ArtistScreen Albums
- [x] Add Menu Button to Album/Playlist/Artist/GenreScreens to open long-press menu from within the screen

---
these todos will be handled later in a second PR:

- [ ]  "See All" Artist Top Tracks
- [ ] Make Top Tracks continue to play
- [ ] Introduce Sorting Options for the Artist Album List (and playback)
- [ ] add optional setting to restore/remember shuffle queue order when starting new queue by tapping on track in album/playlist/artist-top-tracks
- [ ] Fix Snackbar message (swipe action on album track) saying that the entire "album [was] added to next up" instead of track.

---
- [x] I added new Translations / Strings
- [x] I added new Settings
---

## Related Issues

- will close [[Redesign] Play Artist Albums Chronologically instead of Alphabetically](https://github.com/jmshrv/finamp/issues/609)

----
the second follow-up PR to this one will then close:
- will close [[Feature Request] Save loop mode: sequential/shuffle](https://github.com/jmshrv/finamp/issues/818) (if @Chaphasilor s [decision](https://github.com/jmshrv/finamp/issues/818#issuecomment-3154795542) there is final ;))
- will close [Suggestion: All Tracks Listed under Artist](https://github.com/jmshrv/finamp/issues/982) 
- will close [Artist Page Tracks Section only starts queue with visible](https://github.com/jmshrv/finamp/issues/1336)


Komodo5197:
So I have a couple thoughts:
- Queue restore always comes back as linear.  If you want to more reliably track what shuffle mode was last used, you may need a separate variable or more complex logic.  We do save whether the queue was shuffled or not, but the player will not reflect that.
- I'm not sure this actually addresses the Save Loop Mode issue?  The original post seems to be requesting the ability to shuffle when clicking a specific track, which seems strange to me, but I digress.  Any changes to the playback action row or the top album menu seem unrelated.
- We should maintain access to the album menu from the album screen.  It has options that are not accessible elsewhere, and forcing users to back out of the item to access them is silly.
- This is further exposing what I would consider an existing design flaw with the playback action row, namely that the two most important/most used buttons, those being play and shuffle, are on different pages.  It's possible other people have different usage patterns, but considering the previous design of the album menu, I doubt that.  Using the player shuffle mode or the last option selected is better than always starting with play, but is still not great.  Whether I want the newly created queue to be shuffled is fairly uncorrelated with whether the previous one is shuffled.  I would strongly prefer a design that maintains immediate access to both of those buttons without needing to frequently switch pages, and maybe also the shuffle albums button for artists/genres as well.

lukaslindnermusic:
> - I'm not sure this actually addresses the Save Loop Mode issue?  The original post seems to be requesting the ability to shuffle when clicking a specific track

Ah true, I forgot to add this one to the TODO section of this PR, but I will tackle it as well.

> - We should maintain access to the album menu from the album screen.  It has options that are not accessible elsewhere, and forcing users to back out of the item to access them is silly.

Are you refering to the "Add to Playlist" and "Add to Mix" options?
Where would you add these? Would you rather have a button to open the menu that normally opens with a long-press, or add those two options directly to the AlbumScreen?

> [...] the two most important/most used buttons, those being play and shuffle, are on different pages.  It's possible other people have different usage patterns, but considering the previous design of the album menu, I doubt that. 

Thats a valid point!
So, how about changing the Row-pages to `New Queue`, `Next Up` and `Append`? So that the different pages now are adressing the "destination" for the action?
Then we could do:
- New Queue: `Play`, `Shuffle` and `Shuffle Albums` (Artists/Genres only)
- Next Up: `Play next`, `Add to Next Up`, `Shuffle to Next Up`, `Shuffle Albums to Next Up` (Artist/Genre only)
- Append: `Add to queue`, `Shuffle to queue` and `Shuffle Albums to Queue` (Artists/Genres only)


This would improve (or rather persist) the access to start both linear and shuffle play, while it also can remember the last used page, which imo would be nice as I most of the time use Next Up.

(The "Shuffle Albums" for Genres is a TODO I did not explicitly mention in the PR description but (together with the shuffle-limit) is the reason I've split it to a separate checklist-item. The Limit-Count we apply to the Genre Shuffle could be shown to the right of the pagination buttons if it is smaller than the total count of genre-tracks. Genres would still hide the linear play options.)

Komodo5197:
- How would you handle shuffle when tapping a track?  I guess we could just add it to settings, and then if you turn it on and want linear play from a point, you just unshuffle once started in a similar way to how the user currently gets shuffle play.  And you can always play or shuffle easily from the start as well.  I've also noticed that if you want using initialIndex and a shuffled order to work on desktop, you'll need to update the index with the new track position while preshuffling, but it might all be fine on mobile?
- I'd lean more towards adding a button to bring up the menu myself, because this guarantees we always have all options if things change in the future, and I don't think either of those options are really that important, per se.  But I don't have particularly strong feelings on that.
- Using that page order makes sense to me.  If I primarily use the main play buttons, and you primarily use next up, that does seem to be a better axis to split the pages down.  Hopefully most other people also lean heavily into one of these queue management styles.  I think the biggest question of this is what to do with the next up tab - If we have the full play/shuffle/shuffle albums for both play next and add to next up, that gets us 6 options, so you probably need to split that up in some manner.  I note you only listed four - would you consider the two shuffle next options less important?  But as I use the next up queue relatively infrequently, I don't have many thoughts on this.

lukaslindnermusic:
> - How would you handle shuffle when tapping a track?  I guess we could just add it to settings, and then if you turn it on and want linear play from a point, you just unshuffle once started in a similar way to how the user currently gets shuffle play.

Afaik, it currently always starts a new linear queue when you have an existing queue and then tap on a different track - either on the same album/playlist or a different one.
I understood the issue like this: When starting a new queue by tapping a track, the user wants it to keep using shuffle order... but starting with the tapped track.
I would've just checked the setting first, then, if enabled, we check the current queue order. Then we proceed with the already existing logic of starting the new queue. Directly after the playback was started, when the setting is enabled and the previous playback order was shuffle, we just change the order back to shuffle again. This will basically simulate what the user is currently doing manually, and I think it is what they want and is the most elegant way of handling it.

> I've also noticed that if you want using initialIndex and a shuffled order to work on desktop, you'll need to update the index with the new track position while preshuffling, but it might all be fine on mobile?

Ah, with initialIndex, I was just referring to the initial page of the playbackActionRow. I don't think that any changes to the queue restore are necessary.
(I've also not answered to your first point of storing the current playbackorder in a variable as it resets on queue restore, as my suggestion changes the pagination of the actionrow from depending on it anyway😊 But of course, thanks for pointing it out, I missed that.)

> - [...] I don't think either of those options are really that important, per se.

But with stuff that is only accessible through the long press menu, you've meant these options, correct? Because I think, everything else is accessible through the dedicated Screens already?

> - [...] I note you only listed four - would you consider the two shuffle next options less important?

I just used all the options currently available.
There currently is no "Shuffle play next" or "Shuffle Albums play next", only a "play next", which leaves us with 2 up to max. 4 options for the "Next Up" (which I think is still okay from a horizontal space perspective).
I understand that these two missing options would do specific things currently not covered, but I also think that these things are highly special (especially the "Shuffle Albums next up" thats only available for Artists and Genres and when there already is a Next Up queue). I doubt that these functions would be frequently used, quite in opposite to "Play Next" (which I personally often use to add a very specific track, maybe soooometimes an album, to the queue that I quickly want to listen to RIGHT NOW but dont want to lose the remaining queue).
In the very special case where you have to Shuffle all Albums not only to an existing next up (remember, it will already be covered by the other button if next up is empty) but play just now, you always can just open up a new queue and then later restore the previous one.
Also, to quite a few people, all these different next up buttons might be a little bit overwhelming and confusing instead of encouraging them to listen to music - I personally was overwhelmed for a few seconds the first time I discovered the additional "Shuffle Albums" option, back then with the dropdown.

So here, I would vote for simplicity and straightforward UI instead of giving every little edge-case option.

lukaslindnermusic:
I would first go without these two special options.

If users at some point will create issues asking for it, we could add a dedicated "Play Next" page that only appears in addition to "Next Up" if there currently are other tracks in next up.
But I would wait with that.
We can always add it later.

lukaslindnermusic:
Ignore what I said earlier :P I gave it a try and split Next and Next Up.
I'm not quite sure yet if the names are clear enough, but "append to next up" is too long. 
Also, I made "Play next" the default and "Append to Next Up" only appear when there are already tracks in "Next Up", as I believe this makes more sense.
I also added a little shuffle icon badge to the playback action icons, as they now were all the same on each page.

Not yet implemented is the remembering of the last used page for an action, because this commit should be just a first sketch of what it could be, to compare it to the other playback action sorting, before deciding for one and adding new Hive IDs.

Chaphasilor:
I'll try to respond to as many points as I can here:

1.  I'd be interested in an improved queue restore, but I'm not sure if that works with the current just_audio limitations. Maybe we could make further modifications to the ShuffleOrder (we could store a reference to the instance before passing it to the audio source / player), but it'll probably be ugly? Also unrelated but now every restored queue is just called "Saved Queue", I'm not sure why we don't properly restore the queue source anymore...
2. We definitely need a menu button on the album screen. It's unfortunate that the current design had the menu next to the play buttons, because this means we need to find it a new home. The original redesign placed it in the top right, which probably makes sense for now, given the other options in the appbar.  
  FYI the missing options are indeed the mix and add to playlist options. But I'm sure more options are coming eventually.
3. While I agree that the separation of play and shuffle isn't ideal, the original reasoning was that dividing into play / next up / queue would have those 6 options mentioned, which wouldn't fit the concept. @lukaslindnermusic there actually are "Shuffle Next" and "Shuffle Albums Next", but those only show up if you have tracks in the Next Up queue.  
  I think a good solution would be to add a dedicated page for the 3 "Play Next" options, and conditionally hide/show that page (or disable it). Then we have 3-4 pages with 2-3 options each. We can fall back to the Next Up page if Play Next is disabled.  
  For what it's worth I also prefer Next Up over replacing the queue, I rarely start new queues but frequently add new tracks midway through.
  - Also just saw your last message, guess you already handled it. But please call the tabs "Play Next" and "Append Next" (or "Append to Next Up") for clarity. Maybe also "Queue" instead of "Append"? I know there's not a lot of space, so the terminology probably won't be perfect. We could also consider aligning the tabs to the left and letting them overflow on the right, with horizontal scrolling to bring more tabs into view. Not idea, but maybe clearer + we remember the last used tab anyway.
4. I agree about starting the queue and then enabling shuffle. I think setting the order to shuffled when starting would not allow us to keep a specific track as the first, but maybe that could be added instead of waiting for the queue to start and then updating the order? Or at least handle it directly in queue_service and not outside.
5. Can we reduce the height of the PlaybackActionsRow please? I think we can reduce the vertical padding both of the row and of the icons.

Thanks for doing this!

Komodo5197:
1. Sorry, with initialIndex I was referring to _replaceWholeQueue in the queue service.  We should be able to start with a specific song when shuffling without needing a multi-step process.
2. The restored queues always being called 'Saved Queue' is because that message uses the per-track source, not the initial source, and we don't actually save that.  We previously applied that source to every track when restoring, but that was inaccurate and messed up the currently playing album highlighting.
3. If we're trying to reduce the vertical space usage, which I do agree with, it might be worth trying to make variants of the actions which have the icons beside the label instead of above.  I'm not 100% sure this would fit well on genres/artists, but albums/playlists will only ever have two options per page when these rows were designed for up to four.
4. I notice you save page state when an option is selected.  It might make more sense, and be a bit simpler, to just save on page change?

lukaslindnermusic:
> 1. Sorry, with initialIndex I was referring to _replaceWholeQueue in the queue service.  We should be able to start with a specific song when shuffling without needing a multi-step process.

Ah, I will look into this, thanks!

> 3. If we're trying to reduce the vertical space usage, which I do agree with, it might be worth trying to make variants of the actions which have the icons beside the label instead of above.  I'm not 100% sure this would fit well on genres/artists, but albums/playlists will only ever have two options per page when these rows were designed for up to four.

I always thought of that, but I personally don't like it if its different from albums/playlists to artists/gernes (because yes, I think it would not work with 3+ buttons, too), and also different to the long-press menu. 

> 4. I notice you save page state when an option is selected.  It might make more sense, and be a bit simpler, to just save on page change?

Yeah I also thought of that, but decided to go with storing the actual last used page as I thought this might make more sense. But after using my test build for a while, I agree with you... it would be better to save every page change. I will switch to that 😊👍

lukaslindnermusic:
@Komodo5197 I implemented "Shuffle some Albums" for Genres and added the PlaybackActionRow to genres as well (my changes are not pushed yet). This all also applies the track Limit of default 250.

That's when I noticed that in the queueService, `addNext` takes a significant amount longer than `addToNextUp`.
`addToNextUp` doesn't has great performance either, but `addNext` is worse.
Best performance is with either replacing the entire queue or appending to the queue.
I think this could be because we prepend each track one by one with `await _queueAudioSource.insert(`, and this has to shift the entire queue by one each time, and the number that needs to get shifted obviously is lower the less tracks are queued after the insertion index.

Thats when I noticed that just_audio also has a `insertAll` which takes a list instead of a single `AudioSource`.
It says "(Untested)" in the comment, but it says the same for the `insert` method we are already using.

So I was wondering... is there any reason that I'm not aware of why we don't use `insertAll`? 
If not, I would try to switch to it as this hopefully improves performance with adding a lot of tracks at the same time.

Edit: So basically, the default "Shuffle All" limit was fine because it always replaced the entire queue, but insertion is way more intensive and therefore more likely to crash the app, and therefore the default "Shuffle All" limit is too high for our current implementation of adding to the queue/next up.

If there is a reason to not use `insertAll`, (but also if there is none and we use it), would you suggest to split the `AudioSource` list into smaller chunks inside of the methods in `queue_service.dart` and then adding the other chunks with 1-2 second gaps inbetween? 
I know we already discussed this in a similar manner and you mentioned concerns because of handling shuffeling, but with this insertion, we are not requesting additional BaseItemDtos later that should not contain duplicates, but we already have all our BaseItemDtos there, pre-shuffled. The performance-bottlenecks this time are not the api calls but the just_audio insertions. So I think in this case the batch-by-batch-insertion could work?

I guess, the easy but also imo lazy and last-resort alternative would be to just introduce a way smaller limit for queue insertions.

lukaslindnermusic:
Update: I just tried to use `insertAll`. In the simulator, the UI stutters with 250 tracks, but on my iPhone, it works perfectly fine and smooth 👍

Chaphasilor:
I see no problem with batching, since we already have all the BaseItemDtos that we want to insert. If I chose `insert` over `insertAll` it could be related to our custom ShuffleOrder not playing well with it (try inserting multiple items when there are already tracks in Next Up and shuffle is active?). But if there are no adverse effects it's probably fine.

lukaslindnermusic:
> I see no problem with batching, since we already have all the BaseItemDtos that we want to insert. If I chose `insert` over `insertAll` it could be related to our custom ShuffleOrder not playing well with it (try inserting multiple items when there are already tracks in Next Up and shuffle is active?). But if there are no adverse effects it's probably fine.

I tried batching, but `insertAll` handled it perfectly already (as mentioned, not on Simulator, but with a real device it worked perfectly).
I just compared the behavior with Shuffle from my build to the current TestFlight version, and there is no difference. The ShuffleOrder is not affecting NextUp in both builds (which I think makes sense, as the way I see NextUp is as a sort of manual pre-queue).

Chaphasilor:
I'm a little worried that it works fine without batching on a powerful iPhone, but not as well on weaker hardware. But since the queue stuff is worse on iOS anyway it's probably fine, or at least not worse than replacing the queue.  
And while you're right that semantically Next Up is a separate queue, it's not really separate in the actual `ConcatenatingAudioSource`. That's why we need the custom shuffle order, to avoid shuffling anything in Next Up. So the system is a bit fragile.

lukaslindnermusic:
@Chaphasilor what smartphone do you have? Regardless of how powerful your smartphone is: Maybe you could do a quick Android test with the current PR commit? Just try shuffling a big genre to NextUp. Thanks <3

Chaphasilor:
I've got an ASUS Zenfone 10, with a Qualcomm Snapdragon 8 Gen 2, so pretty performant. But that's not really important, since there might be people with much older and slower phones, with less RAM. But I'm not *that* worried, as I said inserting shouldn't be more problematic than regular playback.

Regarding functionality, seems to work well without any substantial load times or performance hiccups, although the "Shuffle Some Albums Next" option only added like 30 tracks? The "Shuffle some Next" added much more (I guess the configured limit). Is that intended behavior?

lukaslindnermusic:
> although the "Shuffle Some Albums Next" option only added like 30 tracks? The "Shuffle some Next" added much more (I guess the configured limit). Is that intended behavior?

well, sort of.
The track limit (default 250) also applies to `Shuffle some Albums`. Basically, we want to avoid loading all the genre tracks, so we would have to pick random tracks with the applied limit. But as here we want **entire albums**, we can't just fetch random tracks (because then we only would get fragmented albums after grouping the result-tracks by album). We first have to fetch some albums, and then the tracks for those albums. But as there potentially also might be thousands of albums, I thought it would be good to also apply an `albumLimit`, which I calculate from the trackLimit by assuming 15 tracks per album (keeping the returned tracks tending to the lower side, as this is better than the other way around). So the albumLimit will get calculated by totalTrackLimit devided by 15. 

In the next step, we fetch all tracks of these albums... I noticed the api parameter "albumIds" which allows us to fetch the tracks of multiple albums with just one query... nice ;) 

With the result, we check if we have exceeded the totalTrackLimit, and if yes, we remove one album and check again, until we are below the limit of default 250.

BUT: If you now only have albums with only 1 or 2 tracks, there of course only will be around 30 tracks (as the default of trackLimit 250 will result in roughly 16-17 albums that get fetched in step 1). 

So, conclusion: My estimate with 15 tracks per album was way too high, and I simply should decrease it so that even with albums of around 1-2 tracks, you will get a decent queue with around 100 tracks.... (and of course max. 250).
Maybe 7 is a sweet spot? Resulting in roughly 35-36 albums with the default trackLimit 😊

Komodo5197:
Why not just request up to 250/all albums, and then just use the childCount to determine how many of those we fetch the tracks from?

lukaslindnermusic:
because I think the childCount is not reliable.
we could however just apply the totalTrackLimit+1 (to see if the 250th track was the last of the last album) to the tracks-api-call, and then follow with my existing logic of comparing with the totalTracklimit and remove albums until below.

But then we still fetch 500 items from the api in a short amount of time... with a high probability of fetching more albums than necessary...  thats why I wanted to reduce the album-fetch.

Chaphasilor:
Hmm, so yeah I mostly have singles, so that checks out. I'd go with fetching 250 (or whatever the limit is) albums, and and then either going by the child count (is what way is this unreliable? too high? too low? unpredictable? in my testing the count seemed to be correct for albums, it might be that it's only incorrect for artists and genres?), or by fetching tracks from albums in chunks, and stopping + trimming once we surpass the limit.

lukaslindnermusic:
> Hmm, so yeah I mostly have singles, so that checks out. I'd go with fetching 250 (or whatever the limit is) albums, and and then either going by the child count (is what way is this unreliable? too high? too low? unpredictable? in my testing the count seemed to be correct for albums, it might be that it's only incorrect for artists and genres?), or by fetching tracks from albums in chunks, and stopping + trimming once we surpass the limit.

I am getting a 414 error (Request URI too long) in step 2 (fetching the album tracks) when I have too many albums shuffled in step 1 (for example 250).
I am trying to find the sweet spot and then I will add a fixed upper limit for that.

lukaslindnermusic:
Sorry for letting this in this unfinished state for so long. 
I was moving to another country for my master's, and also I was busy in general with my main work as film composer.

This is ready to merge now.

I know I know, there are still a lot of open todos, but they are functionally different from what I already did, and there is an upcoming beta release. 
So I decided to split this PR up in another PR (that I will open right after the new beta version drops) so that the already finished playback buttons are already in the new version.

lib/menus/components/playbackActions/playback_action_row.dart
```diff
@@ -49,7 +62,7 @@ class PlaybackActionRow extends ConsumerWidget {
                   ? PlaybackActionRowPage.playNext
                   : PlaybackActionRowPage.appendNext;
 
-              final pageMap = queueService.getQueue().nextUp.isEmpty
+              final pageMap = ref.read(QueueService.queueInfoStreamProvider).value?.nextUp.isEmpty ?? true
```
Komodo5197, line 65: Why don't we just directly read the queue here instead of the provider?

lib/menus/components/playbackActions/playback_action_row.dart
```diff
@@ -14,18 +15,30 @@ class PlaybackActionRow extends ConsumerWidget {
...
   final PageController controller;
-  final Map<String, Widget> playbackActionPages;
+  final PlayableItem item;
+  final bool popContext;
   final bool compactLayout;
+  final BaseItemDto? genreFilter;
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+    ref.watch(QueueService.queueInfoStreamProvider);
```
Komodo5197, line 32: I don't like watching a value without actually using it.  Could we pass a hasNextUp flag to getPlayBackActionPages or something to guarantee the caller is handling the dependency?

lib/components/AlbumScreen/album_screen_content.dart
```diff
@@ -269,7 +275,7 @@ class _AlbumScreenContentState extends ConsumerState<AlbumScreenContent> {
             forceAlbumArtists: (widget.parent.type == "Playlist" && playlistSortBy == SortBy.albumArtist),
           )
         else
-          SliverFillRemaining(child: Center(child: CircularProgressIndicator.adaptive())),
+          SliverFillRemaining(child: CircularProgressIndicator.adaptive()),
```
Komodo5197, line 278: Why'd you get rid of this for all the screens?  It makes the loading spinner look incorrect.

lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
```diff
@@ -44,104 +41,60 @@ class AlbumScreenContentFlexibleSpaceBar extends StatelessWidget {
   final void Function(BaseItemDto?)? updateGenreFilter;
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
     GetIt.instance<AudioServiceHelper>();
```
Komodo5197, line 45: This seems to be unused.

Chaphasilor:
alright, thanks for getting this in on short notice, and don't forget to move the remaining todos and related issues over to the new PR! 😁  
Ideally also edit this PR to remove the unrelated issues :)

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -150,7 +151,7 @@ class _ArtistScreenContentState extends ConsumerState<ArtistScreenContent> {
             // 125 + 116 is the total height of the widget we use as a
```
Komodo5197, line 151: This comment seems to be outdated.

lib/components/ArtistScreen/artist_screen_content_flexible_space_bar.dart
```diff
@@ -45,101 +43,66 @@ class ArtistScreenContentFlexibleSpaceBar extends StatelessWidget {
   final void Function(BaseItemDto?)? updateGenreFilter;
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
     GetIt.instance<AudioServiceHelper>();
```
Komodo5197, line 47: Unused.

lib/components/ArtistScreen/artist_screen_content_flexible_space_bar.dart
```diff
@@ -45,101 +43,66 @@ class ArtistScreenContentFlexibleSpaceBar extends StatelessWidget {
...
-            type: QueueItemSourceNameType.preTranslated,
-            pretranslatedName: parentItem.name ?? AppLocalizations.of(context)!.placeholderSource,
-          ),
-          id: parentItem.id,
-          item: parentItem,
-        ),
-        order: FinampPlaybackOrder.shuffled,
-      );
-    }
+    final lastUsedPlaybackActionRowPage = ref.read(finampSettingsProvider.lastUsedPlaybackActionRowPage);
```
Komodo5197, line 50: This logic seems to be duplicated across all the PlaybackActionRow usages - we should probably bake it into the widget.

lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
```diff
@@ -44,104 +41,60 @@ class AlbumScreenContentFlexibleSpaceBar extends StatelessWidget {
...
-                    crossAxisAlignment: WrapCrossAlignment.center,
+        bottom: false,
+        child: Stack(
+          fit: StackFit.expand,
+          clipBehavior: Clip.hardEdge,
+          children: [
+            Positioned(
+              left: 8,
+              right: 8,
+              bottom: MediaQuery.of(context).padding.bottom,
```
Komodo5197, line 67: Why are we using the mediaquery padding here?  This also applies to the artist screen.

lib/menus/artist_menu.dart
```diff
@@ -39,8 +43,17 @@ Future<void> showModalArtistMenu({
   (double, List<Widget>) getMenuProperties(BuildContext context) {
     final menuEntries = getMenuEntries(context);
     final stackHeight = ThemedBottomSheet.calculateStackHeight(context: context, menuEntries: menuEntries);
+    final ref = GetIt.instance<ProviderContainer>();
```
Komodo5197, line 46: Why are we getting ref here?  If you don't want reactivity, you can just read the settings directly.  Although as another comment mentioned, this initial page logic should be inside playbackactionrow.

lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -5,113 +5,225 @@ import 'package:finamp/l10n/app_localizations.dart';
...
-Map<String, Widget> getCommonPlaybackActionPages({required BuildContext context, required PlayableItem item}) {
+Map<String, Widget> getPlaybackActionPages({
+  required BuildContext context,
+  required PlayableItem item,
+  bool popContext = true,
+  bool compactLayout = false,
+  BaseItemDto? genreFilter,
+}) {
   final queueService = GetIt.instance<QueueService>();
+  final ref = GetIt.instance<ProviderContainer>();
```
Komodo5197, line 24: I'd prefer directly reading FinampSettingsHelper.finampSettings instead of using the global provider ref.  This applies across several widgets.

lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -5,113 +5,225 @@ import 'package:finamp/l10n/app_localizations.dart';
...
 
   final PlayableItem item;
+  final bool popContext;
+  final bool compactLayout;
+  final BaseItemDto? genreFilter;
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final queueService = GetIt.instance<QueueService>();
+    final preferNextUpPrepending = ref.read(finampSettingsProvider.preferNextUpPrepending);
```
Komodo5197, line 204: Prefer ref.watch inside widgets for reactivity.

lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -231,50 +414,88 @@ class ShuffleNextPlaybackAction extends ConsumerWidget {
...
+            source: QueueItemSource.fromPlayableItem(item, type: QueueItemSourceType.nextUpAlbum),
+            order: FinampPlaybackOrder.shuffled,
+          );
+
+          GlobalSnackbar.message(
+            (scaffold) => AppLocalizations.of(scaffold)!.confirmShuffleToNextUp,
+            isConfirmation: true,
+          );
+
+          if (popContext) {
```
Komodo5197, line 459: We should pop the context before loading the child items, not after.  This applies to all the actions.

lib/services/item_helper.dart
```diff
@@ -106,20 +126,123 @@ Future<List<BaseItemDto>?> loadChildTracksOffline({
...
+      // Add the tracks and decrease the total limit
+      newItems.addAll(playableAlbumTracks);
+      totalTrackLimit -= playableAlbumTracks.length;
+    }
+  } else {
+    // Online Mode
+    List<BaseItemDto>? genreAlbums =
+        await jellyfinApiHelper.getItems(
+          parentItem: finampUserHelper.currentUser?.currentView,
+          includeItemTypes: [BaseItemDtoType.album.idString].join(","),
```
Komodo5197, line 205: Why are we joining a list of one element?

lib/menus/components/playbackActions/playback_action_row.dart
Chaphasilor: That was from when I put the .watch at the wrong spot (in this callback), but I though it might be good to use the provider consistently? It's just much nicer for reactive UI, even if it's not specifically needed in this case.

lib/menus/components/playbackActions/playback_action_row.dart
Chaphasilor: I guess we could. Out of curiosity, why is this important?

lib/menus/components/playbackActions/playback_actions.dart
Chaphasilor: I guess this has the same motivation as https://github.com/jmshrv/finamp/pull/1360#discussion_r2404660926 ?  
I think it's easier to use the providers as much as possible instead of using both methods, depending on context.

lib/services/item_helper.dart
Chaphasilor: So that if we want to include additional item types we can just add the type. Probably a copy-paste relic, but I don't see an issue with this.

Chaphasilor:
Thanks for the review. I'll address the points in a bit, since I don't think @lukaslindnermusic has the capacity at the moment :)

lib/services/item_helper.dart
Komodo5197: I guess?  Doesn't seem likely for a field labeled genre albums, but I guess it doesn't really hurt.

Chaphasilor: I'd keep it then. I added this in a few placed when I got annoyed by the use of raw strings with the commas embedded in the string, and change it to this (that was for instances where multiple types were used).

lib/menus/components/playbackActions/playback_action_row.dart
Komodo5197: I mean, it's not doing much for reactivity if you read it instead of watching.  It also introduces additional complications.  Note that queueinfostreamprovider is an async provider, so you have all this handling for a missing value that we wouldn't need if we directly accessed the queue.  Also, async providers cannot initialize in one frame, so if this the first call to this provider it would read null l, and the unreactive read would never update with the true value.  I think the other usage is guaranteed to fire first, but that's still an uneeded complication.  In general, I think any use of .read() needs to be fairly thought out, which is why I'd prefer to avoid using it, or getting the provider container and introducing the temptation to use it, when not nessecarry.

lib/menus/components/playbackActions/playback_action_row.dart
Komodo5197: It just seems like bad practice.  Right now the function has a hidden dependency you just need to know about, and which could very easily be missed.  Using the value establishes a clear and unmissable link between the rebuild dependency and actual reason we need to rebuild in both directions.

lib/menus/components/playbackActions/playback_actions.dart
Komodo5197: Using the provider container in a widget context instead of a service one is probably a bad idea.  I think the idea of wanting to use the provider for everything might actually be correct though.  But instead of reading the container, we should either be passing in the needed setting values from a watch call in the widget, or taking a widgetref so we can watch the setting here in the function.

lib/menus/components/playbackActions/playback_actions.dart
lukaslindnermusic: Just out of curiosity: Why?

lukaslindnermusic: ah because of the await before? 
maybe thats just me, but if the menu closes right away but nothing plays for a while because the server is busy and or the request takes long because it returns a lot of tracks, this feels much more buggy/clumpsy than if you have to wait. But I agree, it isn't ideal either to wait... maybe we can add here and at other relevant places a little loading indicator, signaling the user that the server is still processing the request?

Chaphasilor: Ideally we'd show something like the restore state of the now playing bar. But I guess that requires deeper integration. So for now I'd probably close the menu as feedback that something happened, and then even if the user gets impatient and re-opens the menu it will take them longer to make a second request, so chances are playback starts before that.

Komodo5197: Yeah, loading the child tracks can take a second or two, so you need some form of immediate feedback when the button is pressed.  Maybe the menu closing before the player changes is odd, but having no idea if you even hit the button or if you misclicked is worse.  It also leaves an annoying loading gap where you just need to wait for the menu to close itself even if you do know it could take awhile.  We already have a notification for when the change actually goes through.  Maybe we could integrate an indicator into the player at some point, but I don't think that's a priority.  For all the actions except play/shuffle, you can't even see when they actually happen because they don't affect the current song.

lib/menus/components/playbackActions/playback_action_row.dart
Chaphasilor: Okay. And I guess this doesn't apply to "native" providers, since they have their own internal state that we can't expose in another way?

Komodo5197: I'm not entirely sure what you mean by "native" providers?  If you're talking about the underlying StreamProvider, I believe the delay is due to using futures because not all streams can be read synchronously.  We could theoretically make a queueinfostreamprovider based off a normal, non async Provider where we manually watch the stream and the initial queue value to work around this, if we really needed to.

Chaphasilor: I was more talking about the theme providers, favorite providers, etc., which are not based on any underlying class or stream (like QueueService, FinampSettings, FinampUser would be).

lib/components/AlbumScreen/album_screen_content.dart
Chaphasilor: from which other screens was this removed? I can only find this one instance, for the album/playlist screen...

lib/menus/components/playbackActions/playback_actions.dart
Chaphasilor: sounds good. things are *slowly* starting to come together in my head :D

lib/components/AlbumScreen/album_screen_content_flexible_space_bar.dart
Chaphasilor: just poor layout in general, probably for historic reasons. switched to using Align + Padding

Chaphasilor:
Addressed the review in 7f6aa4a891d4104f2ef6e6cc9dbfcd349ea104c7 :)

lib/menus/components/playbackActions/playback_action_row.dart
Komodo5197: Yeah, there's not any way around using read for those providers if you need the value outside a widget context.  You still need to be careful though, because all the hangups with read still apply.  I think a good example of this is the albumImageProvider, which usually returns the final result immediately, except there's one path which actually returns a temp value and then rebuilds later, which could easily trip you up if you aren't careful.  In a lot of these cases, it's actually a good idea to place all the logic into another provider so you can watch and rebuild, although of course that's not always feasible.

lib/components/AlbumScreen/album_screen_content.dart
Komodo5197: I could've sworn there was another screen where this same thing happened, but I guess I was wrong.  It looks like this is everything.

Komodo5197:
The fix commit looks good to me.  While reading over it, I did notice that the nextup actions seem to be doing pointless nonreactive visibility calculations that we could clean up, because the action row handles hiding them, but their presence or absence shouldn't affect functionality in any way, so you can remove those if you feel like it.

lib/components/AlbumScreen/album_screen_content.dart
Chaphasilor: Maybe you got confused by the album and playlist screen being the same screen even though it's different tabs? Either way, no worries, better save than sorry. That uncentered adaptive spinner was hideous 😅

Chaphasilor:
Good catch. While checking that out I also noticed that the track menu manually inlines the actions and therefore the `preferNextUpPrepending` setting wasn't considered yet, I've fixed that as well (ee13be87f2571b1aea0ff44e05a873ab58bf1461)

lib/components/AlbumScreen/album_screen_content.dart
lukaslindnermusic: I have no idea why I removed it. 🤷🏻‍♂️

maybe there was some reasoning behind it that I don't remember anymore, or it was just an accident.

PR #1369, opened 2025-09-05
[Redesign] Track list item rebuild tweaks.

### LINKED ISSUE #899: Queue does not resume after stop in media notification
Steps to reproduce:

1. Enable the stop button in Layout & Theme > Customization > enable here.
2. Play a song/playlist/album
3. Close the app
4. Stop queue with media notification
5. Press play in media notification

Excepted behavior after stop button is pressed is to either 

- remove the media notification when the stop button is pressed or
- attempt to resume last queue or restart last queue from the beginning when play button is pressed.

### LINKED ISSUE #1340: Timestamp in filename of save logs button

### Type of Feature

- [ ] Something new
- [ ] None of the above

### Description

The "save logs" button tries to save a file with the same name every time. It would be nice if it saved a different name every time (e.g. have a timestamp in the default filename) so I won't have to manually rename it or overwrite the old file every time.

### Reasoning

_No response_

### Implementation Notes

_No response_

### Additional Information

_No response_

### LINKED ISSUE #1345: MdbxError

### Type of Report

- [x] The app is unusable
- [ ] Something isn't quite right
- [ ] Accessibility
- [ ] None of the above

0.9.19

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

_No response_

Open the app. This screen appears. 

### Logs

[finamp-logs.txt](https://github.com/user-attachments/files/21871243/finamp-logs.txt)

Image

<details>

  <summary>Logs</summary>
  
</details> 

### Additional Information

Tried opening the app and I always get this screen that doesn't let me advance.

### LINKED ISSUE #1349: Sync song deletions.
Currently when a downloaded song is deleted in the upstream jellyfin server, the app gets a 404 error because it's trying to fetch deleted tracks.

But it will still show up in offline mode.

I think the app should sync deletions and remove deleted songs from offline storage.

### LINKED ISSUE #1364: [Beta] download stuck in enqueue forever

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

Lates beta

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

Hiby M300 DAP

On my hiby M300 when I try to download some albums or artist with a custom location set to the download folder on my microsd card download stay in enqueue  state and never download even if leave it overnight

[finamp-logs.txt](https://github.com/user-attachments/files/22091720/finamp-logs.txt)

 

### Logs

[finamp-logs.txt](https://github.com/user-attachments/files/22091798/finamp-logs.txt)

### Additional Information

[finamp-logs.txt](https://github.com/user-attachments/files/22091798/finamp-logs.txt)

### LINKED ISSUE #1387: Can Barely Scroll Albums Tab Because Hamburger Menu Keeps Opening

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.9.19 (beta)

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

Pixel 7a

The sensitivity for swiping to open the hamburger menu from the Albums tab is insane. I'm not a robot. Sometimes swiping up isn't _perfectly straight up with no deviation_...

### Logs

N/A

### Additional Information

I tried finding an Issue for this but didn't see one. Apologies if this is a duplicate.

---

Edit: Also: I don't want to disable all gestures. This is the only gesture that seems to cause issues.

Komodo5197:
An attempt to reduce the amount of unneeded widgets rebuilds occurring, primarily in the track list tiles.  This was mostly dealing with the excessive rebuilds when resizing the window on desktop, but I also caught a few other sources.  This should probably be tested to ensure something didn't get lost in the refactor.  Also:
- Changed overflow menu on track list tiles to depend on tile width instead of platform.  
- Changed the default value of the 'show text in grid view' value to false, but made text always be shown if no image is found for an item.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -2804,5 +2804,17 @@
   "syncingFailed": "Sync failed.  This download has likely been removed from the server, and should be deleted.",
   "@syncingFailed": {
     "description": "Message shown for downloads with the sync failed status, which is generally caused by the server-side item being removed."
+  },
+  "autoplayRestoredQueueTitle": "Automatically Play Restored Queues",
+  "@autoplayRestoredQueueTitle": {
+    "description": "Title for the setting that controls if restored queues automatically play."
+  },
+  "autoplayRestoredQueueSubtitle": "Queues restored automatically on startup or via the queue restore screen will begin playing automatically.  Queues which fail to restore all tracks will never autoplay, regardless of this setting.",
+  "@autoplayRestoredQueueSubtitle": {
+    "description": "Subtitle for the setting that controls if restored queues automatically play."
+  },
+  "missingBlurhashWarning": "Warning: The Jellyfin server appears to be misconfigured and is not calculating blurhashes.  This prevents Finamp from de-duplicating image downloads.  Please correct the server and run a downloads repair.",
+  "@missingBlurhashWarning": {
+    "description": "A warning message displayed on the downloads screen if the jellyfin server is missing blurhashes."
   }
 }
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -2243,6 +2243,10 @@
   "@emptyTopTracksList": {
     "description": "Message shown as a placeholder when the top tracks list for an artist is empty"
   },
+  "emptyAlbum": "No tracks.",
+  "@emptyAlbum": {
+    "description": "Message shown as a placeholder when an album or playlist is empty."
+  },
   "emptyFilteredListTitle": "No items found",
   "@emptyFilteredListTitle": {
     "description": "Message shown when a list of items is filtered and no items match the filter"
@@ -2772,5 +2776,17 @@
   "dragToReorder": "Drag to Reorder",
   "@dragToReorder": {
     "description": "Label shown on drag handles of track list  tiles in the queue list"
+  },
+  "addDownloadLocationsErrorPrompt": "Writing to selected directory failed.  You should modify permissions or choose another location.",
+  "@addDownloadLocationsErrorPrompt": {
+    "description": "Label shown on popup dialog when adding a download location when writing songs fails"
+  },
+  "androidImageErrorPrompt": "WARNING: We are unable to write images to the android Music folder due to OS restrictions.  You should choose another location.",
+  "@androidImageErrorPrompt": {
+    "description": "Label shown on popup dialog when adding a download location when writing songs succeeds but writing images fails, which can occur on android."
+  },
+  "addDownloadLocationsErrorButton": "Add anyway",
+  "@addDownloadLocationsErrorButton": {
+    "description": "Label shown on button for adding a download location we suspect to be buggy"
   }
 }
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -2788,5 +2788,21 @@
   "addDownloadLocationsErrorButton": "Add anyway",
   "@addDownloadLocationsErrorButton": {
     "description": "Label shown on button for adding a download location we suspect to be buggy"
+  },
+  "syncFailedWarningShort": "Warning: Failed to synchronize some downloads with server.",
+  "@syncFailedWarningShort": {
+    "description": "Simple warning message for when download sync fails due to items being deleted from the server."
+  },
+  "syncFailedWarningLong": "Some downloads failed to synchronize with the server.  The matching server item was likely deleted.  The failed download should be deleted to clean up the outdated files.",
+  "@syncFailedWarningLong": {
+    "description": "Expanded warning message for when download sync fails due to items being deleted from the server."
+  },
+  "openDownloads": "Open Downloads",
+  "@openDownloads": {
+    "description": "Label on a button to open the downloads screen after a sync fails."
+  },
+  "syncingFailed": "Sync failed.  This download has likely been removed from the server, and should be deleted.",
+  "@syncingFailed": {
+    "description": "Message shown for downloads with the sync failed status, which is generally caused by the server-side item being removed."
   }
 }
--- a/lib/services/finamp_logs_helper.dart
+++ b/lib/services/finamp_logs_helper.dart
@@ -92,7 +92,7 @@ class FinampLogsHelper {
   /// Write logs to a file and share the file
   Future<void> shareLogs() async {
     final tempDir = await getTemporaryDirectory();
-    final tempFile = File(path_helper.join(tempDir.path, "finamp-logs-${DateTime.now().millisecondsSinceEpoch}.txt"));
+    final tempFile = File(path_helper.join(tempDir.path, _logExportName));
     tempFile.createSync();
 
     tempFile.writeAsStringSync(await getFullLogs());
@@ -106,10 +106,13 @@ class FinampLogsHelper {
   /// Write logs to a file and save to user-picked directory
   Future<void> exportLogs() async {
     await FilePicker.platform.saveFile(
-      fileName: "finamp-logs-${DateTime.now().millisecondsSinceEpoch}.txt",
+      fileName: _logExportName,
       // initialDirectory is ignored on mobile
       initialDirectory: (await getApplicationDocumentsDirectory()).path,
       bytes: utf8.encode(await getFullLogs()),
     );
   }
+
+  String? get _logExportName =>
+      "finamp-logs-${DateTime.now().toIso8601String().replaceAll(RegExp(r'[/?<>:*|.\\"]'), "-")}.txt";
 }
--- a/lib/services/finamp_logs_helper.dart
+++ b/lib/services/finamp_logs_helper.dart
@@ -65,6 +65,7 @@ class FinampLogsHelper {
     final logMeta = await EnvironmentMetadata.create();
 
     // Prepend this metadata to the logs
+    fullLogsBuffer.writeln("=== METADATA ===");
     fullLogsBuffer.writeln(logMeta.pretty);
     fullLogsBuffer.writeln("=== LOGS ===");
 
@@ -91,7 +92,7 @@ class FinampLogsHelper {
   /// Write logs to a file and share the file
   Future<void> shareLogs() async {
     final tempDir = await getTemporaryDirectory();
-    final tempFile = File(path_helper.join(tempDir.path, "finamp-logs.txt"));
+    final tempFile = File(path_helper.join(tempDir.path, "finamp-logs-${DateTime.now().millisecondsSinceEpoch}.txt"));
     tempFile.createSync();
 
     tempFile.writeAsStringSync(await getFullLogs());
@@ -105,7 +106,7 @@ class FinampLogsHelper {
   /// Write logs to a file and save to user-picked directory
   Future<void> exportLogs() async {
     await FilePicker.platform.saveFile(
-      fileName: "finamp-logs.txt",
+      fileName: "finamp-logs-${DateTime.now().millisecondsSinceEpoch}.txt",
       // initialDirectory is ignored on mobile
       initialDirectory: (await getApplicationDocumentsDirectory()).path,
       bytes: utf8.encode(await getFullLogs()),
--- a/lib/services/queue_service.dart
+++ b/lib/services/queue_service.dart
@@ -11,8 +11,6 @@ import 'package:finamp/models/finamp_models.dart';
 import 'package:finamp/models/jellyfin_models.dart' as jellyfin_models;
 import 'package:finamp/services/album_image_provider.dart';
 import 'package:finamp/services/current_album_image_provider.dart';
-import 'package:finamp/services/current_track_metadata_provider.dart';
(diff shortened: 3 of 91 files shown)


Komodo5197:
Also, I ended up refactoring the lyrics screen, so that should probably be tested too, because I have minimal test cases for it.

lib/services/theme_provider.dart
```diff
@@ -39,13 +39,18 @@ class PlayerScreenTheme extends StatelessWidget {
           if (item == null) {
             return null;
           }
-          return ThemeInfo(item, largeThemeImage: true, useIsolate: false);
+          // Setting useIsolate to true provides negligible speedup for player images and induces lag.
```
Chaphasilor: the comment should say "false"

lib/services/finamp_logs_helper.dart
```diff
@@ -105,7 +106,7 @@ class FinampLogsHelper {
   /// Write logs to a file and save to user-picked directory
   Future<void> exportLogs() async {
     await FilePicker.platform.saveFile(
-      fileName: "finamp-logs.txt",
+      fileName: "finamp-logs-${DateTime.now().millisecondsSinceEpoch}.txt",
```
Chaphasilor: how about we use an ISO-8601 string for this instead? that's more human-readable and just as sortable.

Chaphasilor (commented):
Thanks, and sorry for not taking a look earlier. That's a lot of neat fixes!

Re: overflow based on tile width, maybe it would make sense to keep the button on all desktop platforms? Or were you annoyed by the button taking up space on your desktop?  
Either way, for me on windows the menu button is never shown, although the clamp should allow larger widths than the overflow threshold. Not sure what's going on there?
  
That OverflowBuilder seems definitely useful, also for the updated action bar for the home screen changes. Same for the CachedBuilder.

Is that `MediaQuery.xyzOf()` pattern more efficient than the `MediaQuery.of().xyz` pattern? Or do you just prefer the code style?

The non-active lyric lines now use a smaller font than the active line, making the layout jump / wrapping changes worse than before. Did you decrease the font size on purpose?
image
image

For word-by-word lyrics, the non-active word highlighting color is also incorrect in light mode (white), but it wouldn't surprise me if I introduced that bug myself...

One thing I noticed while testing this is that if playlists were downloaded previously, but were then deleted from the server, the download repair will not delete them.  
While I think that is probably not a bad thing pre-se, we might wanna show some kind of prompt that there are downloaded items which were deleted on the server, prompting the user to review them? At least for a full repair where the general intention is to be consistent with the server and get rid of download errors (which these deleted playlists show up as). What do you think about that?

Chaphasilor:
I hope that I resolved the conflict correctly, used the web editor because I was looking at another PR. Formatting is probably off, oh well...

lib/services/finamp_logs_helper.dart
Komodo5197: That was what I did at first, but windows doesn't allow colons in filenames and I didn't want to bother with manual date formatting.

Komodo5197:
- The menu buttons were taking up space that wasn't always available.  This was very noticeable in the queue, which is already fairly cramped if the player sidebar is reasonably narrow.  Loosing any functional space for a button that is relativly unessecarry seems incorrect, so I've set the overflow cutoff fairly wide.  I'm not sure why the queue never shows the button because as you say, it should be able to get just wide enough.  I'll look into this.
- The mediaQuery.<attribute>of() pattern is indeed more efficent.   It's similar to using select() on providers - mediaquery.of() will rebuild for any mediaquery change, but mediaQuery.<attribute>of() will only cause widget rebuilds if the relevant attribute changes.  This is especially relevant as MediaQuery.of() will cause rebuilds every frame when resizing windows due to exposing the size.
- I attempted to match the existing styling exactly, so I probably overlooked something.  The recent lyrics PR completly removed all the sizing code though, so I guess this is moot.  I don't really know how the lyrics are expected/desired to look, so feel free to tweak that if there's issues.  I think the more explicit style code should make this sort of thing a bit more straightforward.
- What sort of notification were you thinking?  We could probably just assume anything in syncFailed is likley to have been deleted to just show something on the downloads page.

Komodo5197:
The queue not showing buttons is just because the maxWidth is only calculated when opened, so if you open on a small player and then expand the window, you never actually get beyond 640.  I've corrected this.

Chaphasilor:
Okay, thanks for explaining the performance improvements. I agree that the dedicated button is not ideal for narrow views, and I guess resizing is easy enough on desktop if someone wants to see more buttons. It's not the usual pattern (I've never seen an overflow button *appear* when the available space becomes *wider*), but let's roll with it and see if anyone complains.

For the notification I was thinking of a actionable snackbar that had a button for opening the downloads screen or failed downloads screen. If you think that this most likely only occurs when something got deleted on the server we could also have a short description/explanation text at the top of the failed downloads screen?

I'll look into the lyrics styling, although probably not before tomorrow evening :)

lib/services/finamp_logs_helper.dart
Chaphasilor: Ah, right. I think we can just remove all colons and be done with it?

Komodo5197:
It's not really an overflow button, though.  The other possible options like the like button or drag handle never overflow into it, and you can always easily access the menu through the fairly intuitive right click/long press.  It's just a gesture backup button to increase discoverability, like the queue button on the player.

I could add some sort of snackbar, although I'm concerned about spamming the user if multiple things get deleted or you switch servers with existing downloads.  It could also get annoying if someone tries to keep the downloads in place for a while, although that might not be something we want to put thought into.  I think that regardless, we can replace the file size message for a sync failed item with a suggestion to delete.

Chaphasilor:
You might be right about the overflow button, although it does open a menu with more option, like the like button. The drag handle is separately anyway, since it's related to the list, not the item. But it's fine the way it is now ^^  
I just mentioned it because the reason I added this button was because some desktop user couldn't figure out how to open the menu.

For the snackbar I was thinking to only show one "summary" snackbar at the end of a full repair. That should avoid spamming and lead to the same result. Just set a boolean when you detect anything that has probably been deleted, and then launch the snackbar if the boolean is true at the end of the repair?  
I wouldn't mind adding the delete suggestion as a separate line, and still show the size. That might influence the decision here, if the download is just a few MB they might be inclined to keep it just in case, but if it's larger they might want to delete it. Or if they accidentally deleted GBs of music from their server that they had downloaded to their phone, they might catch that here (like that one person on iOS who's server died and they managed to extract the files from Finamp's download directory).

Chaphasilor:
I didn't notice that you fixed the lyrics line size in dbb6c0278b9a5d735974d19b163d3e387a55fc29, but it's fine now. Just wanted to take a look and thought I was going insane first xD  
Seems like you also got rid of the font weight difference altogether, thanks.

Komodo5197:
In every comment I saw about the not knowing the menu existed, it was a macOS user who ended up with the iOS version without noticing and thus was unable to right click.  We really should make the macOS version more prominent, because it doesn't have the player issues the other desktop builds have, so it should be strictly superior to the ios app.  And nothing actually overflows into the menu button, so I still would not consider it an overflow button.

I could try to show both the sync failed message and the size, I guess.  Sync failed overwrites the download complete state so things will always be a bit funky.  I believe individually downloaded tracks will be unplayable, for example, but collections should still work.  I think one snackbar at the end of the sync should be doable.

All the lyric changes are just a proper merge of the recent PR, which is where the font size changes were actually removed.

Chaphasilor:
Yeah we'll start attaching the macOS build as an asset starting with the next release. Right now the only way to get it is to build from source, which isn't ideal.  
You're amazing, btw. Thanks again for all these fixes. It seems like you're reading everything, here and on Discord!

Komodo5197:
I do read pretty much everything.  It's not a particularly large amount to keep up with.

Komodo5197:
@Chaphasilor Okay, this suffered a bit of scope creep, but I'm finally finished with every change I wanted to do.  Feel free to look it over, test things, and/or merge it.

lib/l10n/app_en.arb
```diff
@@ -2772,5 +2776,17 @@
   "dragToReorder": "Drag to Reorder",
   "@dragToReorder": {
     "description": "Label shown on drag handles of track list  tiles in the queue list"
+  },
+  "addDownloadLocationsErrorPrompt": "Writing to selected directory failed.  You should modify permissions or choose another location.",
+  "@addDownloadLocationsErrorPrompt": {
+    "description": "Label shown on popup dialog when adding a download location when writing songs fails"
+  },
+  "androidImageErrorPrompt": "WARNING: We are unable to write images to the android Music folder due to OS restrictions.  You should choose another location.",
```
Chaphasilor, line 2784: If our plan is to keep this "supported" (as in possible), can we eventually ignore any image-related errors for such locations?  
Alternatively we could offer a choice to either not download images (since they would fail either way), or download them into a separate location (either external or internal), although that's a more involved solution I guess.  
Not blocking, just want to discuss this :)

lib/services/queue_service.dart
```diff
@@ -448,7 +448,9 @@ class QueueService {
           initialIndex: items["current"]!.isNotEmpty || items["queue"]!.isNotEmpty ? items["previous"]!.length : 0,
           // Always restore queues as linear to prevent them being reshuffled while restoring
           order: FinampPlaybackOrder.linear,
-          beginPlaying: isReload && (_audioHandler.playbackState.valueOrNull?.playing ?? false),
+          beginPlaying:
+              (isReload && (_audioHandler.playbackState.valueOrNull?.playing ?? false)) ||
+              (FinampSettingsHelper.finampSettings.autoplayRestoredQueue && droppedTracks == 0),
```
Chaphasilor:
```suggestion
              isReload ? (_audioHandler.playbackState.valueOrNull?.playing ?? false) :
              (FinampSettingsHelper.finampSettings.autoplayRestoredQueue && droppedTracks == 0),
```
This ensures queues don't suddenly start playing when they're paused, network changes, and the queue is either manually or automatically reloaded.

Chaphasilor (commented):
I actually assumed this was a "miscellaneous" PR where you'd keep adding things until I get around to merge it, but good to know you actually had a specific goal. In that case I'll make sure to wait for your okay in the future.

Only found one minor issue, but I'll give this a try until tomorrow when I'll merge it unless something comes up.  
Many thanks for these improvements! If you wouldn't mind, could you mention either here or on Discord which reported bugs you've fixed? It's a bit hard to keep track sometimes, although I'm definitely thankful that you're fixing stuff!

Komodo5197:
It originally had a more specific goal, but then that expanded into a big list of things to fix that I either noticed myself or saw in the issues/discord.  I'll try to get a comment in for all the relevant issues sometime tomorrow.

lib/l10n/app_en.arb
Komodo5197: I don't really want to add code supporting this sort of usage.  I was hesitant to block something that was theoretically doable before, but maybe we should just prevent this completely if its basically guaranteed to spam errors?

Chaphasilor: I agree, and this is the kind of change I'd like to make before releasing to stable. It will probably still lead to errors when users migrate from current stable to the redesign. Maybe we could prompt them to migrate to another directory, moving the existing files instead of requiring a redownload?

For newly added download locations we could suggest people to simply create a "Finamp" or "Music" directory within the "Downloads" directory, that works fine.

Komodo5197:
I've linked the PR to most of the issues I could find.  I know the following fixes came from complaints I saw but I don't know where.  They might have been in the discord general thread or something.
- Making the lyrics scrollbar not fade out at the top/bottom.
- Making the lyrics scroll reset to the top when changing songs with unsynced lyrics.
- Use the original images for the player instead of reencoding them as jpg.
- Hide the OS player notification when the queue is cleared.
- Adding the timestamp to logs exports.
- The download location testing was due to the various issues we've had pop up related to the Music folder.
- The thread on syncing deleted downloads is hopefully fully covered by this, but I haven't actually done that much testing.
- I believe I first started looking into how the OS-side player images were fetched due to a complaint about data usage being higher than expected from someone using a fairly tight transcode rate.
- There was an original complaint about the finamp/finamp folder structure somewhere.

Chaphasilor:
Thanks. I'll try to pinpoint the sources on Discord, but even if I can't find them it will be in the release notes anyway.

lib/l10n/app_en.arb
Chaphasilor: @Komodo5197 so we'll just throw an error if the write check fails, instead of showing a warning?

PR #1371, opened 2025-09-05
lyrics_screen.dart: disable the scrollbar

ios7jbpro:
## Changes
Disables the scrollbar on the lyrics screen. The reason to disable this is because of the fact that the scrollbar shows up every time the screen switches to the next line which is distracting(at least in my opinion), shown in the video;

https://youtube.com/shorts/bKcBVsliy5Y

After;

https://youtube.com/shorts/i6ekIh1IKRs

sidenote: @Chaphasilor gave me the solution for this, im not familiar with dart

### DIFF
--- a/lib/screens/lyrics_screen.dart
+++ b/lib/screens/lyrics_screen.dart
@@ -46,7 +46,8 @@ class _LyricsScreenContent extends ConsumerStatefulWidget {
   const _LyricsScreenContent();
 
   @override
-  ConsumerState<_LyricsScreenContent> createState() => _LyricsScreenContentState();
+  ConsumerState<_LyricsScreenContent> createState() =>
+      _LyricsScreenContentState();
 }
 
 class _LyricsScreenContentState extends ConsumerState<_LyricsScreenContent> {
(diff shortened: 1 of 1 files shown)


Maxr1998 (approved):
Changes look good, but would you mind giving it a quick reformat with `dart format path/to/file` to ensure the indentation of the updated code is correct?

Komodo5197:
This PR got me playing around with things a bit, and I came up with this alternative scroll behavior which might be more desirable than just unconditionally disabling the scrollbar:
```
class LyricsScrollBehavior extends MaterialScrollBehavior {
  @override
  Widget buildScrollbar(BuildContext context, Widget child, ScrollableDetails details) {
    final controller = details.controller;
    switch (axisDirectionToAxis(details.direction)) {
      case Axis.horizontal:
        return child;
      case Axis.vertical:
        assert(controller != null);
        return Scrollbar(
          controller: controller,
          notificationPredicate: (notification) {
            if (notification.depth != 0) return false;
            if (controller is AutoScrollController && controller.isAutoScrolling) {
              if (notification is ScrollUpdateNotification && notification.scrollDelta == null) {
                return true;
              }
              return false;
            }
            return true;
          },
          child: child,
        );
    }
  }
}
```

ios7jbpro:
@Maxr1998 done, reformatted

Chaphasilor:
Thanks for that @ios7jbpro!  
I agree with what @Komodo5197 suggested, where the scrollbar is hidden for auto-scroll events, but is shown when manually scrolling up and down (where auto-scroll gets disabled anyway). Would that solution work for you?  
In that case I'd probably merge your PR to give you the attribution, but then overwrite your changes with Komodo's approach :)

ios7jbpro:
@Chaphasilor i mean yes, that is certainly a way better solution than just disabling it entirely

Chaphasilor:
Well in that case, thanks for getting this started!

PR #1399, opened 2025-10-03
Fix failing build due to incomplete null check

Maxr1998:
## Changes
As the property is mutable, it cannot smart cast to non-null.

This may only be happening since Flutter 3.35, to which I upgraded a little earlier.

### DIFF
--- a/lib/menus/components/icon_button_with_semantics.dart
+++ b/lib/menus/components/icon_button_with_semantics.dart
@@ -28,8 +28,9 @@ class IconButtonWithSemantics extends ConsumerWidget {
           icon: Icon(icon),
           visualDensity: VisualDensity.compact,
           onPressed: () {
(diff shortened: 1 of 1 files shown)


Maxr1998:
I can probably just merge this myself, then I can rebase my other PRs already :smile:

Chaphasilor:
Yeah this stopped working due to some update. I built a version with this commit before pushing it, and soon after every build started failing 🙃  
Thanks for the fix, completely missed the PR until now (although I noticed the commit some time ago)

Maxr1998:
I'm pretty sure it was some flutter update, because it did work for me as well before I updated to the latest version in fvm. But it absolutely makes sense, and is very similar to how Kotlin handles smart casts. For class properties, you simply can't be sure it didn't change in between.

PR #1400, opened 2025-10-03
Optimize link handler code

Maxr1998:
## Changes
As suggested by @Komodo5197, I replaced the retry loop with a SchedulerBinding frame callback to ensure the app is ready before handling links. Tested locally with the Linux build.

## Related Issues
Originally discussed in 254dc41d48d3065798d5928c7a636942b9ccc2a7.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -513,21 +513,19 @@ class _FinampState extends State<Finamp> with WindowListener {
   @override
   void initState() {
     super.initState();
-    // Subscribe to all events (initial link and further)
-    _uriLinkSubscription = AppLinks().uriLinkStream.listen((uri) async {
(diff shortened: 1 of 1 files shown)


Chaphasilor:
Neat. Does clicking a `finamp://` link on Linux work if there's already a Window opened? Seems like that broke at some point (for Windows at least)...  
If not, mind throwing that in?

Maxr1998:
> Neat. Does clicking a `finamp://` link on Linux work if there's already a Window opened? Seems like that broke at some point (for Windows at least)...  

I'll have to check again, but I thought it should work already.

Maxr1998:
@Chaphasilor yes, still works great!

Chaphasilor:
awesome, thanks!

PR #1401, opened 2025-10-04
Allow to fast scroll by album artists on the albums tab

### LINKED ISSUE #1395: FastScroller should be visible on album view when sorted by album-artist and should allow jumping to album-artists

### Type of Feature

- [ ] Something new
- [ ] None of the above

### Description

Fast scroller is working great when sorting by name (example album view).
When you switch to sorting by album-artist the fast scroller disappears.. I think it should stay visible and allow fast scrolling to specific first letters of album-artists.

### Reasoning

_No response_

### Implementation Notes

_No response_

### Additional Information

_No response_

Maxr1998:
## Changes
Updates the `scrollToLetter` function to allow fast scrolling to a specific album artist if the sort by is set to that option.
Also makes the fast scroller visible for the album artist sort type.

## Related Issues
Fixes #1395

### DIFF
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -148,8 +148,9 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
         } else {
           _pagingController.appendPage(newItems, pageKey + newItems.length);
         }
-        if (letterToSearch != null) {
-          scrollToLetter(letterToSearch);
+        var localLetterToSearch = letterToSearch;
+        if (localLetterToSearch != null) {
+          scrollToLetter(localLetterToSearch);
         }
       }
     } catch (e) {
@@ -222,8 +223,9 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
     if (localRefreshCount == refreshCount && mounted) {
       _pagingController.appendLastPage(items);
       fullyLoadedRefresh = localRefreshCount;
-      if (letterToSearch != null) {
-        scrollToLetter(letterToSearch);
+      var localLetterToSearch = letterToSearch;
+      if (localLetterToSearch != null) {
+        scrollToLetter(localLetterToSearch);
       }
     }
   }
@@ -261,12 +263,11 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
 
   // Scrolls the list to the first occurrence of the letter in the list
   // If clicked in the # element, it goes to the first or last one item, depending on sort order
-  void scrollToLetter(String? clickedLetter) async {
-    String? letter = clickedLetter ?? letterToSearch;
-    if (letter == null || letter.isEmpty) return;
+  void scrollToLetter(String letter) async {
+    if (letter.isEmpty) return;
 
     letterToSearch = letter;
-    var codePointToScrollTo = letterToSearch!.toLowerCase().codeUnitAt(0);
+    var codePointToScrollTo = letter.toLowerCase().codeUnitAt(0);
 
     // Max code point is lower case z to increase the chance of seeing a character
     // past the target but below the ignore point
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -277,9 +277,21 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
     }
 
     //TODO use binary search to improve performance for already loaded pages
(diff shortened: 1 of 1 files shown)


lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -514,13 +526,15 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
...
     return RefreshIndicator(
       onRefresh: () async => _refresh(),
       child:
-          ref.watch(finampSettingsProvider.showFastScroller) &&
+          showFastScroller &&
               (widget.sortByOverride == SortBy.sortName ||
-                  (widget.sortByOverride == null &&
-                      ref.watch(finampSettingsProvider.tabSortBy(widget.tabContentType)) == SortBy.sortName))
+                  widget.sortByOverride == SortBy.albumArtist ||
+                  (widget.sortByOverride == null && tabSortBy == SortBy.sortName || tabSortBy == SortBy.albumArtist))
```
Komodo5197: I think we need another parentheses here?

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -277,9 +278,20 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
     }
 
     //TODO use binary search to improve performance for already loaded pages
+    var itemList = _pagingController.itemList!;
```
Komodo5197, line 281: What happens if the fast scroller is tapped while the page is still loading?  We should make sure to not throw an error.

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -277,9 +278,20 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
...
+    SortBy? tabSortBy = FinampSettingsHelper.finampSettings.tabSortBy[widget.tabContentType];
     bool reversed = FinampSettingsHelper.finampSettings.tabSortOrder[widget.tabContentType] == SortOrder.descending;
-    for (var i = 0; i < _pagingController.itemList!.length; i++) {
-      int itemCodePoint = _pagingController.itemList![i].nameForSorting!.toLowerCase().codeUnitAt(0);
+    for (var i = 0; i < itemList.length; i++) {
+      String sortName;
+      switch (tabSortBy) {
+        case SortBy.albumArtist:
+          sortName = itemList[i].albumArtist ?? "";
```
Komodo5197, line 288: These defaults don't work, because I believe codeUnitAt(0) will throw if it recieves an empty string.

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -148,6 +148,7 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
         } else {
           _pagingController.appendPage(newItems, pageKey + newItems.length);
         }
+        var letterToSearch = this.letterToSearch;
```
Komodo5197: Could you use a different variable name, like localLetterTo Search or something?  I don't like the variable shadowing.

lib/components/MusicScreen/music_screen_tab_view.dart
Maxr1998: Yes, you're right, the parentheses were positioned incorrectly. I'll fix it.

lib/components/MusicScreen/music_screen_tab_view.dart
Maxr1998: Works just fine. To be fair, the old code also already used the `!` operator a line below. So it's basically the same, just cached for better code readability. But should we still add some error handling?

lib/components/MusicScreen/music_screen_tab_view.dart
Maxr1998: Right, I'll rewrite that part then.

Maxr1998: Added an extra empty check.

lib/components/MusicScreen/music_screen_tab_view.dart
Maxr1998: Updated :white_check_mark:

Chaphasilor:
I guess this is ready to merge then?

Maxr1998:
I was waiting for @Komodo5197 to review my adjustments and merge it, but from my side it's ready.

Komodo5197:
It looks good to me.

Chaphasilor:
Perfect, thanks!

Maxr1998:
Thanks for the reviews and merge!

PR #1403, opened 2025-10-05
Ensure timestamps of offline listen log are accurate

Maxr1998:
Previously, when sending the playback stop event failed, the offline timestamps were recorded after the network timeout occurred and thus were off by 10 seconds. Now, the timestamps are recorded before attempting the request.

### DIFF
--- a/lib/services/offline_listen_helper.dart
+++ b/lib/services/offline_listen_helper.dart
@@ -33,13 +33,13 @@ class OfflineListenLogHelper {
   /// Logs a listen to a file.
   ///
   /// This is used when the user is offline or submitting live playback events fails.
-  Future<void> logOfflineListen(MediaItem item) async {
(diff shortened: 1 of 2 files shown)


Chaphasilor:
good catch!

Maxr1998:
Well, I think I also introduced that bug, sooo 😁
I discovered it a while back but didn't get around to fixing it. Thanks for the merge!

Chaphasilor:
Max over here accumulating bugs until it's October 💀

lukaslindnermusic:
Introducing bugs to fix them in hacktober to get more PRs :P 

just kidddiiiiiing <3

Maxr1998:
Exposing me like that 😭

No, but I would have filed a ticket but I also felt like fixing it would take the same time and then I ended up not doing it for months, yikes.

#1400 was also stuck in my to-dos for wayyyy to long 🥲

Issue #919, opened 2024-10-13
[FEATURE REQUEST] Custom accent colors


7eventy7:
Summary:
Allow users to customize the app's accent color, impacting all color-themed UI elements for a personalized and cohesive user experience.

Description:
It would be great to have the option to change the accent color within Finamp. This customization could apply to:

- All colored text
- On/off sliders and toggle switches
- UI elements on the "Now Playing" screen (e.g., playback controls, progress bar, etc.)
- Any other color-themed components across the app interface

Overall this is a pretty low priority request, but I think it would help improve the UI a lot! With IOS 18 it would look really nice alongside tinted app icons as well :D

Chaphasilor:
I agree this would be nice to have but has low priority.  
I'd be happy to accept and help with a PR to add this, but I probably won't be adding this anytime soon...

PR #1406, opened 2025-10-08
feat: app accent color

likhithpraveenk:
## Changes

Adds Accent Color setting which generates colorscheme

## Todo before merging

- [x] Translations

- [x] Reset Settings

## Related Issues

feat #919
This is a small feature and the accent obviously won't override the dynamic now playing screen elements. Is that required? I'm not very good at UI/UX so suggestions are welcome.

Screenshot_1759958587

Screenshot_1759958598
Bottom Sheet


flloschy:
I would reorganize the layout to something like this
```
 /'''\   |''''''''''''''''''''|
|color|  | #009688            |
 \___/   |____________________|
```

Where the circle is clickable to select preset or saved colors but also shows the current HEX value, and then this could be placed into the settings screen directly without an popup :)

Also a warning if contrast ratio is too bad could be neat, kinda like realtimecolors.com does. though idk how that calculation actually works

likhithpraveenk:
I could remove the bottom sheet and place everything in the settings tile but there would be no space to mention "Accent Color" or something of that sort. 

Regarding contrast ratio, flutter generates the color scheme based on given color [Color scheme.fromSeed](https://api.flutter.dev/flutter/material/ColorScheme/ColorScheme.fromSeed.html)

lib/services/accent_color_helper.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:hive_ce_flutter/hive_flutter.dart';
+import 'package:hive_ce_flutter/adapters.dart';
+
+/// A helper class for persisting and retrieving the app's accent color
```
Chaphasilor: is there a reason why this isn't part of `finamp_models.dart`?

lib/services/accent_color_helper.dart
```diff
@@ -0,0 +1,54 @@
...
+    final r = ((color.r * 255.0).round() & 0xFF);
+    final g = ((color.g * 255.0).round() & 0xFF);
+    final b = ((color.b * 255.0).round() & 0xFF);
+
+    final alpha = a.toRadixString(16).padLeft(2, '0');
+    final red = r.toRadixString(16).padLeft(2, '0');
+    final green = g.toRadixString(16).padLeft(2, '0');
+    final blue = b.toRadixString(16).padLeft(2, '0');
+
+    return '${includeAlpha ? alpha : ''}$red$green$blue'.toUpperCase();
```
Chaphasilor: You should add this method to our existing extensions for `Color`, where `atContrast()` and such are defined:
https://github.com/jmshrv/finamp/blob/redesign/lib/at_contrast.dart

You could add the `toHex()` extension below and rename the file to be more generic.

Also, it might be possible to simplify your implementation:
```dart
"#${(Colors.amber.toARGB32() & 0xFFFFFF).toRadixString(16).padLeft(6, '0')}"
```
could work (not tested), or check out this answer:
https://stackoverflow.com/a/61186997

Same for the other helper method.

lib/color_schemes.g.dart
```diff
@@ -84,3 +84,14 @@ const darkColorScheme = ColorScheme(
   outlineVariant: Color(0xFF41484D),
   scrim: Color(0xFF000000),
 );
+
+/// If [color] is provided -> returns a generated color scheme
+/// otherwise falls back to default color schemes
+/// [lightColorScheme] or [darkColorScheme]
+ColorScheme getColorScheme(Color? color ,{bool dark = false}) {
```
Chaphasilor: Please just pass the `Brightness` as a required parameter, instead of `dark` as a boolean.

lib/services/accent_color_helper.dart
Chaphasilor: I guess it's because the theme needs to wrap the entire app. But maybe you can use a different approach than adding a `ValueListenableBuilder`:

1. Just use a the `finampSettingsProvider` and watch the specific accent color setting, i.e. `ref.watch(finampSettingsProvider.accentColor)`, then pass that value to `getColorScheme()`

2. Or an approach similar to this one:  
https://github.com/jmshrv/finamp/blob/493078033c098cb2668f7facfc4a13ac072caa35/lib/main.dart#L346-L348

lib/components/LayoutSettingsScreen/accent_color_selector.dart
```diff
@@ -0,0 +1,152 @@
...
+        );
+      },
+    );
+  }
+
+  void _showAccentColorSheet(BuildContext context, Color? currentColor) {
+    final controller = TextEditingController(text: AccentColorHelper.toHex(currentColor));
+    Color? previewColor = currentColor ?? Colors.transparent;
+
+    final suggestedColors = [Colors.pink, Colors.teal, Colors.amber, Colors.cyan];
```
Chaphasilor: Not sure how useful these suggestions are. Might also be nice to show previously selected colors here (those that were actually confirmed/saved).

lib/components/LayoutSettingsScreen/accent_color_selector.dart
```diff
@@ -0,0 +1,152 @@
...
+                    width: double.infinity,
+                    height: 56,
+                    decoration: BoxDecoration(
+                      color: previewColor ?? Colors.transparent,
+                      borderRadius: BorderRadius.circular(12),
+                      border: Border.all(color: Theme.of(context).colorScheme.outlineVariant),
+                    ),
+                    alignment: Alignment.center,
+                    child: previewColor == null ? Text(AppLocalizations.of(context)!.invalidColorCode) : null,
+                  ),
```
Chaphasilor: I'm missing a proper color picker here. From the screenshot I assumed this container could be clicked to open an RGB or HSL color picker, but that's not the case. Only being able to choose form presets or manually enter hex codes seems a bit limited.  
I'm not opposed to depending on another package for this.

Alternatively (or additionally, depending on what the package offers), it would be nice to have HSL (hue, saturation, lightness) sliders to control the color more deliberately. I find that RGB "rainbow pickers" are not great for making small adjustments to the selected color.

Chaphasilor (commented):
Hey, thanks for the PR!

Those changes already look pretty good; I've left a few comments on things that could be improved.  
I like the approach of using a bottom sheet to configure this setting, keeping the settings list tile height at a single line.

Chaphasilor:
Oh and one additional thing:  
It would of course be awesome to base the color theme dynamically on the device theme (e.g. Material You on Android, not sure if there's now an equivalent for iOS).  
This is completely **optional** for this PR though, I'm just mentioning it for completeness' sake, in case you're interested.

likhithpraveenk:
Thanks for good feedback. I'll make the suggested changes.

lib/services/accent_color_helper.dart
likhithpraveenk: I'm not sure on how to use the ref.watch without moving the MaterialApp to a ConsumerWidget. I could use a listener and a local variable but that requires set state and obviously its not a good idea to use set state in **_Finamp_** root app. 
Another idea is to migrate the hive box ThemeMode to dynamic and store our accent color in it.

Komodo5197: You can use the Consumer widget for this.

Komodo5197: If you feel like cleaning things up you could also add providers in the LocaleHelper and ThemeModeHelper classes and then watch them in the same consumer, eliminating the need for the ValueListenableBuilders.

```
static Provider<Locale?> localeProvider = Provider((ref) {
    final listener = localeListener;
    void callback() => ref.state = locale;
    listener.addListener(callback);
    ref.onDispose(() => listener.removeListener(callback));
    return locale;
  });
```

```
static Provider<ThemeMode> themeModeProvider = Provider((ref) {
    final listener = themeModeListener;
    void callback() => ref.state = listener.value.get("ThemeMode") ?? ThemeMode.system;
    listener.addListener(callback);
    ref.onDispose(() => listener.removeListener(callback));
    return listener.value.get("ThemeMode") ?? ThemeMode.system;
  });
```

likhithpraveenk:
@Chaphasilor I think i made all the required changes. Adding HSL picker is the only thing left. [flutter_colorpicker](https://pub.dev/packages/flutter_colorpicker) this looks promising.

Screenshot_1760023800
Screenshot_1760023817


@Komodo5197 Yeah so i cleaned up the themeMode and locale to settings as well. I did the following minimal testing with them and they work
- changed device theme and in settings 
- changed device language and in settings 
- reset settings

lib/components/LanguageSelectionScreen/language_list.dart
```diff
@@ -1,20 +1,21 @@
...
 import 'package:flutter/material.dart';
 import 'package:finamp/l10n/app_localizations.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:locale_names/locale_names.dart';
 
-import '../../services/locale_helper.dart';
 import '../padded_custom_scrollview.dart';
 
-class LanguageList extends StatefulWidget {
+class LanguageList extends ConsumerStatefulWidget {
```
Komodo5197: It looks like locales never changes, so LanguageList can be converted to a stateless widget with locales calculated during build.

lib/components/LanguageSelectionScreen/language_list.dart
```diff
@@ -25,61 +26,54 @@ class _LanguageListState extends State<LanguageList> {
...
-    return ValueListenableBuilder(
-      valueListenable: LocaleHelper.localeListener,
-      builder: (_, __, ___) {
-        return PaddedCustomScrollview(
-          slivers: [
-            // For some reason, setting the null (system) LanguageListTile to
-            // const stops it from switching when going to/from the same
-            // language as the system language (e.g., system to English on a
-            // device set to English)
+    ref.watch(finampSettingsProvider.locale);
```
Komodo5197: This shouldn't be needed, as the LanguageListTiles now watch the setting on their own.  This also means the const constructor exception shouldn't be needed, as that was presumably there to make sure this valueListener rebuild propagated down to the no argument child.

lib/screens/settings_screen.dart
```diff
@@ -189,7 +188,9 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
           ListTile(
             leading: const Icon(Icons.language),
             title: Text(AppLocalizations.of(context)!.language),
-            subtitle: Text(LocaleHelper.locale?.nativeDisplayLanguage ?? AppLocalizations.of(context)!.system),
+            subtitle: Text(
+              ref.read(finampSettingsProvider.locale)?.nativeDisplayLanguage ?? AppLocalizations.of(context)!.system,
```
Komodo5197: I wouldn't expect this to actually change, but this should be ref.watch instead of ref.read.

lib/utils/locale_helper.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:finamp/services/finamp_settings_helper.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:get_it/get_it.dart';
+
+// Only to be used when you have no WidgetRef
+Locale? getLocale() {
```
Komodo5197: I don't think this function is needed, you should be able to just do FianmpSettingsHelper.finampSettings.locale if you don't want reactivity.

lib/services/widget_bindings_observer_provider.dart
```diff
@@ -56,8 +54,9 @@ class _FinampProviderBuilderState extends ConsumerState<FinampProviderBuilder> w
 
   @override
   void didChangePlatformBrightness() {
-    var theme = switch (Hive.box<ThemeMode>("ThemeMode").get("ThemeMode")) {
-      null || ThemeMode.system => View.of(context).platformDispatcher.platformBrightness,
+    final themeMode = ref.read(finampSettingsProvider.themeMode);
```
Komodo5197: I'd prefer FinampSettingsHelper.finampsettings.themeMode

lib/services/widget_bindings_observer_provider.dart
```diff
@@ -41,7 +40,6 @@ class _FinampProviderBuilderState extends ConsumerState<FinampProviderBuilder> w
   void initState() {
     super.initState();
     WidgetsBinding.instance.addObserver(this);
-    ThemeModeHelper.themeModeListener.addListener(didChangePlatformBrightness);
```
Komodo5197, line 44: We still want to react to themeMode changes.

`ref.listenManual(finampSettingsProvider.themeMode,(_,_)=>didChangePlatformBrightness());`

lib/main.dart
```diff
@@ -269,13 +267,12 @@ Future<void> setupHive() async {
...
   Hive.registerAdapters();
   Hive.registerAdapter(ThemeModeAdapter());
+  Hive.registerAdapter(ColorAdapter());
   Hive.registerAdapter(LocaleAdapter());
 
   await Future.wait([
     Hive.openBox<FinampSettings>("FinampSettings", path: dir.path),
-    Hive.openBox<ThemeMode>("ThemeMode", path: dir.path),
     Hive.openBox<FinampStorableQueueInfo>("Queues", path: dir.path),
-    Hive.openBox<Locale?>(LocaleHelper.boxName, path: dir.path),
```
Komodo5197, line 278: I do like moving the theme and locale settings into the standard settings object, but I think we need to add a migration.

likhithpraveenk: I'm not familiar with migration very much. is it something like this?
```dart
Future<void> _migrateThemeModeLocale() async {
  Box<ThemeMode> oldThemeModeBox = Hive.box<ThemeMode>("ThemeMode");
  Box<Locale?> oldLocaleBox = Hive.box<Locale?>("Locale");

  final finampSettings = FinampSettingsHelper.finampSettings;

  bool needsUpdate = false;

  if (oldThemeModeBox.isNotEmpty) {
    var oldTheme = oldThemeModeBox.get("ThemeMode");
    if (oldTheme != finampSettings.themeMode) {
      finampSettings.themeMode = oldTheme!;
      needsUpdate = true;
    }
  }

  if (oldLocaleBox.isNotEmpty) {
    var oldLocale = oldLocaleBox.get("Locale");
    if (oldLocale != finampSettings.locale) {
      finampSettings.locale = oldLocale;
      needsUpdate = true;
    }
  }

  if (needsUpdate) {
    FinampSettingsHelper.overwriteFinampSettings(finampSettings);
  }

  await oldThemeModeBox.deleteFromDisk();
  await oldLocaleBox.deleteFromDisk();
}
```

Komodo5197: The way we've been doing most of the migrations is to just have a flag for them.  So you create a new bool setting like hasCompletedThemLocaleMigration which has a DefaultSetting of true, but has false in the HiveField default ( see hasCompletedDownloadsFileOwnerMigration for an example).  And then if that's false, just open the old boxes, set the values with finampSetter, delete the old boxes, and then set the migration flag.  I personally wouldn't bother putting this in its own function and would keep it inside _setupHive, but that's a style choice.

likhithpraveenk:
found this awesome [color picker](https://rydmike.com/flexcolorpicker/) 


Screenshot_1760043939

likhithpraveenk:
Any other changes? @Komodo5197 @Chaphasilor

Komodo5197:
The code looks good to me.  I did notice that collection tiles are incorrectly setting a background color, which was hidden by that color matching the background in our themes.  You should remove the tileColor setting in the ListTile in item_collection_list_tile.dart to fix this. It would be nice if you extended the accent picker open button to cover the hex and not just the color, because I kept clicking that.  I think it would be easier to dial in a theme if there was a preview in the color picker, because what a theme will look like is not always obvious.  I think the cleanest way to do this would be to retheme the color picker sheet itself with the preview color, but you could also add a preview widget.

Also, while looking over this I noticed a few issues with how our existing theming code worked when player background images were disabled, and because you had changes to related areas I've taken the liberty of pushing onto your branch.  I hope you don't mind.

Komodo5197:
While looking over the changes I noticed that I forgot to apply the text color changes during withColorScheme, so I've fixed that.  I noticed you were passing a WidgetRef into the picker modal, which I'd consider bad practice because that ref can't rebuild the picker, so I moved all that code out into a ConsumerStatefulWidget.  Anyway, everything looks ready to merge from my side.  @Chaphasilor Any thoughts?

Chaphasilor:
Changes look good to me, thanks to both of you. I've noticed that the accent color-based color scheme was also used on the logs view, which didn't work well, so I've hard-coded the signal colors there.  
I'll merge this once the checks have passed :)

PR #1419, opened 2025-10-21
Dynamic Accent Color & Colored Monochrome In-App-Icon

### LINKED ISSUE #1453: Material 3 (and expressive) color support

### Type of Feature

- [ ] Something new
- [ ] None of the above

### Description

The app has a material 3 UI, however it doesn't match the system color scheme.
Possibly a built in way to edit more than just the accent color, but that's a little much.

### Reasoning

It would make the app blend more with the system. I use a daily wallpaper and the color scheme changes every day, it would be cool if finanp matches it.

### Implementation Notes

An addition option in settings for more color options.

### Additional Information

_No response_

flloschy:
## Changes
Swapped all In-app Finamp Logos to an dynamically colored Logo tinted with the active accent color.
If no Accent color is set, the normal finamp logo gets used

image

image

image

## Todo before merging

- [x] Get the icon to adapt
- [x] Setting to allow automatic theme changing. This should then disable the custom input
    - [x] Translations
    - [x] Reset Settings
- [x] (Maybe) Material You integration
- [x] (Very Maybe) Investigate if finamp can change accent color via a command for dynamic color themes on linux

#### Note to my self
Useful article for material color https://fivedottwelve.com/blog/themes-in-flutter-and-how-to-use-material-you-dynamic-colors/
useful post for cmd args https://stackoverflow.com/questions/55004302/how-do-you-pass-arguments-from-command-line-to-main-in-flutter-dart // https://dart.dev/tutorials/server/cmdline // https://stackoverflow.com/questions/77057678/communication-between-two-apps

### DIFF
--- a/lib/color_schemes.g.dart
+++ b/lib/color_schemes.g.dart
@@ -1,4 +1,6 @@
 import 'package:flutter/material.dart';
+import 'package:dynamic_color/dynamic_color.dart';
+import 'package:finamp/services/finamp_settings_helper.dart';
 
 const jellyfinBlueColor = Color(0xFF00A4DC);
 const jellyfinPurpleColor = Color(0xFFAA5CC3);
@@ -98,4 +100,17 @@ ColorScheme getColorScheme(Color? color, Brightness brightness) {
   }
 
   return brightness == Brightness.dark ? darkColorScheme : lightColorScheme;
+}
+
+Future<void> fetchSystemPalette() async {
+  final settingsHelper = FinampSettingsHelper.finampSettings;
+  if (!settingsHelper.useSystemAccentColor) return;
+  final palette = await DynamicColorPlugin.getCorePalette();
+  if (palette == null) return;
+  final scheme = palette.toColorScheme();
+  final primary = scheme.primary;
+  final current = settingsHelper.systemAccentColor;
+  if (primary != current) {
+    FinampSetters.setSystemAccentColor(primary);
+  }
 }
--- a/lib/color_schemes.g.dart
+++ b/lib/color_schemes.g.dart
@@ -101,23 +101,3 @@ ColorScheme getColorScheme(Color? color, Brightness brightness) {
 
   return brightness == Brightness.dark ? darkColorScheme : lightColorScheme;
 }
-
-Future<void> fetchSystemPalette() async {
-  // dont need to run when setting isnt enabled
-  final settingsHelper = FinampSettingsHelper.finampSettings;
-  if (!settingsHelper.useSystemAccentColor) return;
-
-  // skip if invalid palette
-  final palette = await DynamicColorPlugin.getCorePalette();
-  if (palette == null) return;
-
-  // get accent color
-  final scheme = palette.toColorScheme();
-  final primary = scheme.primary;
-  final current = settingsHelper.systemAccentColor;
-
-  // only update when needed (color transitions is laggy, no need to run it when not required)
-  if (primary != current) {
-    FinampSetters.setSystemAccentColor(primary);
-  }
-}
--- a/lib/color_schemes.g.dart
+++ b/lib/color_schemes.g.dart
@@ -103,13 +103,20 @@ ColorScheme getColorScheme(Color? color, Brightness brightness) {
 }
 
 Future<void> fetchSystemPalette() async {
+  // dont need to run when setting isnt enabled
   final settingsHelper = FinampSettingsHelper.finampSettings;
   if (!settingsHelper.useSystemAccentColor) return;
+
+  // skip if invalid palette
   final palette = await DynamicColorPlugin.getCorePalette();
   if (palette == null) return;
+
+  // get accent color
   final scheme = palette.toColorScheme();
   final primary = scheme.primary;
   final current = settingsHelper.systemAccentColor;
+
+  // only update when needed (color transitions is laggy, no need to run it when not required)
   if (primary != current) {
     FinampSetters.setSystemAccentColor(primary);
   }
--- a/lib/components/LayoutSettingsScreen/accent_color_selector.dart
+++ b/lib/components/LayoutSettingsScreen/accent_color_selector.dart
@@ -13,16 +13,18 @@ class AccentColorSelector extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final useSystemColor = ref.watch(finampSettingsProvider.useSystemAccentColor);
+    final useSystemAccent = ref.watch(finampSettingsProvider.useSystemAccentColor);
     final color = ref.watch(finampSettingsProvider.accentColor);
     final isSet = color != null;
 
     return ListTile(
-      enabled: !useSystemColor,
-      subtitle: useSystemColor ? Text(AppLocalizations.of(context)!.systemAccentColorHasPriorityInfo) : null,
+      enabled: !useSystemAccent,
+      subtitle: useSystemAccent ? Text(AppLocalizations.of(context)!.systemAccentColorHasPriorityInfo) : null,
       title: Text(AppLocalizations.of(context)!.accentColor),
       trailing: GestureDetector(
         onTap: () {
+          // "Disable" when using systemAccent
+          if (useSystemAccent) return;
           showModalBottomSheet<void>(
             context: context,
             isScrollControlled: true,
--- a/lib/components/LayoutSettingsScreen/accent_color_selector.dart
+++ b/lib/components/LayoutSettingsScreen/accent_color_selector.dart
@@ -13,9 +13,13 @@ class AccentColorSelector extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+    final useSystemColor = ref.watch(finampSettingsProvider.useSystemAccentColor);
     final color = ref.watch(finampSettingsProvider.accentColor);
     final isSet = color != null;
+
     return ListTile(
+      enabled: !useSystemColor,
+      subtitle: useSystemColor ? Text(AppLocalizations.of(context)!.systemAccentColorHasPriorityInfo) : null,
       title: Text(AppLocalizations.of(context)!.accentColor),
       trailing: GestureDetector(
         onTap: () {
--- a/lib/components/LayoutSettingsScreen/accent_color_selector.dart
+++ b/lib/components/LayoutSettingsScreen/accent_color_selector.dart
@@ -22,9 +22,7 @@ class AccentColorSelector extends ConsumerWidget {
       subtitle: useSystemAccent ? Text(AppLocalizations.of(context)!.systemAccentColorHasPriorityInfo) : null,
       title: Text(AppLocalizations.of(context)!.accentColor),
       trailing: GestureDetector(
-        onTap: () {
-          // "Disable" when using systemAccent
-          if (useSystemAccent) return;
+        onTap: useSystemAccent ? null : () {
           showModalBottomSheet<void>(
             context: context,
             isScrollControlled: true,
--- a/lib/components/LayoutSettingsScreen/accent_color_selector.dart
+++ b/lib/components/LayoutSettingsScreen/accent_color_selector.dart
@@ -22,19 +22,21 @@ class AccentColorSelector extends ConsumerWidget {
       subtitle: useSystemAccent ? Text(AppLocalizations.of(context)!.systemAccentColorHasPriorityInfo) : null,
       title: Text(AppLocalizations.of(context)!.accentColor),
       trailing: GestureDetector(
-        onTap: useSystemAccent ? null : () {
-          showModalBottomSheet<void>(
-            context: context,
-            isScrollControlled: true,
-            useSafeArea: true,
-            builder: (context) {
-              return ScrollConfiguration(
-                behavior: ScrollConfiguration.of(context).copyWith(scrollbars: false),
-                child: SingleChildScrollView(child: const AccentColorPopup()),
-              );
-            },
-          );
-        },
+        onTap: useSystemAccent
+            ? null
+            : () {
+                showModalBottomSheet<void>(
+                  context: context,
+                  isScrollControlled: true,
+                  useSafeArea: true,
+                  builder: (context) {
+                    return ScrollConfiguration(
+                      behavior: ScrollConfiguration.of(context).copyWith(scrollbars: false),
+                      child: SingleChildScrollView(child: const AccentColorPopup()),
+                    );
+                  },
+                );
+              },
         child: Row(
           mainAxisSize: MainAxisSize.min,
           mainAxisAlignment: MainAxisAlignment.spaceBetween,
--- a/lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
+++ b/lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
@@ -23,16 +23,12 @@ class AutomaticAccentColorSelector extends ConsumerWidget {
         mainAxisSize: MainAxisSize.min,
         mainAxisAlignment: MainAxisAlignment.spaceBetween,
         children: [
-          
           Container(
             width: 20,
             height: 20,
             margin: EdgeInsets.fromLTRB(0, 0, 2, 0),
-            decoration: BoxDecoration(
-              color: sysColor ?? Colors.transparent,
-              borderRadius: BorderRadius.circular(13),
-            )
-          ),  
+            decoration: BoxDecoration(color: sysColor ?? Colors.transparent, borderRadius: BorderRadius.circular(13)),
+          ),
           SizedBox(width: 16),
           Switch.adaptive(
             value: ref.watch(finampSettingsProvider.useSystemAccentColor),
--- a/lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
+++ b/lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
@@ -13,7 +13,6 @@ class AutomaticAccentColorSelector extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final sysColor = ref.watch(finampSettingsProvider.systemAccentColor);
-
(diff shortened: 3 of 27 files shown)


flloschy:
Got Android System theme to work :D
~~Sadly it does not update while the app is running :(~~
I got it to update, but only by forcing it to refetch the theme (if needed) via AppLifecycleListener onShow+onRestart+onResume

flloschy:
~~How is it ensured that only one finamp instance is running? And how to circumvent that? xd~~ Guess imma dance around with cpp and c code...

###### Im trying to add functionality to be able to call `finamp --set-accent "#FF00FF"` Goal of this is to either a) Send a piece of information to the existing process, telling it to update the color and exit    b). If no Process is running, start a super minimal hive version, change accent color, exit
###### (i) My goal is to still *avoid* having two finamps running / finamps editing the database at the same time

Why?
The dynamic_color plugin gets theme color on linux via GTK, and personally i dont like GTK themes, I much rather have it the way Im trying right now (while keeping automatic theming via GTK)

flloschy:
https://github.com/user-attachments/assets/7705e1c2-6c88-460c-b500-1b1544a4ad23
ayoo :D

Chaphasilor:
Awesome!  
Dynamic color refresh only when re-entering Finamp is absolutely fine, it doesn't change that frequently anyway, so this is better than expected.  
The linux stuff via dbus is nice too, great thinking. Would be nice to have this on the other platforms too, but I guess the target audience doesn't really need it there.  
Let me know when this ready to review, or are the TODOs outdated?

flloschy:
I reverted from sending the color straight to finamp, instead ill just send an request to refetch the theme. But maybe ill also re-revert again, will have to find that out
Currently im struggling to confirm if this actually works, i cant seem to figure out how to customize colors for GTK
Ill ping you when its review time :)

lib/components/finamp_icon.dart
```diff
@@ -0,0 +1,29 @@
...
+  final double width;
+  const FinampIcon(this.width, this.height, {super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final accent = ref.watch(finampSettingsProvider.accentColor);
+
+    final icon = SvgPicture.asset("images/finamp_cropped.svg", width: width, height: height);
+
+    if (accent == null) {
```
Komodo5197: It might be worth adding a setting to disable this recolor?  It doesn't always look that great.

flloschy:
@Chaphasilor here the promised ping

(Obviously couldn't test theme on Windows/Mac)

lib/main.dart
```diff
@@ -612,7 +616,11 @@ class FinampApp extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final accentColor = ref.watch(finampSettingsProvider.accentColor);
+    final useSystemTheme = ref.watch(finampSettingsProvider.useSystemAccentColor);
+    // System Accent has priority over custom Accent
```
Komodo5197, line 621: We could hide/disable the custom accent color setting when using the system theme to make this clear to the user.

lib/color_schemes.g.dart
```diff
@@ -99,3 +101,23 @@ ColorScheme getColorScheme(Color? color, Brightness brightness) {
...
   return brightness == Brightness.dark ? darkColorScheme : lightColorScheme;
 }
+
+Future<void> fetchSystemPalette() async {
+  // dont need to run when setting isnt enabled
+  final settingsHelper = FinampSettingsHelper.finampSettings;
+  if (!settingsHelper.useSystemAccentColor) return;
+
+  // skip if invalid palette
+  final palette = await DynamicColorPlugin.getCorePalette();
```
Komodo5197: The DynamicColorPlugin suggests that this method is android only, and you need getAccentcolor on other platforms.

lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'package:finamp/services/finamp_settings_helper.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class AutomaticAccentColorSelector extends ConsumerWidget {
+  const AutomaticAccentColorSelector({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final sysColor = FinampSettingsHelper.finampSettings.systemAccentColor;
```
Komodo5197: Shouldn't this be watching the system color instead of reading it?

lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
```diff
@@ -0,0 +1,48 @@
...
+            style: TextStyle(
+              fontWeight: FontWeight.w400,
+              color: Theme.of(context).textTheme.bodySmall?.color,
+              fontSize: 16,
+            ),
+          ),
+          SizedBox(width: 16),
+          Switch.adaptive(
+            value: ref.watch(finampSettingsProvider.useSystemAccentColor),
+            activeThumbColor: sysColor,
```
Komodo5197: What does setting both inactive and active color to the system theme color actually look like?  It might be better to have the color as a dedicated icon like the custom color oval.

lib/color_schemes.g.dart
```diff
@@ -99,3 +101,23 @@ ColorScheme getColorScheme(Color? color, Brightness brightness) {
 
   return brightness == Brightness.dark ? darkColorScheme : lightColorScheme;
 }
+
+Future<void> fetchSystemPalette() async {
```
Komodo5197: It might make more sense to put this into theme_provider.dart.  I'm not sure we should actually be adding code to this ostensibly auto-generated file.

lib/color_schemes.g.dart
```diff
@@ -99,3 +101,23 @@ ColorScheme getColorScheme(Color? color, Brightness brightness) {
...
+  // dont need to run when setting isnt enabled
+  final settingsHelper = FinampSettingsHelper.finampSettings;
+  if (!settingsHelper.useSystemAccentColor) return;
+
+  // skip if invalid palette
+  final palette = await DynamicColorPlugin.getCorePalette();
+  if (palette == null) return;
```
Komodo5197: Do we want to guard for null like this?  If there's no color scheme set, we should probably propagate this.

lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
```diff
@@ -0,0 +1,48 @@
...
+
+class AutomaticAccentColorSelector extends ConsumerWidget {
+  const AutomaticAccentColorSelector({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final sysColor = FinampSettingsHelper.finampSettings.systemAccentColor;
+
+    // Safe to assume that the System does not have Color Theme Support
+    if (sysColor == null) return SizedBox.shrink();
```
Komodo5197: We need a check that useSystemAccentColor is false here before returning to prevent this from getting stuck enabled.  Also, do we even want to completely hide this switch when null, as having no theme set is a valid setting for systems with theming support?

lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
```diff
@@ -0,0 +1,48 @@
+import 'dart:async';
+
+import 'package:finamp/color_schemes.g.dart';
+import 'package:finamp/extensions/color_extensions.dart';
+import 'package:finamp/l10n/app_localizations.dart';
+import 'package:finamp/services/finamp_settings_helper.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class AutomaticAccentColorSelector extends ConsumerWidget {
```
Komodo5197, line 9: Personally, I don't think these switch widgets deserve their own files and just put all of them into the relevant screen files instead.

flloschy:
valid points, will do tomorrow :)

lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
flloschy: I didnt even test how it looked, I just assumed it would be better, i think best of both worlds is.. well the best option here
By that I mean only coloring the inactiveThumb, then it looks like this:
image
image

lib/color_schemes.g.dart
flloschy: Oh wow how did i miss that

Using `DynamicColorPlugin.getAccentColor()` still doesnt read my system theme though and there is an [issue with that problem](https://github.com/material-foundation/flutter-packages/issues/658), so i commented and am waiting for a reply. But Im not that hopeful that there will be one, at least not any time soon

lib/l10n/app_en.arb
```diff
@@ -724,6 +724,22 @@
   "@accentColorTitle": {
     "description": "Title for the accent color picker sheet"
   },
+  "systemAccentColor": "Use {type, select, android{Material You Theme} linux{GTK+ Theme} macOS{App Accent Color} windows{Accent Color} other{System Theme}}",
```
Komodo5197: I don't think all of these are descriptive enough.  Windows should be 'System Accent Color' at a minimum, and I'm not sure how great the mac name is either.

lib/components/finamp_icon.dart
```diff
@@ -0,0 +1,27 @@
...
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final icon = SvgPicture.asset("images/finamp_cropped.svg", width: width, height: height);
+    final useMonochromeIcon = ref.watch(finampSettingsProvider.useMonochromeIcon);
+    if (!useMonochromeIcon) return icon;
+
+    final accent = ref.watch(finampSettingsProvider.accentColor);
+    final brightness = ref.watch(brightnessProvider);
+    final colorScheme = getColorScheme(accent, brightness);
+    final color = colorScheme.onSurface;
```
Komodo5197: Do we really want to use onSurface?  That's basically a text color which is always going to be close to white/black.  It might be better to just use the primary?  Or is matching the text what you're going for?

lib/components/finamp_icon.dart
```diff
@@ -0,0 +1,27 @@
...
+  final double width;
+  const FinampIcon(this.width, this.height, {super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final icon = SvgPicture.asset("images/finamp_cropped.svg", width: width, height: height);
+    final useMonochromeIcon = ref.watch(finampSettingsProvider.useMonochromeIcon);
+    if (!useMonochromeIcon) return icon;
+
+    final accent = ref.watch(finampSettingsProvider.accentColor);
```
Komodo5197: I think we want more complex logic here to use the correct system theme color that's actually in use.  It might be easier to just read the context theme?

lib/models/finamp_models.dart
```diff
@@ -234,6 +234,8 @@ class DefaultSettings {
   static const preferNextUpPrepending = true;
   static const rememberLastUsedPlaybackActionRowPage = true;
   static const lastUsedPlaybackActionRowPage = PlaybackActionRowPage.newQueue;
+  static const useSystemAccentColor = false;
+  static const useMonochromeIcon = true;
```
Komodo5197: do we really want a monchrome icon by default, even with default theming?

lib/color_schemes.g.dart
Komodo5197: I can confirm it does seem to work properly on windows, at the very least.

lib/services/theme_provider.dart
```diff
@@ -490,3 +498,24 @@ class _ThemeTransitionCalculator {
         : Duration.zero;
   }
 }
+
+Future<void> fetchSystemPalette() async {
+  // dont need to run when setting isnt enabled
+  final settingsHelper = FinampSettingsHelper.finampSettings;
+  if (!settingsHelper.useSystemAccentColor) return;
```
Komodo5197: Because we don't update the setting when the setting is disabled, the preview color applied to the system theme switch is inaccurate.

lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
Komodo5197: I don't really like the singular switch not matching the theme, so I'd still lean towards a separate indicator.  But the current inactive only coloring is fine enough, I guess.

lib/components/finamp_icon.dart
flloschy: image
image

I personally like text color more*

*\*with most colors, not all look better*

lib/components/LayoutSettingsScreen/automatic_accent_color_selector.dart
flloschy: Idea was to have an idea what the color looks like while not having a too messy UI
Having another Oval there feels too much imo, but im no designer xd

Chaphasilor: I don't think a preview is needed at all. But we could just get rid of the hex value for the system color and replace it with a proper preview, since I don't think a hex string adds any value for the user.

lib/components/finamp_icon.dart
Chaphasilor: so which is which in your screenshots? xD

Chaphasilor:
Just took a quick look (after forgetting that this PR existed):

- Why is the default app icon color purple? 😅
- Using the material theme works great!
- You should probably disable the onTap handler for the custom accent color if the rest of the tile is disabled
- The "Use monochrome icon" settings isn't super clear, but I can reword that tomorrow
- I think the monochrome setting should only be enabled and apply if a custom (or system) accent color is used, and not by default
- Having a setting "Theme" with value "System" and a "Use System Theme" setting is a bit confusing. Maybe it's time to rename "Theme" to "Theme Brightness" or something like that?

lib/components/finamp_icon.dart
flloschy: The first is primary color
The second is text color

I like the text color :)

flloschy:
Icon could have been purple because I pushed a bad accent color :/  Fixed it back to onSurface :)

Monochrome is already defaulted to off tho?

Hu? What do you mean with the last point? Putting the system theme toggle inside the accent color drawer?

Chaphasilor:
You're right about the default. Perhaps there was still some old Hive field left over that coincided with it.  
Last point is just about renaming the setting that has a dropdown for picking between light, dark, and system brightness. Right now it's just called "Theme", but your new setting is also called "Use system theme". The first is about light/dark, and the second about the accent colors and such. So there should be a clearer distinction.

flloschy:
Ive changed my theme setting to "use system accent", I think that should clear up any confusion

Komodo5197:
You forgot to run codegen, so hive is still using the old default for the monochrome icon setting.

lib/components/LayoutSettingsScreen/accent_color_selector.dart
```diff
@@ -13,16 +13,18 @@ class AccentColorSelector extends ConsumerWidget {
...
     return ListTile(
-      enabled: !useSystemColor,
-      subtitle: useSystemColor ? Text(AppLocalizations.of(context)!.systemAccentColorHasPriorityInfo) : null,
+      enabled: !useSystemAccent,
+      subtitle: useSystemAccent ? Text(AppLocalizations.of(context)!.systemAccentColorHasPriorityInfo) : null,
       title: Text(AppLocalizations.of(context)!.accentColor),
       trailing: GestureDetector(
         onTap: () {
+          // "Disable" when using systemAccent
+          if (useSystemAccent) return;
```
Chaphasilor: you disable it by passing `null` to the actual `onTap` handler. So it should be

```dart
onTap: useSystemAccent ? null : () {...}
```

Chaphasilor:
One last comment from me, then I can give this a final test and we can merge it! I'd like to base some home screen modifications on this :D

flloschy:
ready to merge o7

Chaphasilor:
Alright, looks good, thanks again! Just made some minor fixes.  
Would you mind *briefly* documenting the ways to programmatically change accent color for the release notes? Just for those interested, nothing detailed :)

flloschy:
My Draft for a release note:
______


Dynamic Themes are now sported on most platform (not iOS). Finamp tries to fetch System Accent Color when starting and when focusing the app.
This Setting can be enabled under: Settings > Layout & Theme > Use System Accent
**Note**: System Accent will have priority over your custom Accent Color


<details>
<summary>More Info for Linux Users</summary>


For Linux the GTK themes can be updated at anytime using the following command [1]. You can use this command after updating your System Theme to apply it _without_ interacting with Finamp.
The active GTK theme might not result in the wanted effect though, so in addition to force-updating, you can also force-set the accent Color with this command [2]!
Please note that these DBus commands will only work _while_ Finamp is running.

<details>
  <summary>1. Reload the system Accent from GTK</summary>
  
```sh
gdbus call \
    --session \
    --dest 'com.unicornsonlsd.FinampSettings' \
    --object-path '/com/unicornsonlsd/Finamp' \
    --method 'com.unicornsonlsd.Finamp.updateAccentColor'
```

> [!IMPORTANT]  
> "Use System Accent" needs to be **enabled**.

</details>


<details>
  <summary>2. Overwrite the accent color</summary>
  
```sh
gdbus call \
    --session \
    --dest 'com.unicornsonlsd.FinampSettings' \
    --object-path '/com/unicornsonlsd/Finamp' \
    --method 'com.unicornsonlsd.Finamp.setAccentColor' \
    '#ff0000' # you can also send "default" to clear the accent color
```

> [!IMPORTANT]  
> "Use System Accent" needs to be **disabled**.
</details>

</details>

Chaphasilor:
Awesome, thanks! I updated the section in the readme a bit, but that's definitely a good base for the release notes!  

Btw, you might wanna create feature branches in your fork in the future, instead of making a PR based on your fork's `redesign` branch. That way you could have multiple PR open at once, and keep your `redesign` branch up-to-date at all times.

PR #1421, opened 2025-10-22
Instant Mix include tapped item

### LINKED ISSUE #1416: Tapping a track plays a different one / Instant Mix does not include tapped item

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.9.20 beta

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [x] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

_No response_

I have two libraries, one of which is downloaded (A), and the other is not (B). When I tap a Track from the Tracks tab, it plays a completely random song. Issue existed on 0.9.19 too, but coincidentally I added my second library just as 0.9.20 was released and issue persists (inspite of a similar issue being closed)

### Logs

[finamp-logs-2025-10-21T01-15-31-994097.txt](https://github.com/user-attachments/files/23010287/finamp-logs-2025-10-21T01-15-31-994097.txt)
 

### Additional Information

_No response_

DhruvaSambrani:
Fixes #1416

## Todo before merging

- [ ] Translations

- [ ] Reset Settings

### DIFF
--- a/lib/services/audio_service_helper.dart
+++ b/lib/services/audio_service_helper.dart
@@ -78,7 +78,7 @@ class AudioServiceHelper {
 
     try {
       items = await _jellyfinApiHelper.getInstantMix(item);
-      if (items != null) {
+      if (items != null && items[0].id != item.id) {
         items.insert(0, item);
         await _queueService.startPlayback(
           items: items,
--- a/lib/services/audio_service_helper.dart
+++ b/lib/services/audio_service_helper.dart
@@ -79,6 +79,7 @@ class AudioServiceHelper {
     try {
       items = await _jellyfinApiHelper.getInstantMix(item);
       if (items != null) {
+        items.insert(0, item);
         await _queueService.startPlayback(
           items: items,
           source: QueueItemSource(


Chaphasilor:
@DhruvaSambrani hey thanks, but I'm not sure this is intended behavior? On all prior versions the Instant Mix *did* include the source track, and in the release notes there only was a comment about "similar tracks" not returning the source track anymore (which I assumed to be related to the `/Similar`) endpoint.  

Let me do some digging and asking, because this fix would cause the source track to appear twice in 10.10 and earlier. We should at least include a dynamic check that ensures the source track is included exactly once at the start :)  
But maybe it's just a regression and will be fixed server-side.

Chaphasilor:
This is the PR I saw mentioned in the release notes, which would match the behavior we're seeing but doesn't appear to have intended to change the Instant Mix endpoint:  
https://github.com/jellyfin/jellyfin/pull/14686

lib/services/audio_service_helper.dart
```diff
@@ -79,6 +79,7 @@ class AudioServiceHelper {
     try {
       items = await _jellyfinApiHelper.getInstantMix(item);
       if (items != null) {
+        items.insert(0, item);
```
HeroBrine1st, line 82: I'd suggest checking for duplicates so 10.10 jellyfin is not regressed

```suggestion
        if (items[0].ItemId != item.ItemId) items.insert(0, item);
```

UPD: Oops, already mentioned here xD

Chaphasilor: also, don't commit this suggestion, it should be `item.id` in both cases :)

Chaphasilor:
So the consensus seems to be that this is indeed a bug in Jellyfin and can be fixed in 10.11.z. We *could* include this fix (with the duplicate check) in a hotfix, which might be appropriate given that some people have already upgraded to 10.11.0. But that would need to be pushed quickly (which I don't have a good track record with).  
But unless this can be fixed within a few days on Jellyfin's side this might be a good move.

Chaphasilor:
We have a PR! 😁  
https://github.com/jellyfin/jellyfin/pull/15133  
So I guess we sit this one out :)

DhruvaSambrani:
I added the fix anyway so its ready to merge if PR or next release takes too long

Chaphasilor:
@DhruvaSambrani still, thanks for this PR! You'll get a shout-out in the next release notes!

PR #1422, opened 2025-10-22
Fix shown index in playlist reordering screen

hannesbraun:
When editing a playlist, the shown indices were off by one. I guess regular users expect the list to start at 1. (It was just an operator precedence issue.)


Chaphasilor:
Awesome, yeah @lukaslindnermusic pointed that out, I completely missed it 🙈  
thanks for the fix!

Issue #979, opened 2024-12-18
MacOS App quits when window closed


realtofuine:
The MacOS version of the app quits the app when the window is closed.

Usually, Mac apps do not quit when the red X button is quit, so can the app move to the background instead of fully quitting?

AverageHelper:
Considering the app plays audio, not quitting on window close (continuing to play media in the background) seems to make sense imo

Chaphasilor:
@realtofuine is this still happening after the latest update? There was an issue with the app crashing on iOS, maybe it's related...

AverageHelper:
This also happens on Linux ([Flathub](https://flathub.org/apps/com.unicornsonlsd.finamp) version on Fedora/KDE), tho I'm not sure what the usual behavior there is for music applications. Several apps that do background stuff (like [Signal](https://github.com/signalapp/Signal-Desktop), [Element](https://github.com/element-hq/element-desktop), [VLC](https://code.videolan.org/videolan/vlc), etc.) will let the user close the window but leave an icon in the System Tray widget, so media etc. can continue and users can either select that icon or launch the application again in the usual ways to open the existing process. (Meantime, I just minimize the app lol)

KDE does have a Media Player widget, but it doesn't seem to have a way to focus the app that's playing music (maybe that's configurable on Finamp's side? idk..)

TeamLinux01:
As a Linux user, specifically KDE Plasma, it would be nice to have a minimize/close to system tray as an option.

PR #1430, opened 2025-10-26
feat: add tray functionality

F-4Dev:
## Changes
This PR add tray functionality for desktop (Tested on Linux) using the [tray_manager](https://pub.dev/packages/tray_manager) Library. Inside the tray it is possible to hide the window and show it again. 

image

## Todo before merging

- [ ] Minimize to tray on close (setting)
- [ ] play/pause button
- [ ] favorite playlists shortcut
- [ ] pause indicator on icon
- [ ] Translations

If you have any ideas about what else could be added, please leave a comment :)

## Related Issues

https://github.com/jmshrv/finamp/issues/979

ran into this error message but found a [temporally fix](https://github.com/leanflutter/tray_manager/issues/67#issuecomment-2737796286)
```
Building Linux application...                              
~/Projects/Finamp-Dev/fintray/linux/flutter/ephemeral/.plugin_symlinks/tray_manager/linux/tray_manager_plugin.cc:118:17: error: 'app_indicator_new' is deprecated [-Werror,-Wdeprecated-declarations]
Building Linux application...                                           
Error: Build process failed
```

Maybe there is a better way to handle this. I'm open for suggestions 😅


Chaphasilor:
@F-4Dev sorry for not taking a look earlier. Definitely an interesting feature, although I don't tend to use tray shortcuts anymore. I guess the related settings could go into the "Interaction" settings.  
There's a migration warning for that package now, although that seems to be internal refactoring and nothing that we should do, so it's probably fine.  

Do you need any help with the functionality, beyond Linux stuff? :)

F-4Dev:
@Chaphasilor Oh okay, then there will probably some changes to be made at some point, but we'll see.
I think I need help with the closing behavior of Finamp. I'd like to close the App, but keep living in the background (tray) somehow. What I mean is, when closing the window, it should only be hidden and not fully closed if that makes sense "^^
I know that Local Send is also Flutter and it has the behavior and I tried handling closing the same but I couldn't get it to work sadly.

and don't worry about answering late :)

notjiatan:
Hi, how do I fix this, is the build error just for linux?
Might try to fix & merge this PR

PR #1437, opened 2025-10-29
Allow moving tracks within queue using track menu

### LINKED ISSUE #1434: Move tracks within queue instead of creating new instance when adding to next up from queue

### Type of Feature

- [ ] Something new
- [ ] None of the above

### Description

When adding something to Next Up via "Play Next" or "Add to Next Up", a new instance of the track gets added to the "Next Up" sub-queue.  
This is *also* the case when using the track menu from within the queue list, which leads to duplicate instances of the track within the queue. Instead of this, the track should simply be moved when the menu is accessed via the queue list, possibly including setting the updated queue source.

### Reasoning

If a track is already in the queue, the motivation behind selecting "Play Next" is most likely to move that track to the top of the queue instead of including in in the queue twice.  
The only situation I can think of where someone might want to manually duplicate tracks in the queue is when testing a new feature or debugging.

### Implementation Notes

- Set the queue source when moving so that other tracks can be appended to next up (for seamless reordering of multiple tracks, e.g. for track requests at a party)
- It should be able to tell if the menu is invoked from the queue list or not
- Rename "Add to Next Up" as "Move to Next Up" in this case. "Play Next" can stay as it is

### Additional Information

Discussed on Discord: https://discord.com/channels/1064474779043770408/1064571023212691577/1432742785005522974

Chaphasilor:
Implements #1434

## Changes

- Added `PlaybackActionRow` if needed
- Implemented move logic via removing and re-adding
  - Simply reordering doesn't work since we can't change a `FinampQueueItem`'s `QueueItemQueueType` after it was added to the queue

## Todo before merging

- [x] Translations

## Related Issues

- closes #1434

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -438,9 +438,9 @@
   "@playbackActionPageMoveWithinQueue": {
     "description": "Label for the playback action page that contains all actions related to moving tracks within the queue"
   },
-  "playbackActionPageDuplicate": "Duplicate",
-  "@playbackActionPageDuplicate": {
-    "description": "Label for the playback action page that contains all actions related to duplicating tracks within the queue"
+  "playbackActionPageRegularTrackOptions": "Play",
+  "@playbackActionPageRegularTrackOptions": {
+    "description": "Label for the playback action page that contains all actions of the normal track menu, like starting a new queue or duplicating the track within the existing queue"
   },
   "playbackActionPageNewQueue": "New Queue",
   "@playbackActionPageNewQueue": {
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -434,6 +434,14 @@
       }
     }
   },
+  "playbackActionPageMoveWithinQueue": "Move Within Queue",
+  "@playbackActionPageMoveWithinQueue": {
+    "description": "Label for the playback action page that contains all actions related to moving tracks within the queue"
+  },
+  "playbackActionPageDuplicate": "Duplicate",
+  "@playbackActionPageDuplicate": {
+    "description": "Label for the playback action page that contains all actions related to duplicating tracks within the queue"
+  },
   "playbackActionPageNewQueue": "New Queue",
   "@playbackActionPageNewQueue": {
     "description": "Label for the playback action page that contains all actions related to starting a new queue"
@@ -1069,6 +1077,18 @@
   "shuffleSomeToQueue": "Shuffle some to Queue",
   "@shuffleSomeToQueue": {
     "description": "Used for shuffling some tracks to the end of the regular queue, to play after all prior tracks from the queue have played"
+  },
+  "movePlayNext": "Play next",
+  "@movePlayNext": {
+    "description": "Used for moving a track to the \"Next Up\" queue at the first position, to play right after the current track finishes playing"
+  },
+  "moveAddToNextUp": "Move to Next Up",
+  "@moveAddToNextUp": {
+    "description": "Used for moving a track to the \"Next Up\" queue at the end, to play after all prior tracks from Next Up have played"
+  },
+  "moveAddToQueue": "Move to end",
+  "@moveAddToQueue": {
+    "description": "Used for moving a track to the end of the queue, to play after all prior tracks from the queue have played"
   },
   "confirmPlayNext": "{type, select, track{Track} album{Album} artist{Artist} playlist{Playlist} genre{Genre} other{Item}} will play next",
   "@confirmPlayNext": {
--- a/lib/menus/components/playbackActions/playback_action_row.dart
+++ b/lib/menus/components/playbackActions/playback_action_row.dart
@@ -36,20 +36,13 @@ class _PlaybackActionRowState extends ConsumerState<PlaybackActionRow> {
 
   @override
   Widget build(BuildContext context) {
-    final nextUpNotEmpty = ref.watch(QueueService.queueInfoStreamProvider).valueOrNull?.nextUp.isNotEmpty ?? false;
-    final lastUsedPlaybackActionRowPage = ref.watch(finampSettingsProvider.lastUsedPlaybackActionRowPage);
-    final lastUsedPlaybackActionRowPageIndex = lastUsedPlaybackActionRowPage.pageIndexFor(
-      nextUpIsEmpty: !nextUpNotEmpty,
-    );
-    final initialPageViewIndex = ref.watch(finampSettingsProvider.rememberLastUsedPlaybackActionRowPage)
-        ? lastUsedPlaybackActionRowPageIndex
-        : 0;
-    controller = PageController(initialPage: initialPageViewIndex);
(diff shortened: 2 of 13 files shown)


Chaphasilor:
@Komodo5197 I assume no comments from you means this can be merged?

lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -19,12 +19,65 @@ Map<String, Widget> getPlaybackActionPages({
...
+              MoveAddToQueuePlaybackAction(
+                item: queueItem,
+                popContext: popContext,
+                compactLayout: compactLayout,
+                genreFilter: genreFilter,
+              ),
+            ],
+          ),
+          // New Queue
+          AppLocalizations.of(context)!.playbackActionPageNewQueue: Row(
```
Komodo5197: Is it really worth it to give this option its own page?  This is also inconsistent with the page layout of the same four items for non-queue items.

lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -19,12 +19,65 @@ Map<String, Widget> getPlaybackActionPages({
...
   bool compactLayout = false,
-  bool preferNextUp = false,
+  bool preferPrependingToNextUp = false,
   BaseItemDto? genreFilter,
+  FinampQueueItem? queueItem,
 }) {
   final BaseItemDtoType? itemType = item is BaseItemDto ? BaseItemDtoType.fromItem(item) : null;
 
-  return {
+  return queueItem != null
```
Komodo5197: We seem to be using queueItem as a proxy for whether we have a track.  I'd prefer to directly check if the item is a track, and then show or hide the move page based off the queueItem, to make this more understandable and robust.

lib/menus/track_menu.dart
```diff
@@ -476,13 +481,15 @@ class _TrackMenuState extends ConsumerState<TrackMenu> with TickerProviderStateM
       MenuMask(
         height: MenuItemInfoSliverHeader.defaultHeight,
         child: SliverToBoxAdapter(
-          child: Row(
+          child: widget.queueItem != null
```
Komodo5197: I'd prefer to just use the PlaybackActionRow for non-queue items as well, for more consistency.  We'd need to hide the page headers for one-page rows, though, so you can leave them seperate if you'd prefer.  I would like the action page structure to be consistent though.

lib/models/finamp_models.dart
```diff
@@ -2155,6 +2155,23 @@ class FinampQueueInfo {
     return null;
   }
 
+  FinampQueueItem? getQueueItemByOffset(int offset) {
```
Komodo5197, line 2178: Is this actually used?

lib/models/finamp_models.dart
```diff
@@ -2155,6 +2155,23 @@ class FinampQueueInfo {
...
+    final index = currentTrackIndex + offset;
+    if (index < 0 || index >= fullQueue.length) {
+      return null;
+    }
+    return fullQueue[index];
+  }
+
+  int? getOffsetForQueueItem(FinampQueueItem item) {
```
Komodo5197, line 2186: It might make more sense to just directly add a removeQueueItem method to the queue service?

lib/l10n/app_en.arb
```diff
@@ -1054,6 +1062,18 @@
   "@shuffleSomeToQueue": {
     "description": "Used for shuffling some tracks to the end of the regular queue, to play after all prior tracks from the queue have played"
   },
+  "movePlayNext": "Play next",
```
Komodo5197, line 1081: Could we really not come up with a name that isn't completely identical to the insert variant?  although I'm not coming up with anything that good either...

lib/menus/components/playbackActions/playback_action_row.dart
```diff
@@ -51,7 +53,8 @@ class _PlaybackActionRowState extends ConsumerState<PlaybackActionRow> {
       popContext: widget.popContext,
       compactLayout: widget.compactLayout,
       genreFilter: widget.genreFilter,
-      preferNextUp: ref.watch(finampSettingsProvider.preferNextUpPrepending),
```
Komodo5197, line 54: You should create a separate last used page setting for queue tracks, because the page types don't really line up with the other action rows.

Komodo5197:
I think I just never got around to reading through this thoroughly?  I've got a couple minor issues and nitpicks, but the big thing is that I think this needs to remember its last used page independently of the other playback rows.

lib/menus/track_menu.dart
Chaphasilor: good point, we can just hide the header instead of doing completely manually 👍

lib/l10n/app_en.arb
Chaphasilor: Question is: do we need a different name at all? Semantically it's different, sure, but the item will actually play next, so...

Chaphasilor:
Alright, thanks! Will try to get this finished up tonight

lib/models/finamp_models.dart
Chaphasilor: no, I just included it for completeness' sake, I think. Or I removed the code that used it. I wouldn't mind keeping it around though, might be useful

Chaphasilor:
Okay, I think I addressed all the points now, sorry for the delay. I also tried to simplify the page->index and index->page lookups, and I believe the behavior is still the same/desired one.  
Right now the track menu uses 3 or 4 pages instead of the single page we had before, while this is more consistent with the other options, it potentially leads to more taps needed to perform an action, so I'm not yet sure if we should change that. I guess we could release it and see if people want the "old" behavior back?

Komodo5197:
I was talking more about similarity between track menu in the queue, with the move option, and the track menu outside the queue without it.  I don't think trying to be consistent with non-tracks is worthwhile because tracks are to different, as they don't have the shuffle options.  The design I'd go with would have the move page, if applicable, and then a page with all four other track options.  I honestly think our playback rows are already too split up for albums and whatnot, with only two options per row, but I don't think we have a way use less groups and still have meaningful divisions.  Anyway, I don't have a catchy name for the track monopage, but I think we should keep it.  Also, I think there's something wrong with the stackSized calculation in the track menu, because I'm occasionally getting scrollbars.

Chaphasilor:
@Komodo5197 okay, yeah that makes sense, I agree. That has been reverted now, and I think I've also fixed the height issue for the track menu. I think by "getting scrollbars" you meant the bottom sheet not being tall enough to fix all options? If so, that should be fixed.

For the second page that has all the regular options for the track menu, I've decided to just call it "Play", because while I originally wanted to differentiate between "duplicate" (i.e., play next and add to Next Up) and "new queue" (i.e., play), that doesn't work for a single page, and is also not that important. Having "move" and "play" as options should be clear enough.

Chaphasilor:
@Komodo5197 did you have a look at the new changes yet? Can we merge?

Komodo5197:
I haven't bothered redoing any testing, but it all looks good from the code.  With the just_audio PR merged, make sure this doesn't accidentally disable any of the next up playback options that should be hopefully working on desktop now.

PR #1442, opened 2025-11-01
Migrate to just_audio 0.10

### LINKED ISSUE #772: Shuffle Broken on Desktop
Shuffle works different on Android and the Linux/Flathub build.

I would say that the behavior on Android is "normal": if shuffle is selected, the tracks are placed in the queue in a random way.

On Linux/Flathub it works differently: if shuffle is selected, the next track gets randomly selected and placed in the queue. Then the queue gets filled by the other songs of the playlist in order.

I think on all platform the first behavior should be in place.

### LINKED ISSUE #907: Repeat Button Broken on Desktop
Hi, Finamp on Fedora Linux mostly works pretty well and seems to be 1:1 with the iOS app version at least. Finamp version 0.9.11-beta is installed via flatpak on fedora linux https://flathub.org/apps/com.unicornsonlsd.finamp

But the app doesn't seem to respect/follow the repeat button being off. It will continue to repeat the album/playlist even with the icon for repeat on the far left of the finamp player menu (left of the previous track button)... Screenshot of my finamp player for reference. In other words, the repeat button seems to be broken on linux.

I just tested finamp on iOS and the repeat button being off works just fine and no more playback after the album is over

image

### LINKED ISSUE #1271: "Play next" and "Add to next up" do not work on desktop

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.9.18 (beta)

### Affected Devices/Platform

- [ ] Android
- [ ] Android Auto
- [ ] iOS
- [x] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [x] Windows

### Device

Arch Linux with KDE Plasma 6.5 and 6.4 (Broken on both so probably unrelated)

In order to reproduce the issue, start a queue by clicking on a track, and then navigate to another track, right click, and add to next up or play next. When the next song plays, the title and cover will be correct for the song you picked, but the track which actually plays is the next track in the original queue, ignoring the "next up" list.

For example, here the next track to be played would be Great Grey Wolf Sif, but the cover and title will show Pontiff Sulyvahn.
Image

### Logs

[finamp-logs.zip](https://github.com/user-attachments/files/20887464/finamp-logs.zip)

### Additional Information

Edit: I forgot to mention it completely skips the "Up next" section even if there are multiple tracks.
Also, this is installed from the AUR (finamp-bin)

Chaphasilor:
https://github.com/ryanheise/just_audio/tree/minor/just_audio#migrating-to-010x

Only minor refactoring was needed so far. There are some batch operations that aren't available with the new API, so adding multiple tracks to an existing queue might be less efficient now, but that requires thorough testing.  
Testing on iOS would also be appreciated.

Known bugs/PRs that we might want to wait for:

- https://github.com/ryanheise/just_audio/pull/1555
  - I've switched to this fork for now. It might take a while before the PR is merged, but I see no issue with building the next version based on this instead of the pub.dev version as a temporary fix, instead of waiting for it to get merged
- `_player.playbackEventStream` doesn't emit an event anymore when the shuffle mode is changed, which means the current player index isn't updated properly and the wrong track is shown until the next `playbackEvent` is triggered (e.g. via seeking, pausing)
  - This could be related to the wrong shuffle order being used, but is unaffected by the open PR
  - I've worked around this now by manually fetching the current player index before rebuilding the queue

### DIFF
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -10,6 +10,7 @@ import 'package:finamp/models/finamp_models.dart';
 import 'package:finamp/models/jellyfin_models.dart' as jellyfin_models;
 import 'package:finamp/services/current_track_metadata_provider.dart';
 import 'package:finamp/services/favorite_provider.dart';
+import 'package:finamp/services/finamp_user_helper.dart';
 import 'package:finamp/services/jellyfin_api_helper.dart';
 import 'package:finamp/services/queue_service.dart';
 import 'package:flutter/foundation.dart';
@@ -108,7 +109,7 @@ class PlayerVolumeController {
 
 /// This provider handles the currently playing music so that multiple widgets
 /// can control music.
-class MusicPlayerBackgroundTask extends BaseAudioHandler {
+class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, QueueHandler {
   final _androidAutoHelper = GetIt.instance<AndroidAutoHelper>();
 
   AppLocalizations? _appLocalizations;
@@ -119,14 +120,9 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
   late final List<DarwinAudioEffect> _iosAudioEffects;
   late final AndroidLoudnessEnhancer? _loudnessEnhancerEffect;
 
-  ConcatenatingAudioSource _queueAudioSource = ConcatenatingAudioSource(children: []);
   final _audioServiceBackgroundTaskLogger = Logger("MusicPlayerBackgroundTask");
   final _volumeNormalizationLogger = Logger("VolumeNormalization");
   final _outputLogger = Logger("Output");
-
-  /// Set when creating a new queue. Will be used to set the first index in a
-  /// new queue.
-  int? nextInitialIndex;
 
   // Init the new sleep timer with a length of 0
   // SleepTimer sleepTimer = SleepTimer(SleepTimerType.duration, 0);
@@ -414,15 +410,19 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
     _queueCallbackPreviousTrack = previousTrackCallback;
   }
 
-  Future<Duration?> setAudioSources(
-    List<AudioSource> audioSources, {
+  Future<Duration?> setQueueItems(
+    List<FinampQueueItem> queueItems, {
     bool preload = true,
     int? initialIndex,
     Duration? initialPosition,
     ShuffleOrder? shuffleOrder,
   }) async {
-    _setNextInitialIndex(initialIndex ?? 0);
     try {
+      List<AudioSource> audioSources = [];
+
+      for (final queueItem in queueItems) {
+        audioSources.add(await _queueItemToAudioSource(queueItem));
+      }
       return await _player.setAudioSources(
         audioSources,
         preload: preload,
@@ -443,15 +443,35 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
     return null;
   }
 
-  Future<void> moveAudioSource(int currentIndex, int newIndex) {
+  Future<void> appendFinampQueueItem(FinampQueueItem queueItem) async {
+    return _player.addAudioSource(await _queueItemToAudioSource(queueItem));
+  }
+
+  Future<void> appendFinampQueueItems(List<FinampQueueItem> queueItems) async {
+    return _player.addAudioSources(await Future.wait(queueItems.map(_queueItemToAudioSource)));
+  }
+
+  Future<void> insertFinampQueueItemAt(int index, FinampQueueItem queueItem) async {
+    return _player.insertAudioSource(index, await _queueItemToAudioSource(queueItem));
+  }
+
+  Future<void> insertFinampQueueItems(int index, List<FinampQueueItem> queueItems) async {
+    return _player.insertAudioSources(index, await Future.wait(queueItems.map(_queueItemToAudioSource)));
+  }
+
+  Future<void> moveFinampQueueItem(int currentIndex, int newIndex) {
     return _player.moveAudioSource(currentIndex, newIndex);
   }
 
-  Future<void> removeAudioSourceRange(int start, int end) {
+  Future<void> removeFinampQueueItemAt(int index) {
+    return _player.removeAudioSourceAt(index);
+  }
+
+  Future<void> removeFinampQueueItemRange(int start, int end) {
     return _player.removeAudioSourceRange(start, end);
   }
 
-  Future<void> clearAudioSources() {
+  Future<void> clearFinampQueueItems() {
     return _player.clearAudioSources();
   }
 
@@ -624,7 +644,7 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
       // item, so only pause.
       await pause(disableFade: true);
       // Skipping to zero with empty queue re-triggers queue complete event
-      if (_player.effectiveIndices?.isNotEmpty ?? false) {
+      if (_player.effectiveIndices.isNotEmpty) {
         await skipToIndex(0);
       }
     } catch (e) {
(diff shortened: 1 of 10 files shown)


lib/services/music_player_background_task.dart
```diff
@@ -408,11 +414,22 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
     _queueCallbackPreviousTrack = previousTrackCallback;
   }
 
-  Future<void> initializeAudioSource(ConcatenatingAudioSource source, {required bool preload}) async {
-    _queueAudioSource = source;
```
Komodo5197, line 412: _queueAudiosource should be removed entirely.

lib/services/music_player_background_task.dart
```diff
@@ -965,7 +995,7 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler {
     return skipToIndex(index);
   }
 
-  void setNextInitialIndex(int index) {
+  void _setNextInitialIndex(int index) {
```
Komodo5197: nextInitialIndex should be removed entirely.

lib/services/queue_service.dart
```diff
@@ -724,12 +719,12 @@ class QueueService {
         );
       }
 
-      List<AudioSource> audioSources = [];
-      for (final item in queueItems) {
-        audioSources.add(await _queueItemToAudioSource(item));
-        _queueServiceLogger.fine("Added '${item.item.title}' to queue from '${source?.name}' (${source?.type})");
-      }
-      await _queueAudioSource.addAll(audioSources);
+      await _audioHandler.addQueueItems(
```
Komodo5197: addQueueItems, and some of the other methods you've used, seem to be default empty implementations inherited from BaseAudioHandler and do not actually do anything.

lib/services/queue_service.dart
```diff
@@ -786,18 +781,16 @@ class QueueService {
...
       int offsetLog = offset;
 
-      List<AudioSource> audioSourceList = [];
       for (final queueItem in queueItems) {
-        audioSourceList.add(await _queueItemToAudioSource(queueItem));
         _queueServiceLogger.fine(
           "Prepended '${queueItem.item.title}' to Next Up (index ${adjustedQueueIndex + offsetLog})",
         );
         offsetLog++;
+        await _audioHandler.insertQueueItem(adjustedQueueIndex + offset, queueItem.item);
```
Komodo5197: I think this logic would put the items in backwards?

lib/services/queue_service.dart
```diff
@@ -579,24 +577,21 @@ class QueueService {
...
-      }
-      _audioHandler.setNextInitialIndex(_queueAudioSourceIndex);
-
-      await _audioHandler.initializeAudioSource(_queueAudioSource, preload: true);
+      final Duration? queueDuration = await _audioHandler.setAudioSources(
+        audioSources,
+        initialIndex: initialIndex,
+        preload: true,
+        shuffleOrder: _shuffleOrder,
+        initialPosition: Duration.zero,
```
Komodo5197: If we have an initial position API available, we can expose this so queue restore doesn't need to play and then seek.

lib/services/queue_service.dart
```diff
@@ -854,20 +847,18 @@ class QueueService {
...
       int adjustedQueueIndex = getActualIndexByLinearIndex(_queueAudioSourceIndex);
 
-      List<AudioSource> audioSourceList = [];
       for (final queueItem in queueItems) {
-        audioSourceList.add(await _queueItemToAudioSource(queueItem));
         _queueServiceLogger.fine(
           "Appended '${queueItem.item.title}' to Next Up (index ${adjustedQueueIndex + offsetLog})",
         );
         offsetLog++;
+        await _audioHandler.insertQueueItem(adjustedQueueIndex + offset, queueItem.item);
```
Komodo5197: We can still bulk insert sources if we just put them into a list.

Komodo5197:
Regarding the shuffle bug, I've noticed that even if you start a shuffled queue instead of using the shuffle button, the shuffle order in our queue is not the same as what we get by letting the track advance.  I've also noticed a bug that was reported elsewhere for the main redesign branch, which is that swiping away tracks on the queue doesn't actually remove them.  I noticed that you seem to have accidentally used an unimplemented our AudioHandler doesn't actually have an implementation for.  Maybe we should try to get a few more of those in?  Although I'm not sure what OS would actually use a lot of them.

Komodo5197:
Also, I'm not sure if this was the case before but it seems like the loop mode and shuffle mode streams are basically just views of the main player data stream, so they spam the logs while playing.

Chaphasilor:
Yeah, I'll try to fix a few related bugs.  
At the moment adding things to the queue doesn't seem to work at all, ~~for whatever reason~~ for the reasons you've pointed out in your review. A few methods should probably also be renamed.

I'm not yet sure what exactly you're talking about regarding the loop mode, but I'll check that once I'm back at the computer.

Chaphasilor:
Okay, got most of the issues resolved I think. The shuffle bug persists, and I'm not sure if we are able to work around that one, although I haven't checked yet.  
I'll try to test the remaining queue management more thoroughly in the coming days.

Chaphasilor:
This should be mergeable now. We could put a TODO into the `pubspec.yaml` in case we merge this before the upstream PR is merged.  

I'd appreciate thorough testing from others, especially on the iOS side:
- Adding tracks, albums, etc. to the queue, in original or shuffled order (for collections)
- Removing tracks from the queue
- Reordering the queue when shuffle is disabled
- Trying the different loop modes
- Skipping around in the queue
- Starting playback via various methods

CC @lukaslindnermusic @flloschy @jmshrv  
You can probably just use the app normally for testing, but trying out some more niche queue operations would certainly help!

I might also try to get reordering tracks with shuffle enabled working, some first tests were promising? But that's not a blocking change.

flloschy:
I dont think I have the ability to debug/build ios apps

UnicornsOnLSD:
I'll give it a test tomorrow :)

Chaphasilor:
@flloschy I'm aware, but testing on Android would also be nice! I've done some testing, but the more, the merrier! 😁

Oh and I'll make the improvements to queue restoring before merging too of course 😅

UnicornsOnLSD:
> I'll give it a test tomorrow :)

mb probably tomorrow, had a moderate amount of pints

Chaphasilor:
The shuffle fix seems to mostly work. I've noticed that the Next Up queue isn't properly kept when shuffle is enabled and the player auto-advances. Manually skipping is fine. That could be related to your upstream PR, although I haven't looked into it yet.

Since I was looking at the player screen a lot I also decided to attempt to improve the color contrasts a bit, as discussed on Discord. I think the best strategy here is to keep this in while we continue to test this, and monitor for any problematic accent colors.

Komodo5197:
It looks like shuffling wasn't even related, I was just supplying the wrong player index for a brief period during song transitions which resulted in the next up queue being dropped.  It should be fixed now.

Chaphasilor:
Yep, seems to be fixed, thanks! So then the only thing left to do is:
- [x] Use new API for simpler queue restore
- [x] Test on iOS @jmshrv 

Correct?

UnicornsOnLSD:
I haven't forgotten about this, I'm pretty free tomorrow so I'll get around to it 🫡

UnicornsOnLSD:
Just had a play with it on iOS and queue operations seem stable, playback is working :)

Chaphasilor:
@Komodo5197 just got this error during regular playback on Windows when a new track started playing:
`[MPV/SEVERE] 2025-11-25 14:34:46.339706: lavf: Failed to create file cache.`

Not sure if that's related to your PR at all, but just wanted to let you know about it :)

Komodo5197:
I'm pretty sure I've seen that happen before my changes?  I don't actually know that much about mpv, so I'm not really sure what it means, but it doesn't seem to have hurt much so far.  I might try to look into this at some point.

Chaphasilor:
I also didn't notice anything. I don't use desktop too often, and had never seen that error before. But seems like nothing important then :)

PR #1443, opened 2025-11-01
Add description about running the development setup with flakes

brumik:
## Changes

## Related Issues
none

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ You can ignore the Chrome/Web component, since Finamp is not a web app.
 ### NixOS and Flakes
 
 The project includes a `flake.nix` file that can help you install flutter and rust dependencies. The recommended way of using it is to
-do `nix develop .#fenix` (or you can use `dotenv` with `.envrc` file). 
+use `nix develop`, which allows rustup for Discord RPC dependency. Also there's `nix develop .#fenix` which stubs rustup with fenix, but it is a hack.
 
 To get the application running on linux platform once you are in the development shell:
 - `flutter build linux` - generates the build files in `./build/linux/x64/release/bundle` where the `lib` folder will have all the dynamic libraries.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,13 +16,14 @@ You can ignore the Chrome/Web component, since Finamp is not a web app.
 
 ### NixOS and Flakes
 
-The project includes a `flake.nix` file that can help you install flutter and rust dependencies. The recommended way of using it is to
-use `nix develop`, which allows rustup for Discord RPC dependency. Also there's `nix develop .#fenix` which stubs rustup with fenix, but it is a hack.
+The project includes a `flake.nix` file that can help you install Flutter and Rust dependencies. The recommended way of using it is to
+use `nix develop`, which allows rustup for Discord RPC dependency. Also, there's `nix develop .#fenix` which stubs rustup with fenix, but it is a hack.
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -14,6 +14,16 @@ flutter doctor -v
 This will list out all components used for Flutter development, and if they are installed. This should include either Android or iOS development (depending on your platform), as well as any desktop platform you want to target.  
 You can ignore the Chrome/Web component, since Finamp is not a web app.
 
+### NixOS and Flakes
+
+The project includes a `flake.nix` file that can help you install flutter and rust dependencies. The recommended way of using it is to
+do `nix develop .#fenix` (or you can use `dotenv` with `.envrc` file).
```
HeroBrine1st: I created it mainly for myself and as such it is not entirely maintained, so I won't suggest it for new contributors from the start. Also there's no .envrc file in the repository (except in gitignore) so I didn't understood what you mean.

```suggestion
The project includes a `flake.nix` file that can help you install flutter and rust dependencies. The recommended way of using it is to
use `nix develop`, which allows rustup for Discord RPC dependency. Also there's `nix develop .#fenix` which stubs rustup with fenix, but it is a hack.
```

Feel free to combine envrc with my suggestion back if that's something popular because I'm simply out of mainstream, e.g. I didn't follow mainstream even when I set to install nixos

UPD: Though rustup can require FHS so idk maybe fenix is better or I need to refactor the flake yet again

HeroBrine1st (changes requested):
Otherwise LGTM

CONTRIBUTING.md
brumik: Makes sense, should I move it somewhere in the bottom of the file?

HeroBrine1st: Yeah, I think it being the first option is not optimal. Also, I guess looking at developing without android studio on arch, it is not dependent on NixOS. It's just an mkShell and it should work on any distro, so you can move it just before arcane arts and rename to something like "Setting up a development shell with nix", making a note that it is required for NixOS even with android studio/intellij idea. Also there needs a note that running desktop application from IDE doesn't work due to CWD requirement.

Anyway that's already a ton of info so I'd approve it as is except being the first option.

P.s. while I sound like a gatekeeper, I don't actually merge PRs, so that's just my own opinion

Chaphasilor:
@brumik thanks for adding this. Have you tried running `flutter run` instead of `flutter build + manually launching the app?  
It would probably good to mention that somewhere, either by suggesting it, or by pointing out that it doesn't work on Nix :)  

Aside from that we can merge this of course :)

brumik:
It does not work, though I have a pretty bad workaround by symlinking the .so libraries to the root, then I can use flutter run. Still could not get through VS code hit relapse working with the official flutter plugins. Did play around with devboxes but no luck there either.

Chaphasilor:
Okay. @HeroBrine1st is that also your experience? If so, I'd say we just point that out in the markdown, and then we can merge this :)

HeroBrine1st:
I use Intellij IDEA with flutter plugin (so I don't know what is hit relapse) and I too use `flutter build linux` and then run manually

It is probably either an issue in flutter (ld should not load shared objects from arbitrary folders and that's okay) or specific nixos glibc issue, depending on reproducibility in FHS-based distributions.
Probably nixos but I couldn't quickly find any symlinkJoin or other code related to glibc inclusion into /usr/lib in  buildFHSEnv implementation

Chaphasilor:
Thanks to both of you! :D

PR #1452, opened 2025-11-11
Fix strange container string being shown instead of codec

### LINKED ISSUE #1018: Alac reporting as multiple things at once
When streaming alac Finamp shows this image

While aac and other things in the m4a container seem to report m4a just fine

Chaphasilor:
Just a minor change, not tested yet.

### DIFF
--- a/lib/components/PlayerScreen/feature_chips.dart
+++ b/lib/components/PlayerScreen/feature_chips.dart
@@ -1,3 +1,4 @@
+import 'dart:math';
 import 'dart:ui';
 
 import 'package:collection/collection.dart';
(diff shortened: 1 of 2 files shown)


Chaphasilor:
@Komodo5197 turns out this fix was a bit more complex than I originally thought, since the metadata sourcing is a bit messy. I think I'm now properly differentiating between container and codec, so the issue should be fixed. But please take a look to make sure I didn't break anything else in the process.  
For me at least the behavior seems correct now.

Komodo5197:
Sorry for not getting to this earlier.  Nothing really stands out as incorrect to me.  I'm not really an expert on this part of the codebase, though.

PR #1467, opened 2025-11-26
Misc Fixes

### LINKED ISSUE #271: Instant mix with albums not showing up on Android
LineageOS 12 2022-07-11
FinAmp 0.6.4

On Jellyfin web there is a compass icon for `Instant Mix`. The 0.6.4 release notes list `Instant mix with albums (@faizzzzz)` however when in 0.6.4 and on an album there is no compass icon. Tapping a title, player will open but also there is no compass icon to be found. Not quite sure, how instant mix with albums is supposed to work.

Sidenote: Tapping album or artist works great and is a very useful addition which I love!

PS: Should 0.6.4 be tagged as release (and not pre-release)?

### LINKED ISSUE #1424: [Beta] Playlist Edit Button Overlapping Back Button
Hi,

On the latest version of the beta, the new edit playlist button is overlapping the back button, blocking it from being pressed. My phone is an s21+.

Image

Thanks for the great work!

### LINKED ISSUE #1432: [Beta] Word for word lyric highlight color has a chance to have the same color as the rest of the lyric color on dark theme

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.9.20 (beta)

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [x] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

_No response_

word for word lyric highlight color has a chance to have the same color as the rest of the lyric color on dark theme when:

1. word for word lyric is available
2. color that get picked from the album art is white
3. dark theme is used

example when using dark theme

Image

example when using light theme

Image

### Logs

<details>

  <summary>Logs</summary>
  
  

[finamp-logs.txt](https://github.com/user-attachments/files/23164972/finamp-logs.txt)
at 2025-10-27 20:08:15.146514

  
</details> 

### Additional Information

_No response_

### LINKED ISSUE #1466: No covers in Android Auto album view

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.9.20 (beta)

### Affected Devices/Platform

- [ ] Android
- [x] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

_No response_

Album covers are not showed on Android Auto tabs (albums, artists, playlists, etc).
The current cover image is displayed correctly on the home screen widget and the playback screen.

Image

### Logs

<details>

  <summary>Logs</summary>
  
</details> 

### Additional Information

_No response_

Chaphasilor:
## Changes

Various fixes to close a few more issues with the next release...

## Todo before merging

- [x] Fix contrast issues on current track tile
- [x] Unhide "Max Concurrent Downloads" setting on iOS since it's actually being used now
- [x] ~~Fix MediaSources assertion error~~ Seems to indeed be related to download errors and not a fluke
  - Might be related to deleted `FinampSettings`
- [x] Remove shuffle disabled on desktop string (`desktopShuffleWarning`)
- [x] Check if similar radio really is unavailable after starting a single track (and it stopping at the end)
  - Didn't work on watch, continuous was available
- [x] Add setting to force audio offloading
  - This is now disabled (for compatibility) by default in just_audio 0.10 but could help with reducing battery drain
- [x] Opt in to Impeller again
  - Opt-out is now deprecated, so it should be stable enough and offer some performance benefits for newer devices
- [x] Fix album mix mode running out of tracks
- [x] Open radio source when tapping source on player screen
- [x] Make album mix mode the third mode
- [x] Generate much fewer tracks when starting new queue via radio
  - Especially continuous mode takes a while to generate 30 tracks
- [x] Increase hitbox of back button on player screen
- [x] Try to increase maximum width for titles
- [x] Check if reordering with shuffle active can be enabled now
  - ~~Nope, still broken. Updating the `_shuffleOrder` doesn't persist, and using `_player.moveAudioSource()` seems to overwrite indices, so that un-shuffling isn't possible anymore~~
- [x] Shrink scroll-to-current FAB on lyrics and queue screens
- [x] Fix special download collections showing as "Unlocalized ..."
- [x] Use consistent height for track titles of one and two lines
- [x] Fix don't sync on startup setting not applying
  - The observed behavior stems from `restartDownloads` getting called when opening the downloads screen, so unless that happens downloads don't sync on startup
  - Maybe it would make sense to only actually restart downloads if they were started before, and otherwise wait for `resync` to be called?
- [x] Fix contrast for word-by-word lyrics
- [x] Fix favorite state in notification not being in sync
  - Possibly because the metadata was fetched later, and then the notification wasn't updated?
- [x] Add Instant Mix for all types to release notes
- [x] Remember the current fallback mode for the album mix across generation
  - Otherwise we make a ton of pointless requests
  - Only reset when the radio source changes
- [ ] ~~Try to hook up audio session to system equalizer~~ Will have to wait for another PR
- [x] Mention libraries not being respected by radio
- [x] Remove tooltip for radio, loop mode, and add to playlist button to allow long-pressing 🙄

- [x] Translations

- [x] Reset Settings
- [x] Test upgrading from 0.9.20 before merging


lib/services/queue_service.dart
```diff
@@ -1127,6 +1129,32 @@ class QueueService {
       }
     }
 
+    Uri? artUri = isAndroidAutoOrMediaBrowserRequest
+        ? _providers.read(albumImageProvider(AlbumImageRequest(item: item, maxHeight: 200, maxWidth: 200))).uri
+        : null;
+
+    // use content provider for handling media art on Android
+    if (Platform.isAndroid) {
```
Komodo5197: It might make sense to also check isAndroidAutoOrMediaBrowserRequest here to avoid doing unneeded uri attaches to queue items on android?

lib/services/music_player_background_task.dart
```diff
@@ -238,6 +253,25 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
     }
   }
 
+  static Future<void> configureAudioSession() async {
+    final session = await AudioSession.instance;
+    await session.configure(
+      FinampSettingsHelper.finampSettings.duckOnAudioInterruption
```
Komodo5197, line 261: Isn't this backwards?  Wouldn't we want the old, vanilla music() config if we use the normal ducking behavior?  I feel like there's a mix of two solutions here, the dynamic config changes and the manual ducking, but we only need one.  So if this config swap actually works, we shouldn't need the manual ducking code?

Chaphasilor: You're right, I messed up the boolean. If the setting is true, it's the default behavior.  

The reason why we handle it manually when using the speech config is because we still want to pause on bigger interruptions. I guess we could leave out the ducking settings check there since it's dead code, I just left it in for completeness' sake.

lib/services/radio_service_helper.dart
```diff
@@ -152,36 +152,60 @@ Future<void> maybeAddRadioTracks() async {
...
-    tracks.addAll(await generateRadioTracks(radioTracksNeededForInitialQueue, overrideSeedItem: source));
+    localResult.tracks.addAll(
+      await generateRadioTracks(radioTracksNeededForInitialQueue, overrideSeedItem: source, forNewQueue: true),
+    );
   } catch (e) {
     _radioLogger.warning("Couldn't generate radio tracks: $e");
   }
-  if (tracks.isEmpty) {
+
+  final tracksToAddCount = min(switch (localResult.radioMode) {
```
Komodo5197, line 181: Is clipping the queue like this something we actually want?  I'd imagine this has minimal effect except if you start reshuffle mode, where this clips us from the items native length to 30, which seems undesirable.

lib/services/radio_service_helper.dart
```diff
@@ -152,36 +152,60 @@ Future<void> maybeAddRadioTracks() async {
...
+  if (tracksToAdd.isEmpty) {
     _radioLogger.warning("No tracks generated for radio playback from source '${source.name}'. Aborting.");
     GlobalSnackbar.message((context) => AppLocalizations.of(context)!.radioNoTracksFound);
     return;
   }
 
   toggleRadio(true);
-  invalidateRadioCache();
-  final localResult = _radioCacheStateStream.value!.copyWith(
+  localResult = _radioCacheStateStream.value!.copyWith(
```
Komodo5197, line 198: The logic on this isn't  correct, because our lock could have been preempted in the background, and the cache state never receives the toggleRadio value..  I think instead of the standard identical() check, the correct thing to do here is to just call invalidateRadioCache() a second time?

lib/services/radio_service_helper.dart
```diff
@@ -152,36 +152,60 @@ Future<void> maybeAddRadioTracks() async {
...
+  };
+
+  invalidateRadioCache();
+  var localResult = _radioCacheStateStream.value!.copyWith(generating: true);
+  _radioCacheStateStream.add(localResult);
 
-  final List<BaseItemDto> tracks = [];
   try {
-    tracks.addAll(await generateRadioTracks(radioTracksNeededForInitialQueue, overrideSeedItem: source));
+    localResult.tracks.addAll(
```
Komodo5197: I don't like putting this directly into the cache result, because it should never be used from there.  It should stay in a local list.

lib/services/queue_service.dart
```diff
@@ -829,6 +834,11 @@ class QueueService {
       _savedQueueState = SavedQueueState.saving;
     }
 
+    // avoid radio eagerly adding new tracks from cache (or requesting new tracks) right after the queue is cleared
+    final previousRadioState = FinampSettingsHelper.finampSettings.radioEnabled;
+    FinampSetters.setRadioEnabled(false);
```
Komodo5197, line 838: I don't like changing a persisted setting temporarily like this.  It feels hacky.  I think a better approach would be to make a withRadioLock() method which is basically clearRadioTracks() except it executes a given lambda instead of clearing.  We could also clean up clearRadioTracksLocked() to use that so we can use it directly and don't need clearRadioTracks().

lib/services/radio_service_helper.dart
Chaphasilor: So the reason here was that for modes like Similar, we ended up with 60-90 items because we fetch more than we need and don't have any duplicates yet. That seemed odd to me. I'd like to keep radio queues relatively short, unless they're actually consumed.  

You have a point about reshuffle. It's not like we discard the tracks, but it might look odd.

Komodo5197: Yeah, adding 90 tracks at once does seem a bit high.

lib/services/radio_service_helper.dart
```diff
@@ -301,11 +329,14 @@ Future<List<BaseItemDto>> generateRadioTracks(
...
     }
+    final queueWithoutRadioTracks = currentQueue.fullQueue
+        .whereNot((e) => e.source.type == QueueItemSourceType.radio)
+        .toList();
     // Items originally in the currently playing source (or manually added)
-    final originalQueue = overrideSeedItem != null
-        ? (await loadChildTracksFromBaseItem(baseItem: overrideSeedItem)).map((item) => item).toList()
-        : currentQueue.fullQueue
-              .whereNot((e) => e.source.type == QueueItemSourceType.radio)
+    final originalQueue = actualSeed != null
```
Komodo5197: So I've been thinking a bit more, and neither reshuffle nor random will reliably work properly if chosen via startRadio at the moment.  actualSeed is generally null for reshuffle/random, because if not chosen from startRadio they use the whole queue.  So random will choose 30 random tracks to start, and then just keep returning random from those without ever touching any of the rest from the original source.  And reshuffle will keep doubling the added track list, which could get awkward.  If you add new tracks to next up or something, both will just repeat that indefinitely.

I think the best solution to this is probably to make getActiveRadioSeedProvider check the queue source when in reshuffle/random mode, and if it is of type radio, then return the item, otherwise null.  And then in the generator we use both the child tracks,if we have an item, and non-radio items, if not forNewQueue, as the originalQueue instead of choosing just one of those.  And we never reshuffle radio tracks, because this should always get us something, so if that combo is empty we just throw.

lib/services/radio_service_helper.dart
Chaphasilor: I don't yet understand why we can't check for `identical()` anymore, but I've added the invalidation back

lib/services/queue_service.dart
Chaphasilor: sounds good. by "clean up" you mean moving the radio lock into `clearRadioTracksLocked`?

lib/services/radio_service_helper.dart
Chaphasilor: yeah combining sources seems to make sense for me as well.  
the alternative would probably be something like creating a set of radio tracks to avoid duplicates with each iteration? but that would only fix reshuffle, not random, I guess.

lib/services/queue_service.dart
Komodo5197: Yeah.  I was thinking we rename clearRadioTracksLocked back to clearRadioTracks and it just grabs the radio lock internally.  stopAndClearQueue could also grab the lock internally.  the queueService clearRadioTracks would still need the comment about setting up thge radio state beforehand.

lib/services/radio_service_helper.dart
Komodo5197: I think invalidate is better than the identical check because replacing the queue should have priority, and shouldn't be influenced by the existing radio trying to claim the lock.  Although maybe it shouldn't, and we should have the identical check?  I think this current code is actually vulnerable to a desync if we change radio mode while the new queue is generating. and if something weird like that does occur I'm not sure which result would be preferred.  It probably doesn't matter.  I think the right is probably to force everything through?  So just reset the radio mode setting to currentRadioMode just before the second invalidate.

lib/services/queue_service.dart
Chaphasilor: I guess we can keep using the `withRadioLock` helper, that seems less likely to break compared to full inlining?

Also, what *are* the requirements for the radio state before we call any of this? Might be worth it to list them explicitly

Komodo5197: the actual cache state is private, so yeah.  I just mean the callers shouldn't be worrying about acquiring the lock.

There shouldn't be any limits or anything.  The comment is basically reminding the caller that there's a chance for maybeAddRadioTracks to fire in the async gap after the lock releases before the caller regains execution, so the radio settings should be set up beforehand so that doesn't get immediately invalidated.

Komodo5197:
I noticed your note about rearanging tracks while shuffled.  Can't you just use the same trick as the restore, where you override our shuffle order with updated indices and then call shuffle?  Does player fail to apply the update or something?

Chaphasilor:
You're right, overwriting the `shuffleIndices` does work. I just wasn't familiar with how it worked exactly and didn't give it a try. Thanks for pointing that out!  

Also, I'm not entirely sure if the `fallbackMode` caching I've added is to your taste, so you might wanna take a look at that.  
I'm planning to do a final round of testing tomorrow over the entire day, and then merge this, unless there are issues of course.  
I'll probably try to throw in the system equalizer change if I can get it to work.

lib/services/radio_service_helper.dart
```diff
@@ -321,12 +314,22 @@ Future<List<BaseItemDto>> generateRadioTracks(
     // Items originally in the currently playing source (or manually added)
     final originalQueue = actualSeed != null
         ? (await loadChildTracksFromBaseItem(baseItem: actualSeed)).map((item) => item).toList()
-        // if the queue is purely made up of radio modes (i.e. after using the "Start Radio" option in the menu), we don't filter out radio tracks
-        : (queueWithoutRadioTracks.isEmpty ? currentQueue.fullQueue : queueWithoutRadioTracks)
-              .where((e) => !FinampSettingsHelper.finampSettings.isOffline || e.item.extras?["isDownloaded"] == true)
-              .map((e) => e.baseItem)
-              .nonNulls
+        : <BaseItemDto>[]
```
Komodo5197: I think there's parentheses missing here and in random so that the followedBy applies to the whole ternary result.

lib/services/radio_service_helper.dart
```diff
@@ -428,194 +420,201 @@ Future<List<BaseItemDto>> generateRadioTracks(
...
-    while (filteredSimilarAlbums.isEmpty) {
-      final additionalAlbums = 2 * attempt + pow(attempt, 2.75).toInt(); // ~ 0, 3, 10, 27, 50, 100
-      attempt++;
-      if (attempt > 1) {
-        _radioLogger.warning("No similar albums found. Retrying with $additionalAlbums extra albums.");
-      }
-
-      List<BaseItemDto> similarAlbums;
+    AlbumMixFallbackModes? fallbackMode =
+        _radioCacheStateStream.valueOrNull?.albumMixFallbackMode ??
```
Komodo5197: I believe you need to avoid using this if overrideSeedItem is set?  And maybe we should throw in an assert that actualSeed and the cache seed match.

lib/services/queue_service.dart
```diff
@@ -1103,11 +1103,24 @@ class QueueService {
 
     int adjustedQueueIndex = getActualIndexByLinearIndex(_queueAudioSourceIndex);
```
Komodo5197: This is incorrect.  We want to just use _queueAudioSourceIndex in all places where adjustedQueueIndex is currently being used.

lib/services/radio_service_helper.dart
```diff
@@ -428,194 +420,201 @@ Future<List<BaseItemDto>> generateRadioTracks(
...
-        final randomIndex = _radioRandom.nextInt(min(filteredSimilarAlbums.length, randomnessExtraAlbums));
-        selectedAlbum = filteredSimilarAlbums[randomIndex];
-        filteredSimilarAlbums.removeAt(randomIndex);
+        throw Exception("No suitable similar albums found for album mix radio. Returning empty track list.");
+      }
+    } catch (e) {
+      rethrow;
+    } finally {
+      final cacheState = _radioCacheStateStream.valueOrNull;
+      if (cacheState?.isStillValid() ?? false) {
```
Komodo5197: This isn't good enough.  A valid cache state could still be not ours.  I think the easiest thing to do is capture the state we use at the start and just write to that without touching the stream at all.

Chaphasilor: And then check for that existing state if it's still valid? Since there's an async gap in there.

Komodo5197: If we don't touch the state stream and just update the object we gave, it doesn't matter if its the valid stream object or not, because if it isn't its a no-op.  And unless we want to show this in the ui somehow, bumping the stream isn't actually necessary.  Doing the standard identical check and bumping the stream instead wouldn't hurt, I guess.

lib/services/radio_service_helper.dart
Chaphasilor: When did that go missing, hmm. I definitely added it initially o.O

lib/services/queue_service.dart
```diff
@@ -1101,19 +1101,22 @@ class QueueService {
...
+        itemToMove,
+      );
       try {
         _shuffleOrder.overrideShuffle(newShuffleOrder);
         await _audioHandler.shuffle();
       } finally {
         _shuffleOrder.overrideShuffle(null);
       }
     } else {
+      int adjustedQueueIndex = getActualIndexByLinearIndex(_queueAudioSourceIndex);
```
Komodo5197: You don't actually need  getActualIndexByLinearIndex here either - if the player isn't shuffled, it doesn't do anything.

lib/services/queue_service.dart
```diff
@@ -1101,19 +1101,22 @@ class QueueService {
...
+
       //!!! the player will automatically change the shuffle indices of the ConcatenatingAudioSource if shuffle is enabled, so we need to use the regular track index here
```
Komodo5197: Is this comment applicable anymore?

lib/services/queue_service.dart
```diff
@@ -1101,19 +1101,22 @@ class QueueService {
...
 
-    int adjustedQueueIndex = getActualIndexByLinearIndex(_queueAudioSourceIndex);
-
     if (playbackOrder == FinampPlaybackOrder.shuffled) {
       final newShuffleOrder = [..._shuffleOrder.indices];
-      final int itemToMove = newShuffleOrder.removeAt(adjustedQueueIndex + oldOffset);
-      newShuffleOrder.insert(adjustedQueueIndex + newOffset, itemToMove);
+      final int itemToMove = newShuffleOrder.removeAt(_queueAudioSourceIndex + oldOffset);
+      newShuffleOrder.insert(
+        newOffset < oldOffset ? _queueAudioSourceIndex + newOffset : _queueAudioSourceIndex + newOffset - 1,
```
Komodo5197: I notice that this uses old<new while linear uses new<old, which changes how things are handled if new==old.  I don't think that's a real situation, but using the non-subtracting code for equals is more technically correct, I think?

lib/services/radio_service_helper.dart
```diff
@@ -611,9 +620,8 @@ Future<List<BaseItemDto>> generateRadioTracks(
     } catch (e) {
       rethrow;
     } finally {
-      final cacheState = _radioCacheStateStream.valueOrNull;
-      if (cacheState?.isStillValid() ?? false) {
-        _radioCacheStateStream.add(cacheState?.copyWith(fallbackMode: fallbackMode));
+      if (localRadioState?.isStillValid() ?? false) {
```
Komodo5197: This still isn't correct.  We can't alter the state stream without using identical().  But more importantly, I've realized that altering the state stream at all here is is invalid, because it steals the lock from maybeAddTracks().  So with this current code, album mix mode doesn't ever actually do anything.  What we need to do is `localRadioState.fallbackmode=fallbackmode;` without ever touching the state stream.  Making that variable on the otherwise immutable state is pretty ugly, though.  Another option would be to move the fallbackmode to its own variable, like _radioCallTimestamp, which would be reset separately in invalidateRadioCache().  But then we would need to capture the cache state and check if identical() to know if writing to that was safe, so I think that might be even messier?

lib/services/queue_service.dart
Chaphasilor: true, true...

lib/services/queue_service.dart
Chaphasilor: I guess keeping it consistent is a good idea, yes.

lib/services/radio_service_helper.dart
Chaphasilor: Ahh right I confused isValid with identical, sorry about that.  
Mutating the properties does seem messy, but maybe we can expose the functionality as a method instead, so `localRadioState.setAlbumMixFallbackMode(fallbackMode)`. That feels a bit more deliberate to me? We could also convert all properties to private ones and only expose them via getters, without providing setters.

Komodo5197: That does feel more deliberate, at least.  I'm not 100% sure we need to go that far, be we should at least remove it from the copy method, I think.  Doing getter/setter stuff for the other variables doesn't matter much, because they're all final anyway.

lib/services/radio_service_helper.dart
```diff
@@ -620,8 +620,8 @@ Future<List<BaseItemDto>> generateRadioTracks(
     } catch (e) {
       rethrow;
     } finally {
-      if (localRadioState?.isStillValid() ?? false) {
-        _radioCacheStateStream.add(localRadioState?.copyWith(fallbackMode: fallbackMode));
+      if (localRadioState != null && identical(localRadioState, _radioCacheStateStream.valueOrNull)) {
```
Komodo5197: We shouldn't be updating the localRadioState if we have an overrideSeed.  But if we don't, we should be able to call .updateAlbumMixFallbackMode unconditionally.

Chaphasilor: But that would mean that if we end up in a fallback for the initial queue generation (which could happen e.g. if the user only has singles by that artist), then we'd "forget" about that and only remember the fallback for the next scheduled fetch of radio tracks? Nothing terrible of course.  
But I guess that won't work anyway since we're invalidating the cache after generating the initial queue track in `startRadioPlayback()`

Komodo5197: Yeah, I don't think there's a way we could persist the fallback mode from new radio queues without just returning the fallback mode and letting the caller handle persisting, because the new queue cache state isn't fully set up early enough.  Doing this would complicate the interface, but it would also let the album fallback persisting be cleanly handled along with the generation lock and would get all cache state back out of generateTracks().  Probably not worthwhile?

Chaphasilor: Let's see how the requirements for radio (new modes, etc.) develop based on feedback, and then we can probably better judge which optimizations are worthwhile and which patterns we actually need to support.

PR #1471, opened 2025-12-01
Transcoding screen redesign

brumik:
## Todo before merging

- [ ] Translations

- [ ] Reset Settings


lib/components/menu_shower_toggleable_list_tile.dart
brumik, line 1: This is a copy paste from a different PR, not super usable for this use case, I think when PR ready this is either has to be rewritten or deleted in favour of a different one.

lib/components/TranscodingSettingsScreen/bitrate_selector.dart
brumik, line 1: This is a fix to the current component, this could be actually merged without the rest as it is a generic fix.

Issue without the fix: if you change the `transcodeBitrate` outside of this component, the component does not update and reflect the changes.

PR #1484, opened 2025-12-16
[Redesign] Add Linux ARM build

### LINKED ISSUE #766: Aarch64/Arm64 Linux build
It would be great if it was possible to build Finamp on Arm64 based Linux devices as well. Might be blocked by https://github.com/isar/isar/issues/876 but perhaps there's a way around it.

Komodo5197:
Adds the creation of a linux ARM build into upload_assets.yml.  Fixes #766.  I have tested that this action produces a functional build which loads isar and runs in QEMU, and a build based on the same general procedure has been tested on a PI 5.  I also fixed the windows zip bundle failing to attach to the release.  The general idea behind getting isar working is that you add arm64 as a supported architecture in the dart code, and then you run an isar core build on arm64 and replace the prebuilt x64 libisar.so with the arm one.  The isar core itself seems to build fine on ARM with passing tests, it's just the packaging infrastructure that was the problem.

I imagine flathub would need some updates to use these builds, but I don't know much about that side of things.


Chaphasilor:
Thanks, completely missed this PR, so sorry about the conflict.  
I see it's already tested and in the pipeline, but did you test if the build action actually works? I see no completed action for the commits prior to moving things to just the release artifact upload.

Also, I guess there's no Windows ARM action runner on GitHub yet? ^^  
That would be neat too.

Komodo5197:
It look like windows ARM runners do exist, but isar doesn't seem to immediately compile on windows runners in general and I don't really feel like trying to debug it.  Regarding the tests of the linux builds, those action results are all in my fork.  The final run before I moved them into the release file is https://github.com/Komodo5197/finamp/actions/runs/20254006518

Chaphasilor:
Perfect, then let's merge this :)

PR #1504, opened 2026-01-06
[Redesign] Various remaining TODOs before ->main

lukaslindnermusic:
## Changes

In my [last PR](https://github.com/jmshrv/finamp/pull/1360), I had a few remaining TODOS that I wanted to implement at a later point, with another PR.
Well, life came inbetween, and unfortunately I could not manage to work on Finamp for quite a while due to time reasons :(
However, I might finally be able to slowly check these off now.

## Todo before merging

From the previous PR:
- [x] Make Top Tracks continue to play
- [x] Fix Snackbar message (swipe action on album track) saying that the entire "album [was] added to next up" instead of track.
- [x]  "See All" Artist Top Tracks
- [x] Introduce Sorting Options for the Artist Album List (and playback)
- [ ] ~add optional setting to restore/remember shuffle queue order when starting new queue by tapping on track in album/playlist/artist-top-tracks~

Additionally:
- [x] Change Shuffle Some to Shuffle All (etc.) on Genre Screen when total genre tracks lower than max shuffle all
- [ ] ~Try to use DiscHeader for the TracksSections, with Popup Menu containing playback actions.~

---
- [x] I edited Translations / Strings
---

## Related Issues

- will close [Artist Page Tracks Section only starts queue with visible](https://github.com/jmshrv/finamp/issues/1336)
- ~will close [[Feature Request] Save loop mode: sequential/shuffle](https://github.com/jmshrv/finamp/issues/818) (if @Chaphasilor s [decision](https://github.com/jmshrv/finamp/issues/818#issuecomment-3154795542) there is final ;))~
- will close [Suggestion: All Tracks Listed under Artist](https://github.com/jmshrv/finamp/issues/982)


Chaphasilor:
I have an additional request: can you check if the amount of tracks on the genre screen actually exceeds the fetch limit, and only show the "Shuffle some" if it does? And otherwise just show "Shuffle"

lukaslindnermusic:
> I have an additional request: can you check if the amount of tracks on the genre screen actually exceeds the fetch limit, and only show the "Shuffle some" if it does? And otherwise just show "Shuffle"

yes, that's actually something I wanted to do already back then when I made those changes, but I forgot.
Thanks for reminding me <3 

And no worries, I will keep working on this PR soon 😊

lukaslindnermusic:
@Chaphasilor I'm a little bit unsure about the sorting of albums now.
Adding it for the album list(s) on the artist screen is no big deal. In fact, it seems like I already implemented the sortBy and sortOrder arguments for the providers in question, so its just a matter of adding a setting and passing its value, and adding a button to change it.
However, I wrote "and playback" in paranthesis. 
This opens up a few different possiblities:
For example, let's assume I sort the albums by SortName and have TopTracks MostPlayed, then hitting the normal Play button of an artist could either:
1. Play all tracks of the albums in order, but albums sorted by SortName
2. Play all tracks of all albums, with the SortName of the individual tracks, so albums shuffled up
3. Stick to the sorting of the TopTracks section

I personally would avoid 3, because for me, the Top Tracks section is a decoupled section with separate options.
Also, with what I did two months ago, you can also play all tracks of an artist in the top tracks sorting/filtering now by just tapping one of the five displayed tracks. it expands its playback with that sorting/filtering applied.

What do you think?

Edit: After thinking about it a little bit longer, I think its best to let it primarily apply to the albums list, and then derive the sorting for the artist tracks from it. 
aka.: 
- sorted by release Date -> current behavior, with added SortOrder toggle.
- sorted by SortName -> sort tracks primarily by albumname, not their own sortName.
- sorted by AlbumArtist-> just sort with the fields we have. if individual track returns no album artist, fall back to artist. in most cases the same anyway.
- sorted by Artist: this one would not exist in that scenario, as the sorting is primarily for the albums list.
- orted by duration: thats a tricky one, because tracks dont store its parents album total duration. I guess, in this case we can just sort the tracks after their duration? 

I think, that are already all the sorting options we could apply to it that make sense?

Chaphasilor:
Hmm, yeah that's a bit tricky. So right now we have three options for playback: Play, Shuffle, Shuffle Albums. Which implies that "Play" and "Shuffle" would work on a track-basis, not an album basis, and therefore work like option 2).  
But at the same time I do believe that if you want to play an artist, you'll want to listen to their albums in the original order (i.e. the tracks within the albums stay in that order), and just the ordering of the albums should vary (release date, alphabetic, etc.).

Another problem I'm seeing is that it's not really obvious that the sort order that is selected for each individual section (top tracks, albums) has anything to do with the Playback Actions Row, and more importantly, even if it's obvious, it's still not clear *which* of the two sections controls the "global" order.  
This is partially solved by switching to the "Disc" header, which then includes a play button, that is more closely linked to the sort order of that section.  

What I would propose (if space allows) is the following: Have 4 options per Playback Action Row page in total, those being "Play (Tracks)", "Shuffle (Tracks)", "Play Albums", "Shuffle Albums".  
Shuffle tracks is a true shuffle, shuffle albums keeps the album tracks in the respective order, and the other two play the tracks/albums in the sort order configured in the respective section.  
Does that make sense?

Komodo5197:
Hey, If you're working on things that touch sorting or the music screen, I'll note that there's a pretty solid amount of changes in these areas in the home screen PR.  So it might be easier to build this PR on top of that rather then the main redesign branch.

lukaslindnermusic:
@Komodo5197 I was already worried about that. Yes, I am currently working on the MusicScreen.
Would you suggest that I close this PR and reopen it to target the `redesign-homescreen` branch, or do you recommend another process?

Komodo5197:
On the Github side, I believe you can just change the target branch without reopening in some way?  I don't remember the details.  I feel like the home screen PR might get merged first, but if that happens I guess you can just change the target back.  On the Git side, if it were me I'd probably just git pull the homescreen branch into this and go from there, but I'm not sure everyone would share this opinion and you can of course do whatever you feel makes the most sense.

lukaslindnermusic:
@Chaphasilor what do you think makes most sense?
 
I don't think that the MusicScreen changes in this PR are that much, but it might get annoying to merge if I start doing the sorting stuff now.
Should we merge my PR first and I do the sort artist albums stuff later once the Home Screen is merged? 
Or should I target the homescreen pr and do the conflict resolution now?
Or should I wait until the Home Screen PR is merged and then do the conflict resolving afterwards? 

What do you say, how long until the HomeScreen PR will get merged?

lukaslindnermusic:
As you guys have introduced a new `ItemFilterType` enum where also the genre flter goes into, and I am also introducing an artistFilter in my PR, I think it probably makes most sense if I indeed merge in the home screen and adjust my PR accordingly, and then we merge my PR directly after the HomeScreen PR.

Unless you have a better idea?

Komodo5197:
Sounds good to me.

Chaphasilor:
So I first thought the merge was messed up somehow, but turns out the home screen PR is almost 14k added lines? 😅  
What happened there 🙈

Anyway, I think we could change the target branch of this PR to be the home screen branch to only show the commit delta and not all 128 commits, that might make reviewing easier? And once the home screen PR gets merged, we can change the target branch again.  
@lukaslindnermusic are you fine with me doing that? (or you can do it yourself, it's in the title editing menu at the top).

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -72,11 +78,14 @@ class TrackListTile extends ConsumerWidget {
...
     this.allowDismiss = true,
     this.highlightCurrentTrack = true,
+    this.genreFilter,
     this.playbackProgress,
   });
 
   final BaseItemDto item;
   final FinampPlayable parentPlayable;
+  final bool lazyAddMoreTracksToQueue;
+  final CuratedItemSelectionType? selectedFilter;
```
Komodo5197, line 93: We should be passing down the correct playable through the track sliver list instead of adding these additional filters.

lib/components/MusicScreen/artist_type_selection_row.dart
```diff
@@ -9,17 +9,21 @@ class ArtistTypeSelectionRow extends StatelessWidget {
...
     required this.tabType,
     required this.defaultArtistType,
     required this.refreshTab,
+    this.singleTabConfig,
   });
 
   @override
   Widget build(BuildContext context) {
-    if (tabType == ContentType.genericArtists) {
+    final isArtistTrackList = tabType == ContentType.tracks && singleTabConfig?.sortConfig.artistFilter != null;
```
Komodo5197, line 24: Does this variable do anything?

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -101,6 +113,82 @@ class TrackListTile extends ConsumerWidget {
       playable = true;
     }
 
+    // We lazyload more tracks here if the user starts a queue from one of the top tracks sections
+    // because for performance-reasons, we first only fetch the data for the 5 tracks we really need
+    Future<void> lazyAddMoreTracks(PlayableSlice slice) async {
```
Komodo5197, line 123: This reminded me that I never re-implemented lazy loading for the home sections, so I've added some infrastructure to help support both that use case and this.  I also think a lot of this code could be stripped out by leveraging the correct playable.

Komodo5197:
Hey, I noticed an issue while playing music and thought you might have some thoughts.  Basically, I was playing a song from a different library via a playlist and then tried clicking the artist, but was given an empty artist screen due to all of that artists tracks being in the other library.  Would showing the artist screen with the source tracks library, if we can even determine that, make any sense?  Or maybe even an unfiltered view?  Or would the divergence between the artist screens in different scenarios be confusing or undesirable?  On the other hand, seeing that the artist I was currently playing seemed to have zero tracks was also a confusing.

Chaphasilor:
I've had that happen to me before as well. I usually know that the issue is, but have to go back to the music screen to change the library, which isn't ideal. I'm thinking we could add the library filter similar to how the genre filter is shown, with a single tap to remove it and show tracks from all libraries? That could be a universal thing I guess. Tapping the chip could bring up the library selector instead.  
Just a thought, since I don't see an ideal way to handle this either.

lukaslindnermusic:
Sorry for not responding earlier @Komodo5197 and @Chaphasilor , I did not see your comments immediately and when I finally saw them, at that moment I was too busy and then I forgot again 😵‍💫🙄🙃

I did not have this issue before, as I don't have multiple libraries in use.

Hmm, and we have no way to directly get the library of each item in a playlist, right?
I would not make the library a filter-chip as well. At least not with our current design. It semantically looks and feels as if we are always inside a specific library.

I would suggest that we introduce a simple: 
if (trackCount == 0 && userHasMoreThanOneLib) then show "No tracks by this artist in your currently selected library.    [ Switch Library  ↓]"

Because afaik it normally should not be possible to have an artist with 0 tracks that has NOT some tracks in some other library. So I think its safe to assume: 0 tracks -> must be something in a different library.

I mean: We have empty space that we can use, so I would just put a library dropdown right there. 
And since to me the current layout really feels as if we are inside that library and I personally would avoid showing cross-library stuff as much as possible, I would not simply automatically show stuff from a different library, or worse, auto-switch.

At least thats my opinion on that.

Chaphasilor:
@lukaslindnermusic I think your suggestion is sensible, just showing an empty state with a button that (for now) simply opens the music screen drawer where people can switch to another library.  
And correct, we have no way to tell which library any given item originates from, sadly.

lukaslindnermusic:
I reopen the discussion... when playing an artist via the normal "Play" option, it should play all their tracks, yes. but how should they be sorted?
should they be grouped by album or not? if yes.. should we borrow the sorting from the "Albums" section, or from the "Appears On" section, or ignore both?

currently, they are sorted like the "Albums" section pre-my-PR.

Chaphasilor:
Erik commented on this a few days ago on Discord, stating that you can't really "play" an artist, since it's not a music piece by itself. So the semantic here is somewhat arbitrary. The sorting for the albums is now user-configurable, right? It would probably be nice to match that sorting, although that would be closer to a "Play albums" option than "Play tracks". So I think it would also be fine to keep the old track sorting, if that's easier?

But how about linking this to the sorting chosen on the artist tracks "see all" screen? I think that might be the most sensible, especially if the sorting for that section is already persisted. That would then probably correspond to the tracks section directly on the top-level artist page too, right?  
That's also one solution you mentioned: https://github.com/finamp-app/finamp/issues/1336#issuecomment-3190768315

lukaslindnermusic:
Yeah but I think, making it depend on the curated sections is wrong as well.. you might have RecentlyPlayed but thats not available offline... or favorite-filtered... 

for now, I let it mirror the current behavior I think, aka. it actually is 'Play Albums' and mirrors the sorting on the "Albums" section.

Chaphasilor:
Okay. I thought there was a automatic fallback for unavailable curation modes already, so I don't really get the problem yet. But none of this is set in stone, we can keep iterating on this, especially based on feedback! 😁

lukaslindnermusic:
sorry for letting this sit for so long... 

I'm not happy with the entire artist tracks things as I think the entire codebase got a little bit confusing, but I guess thats more a me-problem / skill issue... but anyway, I hope it at least works.

@Komodo5197 sorry I just saw your review comments now. 😅 Are they still relevant? I think I switched to using the Playable, but regarding the lazyloading of the tracks, I don't know what the infrastructure is that you mentioned that should support this.

Chaphasilor:
Okay, fixed two bugs I noticed, the rest seems to work. I'm not sure if the track order when pressing play is entirely as you described, but it is grouped by album, and I think that's most important. There's still open discussion around this anyway, so no need to make this perfect yet.

I'll go ahead and merge this so we can get a round of testing from other people, and see if anything comes up before the stable release :)

Thanks a ton for putting in the work and getting this finished up in time! <3

Komodo5197:
Why wouldn't they be relevant?  You didn't actually change anything after they were made.

lib/components/MusicScreen/sort_and_filter_row.dart
```diff
@@ -75,6 +75,16 @@ abstract class SortAndFilterController {
   factory SortAndFilterController.trackSettings(ContentType contentType) =>
       TrackingSortAndFilterController(contentType: contentType);
 
+  factory SortAndFilterController.trackArtistAlbums() => ArtistSectionTrackingSortAndFilterController(
```
Komodo5197, line 78: Is this not just Contenttype.inalbumArtistAlbums?  trackArtistAppearsOn should just be a new ContentType as well.

lib/models/finamp_models.dart
```diff
@@ -942,6 +950,18 @@ class FinampSettings {
   @HiveField(152, defaultValue: "unset") // pre-generation default
   String deviceId;
 
+  @HiveField(153, defaultValue: DefaultSettings.artistScreenAlbumsSortBy)
```
Komodo5197, line 953: These should just be in the standard sort setting with relevant ContentTypes, not independent settings.

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -127,14 +136,28 @@ class _ArtistScreenContentState extends ConsumerState<ArtistScreenContent> {
     );
   }
 
+  List<BaseItemDto> _applySortAndFilterLocally(List<BaseItemDto> items, SortAndFilterConfiguration config) {
```
Komodo5197, line 139: Why isn't this happening in the getArtistAlbums/getPerformingArtistAlbums prioviders?

lib/components/ArtistScreen/artist_screen_content.dart
```diff
@@ -218,15 +244,23 @@ class _ArtistScreenContentState extends ConsumerState<ArtistScreenContent> {
...
+        ? _applySortAndFilterLocally(appearsOnAlbumsList, appearsOnSortConfig)
+        : <BaseItemDto>[];
 
-    var appearsOnAlbums = performingArtistAlbums.where((a) => !albumArtistAlbums.any((b) => b.id == a.id)).toList();
+    final allPerformingArtistTracks = allPerformingArtistTracksAsync ?? [];
 
     // Combine Children to get correct ChildrenCount
     // for the Download Status Sync Display for Artists
-    allChildren = [...albumArtistAlbums, ...allPerformingArtistTracks];
+    allChildren = [...(albumArtistAlbumsAsync ?? []), ...allPerformingArtistTracks];
```
Komodo5197, line 263: Is this intentionally avoiding the sort for some reason?  Why aren't we using albumArtistAlbums?

lib/components/ArtistScreen/artist_screen_content.dart
lukaslindnermusic: I first tried that but I remembered that there were issues before with cached provider data being returned in a different order.

lib/services/music_providers.dart
```diff
@@ -401,14 +401,20 @@ Future<List<Track>> getChildTracks(Ref ref, {required FinampUnpagedDisplayable<T
...
+      final children = await ref
+          .watch(
+            getArtistTracksProvider(
+              artist: item.item,
+              libraryFilter: item.library,
+              genreFilter: item.sortConfig.genreFilter?.id,
+              onlyFavorites: item.sortConfig.favoritesFilter,
+            ).future,
+          )
+          .then((tracks) {
```
Komodo5197, line 415: Why are we sorting separately instead of using the sort option in the provider?

lib/models/finamp_models.dart
lukaslindnermusic: sorry, i was not aware that ContentType can be expanded... especially since the other sort configs are stored in the hive field called "**tab**SortBy" and "**tab**SortOrder" but the artist screens album and appears on sections certainly are not **tabs**.

lib/components/ArtistScreen/artist_screen_content.dart
lukaslindnermusic: it is just used for the track count in the header, but I guess we can also use the sorted data.

lib/services/music_providers.dart
lukaslindnermusic: see what I commented at your comment above.
i remembered there being a cache issue where cached provider data was returned in a different order for some reason.
of course, I could've cleaned up the provider in order to remove not used code. sorry!

lib/components/MusicScreen/sort_and_filter_row.dart
lukaslindnermusic: no... the intention is to have these sections remember their own sort settings and operate independent from the tabs on the MusicScreen... besides the favorte filter that we somehow treat globally everywhere...

lukaslindnermusic:
> Why wouldn't they be relevant? You didn't actually change anything after they were made.

as said, i did not see these comments before.. maybe in the meantime, you changed something again? also, I did switch to passing the Playable instead, but apparently that was not what you meant with your comment... which justifies the question again if they are still relevant.

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -105,6 +118,84 @@ class TrackListTile extends ConsumerWidget {
...
```
Komodo5197, line 123: This code seems to be broken for genres currently - it adds a bunch of nonesense.

lukaslindnermusic: you are right.. weird, it worked before. I did not test it again after merging in the redesign branch again. were there changes regarding the genre filter?

lib/components/AlbumScreen/track_list_tile.dart
Komodo5197: I mostly mean leveraging a PreCachedPlayableSlice instead of a BasePlayableSlice.  This could also be better integrated into the Artist playable so that we can use lazyloading/pagination elsewhere, although that's more complex.  A lot of this code is a likely unnecessary copy of the existing artist/genre loading code.  PreCachedPlayableSlice also already has radio management baked in during startslicePlayback, so you wouldn't need to manually pause the radio.

lib/components/MusicScreen/sort_and_filter_row.dart
Komodo5197: Yeah, we probably need to fix up the favorite filter to not be global.  But anyway, inAlbumArtistAlbums is not a music screen tab, it's exactly this situation.  It was added to support showing the albums of an album artist as a home screen section.

lib/models/finamp_models.dart
Komodo5197: Yeah, I guess I should have renamed the setting?  But ContentType used to be TabContentType, but I ended up expanding it a bit in support of showing non-tab things on the home screen.

lib/components/ArtistScreen/artist_screen_content.dart
Komodo5197: Can you elaborate?

lukaslindnermusic: it is already some time ago so don't quote me on that. but if you look at the code before my PR, there were already sort options for that provider but they were unused as well. The provider loaded all tracks on ArtistScreen open, and then later when you hit play with the playback actions that reused the same provider with the same settings, the sort order wasn't right anymore, which is why i introduced that sortArtistTracks function back then to manually restore the same order after the provider returned the data.

I remembered that there were issues back then, and then I decided to just call it outside of the provider again to be save.

lukaslindnermusic: i think this happened before this PR, back when we replaced all the old play buttons with the playback action row everywhere.

lib/components/AlbumScreen/track_list_tile.dart
Komodo5197: I'm not sure of the state when you were writing, but you seem to have introduced or reintroduced the genreFilter argument on the tracklisttile, which isn't actually used anywhere.  I believe you should be using the generic parentItem instead.  Although really, what should be happening is that we feed the tracklisttile a better parentPlayable that can set up the lazy loading itself.

Komodo5197:
Sorry, I don't think I was ever fully explicit about what I was actually thinking about.  Basically, the current design is that we're passing all the track loading information down that stack to tracklisttile as selectedFilter, genrefilter, and the lazyAddMoreTracksToQueue bool.   Then the tracklisttile has to include a bunch of track loading logic to handle getting the tracks using this information.  However, I don't really think this logic has any place in tracklisttile.  Instead, I think the caller, by which I mean artist_screen_content, should be providing a better FinampPlayable with all this information actually included.  Then the basic getPlayableSlice->startSlicePlayback would be able to handle lazily loading and playing the tracks without the tracklisttile needing to be concerned.  This is how the music screen currently works - note that there is no explicit logic for the tracks list lazy loading inside of tracklisttile.  Right now, the parentPlayable the tile gets is generated in TracksSliverList without any information about where the tracks actually came from, so this would need to change.

Komodo5197:
I guess I can just refactor the loading myself in the future if it seems too complicated - the current code isn't broken or anything, it just has a bunch of duplicate, out of scope work.  I guess apart from the incorrect genreFilter use.  I think there are actually a couple interesting questions relating to how the precached tracks get preserved - we either have a custom playable or try to leverage a reused paging provider - so this isn't particularly straightforward.

lib/components/ArtistScreen/artist_screen_content.dart
Komodo5197: That seems more clean.  But I have noticed that at this point, we do seem to be fully loading all performing and artist tracks just to count them?  This seems very inefficient.

lib/components/ArtistScreen/artist_screen_content.dart
Komodo5197: We would actively need client-side sorting if we're trying to reuse the same load, but in that case I'd just remove the sort options from the provider entirely.  But if we're planning to make artists use paged loading for tracks, like genres do, then we need to reliably do all sorting server-side.

lib/components/ArtistScreen/artist_screen_content.dart
Chaphasilor: Is this the reason why loading huge artists takes a few seconds where it only shows a spinner? Maybe we could show more fine-grained spinners instead.

Chaphasilor:
@Komodo5197 I did see there's still a todo about adding more documentation for the Playable and Slices stuff, so that would probably help with understanding what should be used, and how. I fear I've lost track of how this works myself, so I'm not of much help to others in that regard (even less than usual 😅).

lib/components/ArtistScreen/artist_screen_content.dart
Komodo5197: I think this only affects the header, and the body still loads at the speed of the albums - it does have to wait for both of those full lists, of course.  But I don't see any reason to do this when we could use childCount or totalRecordCount or something.

Komodo5197:
Yeah, I'm sorry I haven't gotten to that yet.  The changes I'm talking about actually get into the weeds a bit more than I'd like, because the whole point is that you should be able to pass the full play context around easily without thinking about what exactly you were handling too much.  The hope is that the original source with full context, say the artist_screen_content, can create a playable, and then all the other widgets like tracklisttile can easily and generically pass it around without worrying about sorting or lazy loading or anything like that.  Eventually, it would be passed into queueservice on tap without any of the intermediate widgets knowing or caring about what they're playing.

lib/components/ArtistScreen/artist_screen_content.dart
lukaslindnermusic: the reason is because we have to combine a unique set of tracks between albumartist albums and performing artist tracks.

so in regards of the providers.. in online mode, it is just that one additional query to get all album artist tracks... and we reuse that same provider for playback anyway, so I would consider it "priming the cache" for if the user hits Play :P but yeah, it intentionally is not blocking the UI rendering, and also, if we want to optimize it, we can now add a different provider that only fetches the total Count.
this provider was introduced when the old play button that directly used this result was still present, so we now have indeed a potential for a little bit of performance improvement.

lukaslindnermusic:
thanks for specifying this, Komodo. It was indeed not directly clear what your intention was/is. But it sounds like a good abstraction to me. 

maybe, once we have finally moved redesign to main, we can work on a big codebase cleanup where we optimize these things... 
one thing that bugs me personally is that by now we have a lot of different places where tracks or other items are fetched in different ways and it is totally messy and not clear why it gets fetched where in what way... for example, some menus use the item helper, while a lot of stuff now uses the music providers, while the musicscreen still uses inline queries etc. 

in general, I feel like you are introducing a lot of architecturally great things, but because the codebase in general is messy and also not really well documented, it is hard to actually know what options exist for what etc.

so maybe this could be a post redesign task for us?

lib/components/AlbumScreen/track_list_tile.dart
Chaphasilor: If the fix is just to use the proper parent item, I can try to get that in before James builds the artifacts! Although I'm busy with a move, so no promises 😅

Komodo5197: Yeah, I think the item used for the genre filter just needs to be switched.  It would also be nice to get rid of the extra settings before they get permanently locked in, but that's a bit more work.  I believe everything else I commented on is closer to code cleanup than anything noticeable by users, so it can wait.

Chaphasilor: I got the genre fix in, hopefully I didn't miss anything.  
About the new settings, there seems to be a bunch of custom sortBy settings that are only partially deprecated, so I think don't really need to avoid adding these now and can just get rid of all the obsolete settings in one go.  
But if you want to keep the settings IDs cleaner @Komodo5197, maybe you still have some time before James gets around to build?

Komodo5197: Okay, I've pushed a commit that moves the sort settings to be with the others.  We'll need to roll it back if any of the builds have already happened.

Regarding the obsolete settings, the ID numbers can never be reused, and I don't see any reason we'd remove migrations that we've already written, so I don't think the obselete settings are going anywhere.  I'm not sure what you mean by partly deprecated?  But anyway, I would like to avoid adding more of them if possible.

PR #1490, opened 2025-12-18
Fix flutter run on flake.nix

brumik:
## Changes

Sorry for running the formatter (this is the default nixpkgs formatting, maybe we can keep it, if not I can revert it).

Main changes:
- Added shellHook to link the .so files so flutter run works
- added `dbus` and `pkg-config` to avoid compilation errors

## Todo before merging

- [ ] Translations

- [ ] Reset Settings

- [ ] Extended CONTRIBUTING.md

PR #1509, opened 2026-01-11
Add setting to reset "previousTracksExpaned" each time a queue is opened

HeroBrine1st:
## Changes

- Update nix flake and CONTRIBUTING.md to allow contributing from NixOS at least with workarounds
- Add setting to reset "previousTracksExpaned" each time a queue is opened (this behavior is disabled by default but the setting is inversed). 
- Add a workflow check for codegen verification

<details>
<summary>Screenshot</summary>
image

</details>

## Todo before merging

- [x] Add a workflow check which checks that code generation was done properly; or
- [ ] check that change from `..writeByte(137)` to `..writeByte(138)` is okay.

- [x] Translations

- [x] Reset Settings

## Related Issues

Not related to any issues

### DIFF
--- a/.github/workflows/pr-healthchecks.yml
+++ b/.github/workflows/pr-healthchecks.yml
@@ -35,3 +35,35 @@ jobs:
           # Format locally generated localization files to prevent false positives
           dart format lib/l10n/ 
           dart format --output=none --set-exit-if-changed lib/
+  check-codegen:
+    name: Ensure correct code generation
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: '17'
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential cmake ninja-build libgtk-3-dev
+      - name: Set up Flutter
+        uses: subosito/flutter-action@v2
+        with:
+          channel: 'stable'
+      - name: Verify Flutter setup
+        run: flutter doctor
+      - name: Get dependencies
+        run: flutter pub get
+      - name: Verify code generation
+        run: |
+          dart run build_runner build --delete-conflicting-outputs
+          git add . # new files are not tracked by git
+          if git diff --cached --name-only | grep -q '\.g\.dart$'; then
+            echo "Code generation is different, diff follows"
+            git diff --cached | cat
+            exit 2
+          fi
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,16 +14,46 @@ flutter doctor -v
 This will list out all components used for Flutter development, and if they are installed. This should include either Android or iOS development (depending on your platform), as well as any desktop platform you want to target.  
 You can ignore the Chrome/Web component, since Finamp is not a web app.
 
-### NixOS and Flakes
+### NixOS
 
 The project includes a `flake.nix` file that can help you install Flutter and Rust dependencies. The recommended way of using it is to
-use `nix develop`, which allows rustup for Discord RPC dependency. Also, there's `nix develop .#fenix` which stubs rustup with fenix, but it is a hack.
+use `nix develop`, which allows rustup for Discord RPC dependency. Also, there's `nix develop .#fenix` which stubs rustup with [fenix](https://github.com/nix-community/fenix/), but it is a hack.
 
 To get the application running on NixOS once you are in the development shell:
 - `flutter build linux` - generates the build files in `./build/linux/x64/release/bundle` where the `lib` folder will have all the dynamic libraries.
-- `cd /build/linux/x64/release/bundle/lib` - flutter cannot find the dynamic libraries by default. Probably works from here only because the fallback is CWD.
-- `../flutter` - start the application.
-- The normally-used `flutter run ...` command does not seem to work on NixOS at this time.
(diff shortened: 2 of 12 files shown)


Chaphasilor:
Great, another test that will lead to failed PR run notifications 🙃  
/s  
Thanks for these changes. The codegen check is probably good, although I wonder if it's really necessary to run this on every commit (then again, it's probably more lightweight than the full builds).  

I'll try to slightly revise the wording for the setting tomorrow before merging.  
Also, does this PR supersede #1490 ?

HeroBrine1st:
> to run this on every commit

I wish there was a way to reuse "images" of actions because, as you can see, I literally copied first job and simply replaced actual check. And it doesn't change much between different revisions

> Also, does this PR supersede https://github.com/jmshrv/finamp/pull/1490 ?

Only parts that add dbus and update flake.lock file, the remaining can be merged but there will be conflict because it also updates flake.lock file. My flake.lock is simply newer so theirs[^1] can be discarded cleanly.

[^1]: I'm not english native so if it's not appropriate to say it like that please tell me :P I actually recalled "accept yours" "accept theirs" in IDEA conflict resolution window

Chaphasilor:
Alright. Yeah Docker-like layers to share would be neat. Maybe there's a way set multiple check results from a single action somehow? We won't be able to consolidate most other actions anyway since they run on different arches and OSs, but for the checks it really is more of just the same, and could all be added to the linux build action...

And "theirs" is absolutely fine :)

---

Regarding the PR, why didn't you put the overwrite at https://github.com/herobrine1st/finamp/blob/285ebcc2859d0c2863b4c52e68e42b24d57d568f/lib/components/PlayerScreen/queue_list.dart#L175? That seems like a much cleaner solution.  
Additionally, the setting toggles if toggling the queue list is remembered - technically that could include people wanting to *always* open the previous tracks list, instead of never? So I'm wondering if instead of the boolean setting, we should just have a dropdown for "When the queue panel is opened, previous tracks are..." and then 3 options, "Shown", "Hidden", "Like before"?  
And then we always track the previous state, but ignore it depending on the setting?

That might be cleaner, simpler to understand, and more flexible?  
But let me know what you think, I'm rather impartial here. I'll change the wording once we've settled on a solution :)

HeroBrine1st:
> Regarding the PR, why didn't you put the overwrite at

Because tunnelvision :-) It could be worse - the first variant used "whenComplete", and I'd send that if I didn't stumble on codegen

I like your thoughts. My use case is just "wow antistress" since it has very satisfying close animation and when it scrolls past previous tracks instead of closing it's hurting.. again, tunnelvision :P but always opening is probably good for.. desktops! At least while there's no dedicated desktop UI

UPD: I probably know why I tunnelvisioned. I didn't see remembering that header is expanded as feature until I opened the code and saw that it is an actual value in settings DTO, and thus a feature, not a bug.

HeroBrine1st:
So it looks like this

<details>
<summary>Screenshot</summary>
image
</details>

Regarding this:

> Regarding the PR, why didn't you put the overwrite at https://github.com/jmshrv/finamp/blob/285ebcc2859d0c2863b4c52e68e42b24d57d568f/lib/components/PlayerScreen/queue_list.dart#L175?

It's not viable here because it is a callback, so overriding the value where it currently is is the most clean as `showQueueBottomSheet` function is a chokepoint between 4 usages and another function from dependency.

It can also be done here:

https://github.com/jmshrv/finamp/blob/fedebf8a02e4450c0fc586de52da3ad14f67fd78/lib/components/PlayerScreen/queue_list.dart#L103-L118

But then it is violation of responsibility I think, because it changes application state while initState is for Widget state I guess

Chaphasilor:
I honestly wouldn't mind having it in `initState`, that seems easier to find to me. But I'm curious what @Komodo5197 thinks about it?  

The dropdown looks fine, thanks! I'll change the wording to be a bit less technical.

Komodo5197:
I feel like initState makes the most sense, because the widget is what actually uses the value, and this is basically just widget state we're persisting.  If the queue was embeded in the player screen, as an example, we'd still want this code to run as the queue loads, and initState would do that.

Chaphasilor:
@HeroBrine1st would you mind adding the change? Otherwise I'll try to do it myself some time tomorrow :)

@Komodo5197 I see I'm not the only one who thought about embedding the queue on the player. So you think it's worth tackling (in a separate PR of course), along with some kind of "maximize" button for the player screen that temporarily overrides the split screen setting?

HeroBrine1st:
> @HeroBrine1st would you mind adding the change? Otherwise I'll try to do it myself some time tomorrow :)

It was postponed due to extreme lack of time actually which will become even worse soon. I can try but if it takes more than an hour I'll postpone it again.

See me in ~16 hours.

Komodo5197:
Yeah, I feel like I've seen requests for the queue screen on the player around, and it does make a lot of sense.  A maximize button also makes sense, especially in combination with the player screen queue.  I've also seen a request to get the lyrics screen working better in landscape which might use similar layout code, but that seems a lot less important.  I've been wanting to do this for a while, but I've haven't had that much free time and my priorities are helping with the home screen PR first, and then trying to get the download sync into a background isolate, so I'm not handling this anytime soon.

HeroBrine1st:
Well, it works. Hive even fixed field id change (from 144 to 145), I expected I'd need to clear data folder.
Idk why codegen fails, lemme add debug output on error.

Chaphasilor:
@HeroBrine1st thanks for taking the time to deal with this. I know it's hard to find enough time for working on the app, so I really appreciate that you managed to finish this up!  
Everything seems to be working, so let's merge this!

PR #1510, opened 2026-01-11
Fix race condition in queue restoration code

HeroBrine1st:
## Changes

Discard failed queue restoration state if user already started playback by that point.

This race condition is possible on slow connections, for example but not limited by, when user went to offline mode on a network with high latency; then, starting local playback leads to queue override by queue restoration. 

This can be possible in online mode too, for example, due to changing networks, or if user retries queue restoration and instantly launches playback from already-loaded album.

### DIFF
--- a/lib/services/queue_service.dart
+++ b/lib/services/queue_service.dart
@@ -437,7 +437,7 @@ class QueueService {
     }
     _queueServiceLogger.finest("Loading stored queue: $info");
 
-    SavedQueueState? finalState = SavedQueueState.failed;
+    SavedQueueState finalState = SavedQueueState.failed;
     try {
       _savedQueueState = SavedQueueState.loading;
       if (info.trackCount == 0) {
@@ -558,7 +558,6 @@ class QueueService {
       int droppedTracks = info.trackCount - loadedTracks;
 
       if (_savedQueueState != SavedQueueState.loading) {
-        finalState = null;
         return Future.error("Loading of saved Queue was interrupted.");
       }
 
@@ -592,16 +591,17 @@ class QueueService {
         }
       }
     } finally {
-      if (finalState != null) {
+      // Do not overwrite state or bump reporting if loading was interrupted by the user starting a new queue
+      if (_savedQueueState == SavedQueueState.loading) {
         _savedQueueState = finalState;
+        if (finalState == SavedQueueState.failed) {
+          _failedSavedQueue = info;
+        }
+        refreshQueueStream();
+        await Future<void>.delayed(const Duration(seconds: 1)).then((_) {
+          unawaited(playbackHistoryService.reportRestoredSessionStatus());
+        });
       }
-      if (finalState == SavedQueueState.failed) {
-        _failedSavedQueue = info;
-      }
-      refreshQueueStream();
-      await Future<void>.delayed(const Duration(seconds: 1)).then((_) {
-        unawaited(playbackHistoryService.reportRestoredSessionStatus());
-      });
     }
   }
 


HeroBrine1st:
Dart format was done from wrong CWD :/

Also race condition was not reproduced but the change is trivial enough.

Komodo5197:
This isn't correct, because replacing an existing queue with a restored one is an intended feature in the queue restore screen, and this completely blocks that.  Can you give me more info on the race condition?  Because the code already attempts to prevent unintended overwrites via the _savedQueueState variable.  It looks like there might be a possible path where a local queue could get overwritten with the failed state if getItems throws, but I'm not sure how you would get an overwrite with a queue that actually restores.

HeroBrine1st:
> existing queue with a restored one is an intended feature in the queue restore screen

Oops forgot about that

> get overwritten with the failed state if getItems throws

Looking at my report somewhere ago, it is really what happens. I simply forgot that it only happens in failing case

HeroBrine1st:
So yes, easily reproducible given it fails, music continues to play though. Lemme debug and fix that properly

<details>

<summary>Patch and screenshot</summary>

```diff
diff --git a/lib/services/queue_service.dart b/lib/services/queue_service.dart
--- a/lib/services/queue_service.dart	(revision 8e5e0f611f87fe73fb933e7d3c424c316a3d11bf)
+++ b/lib/services/queue_service.dart	(date 1768155529713)
@@ -488,6 +488,8 @@
           }
         }
       } else {
+        await Future.delayed(Duration(seconds: 10));
+        return Future.error("stub");
         List<jellyfin_models.BaseItemDto> itemList = await _jellyfinApiHelper.getItems(itemIds: missingIds) ?? [];
         for (var d2 in itemList) {
           idMap[d2.id] = d2;
```

8e5e0f611f87fe73fb933e7d3c424c316a3d11bf is from my different PR but I think it doesn't interfere

image
</details>

HeroBrine1st:
There are two ways to fix that: one of them is committed, the other is slight refactoring (remove `finalState` via extracting `catch` case). I went with minimal change. Looks a bit like band-aid though but fixes the issue.

I also interpret the whole `finally` block as something that is done only if queue restoration task is not cancelled so moved all of it under `if`

P.s. dev pro tip: re-read bug report before fixing

lib/services/queue_service.dart
```diff
@@ -592,16 +591,19 @@ class QueueService {
...
       }
     } finally {
-      if (finalState != null) {
-        _savedQueueState = finalState;
-      }
-      if (finalState == SavedQueueState.failed) {
-        _failedSavedQueue = info;
+      // Don't do anything if it's late
+      if (_savedQueueState == SavedQueueState.loading) {
+        if (finalState != null) {
```
Komodo5197: I think we can make finalState non-nullable now.

lib/services/queue_service.dart
```diff
@@ -592,16 +591,19 @@ class QueueService {
         }
       }
     } finally {
-      if (finalState != null) {
-        _savedQueueState = finalState;
-      }
-      if (finalState == SavedQueueState.failed) {
-        _failedSavedQueue = info;
+      // Don't do anything if it's late
```
Komodo5197: Maybe `// Do not overwrite state or bump reporting if loading was interrupted by the user starting a new queue.`

lib/services/queue_service.dart
HeroBrine1st: I am too reliant on IDEA inspections which aren't available for Dart :-( It was made obvious by mere "show usages"

Should look good now :P

Komodo5197:
looks good to me.  If you're finished up, I'll go ahead and merge it.

HeroBrine1st:
Yes, it's tested for failing case so it's complete

PR #1511, opened 2026-01-12
Fix duplicate listen events on iOS

### LINKED ISSUE #1172: Duplicate AudioPlayback and AudioPlaybackStopped events
The redesigned app seems to be sending an `AudioPlayback` event, immediately followed by an `AudioPlaybackStopped` event, then the real `AudioPlayback` event.

Image

This has a direct impact on the ListenBrainz integration, as the playbacks are double-reported.

Note that this does not happen every time, just most times.

- Jellyfin Server: 10.10.7
- Finamp: 0.9.15 (TestFlight build 115)
- iOS 18.3.2

Coloradohusky:
## Changes
This PR adds a loop-mode check to `playback_history_service.dart`, as described by https://github.com/jmshrv/finamp/issues/1172#issuecomment-3434831790. As described in the issue, the client would report multiple listens for the same song, even when not in loop mode - however, this issue seemed to only appear on iOS devices.

## Related Issues
- closes #1172 
- possibly related to #688

### DIFF
--- a/lib/services/playback_history_service.dart
+++ b/lib/services/playback_history_service.dart
@@ -131,7 +131,7 @@ class PlaybackHistoryService {
             prevItem?.id == currentItem.id &&
                 // current position is close to the beginning of the track
                 currentState.position.inMilliseconds <= 1000 * 10) {
-              if ((prevState.position.inMilliseconds) >= ((prevItem?.item.duration?.inMilliseconds ?? 0) - 1000 * 10)) {
+              if (((prevState.position.inMilliseconds) >= ((prevItem?.item.duration?.inMilliseconds ?? 0) - 1000 * 10)) && (_queueService.loopMode == FinampLoopMode.one)) {
                 // looping a single track
                 // last position was close to the end of the track
                 updateCurrentTrack(currentItem, forceNewTrack: true); // add to playback history


Chaphasilor:
@Coloradohusky thanks for the PR. Please see my comment in the original issue for further info.  

Essentially, this code:

https://github.com/coloradohusky/finamp/blob/cca584dcc651695f513396babbfbdff95bcc98e8/lib/services/playback_history_service.dart#L75-L81
(which handles the playback reporting)

fires before this code:

https://github.com/coloradohusky/finamp/blob/cca584dcc651695f513396babbfbdff95bcc98e8/lib/services/queue_service.dart#L119-L132
(which triggers a rebuild of Finamp's internal queue based on the current player index)

But because the playback reporting uses `QueueService.getCurrentTrack()` before the queue service actually handled the `playbackState` event itself, it obtains the wrong track. I'm guessing the order of playback event changes (e.g. current index, duration progress) are different between iOS and Android, and therefore we only see this bug on one platform.  

Are you willing to work on a proper fix? I can also do that, but I don't know when I'll get around to it :)

Coloradohusky:
Thanks for the notes! I'll take a look tonight and see what I can figure out  
Edit: I'm having issues with the Macbook I use for development, so I'll be delayed in fixing this :(

Chaphasilor:
Seems like this was picked up in #1519 :)

Coloradohusky:
Good point! Thanks for the notice

PR #1141, opened 2025-04-01
✨ Home Screen

Chaphasilor:
Finally had the time to work on this. Finamp will soon™️have an actual home screen!

This closes #612.

The home screen is based the design mockup proposed in https://github.com/jmshrv/finamp/issues/220#issuecomment-1814714778.  
It will be user-configurable content-wise. The two rows of buttons ("1-click play actions" - top row, and "library links" - bottom row) will be configurable, as well as the "sections" below the buttons.  
1-click play actions start playing something right-away (possibly after choosing an option, like the decade to play).  
Library links open a place in the library. They are not meant for linking to specific items (e.g. albums), but it's not off the table yet.  
The sections will contain a dynamic list of items. The items within a section will not be explicitly user-configurable, they will always be powered by some kind of request to the server (or downloads). The sections do however support displaying collections, which could be used to "pin" specific items if desired, while integrating neatly with existing Jellyfin features and allowing to sync the content across devices.

The basic home screen already works, only thing not functional yet are the "show more"/`>` buttons and the "Decade Mix".  
Customization is not there yet, but will most likely be added before this PR is merged.

The new navbar is used on (almost) all other screens, it already works but the navigation experience still needs improvements.

Feedback and suggestions welcome!

Screenshots:

TODO:

- [x] Implement sort bar on tab view (with bottom sheets)
  - [x] Fix buttons
  - [x] Add quick toggle for sort order
- [x] Consolidate fetching server and user info
- [x] Improve home screen settings
  - [x] Add more section presets (crowdsource?)
  - [x] Fix home screen settings layout on desktop
  - [x] Fix edit button for home screen actions
  - [x] Make it clearer that only collections can be selected
- [x] Check list keys to prevent reusing old data (e.g. genre) with new UI element (e.g. tracklisttile)
- [ ] Look into using existing higher-res images if available?
- [x] Fix grid and grid + text
  - [x] Force re-layout of home screen after toggling grid text
  - [x] Consider grid text setting for skeleton loader
  - [x] Show grid text as fallback if no cover exists
  - [x] Separate grid text setting for home screen and other tabs
    - [x] Always ignore setting for tracks
  - [x] Fix overflow on home screen during load (especially if text off)
  - [x] Bring back fixed-width grid tiles?
- [x] Add migrations?
- [x] Set sensible defaults
- [x] Prevent opening collections/items on tap in home screen section collection picker
- [x] Fix not being able to save Jellyfin collections in home screen section collection picker
[...]

PR #1515, opened 2026-01-16
Handle android.intent.action.MUSIC_PLAYER in AndroidManifest

Maxr1998:
This allows starting Finamp from the “Play music” quick settings action on Samsung One UI devices. Currently, this doesn't start any playback, however.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -51,9 +51,12 @@
       </intent-filter>
 
       <intent-filter>
-        <action android:name="android.intent.action.MAIN"/>
-        <category android:name="android.intent.category.APP_MUSIC"/>
-        <category android:name="android.intent.category.LAUNCHER"/>
+        <action android:name="android.intent.action.MAIN" />
+        <action android:name="android.intent.action.MUSIC_PLAYER" />
+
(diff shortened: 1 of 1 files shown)


Chaphasilor:
@Maxr1998 do you think it would make sense to add this to #1141 and then have this intent trigger the `Surprise Me` action I added there (which picks a random track an starts Continuous radio with it)?  
I'd like to actually attach a proper handler if we add the capability.

Maxr1998:
Sure, feel free to just cherry pick it if you prefer. I also think it makes more sense to directly integrate it.

Chaphasilor:
So how would I go about handling the intent? Are you familiar with this in Flutter?

Maxr1998:
The intent from the system would need to be handled in the `onCreate` and possibly `onNewIntent` callbacks within the activity, and then call to the Flutter side from there. This shouldn't be to hard, I can look into it when I have some time.

Chaphasilor:
@Maxr1998 if I implement it, can you help test it?

Maxr1998:
> @Maxr1998 if I implement it, can you help test it?

Sure, I can do that. Unfortunately didn't have much time to look at it myself, sorry 😔

Chaphasilor:
alright. I'll tackle this in #1141 then!

Issue #24, opened 2021-04-14
Carplay support?


JohnGalt1717:
Really need this in my car like PlexAmp.

Would be really great if I can script a shortcut on iOS to start this on carplay connect and resume music where I last left off automatically.

Thanks!

UnicornsOnLSD:
The audio service plugin I'm using says it supports Carplay/Android Auto, but there isn't really any documentation about it (ryanheise/audio_service#593). Once there's documentation, I'll look into it :)

itscaro:
+1 for `Android Auto`

stefan1983:
I would also love to have CarPlay support! Would be nice to have the app be able to run and select the tracks on the car play screen. 

Is there anything planned? 
thanks a lot

UnicornsOnLSD:
Won't be able to fit it in for this release unless I postpone it even longer, so 0.7 it is (if its even possible, CarPlay is weird)

stefan1983:
OK.

Dedicated CarPlay support would be nice, for now the pure playback works so far:

![image](https://user-images.githubusercontent.com/8542806/142018374-9911a0bd-e889-4be3-a328-23caef1581ca.png)

But one has to switch between tracks / albums etc. directly on the phone.

stefan1983:
If you need someone to test and play around, let me know ;)

UnicornsOnLSD:
I'm going to remove this from the 0.7 milestone due to the amount of work it's going to need, there are other issues that need focus right now. I'll still keep this feature in mind - it's a pretty common request :)

kaysond:
Bumping for support!

TCB13:
+1 on this feature! It would be very nice to have a basic UI to pick albums and initiate playback directly from the CarPlay UI instead of having to pick the phone. Thank you.

Luca-vz:
> OK.
> 
> Dedicated CarPlay support would be nice, for now the pure playback works so far:
> 
> ![image](https://user-images.githubusercontent.com/8542806/142018374-9911a0bd-e889-4be3-a328-23caef1581ca.png)
> 
> But one has to switch between tracks / albums etc. directly on the phone.

How are you able to open your app on carplay? My app crashes as soon as I open it. I followed the documentation and even went through the official Carplay Programming Guide

jmshrv:
You can't just open an app directly in Carplay, you have to program a separate UI for it. I'm not sure how it works in detail, but it's definitely a very involved process.

[...]

PR #1474, opened 2025-12-07
[superseded by #1516] Add CarPlay support

llim1:
## Changes

Adds CarPlay support using [flutter_carplay](https://github.com/oguzhnatly/flutter_carplay) package. 

## Todo before merging

- [x] install dependencies & configure iOS project
- [x] basic now-playing screen  
- [x] library tab: albums list 
- [x] library tab: artists list 
- [x] library tab: genres list 
- [x] library tab: playlists list  
- [ ] library tab: tracks list
- [ ] search tab (likely requires changes to flutter_carplay)
- [ ] recently played tab
- [ ] home tab? 
- [x] support offline mode
- [ ] replace static text with localizations 
- [ ] rich 'now-playing' screen inc. queue management + album art (likely requires changes to flutter_carplay)
- [ ] now-playing icon on list entries 

### Known issues
Will cross off issues here once they are addressed. 
- flutter_carplay loads images from web URIs but file URIs are broken despite file paths being listed as supported 
- 'now-playing' screen has inconsistent behaviour on iOS simulator. Need build to physical hardware to test
- 'now-playing' button in root tab view is missing the icon, despite being clickable

- [ ] Translations

- [ ] Reset Settings

- [ ] Extended CONTRIBUTING.md

## Related Issues
- Closes #24

PR #1516, opened 2026-01-17
CarPlay Support w/ llim1 base

APIUM:
Adds CarPlay support using [flutter_carplay](https://github.com/oguzhnatly/flutter_carplay) package.

This PR is based on the work from llim1 https://github.com/jmshrv/finamp/pull/1474 adding the finishing touches to get it ready to merge.

*Todo before merging*

  - [x] install dependencies & configure iOS project
  - [x]  basic now-playing screen
  - [x]  library tab: albums list
  - [x]  library tab: artists list
  - [x]  library tab: genres list
  - [x]  library tab: playlists list
  - [x]  library tab: tracks list
  - [x]  ~~search tab (likely requires changes to flutter_carplay)~~ - Replaced this with Siri support, this can come in a follow up PR (Done in this PR now)
  - [ ]  ~~recently played tab~~ - follow up PR
  - [x]  home tab?
  - [x]  support offline mode
  - [ ]  ~~replace static text with localizations~~
  - [ ]  ~~rich 'now-playing' screen inc. queue management + album art~~ (likely requires changes to flutter_carplay) - follow up PR
  - [ ] ~~now-playing icon on list entries~~ - follow up PR
   
*Known issues*

Will cross off issues here once they are addressed.

  - [x] flutter_carplay loads images from web URIs but file URIs are broken despite file paths being listed as supported
  - [x] 'now-playing' screen has inconsistent behaviour on iOS simulator. Need build to physical hardware to test - Fixed
  - [x] 'now-playing' button in root tab view is missing the icon, despite being clickable
  - [x]  Translations
  - [x]  Extended CONTRIBUTING.md
  
*Related Issues*

Closes https://github.com/jmshrv/finamp/issues/24
Original PR https://github.com/jmshrv/finamp/pull/1474


lib/services/android_auto_helper.dart
```diff
@@ -552,7 +552,8 @@ class AndroidAutoHelper {
     return mediaItems;
   }
 
-  Future<void> playFromMediaId(MediaItemId itemId) async {
+  // optional startingIndex to for carply only: allows specifying starting track for album playback
+  Future<void> playFromMediaId(MediaItemId itemId, {int? index = 0, FinampPlaybackOrder? order = FinampPlaybackOrder.linear}) async {
```
llim1: All changes in this file can be reverted, this is left over from my first attempt where I was trying to reuse the android auto codebase. It is unused now that `carplay_helper.dart` has `playItem()`

pubspec.lock
llim1, line 1: Most of these dependency changes are to older versions than those used on `redesign` latest. Is there a reason for this? If not, consider cleaning the diff to only add the `flutter_carplay` fork and any other new dependencies.

lib/services/queue_service.dart
```diff
@@ -290,13 +290,23 @@ class QueueService {
...
           final packageInfo = await PackageInfo.fromPlatform();
           artUri = Uri(scheme: "content", host: packageInfo.packageName, path: artUri.path);
         }
         if (!force && _audioHandler.mediaItem.valueOrNull?.id != currentMediaItem?.id) return;
-        currentMediaItem = currentMediaItem?.copyWith(artUri: artUri);
+        currentMediaItem = currentMediaItem?.copyWith(
+          artUri: artUri,
+          extras: {
+            ...currentMediaItem?.extras ?? {},
+            'artCacheFile': artCacheFilePath,
```
llim1: is `artCacheFile` actually used anywhere? I don't see any other references and afaik iOS already had full album art support

ios/Runner/RunnerDebug.entitlements
llim1, line 1: This file is from my draft, should be able to remove it as the debug runner profile is already set to point to `Runner.entitlements`

llim1 (commented):
Some comments, also a few notes: 

- known issue "_flutter_carplay loads images from web URIs but file URIs are broken despite file paths being listed as supported_" is marked as fixed but issue is still present in my environment. Some artwork shows a '?' image where visible 
- jellyfin api related changes not reviewed
- siri search not tested

pubspec.lock
APIUM: I have just been running `flutter pub get`, so I had assumed it was just getting the latest, I've run it again here and have some newer ones.

ios/Runner/RunnerDebug.entitlements
APIUM: Done

lib/services/queue_service.dart
APIUM: No, not anymore. Thanks, Done.

lib/services/android_auto_helper.dart
APIUM: Done

APIUM:
Thanks for the review @llim1 

> known issue "flutter_carplay loads images from web URIs but file URIs are broken despite file paths being listed as supported" is marked as fixed but issue is still present in my environment. Some artwork shows a '?' image where visible

I misunderstood this originally, I am now able to recreate it, when there were spaces encoded in the url it broke. I've added a fix for this.

llim1:
> Thanks for the review @llim1
> 
> > known issue "flutter_carplay loads images from web URIs but file URIs are broken despite file paths being listed as supported" is marked as fixed but issue is still present in my environment. Some artwork shows a '?' image where visible
> 
> I misunderstood this originally, I am now able to recreate it, when there were spaces encoded in the url it broke. I've added a fix for this.

Confirmed, images issue is fixed. No further comments from me 👍

Chaphasilor:
@APIUM @llim1 do any of you already have access to the entitlements? I think Logan hasn't created the developer account yet, but what about Algy?

Chaphasilor:
@APIUM you mentioned on Discord that we can go back to using the upstream plugin instead of your fork, right?

APIUM:
> @APIUM you mentioned on Discord that we can go back to using the upstream plugin instead of your fork, right?

Yep I’ll make this change in here

APIUM:
> @APIUM @llim1 do any of you already have access to the entitlements? I think Logan hasn't created the developer account yet, but what about Algy?

I’m waiting on Apple to approve my personal account, they look about 3 days to ask for ID, and it’s been over a week since then. They say it’s a 2 day process…. I’m all set up in the team so I’ll be able to test very quickly once Apple approves.

networkoctopus:
really looking forward to this in Finamp. Thanks for working on it! 

Can I cheekily request having an instant mix button available on CarPlay. Would love if it was easy to start a good mix (especially for those who use the new audiomuse-ai backend).

Haar:
image

Just been running this PR locally on a Simulator before trying it out on actual Hardware. Quick question - what's the difference between Shuffle all and Instant Mix?

APIUM:
> Just been running this PR locally on a Simulator before trying it out on actual Hardware. Quick question - what's the difference between Shuffle all and Instant Mix?

Hi @Haar thanks for having a look. 

Shuffle all is just a standard shuffle of all the music in your library. 
Instant Mix seeds a Jellyfin 'Instant Mix' with a random album from a list of 10. That list is made from 5 random albums, and 5 frequently played albums. I think this behavior will change as I think Chaphasilor is working on some Instant Mix functionality in the home screen redesign. 

@networkoctopus see above for the current Instant Mix situation

Harr please let me know if you find a way to test on real hardware, I am having trouble with profiles and certificates. Still working this out and it's why this PR has stalled.

ios/Runner/AppDelegate.swift
```diff
@@ -23,20 +40,240 @@ import Flutter
...
+
+// MARK: - Siri Media Intent Handling
+// Handles voice commands like "Hey Siri, play [song/artist] on Finamp"
+
+private var siriIntentChannel: FlutterMethodChannel?
+
+extension AppDelegate {
+    func setupSiriIntentChannel() {
+        siriIntentChannel = FlutterMethodChannel(
+            name: "com.unicornsonlsd.finamp/siri_intent",
```
Chaphasilor: I'm not sure how exactly things work on iOS, but could this cause issues with debug and release versions using the same channel?

ios/Runner/AppDelegate.swift
```diff
@@ -23,20 +40,240 @@ import Flutter
...
+            binaryMessenger: flutterEngine.binaryMessenger
+        )
+    }
+
+    // MARK: INPlayMediaIntentHandling
+
+    /// Resolves media items for Siri so it knows the app can handle the request.
+    /// Without this, Siri says "That action is not supported on Finamp."
+    func resolveMediaItems(for intent: INPlayMediaIntent, with completion: @escaping ([INPlayMediaMediaItemResolutionResult]) -> Void) {
+        // If Siri provided media items, accept them
```
Chaphasilor, line 143: I'm guessing this works similar to Google Assistant on AA, where it will try to fetch actual metadata for a request and pass that along? On Android Auto we will first search based on that provided metadata, and then if there is no match, try with the "raw" query as a fallback. We also use the properties of the metadata to determine what kinds of items to search for, e.g. if something has a title, and artist, and an album, it's probably a tracks, and it if has a title and an artists it's probably an album.  
For reference, this is the relevant code:  
https://github.com/lukaslindnermusic/finamp/blob/c2e9fa846ec485572a4560e0795260a0d8c55e37/lib/services/android_auto_helper.dart#L263-L371

You might already be doing this, I haven't looked at the full code yet...

ios/Runner/Info-Debug.plist
```diff
@@ -62,6 +62,18 @@
 	<string>Required for debugging</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>I don't know why this permission is being requested, you should probably deny it. If you can, create a GitHub issue with a screenshot of this message.</string>
+	<key>NSSiriUsageDescription</key>
+	<string>Finamp uses Siri to play music via voice commands like "Play [song/artist] on Finamp".</string>
```
Chaphasilor: Just a quick note, the terminology we try to use it "Track" instead of "Song", since tracks include music that is instrumental. Would be nice if you stayed consistent with that ^^

lib/services/audio_service_smtc.dart
```diff
@@ -1,10 +1,10 @@
 import 'package:audio_service_platform_interface/audio_service_platform_interface.dart';
-import 'package:flutter/cupertino.dart' hide RepeatMode;
-import 'package:smtc_windows/smtc_windows.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:smtc_windows/smtc_windows.dart' as smtc_windows;
```
Chaphasilor: the simpler change (which is already in upstream) is just

```suggestion
import 'package:flutter/cupertino.dart' hide RepeatMode;
import 'package:smtc_windows/smtc_windows.dart';
```

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,798 @@
...
+        id: "carplay-tracks-${DateTime.now().millisecondsSinceEpoch}",
+      ),
+      order: order,
+      startingIndex: order == FinampPlaybackOrder.linear ? index : null,
+    );
+    await FlutterCarplay.showSharedNowPlaying();
+  }
+
+  /// CarPlay-optimized shuffle that fetches a smaller initial batch for fast startup.
+  /// The default shuffleAll() fetches 250 tracks which is slow on poor connections.
```
Chaphasilor: maybe we should pass a `limit` argument to the default shuffle handler instead of reimplementing this from scratch?

lib/services/ios_helpers.dart
```diff
@@ -0,0 +1,301 @@
...
+    return true;
+  }
+
+  /// Shows the player screen in the Flutter app after Siri-initiated playback.
+  static void _showPlayerScreen() {
+    final context = GlobalSnackbar.materialAppNavigatorKey.currentContext;
+    if (context != null && context.mounted) {
+      NowPlayingBar.openPlayerScreen(context);
+    }
+  }
```
Chaphasilor: what's the reasoning behind this? we already have the `autoExpandPlayerScreen` setting for handling this. I don't see why voice interactions should override that setting?

lib/services/ios_helpers.dart
```diff
@@ -0,0 +1,301 @@
...
+
+  static const _siriShuffleLimit = 30;
+
+  /// Fast shuffle (same approach as CarPlay) - fetches 30 random tracks instead of 250.
+  static Future<void> _shuffleAll() async {
```
Chaphasilor: why is this implemented again and not reused?

ios/Runner/AppDelegate.swift
UnicornsOnLSD: Finamp's Bundle ID on iOS is also `com.unicornsonlsd.finamp-ios` (for historical reasons - some people tried to get it onto Testflight before I could, so the proper Bundle ID was taken!)

It'd might make more sense to use `Bundle.main.bundleIdentifier` and append the intent path

Chaphasilor: If that works, sure. Yeah the slightly different ID is also what I was thinking of. But to be fair, we also use the regular ID for all message channels on Android, and I think those are scoped to the app's processes anyway?

ios/Runner/AppDelegate.swift
APIUM: Yeah, have updated this now so it works much closer to AA (trying to build a query which just uses the AA code) which had a much better system on review

ios/Runner/Info-Debug.plist
APIUM: Done

lib/services/audio_service_smtc.dart
APIUM: Done

lib/services/carplay_helper.dart
APIUM: I've realised now that there is a setting for the shuffle size, so I'm not keen on overwriting the users selection. I will try again with the 250, and if it's still slow will add another setting which defaults to 30 or so.

lib/services/ios_helpers.dart
APIUM: Missed that, I had that setting off which is why it wasn't working for me. I'd like it to default to on in the car through, possibly another setting? In the car it's much nicer to have it default to move, where on the phone I don't mind the extra click.

lib/services/ios_helpers.dart
APIUM: I don't remember anymore and can't work out my reasoning, so maybe there was none - have fixed as now using system shuffle with the user-defined limit.

ios/Runner/AppDelegate.swift
APIUM: Done (used `name: "\(Bundle.main.bundleIdentifier!)/siri_intent",`)

lib/services/carplay_helper.dart
Chaphasilor: I think it would be fine to disregard the setting when starting music via CarPlay. But maybe you could add 30 tracks first, and then load another 250 tracks (or however the setting is configured), remove any duplicates, and then add the remaining tracks (until 250) to the queue asynchronously?

lib/services/ios_helpers.dart
Chaphasilor: Alright, yeah I see that point. Iirc Android Auto also always shows the player screen when starting new playback. So maybe it's really best to override the setting here.

APIUM: Ah good, OK I will revert this change then for now and we can split it into a setting if there is pushback.

lib/services/carplay_helper.dart
APIUM: OK good, happy with that too, same as autoExpandPlayerScreen. I haven't put work into it yet, but surely there is some way. I have noticed other players don't seem to do this, maybe there is a technical limitation. I'll revert it back to 30, then like below we can expand later.

30 would generally give 1-4h of playback depending on the type of music in the library, so it's not too onerous to have to hit shuffle again in the meantime.

APIUM: Yeah I tried 250 again today on the build and it was very slow. Over 20s on the 2 bars of 4g outside my house.

lib/services/ios_helpers.dart
APIUM: Done

lib/services/carplay_helper.dart
APIUM: Done

Chaphasilor: 1-4h might be enough for most, but I've heard from people going on road trips several times now (mainly US I guess), where they end up listening for 6-8 hours. And I'd like for the music to just keep going.  
But I guess if they have radio enabled that would also ensure that :)

Chaphasilor:
@APIUM seems like this is coming along nicely! If you resolve the merge conflicts you'll also get CI runs again :)

lib/services/carplay_helper.dart
APIUM: I agree I do intend to fix this, will make a ticket for the work and look at some options

APIUM:
> @APIUM seems like this is coming along nicely! If you resolve the merge conflicts you'll also get CI runs again :)

Done! Needs approval to run

Chaphasilor:
also, can you update the todo list at the top to reflect the current status?  
oh and don't worry if the code gen - splash screen check fails, that's not your fault

APIUM:
> also, can you update the todo list at the top to reflect the current status? oh and don't worry if the code gen - splash screen check fails, that's not your fault

Sorry, didn't see this. Have just pushed the formatting fix.

ios/Runner/AppDelegate.swift
```diff
@@ -23,20 +41,253 @@ import Flutter
...
+
+extension AppDelegate {
+    func setupSiriIntentChannel() {
+        siriIntentChannel = FlutterMethodChannel(
+            name: "\(Bundle.main.bundleIdentifier!)/siri_intent",
+            binaryMessenger: flutterEngine.binaryMessenger
+        )
+    }
+
+    // MARK: INPlayMediaIntentHandling
```
Chaphasilor: Are all these `MARK`s for collapsible sections in your editor?  
If so, maybe consider removing them before merging.

ios/Podfile.lock
```diff
@@ -70,9 +70,9 @@ PODS:
     - DKImagePickerController/PhotoGallery
     - Flutter
   - Flutter (1.0.0)
-  - flutter_discord_rpc (0.0.1):
+  - flutter_carplay (1.1.0):
     - Flutter
-  - flutter_native_splash (2.4.3):
```
Chaphasilor, line 75: This should probably not be removed?

ios/Podfile.lock
```diff
@@ -123,8 +123,8 @@ DEPENDENCIES:
   - DKImagePickerController (= 4.3.4)
   - file_picker (from `.symlinks/plugins/file_picker/ios`)
   - Flutter (from `Flutter`)
+  - flutter_carplay (from `.symlinks/plugins/flutter_carplay/ios`)
   - flutter_discord_rpc (from `.symlinks/plugins/flutter_discord_rpc/ios`)
-  - flutter_native_splash (from `.symlinks/plugins/flutter_native_splash/ios`)
```
Chaphasilor, line 127: same here, and below

lib/models/finamp_models.dart
```diff
@@ -111,6 +111,7 @@ class DefaultSettings {
   static const androidStopForegroundOnPause = true;
   static const onlyShowFavorites = false;
   static const trackShuffleItemCount = 250;
+  static const quickShuffleItemCount = 30;
```
Chaphasilor, line 119: maybe add a description of what this is used for? as in, why do we need this.

lib/services/jellyfin_api.dart
```diff
@@ -172,6 +172,15 @@ abstract class JellyfinApi extends ChopperService {
...
     @Query("CollapseBoxSetItems") bool? collapseMultiDiscAlbums,
+
+    /// Optional. Filter by items whose name is sorted equally than a given input string.
+    @Query("NameStartsWith") String? nameStartsWith,
+
+    /// Optional. Filter by items whose name is sorted equally or greater than a given input string.
+    @Query("NameStartsWithOrGreater") String? nameStartsWithOrGreater,
+
+    /// Optional. Filter by items whose name is equally or lesser than a given input string.
+    @Query("NameLessThan") String? nameLessThan,
```
Chaphasilor, line 183: are those real filters? who came up with these names? 😅

lib/services/queue_service.dart
```diff
@@ -123,7 +123,7 @@ class QueueService {
       _currentQueueIndex = event.queueIndex ?? 0;
 
       // Ignore playback events if queue is empty.
-      if (previousIndex != _currentQueueIndex && _currentTrack != null) {
+      if (_audioHandler.audioSources.isNotEmpty && (previousIndex != _currentQueueIndex || _currentTrack == null)) {
```
Chaphasilor, line 126: Did you change this condition on purpose? the `_currentTrack == null` is not equivalent to the old behavior

lib/services/queue_service.dart
```diff
@@ -742,8 +742,10 @@ class QueueService {
       }
 
       if (Platform.isIOS || Platform.isMacOS) {
-        // Both iOS and macOS will start playing the first queue index if we don't stop first
-        await _audioHandler.stopPlayback();
+        // Both iOS and macOS will start playing the first queue index if we don't stop first.
+        // Use pause() instead of stop() to keep the audio session active and prevent
+        // other apps (e.g. Podcasts) from briefly resuming during Siri handoff.
+        await _audioHandler.pause();
```
Chaphasilor, line 751: Can you confirm that this doesn't mess up playback activity reporting to the server? Just look at the admin dashboard and then start playing a new queue, it should say "has finished playing <old>" and then "started playing <new>"

lib/services/queue_service.dart
```diff
@@ -766,7 +768,10 @@ class QueueService {
       if (beginPlaying) {
         // only open the player screen if we actually start playing, otherwise it would open after startup + queue restore
         if (FinampSettingsHelper.finampSettings.autoExpandPlayerScreen) {
-          unawaited(NowPlayingBar.openPlayerScreen(GlobalSnackbar.materialAppNavigatorKey.currentContext!));
+          final navContext = GlobalSnackbar.materialAppNavigatorKey.currentContext;
```
Chaphasilor: I think there are some other places where we assume the nav context to always be available, have you searched for them?

lib/main.dart
```diff
@@ -412,6 +415,10 @@ Future<void> _setupPlaybackServices() async {
 
   // Begin to restore queue
   unawaited(queueService.performInitialQueueLoad().catchError((dynamic x) => GlobalSnackbar.error(x)));
+
+  if (Platform.isIOS) {
+    GetIt.instance.registerSingleton<CarPlayHelper>(CarPlayHelper());
+  }
```
Chaphasilor, line 449: could you move this to be just below the android auto helper?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+          quality: _carPlayImageQuality,
+        )
+        ?.toString();
+  }
+
+  /// Pre-downloads unique images for a list of items to local cache.
+  /// Returns a map of imageId -> local file URI string.
+  /// Items sharing the same imageId are downloaded only once.
+  Future<Map<String, String>> _preloadCarPlayImages(List<BaseItemDto> items) async {
+    final Map<String, String> imageUriMap = {};
```
Chaphasilor: I think Finamp already has an internal image/artwork cache, could you use that one instead?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+    List<BaseItemDto> items,
+    CPListItem Function(BaseItemDto item, int index) itemBuilder,
+  ) {
+    Map<String, List<CPListItem>> grouped = {};
+
+    for (int i = 0; i < items.length; i++) {
+      final item = items[i];
+      final name = item.name ?? "";
+      String letter = name.isNotEmpty ? name[0].toUpperCase() : "#";
+      if (!RegExp(r'[A-Z]').hasMatch(letter)) {
```
Chaphasilor, line 122: does that match diacritics too?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+  Future<List<BaseItemDto>> getTabItems({required TabContentType tabContentType}) async {
+    final sortBy = FinampSettingsHelper.finampSettings.getTabSortBy(tabContentType);
+    final sortOrder = FinampSettingsHelper.finampSettings.getSortOrder(tabContentType);
+
+    // If we are in offline mode, display all matching downloaded parents
+    if (FinampSettingsHelper.finampSettings.isOffline) {
+      List<BaseItemDto> baseItems = [];
+      for (final downloadedParent in await _downloadsService.getAllCollections()) {
+        if (baseItems.length >= _carPlayOfflineLimit) break;
+        if (downloadedParent.baseItem != null && downloadedParent.baseItemType == tabContentType.itemType) {
```
Chaphasilor: Is there no way to filter this within the download system itself?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+          : _finampUserHelper.currentUser?.currentView,
+      includeItemTypes: tabContentType.itemType.jellyfinName,
+      sortBy: sortBy.jellyfinName(tabContentType),
+      sortOrder: tabContentType == TabContentType.tracks ? null : sortOrder.toString(),
+      limit: _carPlayOnlineLimit,
+    );
+    return items ?? [];
+  }
+
+  // Fetch artist tracks directly via API, bypassing Riverpod providers
```
Chaphasilor: Is there a specific reason you want to avoid the providers?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+
+    await _queueService.stopAndClearQueue();
+
+    if (FinampSettingsHelper.finampSettings.isOffline) {
+      // Offline: instant mix not available, fallback to shuffle
+      await shuffleAllTracks();
+      return;
+    }
+
+    // Fetch 1 random track and start continuous radio from it.
```
Chaphasilor, line 231: If you want to enable radio, then please rename the method. "Instant Mix" is a specific Jellyfin feature and pre-fetches similar tracks in advance. Radio fetches tracks just-in-time, and the Continuous radio mode does not yield the same results as an instant mix. For that, you should use the "Similar" mode instead.

Chaphasilor: On the other hand, if this is for the "Instant Mix" quick action, then you should just fetch the items from the correct API instead, or add an `itemCount` parameter to the existing instant mix method of the audio service helper.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+
+    final albums = await _jellyfinApiHelper.getItems(
+      parentItem: _finampUserHelper.currentUser?.currentView,
+      includeItemTypes: "MusicAlbum",
+      sortBy: "DateCreated",
+      sortOrder: "Descending",
+      limit: limit,
+    );
+    return albums ?? [];
+  }
```
Chaphasilor, line 271: Many of these things can get rewritten to use the new MusicScreenProviders once the home screen PR lands. Are you up for that?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+
+  List<FinampQueueItem> getRecentPlays({int limit = 5}) {
+    return _queueService.peekQueue(previous: limit);
+  }
```
Chaphasilor, line 277: where is this used? If the queue is replaced, this might be an empty list. The playback history service could be another source.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+                items: [baseItem],
+                source: QueueItemSource(
+                  type: QueueItemSourceType.nextUp,
+                  name: QueueItemSourceName(
+                    type: QueueItemSourceNameType.preTranslated,
+                    pretranslatedName: baseItem.name ?? _l10n.tracks,
+                  ),
+                  id: baseItem.id,
+                  item: baseItem,
+                ),
```
Chaphasilor: This seems incorrect. You'd not adding the tracks to Next Up, so the source is wrong. And you probably want to reuse some sort of instant mix or radio helper instead of only playing a single track?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,824 @@
...
+          onPress: (complete, self) {
+            final parentId = MediaItemId.fromJson(jsonDecode(item.id) as Map<String, dynamic>);
+
+            switch (parentId.contentType) {
+              case TabContentType.albums:
+              case TabContentType.playlists:
+              case TabContentType.genres:
+                showAlbumsTemplate(tabType: parentId.contentType);
+              case TabContentType.artists:
+                showArtistsTemplate();
```
Chaphasilor, line 404: traditionally genres in Finamp were closer to artists than to albums. What made you use the albums template for them here?

Chaphasilor (changes requested):
Okay, found some things that should probably be updated, but most things look good from my perspective!

lib/services/carplay_helper.dart
APIUM: Do you mean you'd get a list of tracks rather than albums? I picked albums as it gives the user a lot more choice with the constrained list length on CarPlay

lib/services/carplay_helper.dart
APIUM: Yeah good catch. I have changed online to use startInstantMixforItem() and offline to just play the single.

lib/services/carplay_helper.dart
APIUM: Yes that will be good. Happy to do this. I will make a general ticket for this.

lib/services/carplay_helper.dart
APIUM: Yeah that's fair. It was for 'Instant Mix' on the UI. I would like to avoid the queue size issue again so I have renamed this Radio Mix

lib/services/jellyfin_api.dart
APIUM: Haha to be fair I'm not sure I could have done better!

lib/models/finamp_models.dart
APIUM: Done

ios/Podfile.lock
APIUM: Done

ios/Podfile.lock
APIUM: Missed a flutter pub get after rebase, fixed

ios/Runner/AppDelegate.swift
APIUM: Yep, I have removed them if that's the standard.

lib/services/queue_service.dart
APIUM: Yeah, I was having the queue not rebuilding sometimes on CarPlay and was left with an empty now playing screen

lib/main.dart
APIUM: Done

lib/services/carplay_helper.dart
APIUM: Yeah this does but it's not clear. CarPlay needs URIs of the album art not the image. I have made this clearer.

lib/services/carplay_helper.dart
APIUM: No, comes up under #

lib/services/carplay_helper.dart
APIUM: Edit: on reviewing this again I think you're right I can't work out why it's this way. Will address. 

Edit2: I did try this before, but it make shuffle really slow, it now works differently and I think should be OK

lib/services/carplay_helper.dart
APIUM: I missed the filter on that, there is, have updated to use it

lib/services/carplay_helper.dart
APIUM: Yeah you're right. It's used on the CarPlay home screen. I have changed it to that.

lib/services/queue_service.dart
APIUM: Can confirm. Reporting looks fine in the dashboard after doing this.

lib/services/carplay_helper.dart
APIUM: I didn't really pick up on this, the CarPlay home screen is only built once and there aren't any tracks in there, so it doesn't update. I think I'd rather just make it better in the updated Home Screen so I'm going to leave this as-is (still changing to the playback history service)

lib/services/queue_service.dart
APIUM: Yeah, there are a couple but I didn't want to overstep with this change. I can fix those too if you are happy with the scope increase

lib/services/carplay_helper.dart
APIUM: #1586 

I created a few, I can't add any organisation to them so went with overarching tickets for features to add.

lib/services/carplay_helper.dart
Komodo5197: I think it's better to resolve the URI with something more like _providers.read(albumImageProvider(AlbumImageRequest(item: item, maxHeight: 200, maxWidth: 200))).uri, which is what android auto does.  This covers things like the player image cache, which you missed with this logic.  By the way, have we verified that carplay doesn't have an internal cache if we give identical URIs repeatedly?

ios/Runner/AppDelegate.swift
```diff
@@ -23,20 +41,247 @@ import Flutter
...
 //    Awkwardly make a mutable copy of the dir
     var mutableDir = dir
-    
+
     var values = URLResourceValues()
     values.isExcludedFromBackup = isExcluded
     try mutableDir.setResourceValues(values)
 }
+
+// TODO: This is a workaround because audio_service doesn't set MPNowPlayingInfoCenter.playbackState on iOS.
```
Komodo5197, line 94: It looks like there's already a PR fixing this https://github.com/ryanheise/audio_service/pull/1140, so it might be easier to strip this code out and just switch to that fork until it's merged?

lib/services/ios_helpers.dart
```diff
@@ -0,0 +1,183 @@
...
+    if (arguments == null) {
+      _logger.warning("Siri searchMedia called with null arguments");
+      return;
+    }
+
+    final query = arguments['query'] as String?;
+    _logger.info("Siri searchMedia - query: $query");
+
+    // For now, just play the search result (same as playFromSearch)
+    // In the future, this could navigate to a search results screen
```
Komodo5197: Can we get a proper TODO comment here?

lib/services/ios_helpers.dart
```diff
@@ -0,0 +1,183 @@
...
+      query: query,
+      artist: artist,
+      album: album,
+      genre: genre,
+      mediaType: mediaType,
+    );
+
+    _logger.info("Siri delegating to AA search - rawQuery: $rawQuery, extras: $extras");
+    final androidAutoHelper = GetIt.instance<AndroidAutoHelper>();
+    await androidAutoHelper.playFromSearch(AndroidAutoSearchQuery(rawQuery, extras));
```
Komodo5197, line 115: It's probably cleaner to move the android auto logic into a more general method that takes a bunch of query strings instead of building the exact android auto input format here.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+
+  ProviderSubscription? _userSubscription;
+
+  bool get isUserLoggedIn => _finampUserHelper.currentUser != null;
+
+  final _queueService = GetIt.instance<QueueService>();
+
+  /// Returns the image URI string for a CarPlay list item.
+  /// Checks downloaded images first, then falls back to network URL.
+  String? _getCarPlayImageUrl(BaseItemDto item) {
```
Komodo5197: There seem to be some methods in this class that call this directly, and thus will never contribute to or draw from the carplay image cache.  These should probably all be consolidated into a single path.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+  void disposeCarplay() {
+    _userSubscription?.close();
+    _flutterCarplay.removeListenerOnConnectionChange();
+  }
+
+  void onConnectionChange(ConnectionStatusTypes status) {
+    connectionStatus = status;
+    if (status == ConnectionStatusTypes.connected) {
+      // Resume playback if there's a loaded queue that's paused
+      final audioHandler = GetIt.instance<MusicPlayerBackgroundTask>();
```
Komodo5197, line 96: It might be worth adding a setting for disabling autoplay on re-connection.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+    if (!isUserLoggedIn) {
+      _carPlayLogger.info("User not logged in, showing login prompt on CarPlay");
+      await _showLoginRequiredTemplate();
+      return;
+    }
+
+    // Fetch home sections and library items in parallel
+    final results = await Future.wait([
+      _buildHomeSections(),
+      GetIt.instance<MusicPlayerBackgroundTask>().getChildren(AudioService.browsableRootId),
```
Komodo5197, line 384: getChildren with the root ID is just returning the section headers, not the library items, so it doesn't actually take time to complete.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+    if (_isPushingPageUpdate) {
+      _carPlayLogger.warning("Navigation dropped: already pushing page update");
+      return;
+    }
+    _isPushingPageUpdate = true;
+    try {
+      List<BaseItemDto> mediaItems = await loadChildTracksFromBaseItem(baseItem: parent);
+
+      // Pre-cache images — deduplicates shared album art across tracks
+      final imageMap = await _preloadCarPlayImages(mediaItems);
```
Komodo5197: Does this completely block showing the page until all the images are loaded?  Is there a way to avoid that while still handling image downloading ourselves?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+        emptyViewTitleVariants: [_l10n.login],
+        emptyViewSubtitleVariants: [_l10n.carPlayLoginPrompt],
+        systemIcon: 'person.crop.circle.badge.exclamationmark',
+      ),
+    );
+
+    await _flutterCarplay.forceUpdateRootTemplate();
+  }
+
+  Future<void> showPlaylistTemplate(BaseItemDto parent) async {
```
Komodo5197: This appears to cover both albums and playlists.  This should be renamed, or at least have a header comment added.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+
+      CPListTemplate playlistTemplate = CPListTemplate(sections: [playlistSection], systemIcon: 'gear');
+
+      await FlutterCarplay.push(template: playlistTemplate);
+    } finally {
+      _isPushingPageUpdate = false;
+    }
+  }
+
+  Future<void> showAlbumsTemplate({required TabContentType tabType, BaseItemDto? genreFilter}) async {
```
Komodo5197: This appears to cover what the main app calls the music screen tabs for albums playlists, and genres, and it also has the display of individual genres crammed in.  It needs a better name and more documentation.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+    if (_isPushingPageUpdate) {
+      _carPlayLogger.warning("Navigation dropped: already pushing page update");
+      return;
+    }
+    _isPushingPageUpdate = true;
+    try {
+      List<BaseItemDto> mediaItems;
+      if (genreFilter != null) {
+        if (FinampSettingsHelper.finampSettings.isOffline) {
```
Komodo5197, line 518: You might be able to just use the genreCuratedItems provider for this.  You can at least look at that to see how to retrieve genre albums from getAllCollections without filtering.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+      _carPlayLogger.warning("Navigation dropped: already pushing page update");
+      return;
+    }
+    _isPushingPageUpdate = true;
+    try {
+      List<BaseItemDto> tracks;
+      if (FinampSettingsHelper.finampSettings.isOffline) {
+        tracks = await getTabItems(tabContentType: TabContentType.tracks);
+        if (tracks.length > _carPlayOnlineLimit) {
+          tracks = tracks.sublist(0, _carPlayOnlineLimit);
```
Komodo5197: Shouldn't this be _carPlayOfflineLimit?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+    if (_isPushingPageUpdate) {
+      _carPlayLogger.warning("Navigation dropped: already pushing page update");
+      return;
+    }
+    _isPushingPageUpdate = true;
+    try {
+      List<BaseItemDto> artists;
+      if (FinampSettingsHelper.finampSettings.isOffline) {
+        artists = await getTabItems(tabContentType: TabContentType.artists);
+        if (artists.length > _carPlayOnlineLimit) {
```
Komodo5197: Shouldn't this be _carPlayOfflineLimit?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+      final collections = await _downloadsService.getAllCollections(baseTypeFilter: tabContentType.itemType);
+      final baseItems = collections
+          .where((d) => d.baseItem != null)
+          .map((d) => d.baseItem!)
+          .take(_carPlayOfflineLimit)
+          .toList();
+      return sortItems(baseItems, sortBy, sortOrder);
+    }
+
+    // Fetch the online version if we can't get the offline versions
```
Komodo5197: Inaccurate comment.

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+/// Safe to call after first frame — all CarPlay template setup is deferred via
+/// SchedulerBinding.addPostFrameCallback (see setupCarplay), which ensures the
+/// GlobalSnackbar context is available before any l10n access.
+AppLocalizations get _l10n => AppLocalizations.of(GlobalSnackbar.materialAppScaffoldKey.currentContext!)!;
+
+final _carPlayLogger = Logger("CarPlay");
+
+/// Maximum items to fetch from server for CarPlay lists.
+/// Keeps UI responsive and avoids memory issues on car displays.
+const _carPlayOnlineLimit = 250;
```
Komodo5197, line 32: I notice a lot of the lists seem to just cutoff at this limit, and 250 isn't crazy high.  Is there a way to add pagination, or at least an indicator that we've cutoff the list?

lib/services/carplay_helper.dart
```diff
@@ -0,0 +1,805 @@
...
+    // This starts playback in ~1 API call instead of the previous 3.
+    final randomTracks = await _jellyfinApiHelper.getItems(
+      parentItem: _finampUserHelper.currentUser?.currentView,
+      includeItemTypes: "Audio",
+      sortBy: "Random",
+      limit: 1,
+    );
+
+    if (randomTracks != null && randomTracks.isNotEmpty) {
+      // Pre-cache the first track's image so it's available for the now-playing screen
```
Komodo5197: I don't believe this does anything.  The now playing screen should be using the full-size images submitted to the audio payer by the general purpose player image logic.

ios/Runner/AppDelegate.swift
APIUM: I'm worried it won't get merged, it's 7 months old. I'm happy to make a ticket for this and we can update and remove the workaround when it's merged?

lib/services/ios_helpers.dart
APIUM: Done

lib/services/ios_helpers.dart
APIUM: I agree, I'd like to make a follow up ticket for this as it's a big refactor and I'm making an effort to not make this MR any bigger.

lib/services/carplay_helper.dart
APIUM: Done

lib/services/carplay_helper.dart
APIUM: I'd like to add a ticket for this as a follow up

lib/services/carplay_helper.dart
APIUM: Yeah that's fair, I'm not sure keen to change this it feels fairly readable as conceptually I want both at the same time. Are you OK with that?

lib/services/carplay_helper.dart
APIUM: Yeah it does, I couldn't see a neat workaround without a fair change, I'll add another follow-up ticket for this

lib/services/carplay_helper.dart
APIUM: Done

lib/services/carplay_helper.dart
APIUM: Done

lib/services/carplay_helper.dart
APIUM: Used getAllCollections, thanks

lib/services/carplay_helper.dart
APIUM: Yep thanks, done

lib/services/carplay_helper.dart
APIUM: Done

lib/services/carplay_helper.dart
APIUM: Done

lib/services/carplay_helper.dart
APIUM: Another I'd like to improve later. CarPlay starts to get laggy above around this - AA is the same. I want to be able to load as we go with pagination.

lib/services/carplay_helper.dart
APIUM: Yep dead code, done

ios/Runner/AppDelegate.swift
APIUM: #1590

lib/services/ios_helpers.dart
APIUM: #1591

lib/services/carplay_helper.dart
APIUM: #1592

lib/services/queue_service.dart
Chaphasilor: I mean I think this is part of the PR scope, since the CarPlay instance really seems to be a way to get into these sections of the code without having a valid context available, so it should be handled, yes :)

lib/services/carplay_helper.dart
Chaphasilor: That is probably unexpected - unless Jellyfin behaves that way too?  
There should be a way to compare the "base characters", I'd guess?

lib/services/carplay_helper.dart
Chaphasilor: There's no "Radio Mix" - we have "Instant Mix" and "Start Radio" in the UI.  
I'm guessing you can't access the translation tokens from CarPlay? If you can, just use the same token that's used in the track or album menu ^^

lib/services/carplay_helper.dart
Chaphasilor: Thanks, I handled that part :)

lib/services/carplay_helper.dart
Chaphasilor: So am I wrong in assuming that these templates refer to a screen? So that an album template shows the content of a specific album, and an artist template that of a specific artists?  
Because in that case, an album would contain tracks, and an artist could contain a mix of tracks and albums, with an emphasis on albums. That's what I meant by "traditionally": before the redesign, the artist screen in Finamp was just a list of albums, and so was the genre screen.

But if this template refers to how list items are rendered, then of course the genres tab should could show genres that look like "albums" when rendered.

ios/Runner/AppDelegate.swift
Chaphasilor: If this is not a ton of messy code that could be avoided by this, it's probably better to keep tracking the regular audio_service repo. I'd like to stay up to date with any other fixes there.

lib/services/carplay_helper.dart
Chaphasilor: @APIUM FYI there's an open PR that tries to add pagination to AA, although it's a workaround: #1579  
Might be worth looking into at least, since the native capabilities of AA (and CarPlay from what you've told me) are really limited.

Chaphasilor:
@UnicornsOnLSD is anything needed on the App Store side if we release this? I'd like to avoid being rejected because we forgot to change some configuration here or in the store settings 😅

lib/services/carplay_helper.dart
APIUM: That was the core concept I was planning to use too - when that's merged then we should have the functions I was missing when I was trying to get this in that made me push it out to a follow up PR. I'll put a link to it in my issue.

APIUM: Ah I didn't see they were using NameStartsWith - I was having trouble for this use, I've now forgotten as I think there were a couple of issues, but one is that in CarPlay we display by SortName, but that API call doesn't give SortName it gives the regular name so things would show up in the wrong spot. I think there was a bigger showstopper though, it might have been that this causes issues with special characters?

lib/services/carplay_helper.dart
APIUM: Yeah it was a confusing name (since renamed) - yes it's just the list template, not a specific screen. I'd previously only used it for Albums but now it's 'showBrowsableListTemplate' instead.

lib/services/carplay_helper.dart
APIUM: Done

UnicornsOnLSD:
There shouldn't be, main thing is getting the entitlement, which we do have (and seems to work in TestFlight builds, which is promising :) )

lib/services/carplay_helper.dart
Chaphasilor: I wouldn't be surprised. Also haven't tested that PR yet.

lib/services/carplay_helper.dart
Chaphasilor: @Komodo5197 :)

Komodo5197: This is a lot of excess syntax for something that is functionally identical to await 1; await 2; but I don't really care that much.

lib/services/carplay_helper.dart
APIUM: OK it does, this is technically slightly worse as it does have to do a roundtrip over http for the cache hit, but it's way simpler and not custom so I agree it's the right move. I tested it in the sim throttled to 3g and it was ok.

lib/services/carplay_helper.dart
APIUM: OK I get it now, I was using item.name as you can see but I needed to use sortName as Jellyfin already deals with the diacritics there. Fixed.

lib/services/queue_service.dart
APIUM: Done

lib/services/carplay_helper.dart
```diff
@@ -318,8 +224,8 @@ class CarPlayHelper {
     await FlutterCarplay.showSharedNowPlaying();
   }
 
-  Future<void> startInstantMix() async {
-    _carPlayLogger.info("Starting instant mix");
+  Future<void> startRadioMix() async {
+    _carPlayLogger.info("Starting radio mix");
```
Chaphasilor: this still says "radio mix"

lib/services/carplay_helper.dart
Chaphasilor: hmm, I'm still seeing the wrong wording, check my other comment

lib/l10n/app_en.arb
```diff
@@ -207,6 +207,10 @@
   "@playlists": {},
   "startMix": "Start Mix",
   "@startMix": {},
+  "radioMix": "Radio Mix",
+  "@radioMix": {
+    "description": "Label for the CarPlay radio mix button that starts continuous radio from a random track"
+  },
```
Chaphasilor: please just use the existing `startRadio` key instead of adding the new one here. we can then update it once the home screen PR is merged.

lib/services/carplay_helper.dart
Chaphasilor: I think this is moot since we now let carplay handle the caching?

Chaphasilor (commented):
@APIUM just one tiny detail I'd like to have changed, where we seemed to have a misunderstanding. I can also handle that myself if you don't have the capacity.

lib/l10n/app_en.arb
APIUM: Done

lib/services/carplay_helper.dart
APIUM: Must have missed that, done.

lib/l10n/app_en.arb
APIUM: I can't reply to the comment below so I'll do it here

"I think this moot since we now let carplay handle the caching?"

Yes, this is removed now.

Chaphasilor (approved):
Forgot I already went over this and was waiting until I had more time to look through the code. Oh well, everything looks fine to me now. 

There are merge conflicts now, that's on me. I'll see what I can do about those.

Chaphasilor:
Looks green enough to me!  
Many thanks for your sustained efforts with this @APIUM, from picking this back up based on the old PR to actually following through with the full implementation and testing, and then waiting for us to get around to it!  
This will really ensure Finamp has a solid and even feature base when transitioning to stable!

Chaphasilor:
Also thanks @llim1 for getting this started in the first place!

PR #1519, opened 2026-01-21
Resolve duplicate player stop events

### LINKED ISSUE #1172: Duplicate AudioPlayback and AudioPlaybackStopped events

vasudan:
## Changes

1. Edited the playback_history_service to rely less on the queue_service
   1.  The queue service consumes the same event as the playback_history_service. This can cause an issue where the playback_history_service tries to use queue data that hasn't had the chance to be modified yet. I changed the class to get the current track straight from the event, in the same manner that the queue service does.
   2. I removed the listen event for `getCurrentTrackStream` as it seems to me the `_audioPlayer.playbackState` event can handle all the funcionality, and I was running into a problem where `_currentTrack` would get updated before the `_audioPlayer.playbackState` event  finished
2. Changed `MusicPlayerBackgroundTask._transformEvent` to use the event as the source of the updatePosition instead of the _audioPlayer
   1.  In my tests I found that `_player.position`  was not correct. There is a race condition that occurs because the value of position was the position of the last event instead of the new event. In order to resolve this I set the transform function to use `event.updatePosition` instead of `_player.position` 

## Todo before merging

Verify this thoroughly please. This is my first contribution and my first use of Dart.

- Check that going from track to track produces one stop event to the Jellyfin server
- Check that looping a single track produces one stop event
- Check that rewinding a track produces one stop event
- Check that the playback history is accurate after playing a few tracks

## Related Issues

- fixes #1172


Chaphasilor:
@AdamJohnSwan thanks for the PR!  

Just gave this a quick test, and both looping and the stop event seem to work correctly. One thing I couldn't test without access to iOS is making sure that the event order is always "is playing A" > "has finished playing A" > "is playing B", and not "is playing A" > "is playing B" > "has finished playing A". For me on Android this was the case, but iOS seems to behave differently (hence this PR), so I don't know if that's working there.

What's not working correctly however is **rewinding** (not sent to Jellyfin, not shown in playback history) which should start a new instance of the current track in Finamp and send a stop followed by a start event to the server (just like looping), and also **Finamp's playback history is diverging from the server history** (on the "Activity" panel on the admin dashboard, don't use the activity tab/page since it's not updated in real time) when skipping back, rewinding, looping, etc. The server shows much more changes than Finamp's history for some reason, and the server version seems to be more correct in this case.

What's worse, there seems to be an issue with the current position reporting, where sometimes the currently playing track, even without user interaction, seems to "reset" to zero and counts up the duration from there, while the music continues playing. If this happens, it seems to also get reflected in the Finamp history as two separate plays. Nothing out of the ordinary is sent to the server in this case.  
I haven't figured out how to reproduce this yet, and don't know if it only resets (backwards) or can also jump forward.

vasudan:
I'm back. Thanks for testing my PR @Chaphasilor 

> What's not working correctly however is rewinding 

How is this feature supposed to work? I thought that new stop/start events were only made when there is less than 10 seconds of the track and the track is rewound to the start?

> Finamp's playback history is diverging from the server history

I was comparing the wrong IDs when a user performed a seek, I think that was the problem here.

>  there seems to be an issue with the current position reporting

That's pretty interesting. I'm having a hard time recreating this on Android or iOS. Was this on Android? Also, was the track downloaded?
My thought is that maybe `updatePosition` and `bufferPosition` should both be taken from the event instead of the `AudioPlayer`
I pushed a change, can you give it another shot?

lib/services/music_player_background_task.dart
```diff
@@ -1269,8 +1269,8 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
         ProcessingState.completed: AudioProcessingState.completed,
       }[_player.processingState]!,
       playing: _player.playing,
-      updatePosition: _player.position,
-      bufferedPosition: _player.bufferedPosition,
```
vasudan, line 1273: Anyone have any insight on the history of this? I'm wondering if there is a reason that these values weren't taken from the `_player` event but instead grabbed from the `_player`'s current state.

Chaphasilor: No, there probably was no reason. I think whoever wrote that code (probably me) just grabbed the source for the position they were most familiar with. Using the event seems to make the most sense 👍

Chaphasilor:
Thanks for the changes! I'll take a look in the coming days!

Chaphasilor:
@vasudan sorry for the delay with this.  
Speaking of, it seems like I at least found a pattern that (sometimes) coincides with the position resetting. Whenever it happened, the console showed a log like the following: 
```
W/Looper  (12962): Slow delivery took 10650ms main h=android.os.Handler c=com.ryanheise.just_audio.AudioPlayer$1@8f57756 m=0
```
It seems like there's some kind of hang? I'm not sure what this would have to do with the progress resetting to 0, maybe the hang is another symptom of something else?  
To reproduce, all I did was leave Finamp playing music (in debug mode) with the Playback History screen open. Every now and then, a track would get "duplicated" within the history, showing that it was reset at some point. Here's what that looks like:  
image
Like before, the music kept playing normally and didn't reset, just the progress shown was incorrect. This also seems to be the only situation in which the history page diverges from the server activity, since the server doesn't show seek events.  
However, regular seeking in Finamp doesn't create a new history entry, so the observed behavior here is obviously different from just seeking backwards.

At this point I'm not even sure if this is caused by this PR at all, but it *could* be related to the change you made in `_transformEvent()`:

```diff
--- updatePosition: _player.position,
+++ updatePosition: event.updatePosition,
```

I'll play around with some tweaks to see if I can fix this.

Chaphasilor:
Okay, I've confirmed that the position resets happen when using `event.updatePosition` but don't happen when using `_player.position`, so I've now reverted those changes to again prefer the player state, and added a warning comment.  
All other behavior is now working correctly too. Regarding your question about rewinding, you're correct that Finamp currently only counts a rewind as a new play if the track is almost finished (less than 10s left), since it's treated as a "loop" event. It might make sense to change this in the future, but this is out-of-scope for this PR.

Thanks again for adding this fix! Looking forward to more contributions from you in the future, and I'll do my best to address them more promptly in the future! 😅

vasudan:
Hey @Chaphasilor , thanks for testing it out and merging. I'll look further into why that is happening and create a new PR. I haven't tried yet but the problem in #1172  will probably still occur on iOS until the playback event shenanigans are resolved.

PR #1521, opened 2026-01-23
CI: Add arm to build and unify release

luigi311:
## Changes
Adds workflow_dispatch so a manual build can be executed in the CI page

Adds the arm linux build to the build job based off the one added to the upload-assets recently

Combines the upload assets job with the build job as its just a difference of a single step so we can reduce the maintenance by only having to modify the build process in the build.yaml only. I kept the x64 release files the same without the arch in them to maintain backwards compatibility with any external tools downloading them.

## Todo before merging

- [ ] Translations

- [ ] Reset Settings

## Related Issues

relates to #766


luigi311:
Tested the release workflows here
https://github.com/luigi311/finamp/releases/tag/0.9.23-beta5

.github/workflows/build.yml
```diff
@@ -1,8 +1,11 @@
 name: Build
 
 on:
+  workflow_dispatch:
   pull_request:
   push:
+    tags:
```
Komodo5197, line 8: I think this setup prevents CI from being run at all on non-release commits pushed directly to main/redesign?

.github/workflows/build.yml
```diff
@@ -158,6 +221,25 @@ jobs:
         name: finamp-windows.msix
         path: build/windows/x64/runner/Release/finamp.msix
         compression-level: 0 # no compression
+
+    - name: Prepare release archive
+      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+      run: |
+        cp -r build/windows/x64/runner/Release/ ./Finamp-${{ github.ref_name }}-Windows
```
Komodo5197: This copy needs to be run before `dart run msix:create --install-certificate false` or else the msix installer gets bundled into the zip.

.github/workflows/build.yml
Komodo5197, line 238: There's some comments throughout upload-assets.yml that should probably be migrated to this file.

.github/workflows/build.yml
```diff
@@ -99,32 +113,81 @@ jobs:
...
-        && cp assets/finamp.desktop.m4 build/linux/x64/release/ \
-        && cp assets/com.unicornsonlsd.finamp.metainfo.xml build/linux/x64/release/
+        cp -r assets/icon/linux/ build/linux/${{ matrix.arch }}/release/icons \
+        && cp assets/finamp.desktop.m4 build/linux/${{ matrix.arch }}/release/ \
+        && cp assets/com.unicornsonlsd.finamp.metainfo.xml build/linux/${{ matrix.arch }}/release/
+    
+    # Create dynamic names based on architecture, keep x64 name without arch to preserve backwards compatibility
+    - name: Set archive and checksum names
+      run: |
+        if [ "${{ matrix.arch }}" = "x64" ]; then
```
Komodo5197: I'm not sure who other than the flatpak would be pulling these archives programmatically, so it might make sense to just add the arch to x64 as well.  We could also trim the -release, because I'm not sure why we added that.

luigi311: sounds good, wasnt sure what else was out there so i avoided it. Let me make that change

.github/workflows/build.yml
luigi311: No they are all in addition to each other. To verify i just did a push to my own redesign branch and it kicked off a CI run.

luigi311: Can see my actions here
https://github.com/luigi311/finamp/actions

I added in the same workflow settings on "CI: Test redesign push" and then added a new line and commited that too just to make sure and that was in "CI: Test redesign push 2" so looks like it is working correctly

.github/workflows/build.yml
luigi311: Ok i restructured it a bit and now the msix isnt in the zip file so they are around the same size now

.github/workflows/build.yml
luigi311: Only comment i can see that is still relevant is the 
\# upload original, unzipped directory. GitHub will zip it anyway
so i copied that over. 

Were there more you were thinking of?

.github/workflows/build.yml
Komodo5197: Ah, it seems like the branch whitelist is ORed with the tag whitelist.  I think a comment pointing this out would be nice.

.github/workflows/build.yml
```diff
@@ -99,32 +113,76 @@ jobs:
...
     # copy icon assets and desktop file next to bundle location for easy including in tar archive
     - run: |
-        cp -r assets/icon/linux/ build/linux/x64/release/icons \
-        && cp assets/finamp.desktop.m4 build/linux/x64/release/ \
-        && cp assets/com.unicornsonlsd.finamp.metainfo.xml build/linux/x64/release/
+        cp -r assets/icon/linux/ build/linux/${{ matrix.arch }}/release/icons \
+        && cp assets/finamp.desktop.m4 build/linux/${{ matrix.arch }}/release/ \
+        && cp assets/com.unicornsonlsd.finamp.metainfo.xml build/linux/${{ matrix.arch }}/release/
+    
+    # Create dynamic names based on architecture, keep x64 name without arch to preserve backwards compatibility
```
Komodo5197: This comment is outdated.

.github/workflows/build.yml
Komodo5197: I thought there was more, but it looks like you already got everything worthwhile.

luigi311:
Looks like PR have messed up names since the ref_name includes a slash on them. Let me fix that

.github/workflows/build.yml
luigi311: ok all set

luigi311:
Ok i think i have it set to handle PR names correctly now. For PRs it should now use pr-# which should be a safe name to use

luigi311:
Tested on my fork and created an internal PR and it indeed runs correctly and a release also looks correct

.github/workflows/build.yml
Chaphasilor: The `-release` was probably added when we had some debug builds and some release builds?

luigi311: ahh that makes sense, probably just followed the android builds too which have a -debug on them for the ones built within the build action that arent signed.

Chaphasilor:
@luigi311 thanks for this refactoring! The changes itself look good, there's just one thing I'm worried about:  
Won't these checks get cumbersome if we have more conditions in the future? For example, building debug builds on push but release builds on tag, possibly only for some platforms?  
I'm not sure if consolidating everything into one file is really needed. The reusability is nice of course, to avoid inconsistencies between test builds and release builds, but it doesn't make the CI any more efficient, right?

@HeroBrine1st recently mentioned that Docker-like, reusable layers would be nice. While I don't think that is supported for GitHub actions, maybe we could have one file for reusable base configurations (e.g. setup for each platform), and then other files that extend this setup with specific build commands. packaging. or validation?  
A lesser version of this could be grouping of conditions, so that we only have a single check and the multiple commands run, which should make this easier to parse (e.g. for things like `matrix.arch == 'arm64'` or `github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')` which are used multiple times _in succession_).  
If you think that's not feasible then we can continue with the current approach though.

luigi311:
> Won't these checks get cumbersome if we have more conditions in the future? For example, building debug builds on push but release builds on tag, possibly only for some platforms?

If this is something thats needed then the only change would be adding an if on the flutter build step to check if its a tag then set the flag to release else default to debug.

> I'm not sure if consolidating everything into one file is really needed. The reusability is nice of course, to avoid inconsistencies between test builds and release builds, but it doesn't make the CI any more efficient, right?

That is correct, this is more of being more consistent with the build and tagged releases. Have two separate jobs would mean having to maintain them both when it comes to changes that need to be made.

> recently mentioned that Docker-like, reusable layers would be nice. While I don't think that is supported for GitHub actions, maybe we could have one file for reusable base configurations (e.g. setup for each platform), and then other files that extend this setup with specific build commands. packaging. or validation?

Is this posted somewhere i can take a look at? Github actions has caching support and the flutter has a cache flag we can try, wasnt sure if that was something you guys wanted though, i can take a look after this.

Chaphasilor:
So the "discussion" I was referring to is here: https://github.com/jmshrv/finamp/pull/1509#issuecomment-3736181005  
That was just the idea though, nothing specific.  

If we use caching then of course we'd need to think about when to reset the cache, to keep the CI build reliable (e.g. should break if upstream packages break). But I guess for per-commit runs on feature branches that's not an issue, so resetting the cache on any commit to `redesign` or `main` could work well?

Anyway, it seems like this PR will stay the way it is then, and is ready to merge?

luigi311:
Yes this is ready to go, ill take a look at caching and that discussion for a different PR

Chaphasilor:
Awesome, thanks again!

PR #1338, opened 2025-08-17
Allow to toggle between pressing or holding a button to like a track on player screen

HeroBrine1st:
## Changes

Adds a simple setting:

<details>
<summary>Screenshot</summary>

image

</details>

Allows to toggle between hold-to-like and press-to-like, as well as press-to-open-playlists vs hold-to-open-playlists.

Quickly tested on Linux, nothing should break though

## Todo before merging

- [x Translations
- [x] Reset Settings

## Related Issues

Couldn't find issue but I remember that this was requested

PR #1524, opened 2026-01-25
Add setting for Handling Multichannel tracks

### LINKED ISSUE #1496: Redesign: Downmix during transcoding

### Type of Feature

- [ ] Something new
- [ ] None of the above

### Description

As discussed on Discord.

I'm requesting to change the request to the server so that it downmixes to Stereo during the transcode.
This can either be static or user configurable (for desktop use).

### Reasoning

Currently, when transcoding we're only requesting the format/container and bitrate. This is fine for most uses, however for tracks that are multichannel/surround on the server, this results in audibly low quality transcoded, because the requested bitrate needs to fit (in my case) 6 channels instead of the expected 2.

### Implementation Notes

Currently the `Audio` endpoint is being used:
https://github.com/jmshrv/finamp/blob/7738c967fa61450c01aab454657d41cd89bf9e14/lib/services/music_player_background_task.dart#L1367-L1382

This supports a `maxAudioChannels` parameter to request said downmix, without upmixing potential mono tracks.

Existing Stereo tracks would be unaffected.

### Additional Information

_No response_

mihawk90:
## Changes
Since we're not adapting the bitrate to the number of channels, and have no setting to do so, transcodes will easily bitrate starve for Multichannel Audio (e.g. 5.1).

This adds a setting for the user to choose how they want to handle these Multichannel tracks:
- Downmix to Stereo
- Downmix only lossy transcodes to Stereo (Default)
- Keep all channels, respecting the Bitrate set for all encodes

This implementation leaves an option for future additions, such as a dynamic Bitrate allotment based on channel count or similar.

## Related Issues
- closes #1496

### DIFF
--- a/lib/models/finamp_models.dart
+++ b/lib/models/finamp_models.dart
@@ -140,6 +140,7 @@ class DefaultSettings {
   static const syncPlaybackSpeedAndPitch = false;
   static const autoLoadLastQueueOnStartup = true;
   static const shouldTranscodeDownloads = TranscodeDownloadsSetting.ask;
+  static const multichannelHandlingSetting = MultichannelHandlingSetting.stereoDownmixLossy;
   static const shouldRedownloadTranscodes = false;
   static const resyncOnStartup = true;
   static const fixedGridTileSize = 150;
@@ -302,6 +303,7 @@ class FinampSettings {
     this.downloadTranscodingCodec,
     this.downloadTranscodeBitrate,
     this.shouldTranscodeDownloads = DefaultSettings.shouldTranscodeDownloads,
+    this.multichannelHandlingSetting = DefaultSettings.multichannelHandlingSetting,
     this.shouldRedownloadTranscodes = DefaultSettings.shouldRedownloadTranscodes,
     this.itemSwipeActionLeftToRight = DefaultSettings.itemSwipeActionLeftToRight,
     this.itemSwipeActionRightToLeft = DefaultSettings.itemSwipeActionRightToLeft,
@@ -836,6 +838,9 @@ class FinampSettings {
 
   @HiveField(143, defaultValue: DefaultSettings.forceAudioOffloadingOnAndroid)
   bool forceAudioOffloadingOnAndroid = DefaultSettings.forceAudioOffloadingOnAndroid;
+
+  @HiveField(144, defaultValue: DefaultSettings.multichannelHandlingSetting)
+  MultichannelHandlingSetting multichannelHandlingSetting;
 
   static Future<FinampSettings> create() async {
     final downloadLocation = await DownloadLocation.create(
@@ -3962,4 +3967,19 @@ class FinampStorableQueueInfo extends FinampStorableQueueInfoLegacy {
   String toString() {
     return "previous:${previousTracks.length} current:$currentTrack seek:$currentTrackSeek next:${nextUp.length} queue:${queue.length} order:${packedShuffleOrder == null ? "linear" : "shuffled"} sources $sourceList";
   }
+}
+
+@HiveType(typeId: 111)
+enum MultichannelHandlingSetting {
+  @HiveField(0)
+  stereoDownmixLossy,
+  @HiveField(1)
+  stereoDownmixAll,
+  @HiveField(2)
+  fixedBitrate,
+  /**
+    reserved for potential automatic bitrate calculation
+  @HiveField(3)
+  dynamicBitrate,
+  **/
 }
--- a/lib/models/finamp_models.g.dart
+++ b/lib/models/finamp_models.g.dart
@@ -197,6 +197,9 @@ class FinampSettingsAdapter extends TypeAdapter<FinampSettings> {
         shouldTranscodeDownloads: fields[44] == null
             ? TranscodeDownloadsSetting.ask
             : fields[44] as TranscodeDownloadsSetting,
+        multichannelHandlingSetting: fields[144] == null
+            ? MultichannelHandlingSetting.stereoDownmixLossy
+            : fields[144] as MultichannelHandlingSetting,
         shouldRedownloadTranscodes: fields[46] == null
             ? false
             : fields[46] as bool,
@@ -462,7 +465,7 @@ class FinampSettingsAdapter extends TypeAdapter<FinampSettings> {
   @override
   void write(BinaryWriter writer, FinampSettings obj) {
     writer
-      ..writeByte(137)
+      ..writeByte(138)
       ..writeByte(0)
       ..write(obj.isOffline)
       ..writeByte(1)
@@ -736,7 +739,9 @@ class FinampSettingsAdapter extends TypeAdapter<FinampSettings> {
       ..writeByte(142)
       ..write(obj.duckOnAudioInterruption)
       ..writeByte(143)
-      ..write(obj.forceAudioOffloadingOnAndroid);
+      ..write(obj.forceAudioOffloadingOnAndroid)
+      ..writeByte(144)
+      ..write(obj.multichannelHandlingSetting);
   }
 
   @override
@@ -3099,6 +3104,48 @@ class RadioModeAdapter extends TypeAdapter<RadioMode> {
   bool operator ==(Object other) =>
       identical(this, other) ||
       other is RadioModeAdapter &&
+          runtimeType == other.runtimeType &&
+          typeId == other.typeId;
+}
+
+class MultichannelHandlingSettingAdapter
+    extends TypeAdapter<MultichannelHandlingSetting> {
+  @override
+  final typeId = 111;
+
+  @override
+  MultichannelHandlingSetting read(BinaryReader reader) {
+    switch (reader.readByte()) {
+      case 0:
+        return MultichannelHandlingSetting.stereoDownmixLossy;
+      case 1:
+        return MultichannelHandlingSetting.stereoDownmixAll;
+      case 2:
+        return MultichannelHandlingSetting.fixedBitrate;
+      default:
+        return MultichannelHandlingSetting.stereoDownmixLossy;
+    }
+  }
+
+  @override
+  void write(BinaryWriter writer, MultichannelHandlingSetting obj) {
+    switch (obj) {
+      case MultichannelHandlingSetting.stereoDownmixLossy:
+        writer.writeByte(0);
+      case MultichannelHandlingSetting.stereoDownmixAll:
+        writer.writeByte(1);
+      case MultichannelHandlingSetting.fixedBitrate:
+        writer.writeByte(2);
+    }
+  }
+
+  @override
+  int get hashCode => typeId.hashCode;
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is MultichannelHandlingSettingAdapter &&
           runtimeType == other.runtimeType &&
           typeId == other.typeId;
 }
--- a/lib/screens/transcoding_settings_screen.dart
+++ b/lib/screens/transcoding_settings_screen.dart
@@ -39,6 +39,8 @@ class _TranscodingSettingsScreenState extends State<TranscodingSettingsScreen> {
           const DownloadTranscodeEnableDropdownListTile(),
           const DownloadTranscodeCodecDropdownListTile(),
           const DownloadBitrateSelector(),
+          Divider(),
+          const MultichannelHandlingSelector(),
         ],
       ),
     );
@@ -148,6 +150,29 @@ class StreamingTranscodingFormatDropdownListTile extends ConsumerWidget {
             onSelected: FinampSetters.setTranscodingStreamingFormat.ifNonNull,
           ),
         ],
+      ),
+    );
+  }
+}
+
+class MultichannelHandlingSelector extends ConsumerWidget {
+  const MultichannelHandlingSelector({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return ListTile(
+      title: Text(AppLocalizations.of(context)!.multichannelHandlingTitle),
+      subtitle: FinampSettingsDropdown<MultichannelHandlingSetting>(
+        dropdownItems: MultichannelHandlingSetting.values
+            .map(
+              (e) => DropdownMenuEntry<MultichannelHandlingSetting>(
+                value: e,
+                label: AppLocalizations.of(context)!.multichannelHandlingOption(e.name),
+              ),
+            )
+            .toList(),
+        selectedValue: ref.watch(finampSettingsProvider.multichannelHandlingSetting),
+        onSelected: FinampSetters.setMultichannelHandlingSetting.ifNonNull,
       ),
     );
   }
--- a/lib/screens/transcoding_settings_screen.dart
+++ b/lib/screens/transcoding_settings_screen.dart
@@ -162,17 +162,24 @@ class MultichannelHandlingSelector extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     return ListTile(
       title: Text(AppLocalizations.of(context)!.multichannelHandlingTitle),
-      subtitle: FinampSettingsDropdown<MultichannelHandlingSetting>(
-        dropdownItems: MultichannelHandlingSetting.values
-            .map(
-              (e) => DropdownMenuEntry<MultichannelHandlingSetting>(
-                value: e,
-                label: AppLocalizations.of(context)!.multichannelHandlingOption(e.name),
-              ),
-            )
-            .toList(),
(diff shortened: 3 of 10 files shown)


lib/services/music_player_background_task.dart
```diff
@@ -1373,6 +1373,7 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
         "maxAudioBitDepth": "16",
         "audioBitRate": FinampSettingsHelper.finampSettings.transcodeBitrate.toString(),
         "segmentContainer": FinampSettingsHelper.finampSettings.transcodingStreamingFormat.container,
+        "maxAudioChannels": "2",
```
Komodo5197: I imagine we want to disable this for FLAC transcodes.

Komodo5197:
I'd really prefer to have a setting for down mixing vs upping bitrate with channel count immediately, because someone getting actual use out of transcoded multi-channel audio by just upping the bitrate manually seems conceivable, which would make this a regression.

Chaphasilor:
@Komodo5197 so do you want a setting for enabling/disabling downmixing, with the ability to manually up the transcode bitrate, or do you also want Finamp to detect the channel count, and if downmixing is disabled, add a multiplier to the bitrate for any extra channels automatically?

lib/services/music_player_background_task.dart
Chaphasilor: Mind explaining why? Are channels handled differently in FLACs?

mihawk90: It's not, but there's also no bitrate starving for FLAC.

Then again, I wasn't aware transcoding to FLAC is even an option?

mihawk90: Implemented a check for FLAC exclusion.

Although since I never worked with Dart I am not sure if this will omit the entire parameter or just the value.

Komodo5197:
I was imagining a dropdown where you select either the two channel downmixing, or a mode where it automatically scales the configured bitrate with channel count.  Maybe it could also have an option for the current behavior, although I'm not 100% sure why anyone would want that instead of one of the other modes, so it would probably need a warning.

In regards to upping the bitrate, I'm wondering how that actually scales with channels.  Naively you might imagine 5 ch having 2.5 the data of 2, but the different channels seem ripe for compression so I'm wondering if some/all of the transcode formats will scale in a different way.  This might be worth testing out.

mihawk90:
That's the concern I have. Channels don't scale linearly because there is inter-channel compression going on. That aside, it will also heavily depend on the content of each channel. If there is a large variance between front and back channels for example my bets are it will take considerably more bitrate to achieve the same level of quality.

If we want a setting (which by the way I have no idea how to implement), then I would not want to do any automagic bitrate selection because that just opens a rabbit hole of potential issues down the road. What happens for example if our automagic bitrate isn't enough for a particularly difficult to compress track? What happens when our automagic bitrate is wasting bandwidth/storage space because it's an easy to compress track? We'll be getting bug reports about it and asking for a selection anyway.

But realistically, how many users have *any* multichannel tracks at all, and if they do how many? IMO the time investment to figure out the "right" bitrate is not worth it for the relatively tiny usecase.

Among my couple thousand tracks I have a grand total of 30(!) that are in 5.1. 29 of which are a single concert album which I *also* have in Stereo, the other is one single track I cut from a movie, so it's not even technically a released track.

Long story short, if anything I'd make it an on-off setting, have it on by default (unless you want it off, but I think for most users Stereo is going to be the norm), and if users *really* want a multichannel transcode they can fall back to the current behaviour by turning it off.

Komodo5197:
I do agree with down mixing being the default.  Regarding the comression, now that I think about it more I realize that potentially better/worse quality depending on compressability is already a thing, so it probably doesn't matter that much if its tuned well.  I'm thinking something like `targetBitate=bitrateSetting+max(0,trackNumber-2)*bitrateSetting*0.3;`  If you don't know how to add settings, #1338 is a relativly clean PR to look at, or I guess I could throw one together?

mihawk90:
> it probably doesn't matter that much if its tuned well

That's kind of my core issue with the automagic bitrate though. We don't even know what to tune *for*, much less how to tune it *well*. And while it might be an interesting thing to rabbit hole into, I don't think it's worth the time for the amount of multichannel audio in the wild.

It also effectively takes away control from users that do want to use it. We're hardcoding whatever formula and the eventual user is going to wonder why their bitrate isn't what they set it to just to discover there's some arcane invisible calculations going on.
I'd much rather have it an on-off switch and then the user can decide their desired bitrate on their own, and adjust it based on the resulting quality.

Chaphasilor:
Actually, having a dropdown with 3 options would pretty much solve the issue of lack of "control" you mentioned, because we could have 1) downmix, 2) base bitrate + scale to additional channels, and 3) fixed bitrate.  
I think we could just implement the setting as a dropdown with options 1 and 3 for now, and if we figure out how to test if @Komodo5197's formula is reasonably correct, then we can just add that as a third option later on?  

But it seems like the "scaling with channels" discussion shouldn't block the downmix fix, right?

mihawk90:
So... now that I had a look at it and I'm pretty much done, I stumble on this ... ?

https://github.com/jmshrv/finamp/blob/923825e332f4c721c3aa598763f70127e452f229/lib/models/finamp_models.dart#L2448-L2467

Added 2 years ago in 84c21a560d2739dd6d193896ccb489b0a591aa36, but can't say this is doing anything given the reason for opening #1496 in the first place.

lib/models/finamp_models.g.dart
```diff
@@ -2212,6 +2220,64 @@ class TranscodeDownloadsSettingAdapter
...
 
+/*
+  Multichannel - cloned from TranscodeDownloadsSettingAdapter
+  */
+class MultichannelHandlingSettingAdapter
+    extends TypeAdapter<MultichannelHandlingSetting> {
+  @override
+  /**
+    I *think* this is the HiveType id?
+    last is 110 in FinampStorableQueueInfoAdapter, so 111 it is 🤷
```
Komodo5197: Any file named *.g.dart is autogenerated and shouldn't be edited manually.  Check contributing.md for details on regenerating.

lib/models/finamp_models.dart
```diff
@@ -2496,6 +2496,22 @@ enum TranscodeDownloadsSetting {
   ask,
 }
 
+@HiveType(typeId: 111)
```
Komodo5197: I'd prefer this to be added at the bottom, or at least for a comment to be added there, solely to make the highest used typeId easily apparent during future changes.

lib/services/music_player_background_task.dart
```diff
@@ -1374,6 +1374,13 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
         "audioBitRate": FinampSettingsHelper.finampSettings.transcodeBitrate.toString(),
         "segmentContainer": FinampSettingsHelper.finampSettings.transcodingStreamingFormat.container,
       });
+
+      /** honestly no idea what type this setting even is 🤷 */
+      if (FinampSettingsHelper.finampSettings.multichannelHandlingSetting as String == "stereoDownmix") {
```
Komodo5197: You never actually added multichannelHandlingSetting as a field to the FinampSettings class.  It should be a MultichannelHandlingSetting.  You'll also want to add a default for the setting in DefaultSettings and reset the setting in resetTranscodingSettings() in FinampSettingsHelper.  Once that's done, you can just check == MultichannelHandlingSetting.stereoDownmix.

lib/screens/transcoding_settings_screen.dart
```diff
@@ -39,6 +39,8 @@ class _TranscodingSettingsScreenState extends State<TranscodingSettingsScreen> {
           const DownloadTranscodeEnableDropdownListTile(),
           const DownloadTranscodeCodecDropdownListTile(),
           const DownloadBitrateSelector(),
+          Divider(),
```
Komodo5197, line 42: I don't think adding a divider here makes much sense, because this setting is pretty tightly coupled with the preceding bitrate setting.

lib/models/finamp_models.dart
```diff
@@ -2496,6 +2496,22 @@ enum TranscodeDownloadsSetting {
...
+enum MultichannelHandlingSetting {
+  /**
+    notation taken from FinampTranscodingCodec since I assume that allows using it as a String?
+  **/
+  @HiveField(0)
+  stereoDownmix("stereoDownmix"),
```
Komodo5197: The string version of the enum values is actually always available as MultichannelHandlingSetting.stereoDownmix.name, and so on.  This notation is actually supplying arguments to a constructor for the enum, which this one doesn't have.  You can just do

```
enum MultichannelHandlingSetting {
  @HiveField(0)
  stereoDownmix,

  @HiveField(1)
  fixedBitrate,
  /**
    reserved for potential automatic bitrate calculation
  @HiveField(2)
  dynamicBitrate,
  **/
}
```

Komodo5197:
It looks like the bitrateChannels function you found is exclusively used to scale the expected download size shown for transcodes in the download dialog, but then is never actually used when generating the transcode.  I'll try to clean up this error at some point in the future, I guess?

lib/screens/transcoding_settings_screen.dart
mihawk90: I added the divider because it applies to both Stream and Download transcodes. Having it without a divider would visually imply that it belongs to the Downloads settings only, which wouldn't match the actual setting.

lib/models/finamp_models.dart
mihawk90: Yeah wasn't sure where to put this because not all of them are in order already so I thought I'd keep it together with the other setting, but doesn't matter to me

lib/models/finamp_models.dart
mihawk90: That's what I had at first until I saw the codec selection :melting_face:

lib/services/music_player_background_task.dart
mihawk90: I hope I got everything :thinking:

lib/models/finamp_models.g.dart
mihawk90: ran and should be good

lib/screens/transcoding_settings_screen.dart
Komodo5197: Makes sense.

mihawk90:
Anything else I need to take care of? Otherwise I'll squash and stuff

lib/services/music_player_background_task.dart
```diff
@@ -1374,6 +1374,12 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
         "audioBitRate": FinampSettingsHelper.finampSettings.transcodeBitrate.toString(),
         "segmentContainer": FinampSettingsHelper.finampSettings.transcodingStreamingFormat.container,
       });
+
+      /** honestly no idea what type this setting even is 🤷 */
+      if (FinampSettingsHelper.finampSettings.multichannelHandlingSetting ==
+          MultichannelHandlingSetting.stereoDownmix) {
```
Komodo5197: I think we still need to disable this if we're transcoding to flac.

lib/models/finamp_models.dart
```diff
@@ -837,6 +839,13 @@ class FinampSettings {
   @HiveField(143, defaultValue: DefaultSettings.forceAudioOffloadingOnAndroid)
   bool forceAudioOffloadingOnAndroid = DefaultSettings.forceAudioOffloadingOnAndroid;
 
+  /**
+    Multichannel - cloned from TranscodeDownloadsSetting (44)
+    No idea what these numbers are but.. it's sequential 🫠
+   */
```
Komodo5197: None of these cloned from comments and similar seem worth committing.

lib/services/music_player_background_task.dart
```diff
@@ -1374,6 +1374,12 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
         "audioBitRate": FinampSettingsHelper.finampSettings.transcodeBitrate.toString(),
         "segmentContainer": FinampSettingsHelper.finampSettings.transcodingStreamingFormat.container,
       });
+
+      /** honestly no idea what type this setting even is 🤷 */
+      if (FinampSettingsHelper.finampSettings.multichannelHandlingSetting ==
```
Komodo5197, line 1378: The query parameters for transcoded downloads are actually calculated in getTrackDownloadUrl() inside jellyffin_api_helper.dart.  You should add this there too.

lib/models/finamp_models.dart
mihawk90: I'm aware, they were for personal reference and were going to be removed during squashing.

lib/services/music_player_background_task.dart
mihawk90: I'm not sure I see the benefit in this. The disabling for FLAC from before was done because there was no option to select how to handle multiple channels, so it was assumed that downmixing for FLACs would not be a wanted behaviour.

However, now with the explicit user choice between downmix or not that is no longer the case. What if users want a FLAC transcode, but simply don't want the overhead of additional surround channels? Disabling this on FLAC regardless of setting would make the setting meaningless in that case.

The only limitation for FLAC that we *may* have to impose is [limiting the channel count to 8](https://www.ietf.org/rfc/rfc9639.html#:~:text=FLAC%20files%20and%20streams%20can%20code%20for%20pulse%2Dcode%20modulated%20(PCM)%20audio%20with%201%20to%208%20channels), but I don't know if that is actually required. `ffmpeg` may or may not handle this on its own, but I don't have any tracks with ~~>6~~ >8 channels to test with.

lib/services/music_player_background_task.dart
mihawk90: added

lib/services/music_player_background_task.dart
Komodo5197: Flac is here for when you basically don't care about overhead at all.  I feel like the amount of people who will intentionally downmix flac is ~0, while the amount who set flac but leave this setting on the default of downmix without thinking would be far higher.  If the user cares about bitrate even a little bit, they should probably be using high-bitrate opus instead.  I feel like ignoring this for flac basically lines up with how the bitrate is already ignored for it.

mihawk90: While I do see your point, I don't really feel comfortable taking away user choice just because some might choose the wrong option.

And that usecase outlines above is not hypothetical either, because I have done this exact thing in the past (in fact still have the files somewhere.. I think).

Some devices might not handle downmixing locally well (which was my reason way back when), so having the option to downmix before it gets to the device is a valuable thing to have.

So I'm wondering if we could compromise with the default. Would it be possible to have the setting default to Multichannel when FLAC is set? That would still leave the *option* to downmix while alleviating your concern about unwanted defaults.

mihawk90:
Squashed, removed comments, changed PR description, and also title (which took me 10 minutes to figure out somehow)

lib/services/music_player_background_task.dart
Komodo5197: If you feel there's an actual use case for downmixing flac, I guess we can keep it.  Regarding a different default, there's basically two options.  One would be to add a second setting for downmixing flac specifically with a different default.  The other option would be to add a new option to the existing setting so that we have seperate 'downmix all formats' vs 'downmix non-flac' options, and make the latter the default.  I think the latter option might make the most sense?  And this settings already pretty technical, so making it a bit more complicated probably isn't a dealbreaker.

mihawk90: Sounds good, added

lib/services/music_player_background_task.dart
```diff
@@ -1374,6 +1374,13 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
         "audioBitRate": FinampSettingsHelper.finampSettings.transcodeBitrate.toString(),
         "segmentContainer": FinampSettingsHelper.finampSettings.transcodingStreamingFormat.container,
       });
+
+      if (FinampSettingsHelper.finampSettings.multichannelHandlingSetting ==
+              MultichannelHandlingSetting.stereoDownmixAll ||
+          FinampSettingsHelper.finampSettings.multichannelHandlingSetting ==
+              MultichannelHandlingSetting.stereoDownmixLossy) {
```
Komodo5197: We need to check if the track is FLAC before down mixing with stereodownmixlossy.

mihawk90: Right woops, that's why copy and paste is bad kids.

Chaphasilor:
Okay, just gave this a test (from a UI standpoint, I don't think I have any multi-channel tracks), and fixed a few issues with the localization strings, but nothing major. I'll merge once the tests have passed.  
Thanks for the PR! <3

Issue #1365, opened 2025-09-02
Global search


PANICBUTT0N:
### Are you using the newest Version?

- [x] Yes, the feature is still missing on the newest version
- [x] No, there aren't any open request about this. I checked!

### Type of Feature

- [x] Something new
- [x] Improvement of a feature
- [ ] None of the above

### Description

A search bar that can search across all categories of the library (artists, albums, playlists, tracks, etc.) rather than having an individual search bar for each category. 

### Reasoning

This would allow more flexibility in search by allowing users to make broader searches that still return satisfactory results, rather than requiring them to recall more specific details about the media they are looking for. 

### Implementation Notes

Instead of only each library section having its own search function, a new global search menu could be added that returns results from each category. This feature could coexist with the current per-category search bars, or could replace them entirely by having search filters that still allow the user to search within a specific category/specific categories.

Chaphasilor:
yes, this is planned :)  
there's some discussion in #1141 about it

PR #1529, opened 2026-02-02
Add universal search, AirPlay, and iOS glass UI

obiwantoby:
## Changes

- Adds an iOS‑only native AirPlay route picker button on the player screen.
- Introduces a universal search screen (default) with a settings toggle to fall back to the legacy per‑tab search.
- Applies iOS “glass” styling to app bars and the now‑playing bar.
- Adds a small AirPlay availability note on iOS.

Note: I know there’s already a redesign branch in progress; I made this change to socialize and really selfishly enjoy a universal search by default.

IMG_1715
IMG_1714
IMG_1713

## Related Issues

- Will need to reconcile with redesign branch


flloschy:
Would be great if you can transfer the changes to the redesign branch, else they will be overwritten when the beta becomes the default, and thus merging wont make much sense.
Sorry for that :/

Edit: Screw that, I didnt notice you've [already did that](https://github.com/jmshrv/finamp/pull/1531) lol 

Global search is actually a long standing [feature request](https://github.com/jmshrv/finamp/issues/1365) which seems to be [added with the new homescreen](https://github.com/jmshrv/finamp/pull/1141) anyway

PR #1533, opened 2026-02-07
Add universal search screen with classic search toggle

### LINKED ISSUE #471: Allow searching for songs using artist's name.
Right now it's only possible to search for songs (in the SONGS tab) by using the song's name.

I think it would really improve the user experience by also allowing to search for songs by using the Artist's name while in the "SONGS" tab.

Because right now if I want to find a song named "d4vd - Fly Away", I have to explicitly search by the name of the song, so "Fly Away", but if I for example can't remember the song's name but I can remember the artist's name, well searching for "d4vd" shows no results.

Just as I am writing this I tried it in a web browser and it behaves the same... oh well, do as you see fit.

Thanks!

obiwantoby:
## Changes

- **Universal Search Screen**: New dedicated search screen that queries the Jellyfin API with debounced requests (300ms) across Audio, Albums, Artists, Playlists, and Genres
- **Rich Search Results**: Uses `TrackListTile` for audio items and `ItemCollectionWrapper` for albums/artists/playlists to maintain consistency with existing app UI
- **Classic Search Toggle**: Added `useUniversalSearch` setting (HiveField 144) with a toggle in Layout & Theme settings. Users can switch back to the in-page classic search if preferred
- **Search Integration**: Search icon in the app bar and drawer menu entry now route to universal or classic search based on the user's preference setting
- **iOS 26 Crash Fix**: Added `FLTEnableTouchRateCorrection=false` to Debug, Release, and Profile plists to resolve VSyncClient crash on iOS 26 devices
- **iOS file_picker Fix**: Added Podfile preprocessor rename for `FileUtils` → `FPFileUtils` to resolve class collision with OSAnalytics framework
- **Dependency Bump**: Updated file_picker to ^10.3.10 for compatibility fixes

## Todo before merging

- [x] Translations (English strings added to app_en.arb; other language translations will follow)
- [x] Reset Settings (useUniversalSearch reset support added to resetLayoutSettings)

## Related Issues

- closes #471

Note this did fix an inappropriate pull request to which some other dart code was removed.

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -42,6 +42,13 @@ post_install do |installer|
   installer.pods_project.targets.each do |target|
     flutter_additional_ios_build_settings(target)
 
+    if target.name == 'file_picker'
+      target.build_configurations.each do |config|
+        config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)']
+        config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] << 'FileUtils=FPFileUtils'
+      end
+    end
+
     target.build_configurations.each do |config|
       config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '14.0'
     end
--- a/ios/Runner/Info-Debug.plist
+++ b/ios/Runner/Info-Debug.plist
@@ -22,6 +22,8 @@
 	<string>$(FLUTTER_BUILD_NAME)</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
+	<key>FLTEnableTouchRateCorrection</key>
+	<false/>
 	<key>CFBundleURLTypes</key>
 	<array>
 		<dict>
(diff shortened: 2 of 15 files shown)


Chaphasilor:
Thanks for the (new) PR!  
I haven't tried it out yet, but looking at the code this doesn't support offline mode yet. Could you look into adding this?  

Some more things you might want to consider:  
1. Originally it was planned to add global search along with the new home screen (#1141), but I didn't have time to add both. You could consider updating the target branch to that PR branch, the branches should be mostly compatible (just the Hive IDs might change, and then you will need to run code generation again)
2. Splitting your changes into separate commits would be nice, but you can do that next time. Try to separate different changes as much as possible.
3. There is an existing more or less complete mockup for the global search:  
  image  
  This is just a **rough** mockup, and somewhat outdated. Don't mind how the item tiles look, don't mind the dark mode, don't mind the colors. The important bit would be the tabs / item types to search in, which can be toggled on and off. It would also be good if were obvious which kind of item is being shown.

4. This will now keep the old and new search implementation around, which isn't ideal. I don't see a reason to keep the old search function around, aside from filtering maybe?

Chaphasilor:
@obiwantoby are you planning to make any further changes here? If not, I'll have to close the PR :)

PR #1540, opened 2026-02-22
Fix playback history reporting

StayBlue:
Sometimes, after switching tracks, Finamp would send a `/Sessions/Playing/Progress` update for the previous track.

`generateGenericPlaybackProgressInfo()` could use an outdated `_currentTrack` when the queue changed faster than playback-state events. This patch syncs `_currentTrack` with `QueueService.getCurrentTrack()` before building the progress payload, so reports use the correct `ItemId`.

### DIFF
--- a/lib/services/playback_history_service.dart
+++ b/lib/services/playback_history_service.dart
@@ -548,7 +548,12 @@ class PlaybackHistoryService {
     bool includeNowPlayingQueue = false,
     bool force = false,
   }) {
-    final currentTrack = _currentTrack?.item ?? _queueService.getCurrentTrack();
+    final queueTrack = _queueService.getCurrentTrack();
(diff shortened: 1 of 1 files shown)


Chaphasilor:
@StayBlue I think your fix targets a bug that is also fixed by #1519 which I just merged. The fix in the other PR seems a bit more maintainable than this very specific check, and will probably be more robust.   
I'll go ahead and merge your other fix though, since you've put in the effort of opening this PR :)  

Please take a look at the other PR and let me know if I'm correct in my assumption. If there's still fixing needed, we can add in your change as well, of course!

Chaphasilor:
cherry-picked bbb7b8bd978551a074ab02311ce3ff0768990fc9 as f838122932240af389e97cdc6daca421884385bf

PR #1543, opened 2026-03-01
Add native splash screens for light and dark mode

### LINKED ISSUE #1538: [UI Improvement] iOS loading/splash screen is bright white, despite dark theme of app

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.9.18 beta

### Affected Devices/Platform

- [ ] Android
- [ ] Android Auto
- [x] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

_No response_

1. Launch the app

### Logs

purely a display issue, no logs relevant

### Additional Information

When opening the iOS app, the loading screen (which displays for a few seconds on my 13 Mini) displays bright white even when using a dark theme for finamp. We may want to consider explicitly defining a [launch screen](https://developer.apple.com/documentation/xcode/specifying-your-apps-launch-screen/) to make this screen match the selected theme. We could even add the finamp logo!

I imagine this might impact Android as well, but I don't know for sure.

I don't have time to open a PR for this myself right now, but if nobody is interested I will likely take a stab at it soon[tm].

Haar:
## Changes

Introduced a native splash screen that is displayed whilst the app initially loads. This respects the OS light/dark mode system settings.

I have only tested this in iOS via the Simulator, as that's something I'm familiar with. I'm hoping someone can validate the Android implementation works as expected as well. In theory with this being abstracted by flutter_native_splash, it should be 👍.

### Screenshots

Dark mode splash screen - iOS
Light mode splash screen - iOS

## Related Issues

- Potentially closes #1538 (not exactly - that specifies Finamp theme, this uses OS theme)


Haar:
Setup the Android Emulator, screenshots attached:

Screenshot 2026-03-09 at 22 51 48
Screenshot 2026-03-09 at 22 57 09

CONTRIBUTING.md
```diff
@@ -106,6 +118,18 @@ As mentioned above, Finamp uses Hive for most data storage needs. If you're doin
...
+Native splash screens are generated from `flutter_native_splash.yaml`.
+
+Run:
+
+```bash
+dart run flutter_native_splash:create --path=flutter_native_splash.yaml
+```
+
+Rerun this after changing splash colors, splash icon assets, or light/dark splash behavior. This updates the generated iOS and Android native splash resources.
+
```
Chaphasilor: you seem to have added this section twice. the upper version is probably the one to keep, since the splash screen has nothing to do with Hive :)

Chaphasilor (commented):
@Haar thanks for the PR, and sorry for taking so long to take a look at it.  
One question regarding the splash screen generation, this has to be run manually and is not included in any other build step, like code gen (`dart run build_runner ...`) or the regular build (similar to translations, which are generated during each build), correct? 

Also, while it would be nice to make the splash screen match the *app* theme instead of the system theme, this is probably not really feasible since the database hasn't been opened when the splash screen is shown. So your approach seems like the most sensible solution in that case.

Another question: after running the splash screen generation myself, the `Info*.plist` files had changes, where instead of the `&apos;` escape sequence, a regular `'` was used. Which one is correct? Will using a regular `'` symbol lead to issues with the iOS files?

Finally, would you mind extending existing code gen check (https://github.com/haar/finamp/blob/a73a41eedcb4f1e18b52f6bc08e660098a680142/.github/workflows/pr-healthchecks.yml#L61-L69) to account for changes in the generated splash screen? That would avoid ever making a change to the splash screen without running generation afterwards.

Either way, the adaptive splash screen is working for me on Android just like it should, so thanks for that!

Haar:
Thanks for the review!

> @Haar thanks for the PR, and sorry for taking so long to take a look at it. One question regarding the splash screen generation, this has to be run manually and is not included in any other build step, like code gen (`dart run build_runner ...`) or the regular build (similar to translations, which are generated during each build), correct?

Correct, it's a manual step, not part of build_runner or the regular build. It only needs to be rerun when the splash config or assets change.

> Also, while it would be nice to make the splash screen match the _app_ theme instead of the system theme, this is probably not really feasible since the database hasn't been opened when the splash screen is shown. So your approach seems like the most sensible solution in that case.

👍 

> Another question: after running the splash screen generation myself, the `Info*.plist` files had changes, where instead of the `&apos;` escape sequence, a regular `'` was used. Which one is correct? Will using a regular `'` symbol lead to issues with the iOS files?

A regular `'` symbol should be perfectly fine (I've regenerated the file again myself now), I'm not sure if a stray formatter ran - but have commited the correct file now.

> Finally, would you mind extending existing code gen check (https://github.com/haar/finamp/blob/a73a41eedcb4f1e18b52f6bc08e660098a680142/.github/workflows/pr-healthchecks.yml#L61-L69) to account for changes in the generated splash screen? That would avoid ever making a change to the splash screen without running generation afterwards.

I'll do that shortly and then push up the change 🥳.

Chaphasilor:
@Haar awesome, thanks for the changes! The generated files are now identical for me, and the new CI check also works and agrees with me regarding the files 😁  
Thanks for the PR, looking forward to future contributions from you!

Chaphasilor:
@Haar not sure if you saw the message in Discord, but it seems like the splash screen isn't optimized for large screens yet (logo is way too big).  
Would you mind handling that? 😁
splashscreen_large-screen_no-exif

Haar:
@Chaphasilor Ah I think I must have missed it. 

Happy to sort that out. Is there a maximum size we'd want it to render as?

Haar:
@Chaphasilor I've just had a quick look at the different files. I can swap it to use the same icon I'm using for iOS (512x512), but the issue is that it looks blurry in the Android Emulator:

Screenshot 2026-03-30 at 16 54 40


It might just be the Emulator that looks like that though?



``` diff
--- a/flutter_native_splash.yaml
+++ b/flutter_native_splash.yaml
@@ -8,8 +8,7 @@ flutter_native_splash:

   color: "#FCFDFE"
   color_dark: "#101315"
-  image: assets/icon/icon_foreground.png
-  image_ios: assets/icon/splash_ios_light.png
+  image: assets/icon/splash_ios_light.png
   android_gravity: center
   ios_content_mode: center
   fullscreen: false
@@ -22,4 +21,4 @@ flutter_native_splash:
   android_12:
     color: "#FCFDFE"
     color_dark: "#101315"
-    image: assets/icon/icon_foreground.png
+    image: assets/icon/splash_ios_light.png
```

Then running `dart run flutter_native_splash:create --path=flutter_native_splash.yaml` should be sufficient if you want to test in on that device and see how they compare.

Chaphasilor:
@Haar thanks, it's not my device so I sadly can't test. I was thinking using the high-res logo but constraining the size to a specific amount of physical pixels or a specific physical size (i.e. virtual pixels). Is that possible somehow?

PR #1561, opened 2026-03-20
CI: Cache the pub, dart, gradle and cocopods

luigi311:
## Changes

Adds caching to the pub, dart, gradle and cocopods. All of these together reduces the CI time in half by cutting the build step as much as possible with caching.

After: https://github.com/luigi311/finamp/actions/runs/23389760029

## Todo before merging

- [ ] Translations

- [ ] Reset Settings

## Related Issues

follow up to initial cache discussion on #1521 and [#1509](https://github.com/UnicornsOnLSD/finamp/pull/1509#issuecomment-3736181005)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,11 +27,6 @@ jobs:
         distribution: 'zulu'
         java-version: '17'
   
-    - name: Install dependencies
-      run: |
-        sudo apt-get update
-        sudo apt-get install -y build-essential cmake ninja-build libgtk-3-dev
-
     - name: Free up disk space
       run: |
         # Remove .NET SDKs
@@ -46,10 +41,46 @@ jobs:
         sudo rm -rf /usr/local/share/powershell
         df -h
 
+    # Use the same flutter setup as the linux build to allow sharing the pub cache and it utilizing the same cache key
+    - name: Determine Flutter stable version
+      uses: subosito/flutter-action@v2
+      id: flutter-version
+      with:
+        architecture: 'x64'
+        channel: 'stable'
+        dry-run: true
     - name: Set up Flutter
       uses: subosito/flutter-action@v2
       with:
-        channel: 'stable'
+        # Prebuilt ARM linux builds are not provided, so build flutter from source
+        channel: master
+        flutter-version: "${{ steps.flutter-version.outputs.VERSION }}"
+        cache: false
+        pub-cache: true
+        # Include arch as linux aarch64 include isar overriding
+        # arch can be removed later once isar aarch64 is easily available on pub.dev and overriding is not needed anymore
+        pub-cache-key: "flutter-pub-:os:-:channel:-:version:-:arch:-:hash:"
+
+    
+    - name: Gradle cache
+      uses: actions/cache@v5
+      with: 
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+        # If exact match is not found, fallback to OS match
+        restore-keys: ${{ runner.os }}-gradle-
+
+    - name: Cache Dart build cache
+      uses: actions/cache@v5
+      with:
+        path: |
+          .dart_tool
+          build
+        key: ${{ runner.os }}-android-dart-build-${{ hashFiles('pubspec.lock') }}
+        restore-keys: ${{ runner.os }}-android-dart-build-
+
     - run: flutter doctor
     - run: flutter pub get
 
@@ -73,7 +104,28 @@ jobs:
       uses: subosito/flutter-action@v2
       with:
         channel: 'stable'
-        architecture: x64
+        cache: false
+        pub-cache: true
+
+    - name: Cache CocoaPods
+      uses: actions/cache@v5
+      with:
+        path: |
+          ios/Pods
+          ~/.cocoapods
+        key: ${{ runner.os }}-pods-${{ hashFiles('ios/Podfile.lock') }}
+        restore-keys: ${{ runner.os }}-pods-
+
+    - name: Cache Dart build cache
+      uses: actions/cache@v5
+      with:
+        path: |
+          .dart_tool
+          build
+        key: ${{ runner.os }}-dart-build-${{ hashFiles('pubspec.lock') }}
+        restore-keys: ${{ runner.os }}-dart-build-
+
+
     - run: flutter doctor
     - run: flutter pub get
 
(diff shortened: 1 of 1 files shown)


luigi311:
Should hopefully also stop cases like this https://github.com/UnicornsOnLSD/finamp/actions/runs/23315896092/job/67815169958 where it took 14 mins to setup flutter for some reason.

Edit: No longer applies

luigi311:
Marking as draft for now so i can see if i can cache the actual build steps and see what sizes look like between the two since caching the sdk itself is huge and it makes it easy to reach the github cache storage size limit with not a big benefit compared to what caching the build would do.

luigi311:
Changed what this is now caching, looks like the SDK itself is massive, 2gb, which when you enable caching for all the different options, it takes up a ton of space on github and runs you into the space limitations. Instead i shifted this to caching the pub, dart, gradle and cocopods all of which take up less space than the sdk, except maybe gradle, and has a bigger time savings compared to sdk caching.

After: https://github.com/luigi311/finamp/actions/runs/23389760029

brings the total run time to 7 minutes which basically brings down the run time in half. All the builds are a lot closer in terms of compilation too with around 6-7 mins for everything non linux. Its interesting how fast the linux builds themselves are even with the arm64 build having to build isar itself.

Chaphasilor:
Neat! Any specific reason for pinning the action version? Is that specific version required or is it just to avoid unexpected changes?  
Also, is there a way to invalidate the cache? Probably in the repo/actions settings somewhere?

luigi311:
I pinned the action since they just added in the pub-cache individual flag in the main branch and havent created a release with it yet. 

As for invalidating the cache, most of them are tied to the checksum of the pubspec lock file so it will be updated whenever that is changed. The others are tied to their individual files like the gradle and coco ones. If for whatever reason you want to clear it out yourself you can go to Actions > Caches > Click the Trashcan on whatever caches you want to clear out.

Chaphasilor:
@Komodo5197 any comments before this is merged?

.github/workflows/build.yml
```diff
@@ -204,12 +249,26 @@ jobs:
...
+
     - run: flutter doctor
     - run: flutter pub get
+
+    - name: Cache Dart build cache
+      uses: actions/cache@v5
+      with:
+        path: |
+          build
+        key: ${{ runner.os }}-dart-${{ hashFiles('**/pubspec.yaml') }}
```
Komodo5197: You forgot -build to make this consistent with the other caches.

.github/workflows/build.yml
```diff
@@ -64,32 +40,84 @@ jobs:
         # Remove PowerShell
         sudo rm -rf /usr/local/share/powershell
         df -h
+
     - name: Set up Flutter
-      uses: subosito/flutter-action@v2
+      uses: subosito/flutter-action@4efee115f1f8f612aafcb187ee16db202d7e4376
```
Komodo5197: It looks like the pub-cache flag may have been released already.

.github/workflows/build.yml
```diff
@@ -64,32 +40,84 @@ jobs:
...
       with:
         channel: 'stable'
+        cache: false
+        pub-cache: true
```
Komodo5197, line 59: It looks like pub-cache is including architecture into the cache path, which I don't think is actually needed for the pub cache?  So I think we could save 500MB with pub-cache-key:  ${{ runner.os }}-pub-cache.

.github/workflows/build.yml
```diff
@@ -64,32 +40,84 @@ jobs:
...
+    
+    - name: Gradle cache
+      uses: actions/cache@v5
+      with: 
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-
```
Komodo5197: It might be worth a brief comment on what restore-keys actually does.  I had to look it up.

.github/workflows/build.yml
```diff
@@ -64,32 +40,84 @@ jobs:
...
+
+    - name: Cache Dart build cache
+      uses: actions/cache@v5
+      with:
+        path: |
+          .dart_tool
+          build
+        key: ${{ runner.os }}-dart-build-${{ hashFiles('pubspec.lock') }}
```
Komodo5197: This is going to use the same cache for linux x64, linux arm64, and an android.  I think we need another variable to separate all three of these out, because they should be producing distinct artifacts at this stage.

.github/workflows/build.yml
luigi311: im not too familiar with dart/flutter, if this is true then yes we can override the key to remove the arch section of it.

.github/workflows/build.yml
luigi311: all set

.github/workflows/build.yml
luigi311: ohh sweet, swapped back to v2

.github/workflows/build.yml
luigi311: Added it in

.github/workflows/build.yml
luigi311: yeah that makes sense, distinguished all 3 of the linux dart builds now

.github/workflows/build.yml
luigi311: I made the pub cache shared across all 3 linux platform builds, linux x64, arm64, android. Let me know if i should separate out the android one or if it should be fine to keep it together.

Its also important to note , the CI will log a Failed to save in the Post Set up Flutter jobs now when a new cache is being generate on 2/3 of the runners since the key is gonna be shared and only one runner can populate that cache as caches are immutable. This is not a failure though so you only see that message when you open up a action and see the logs. The run is still marked as successful.

Komodo5197: Yeah, I believe the pub-cache is fully platform-agnostic, because it's a bunch of uncompiled fluter projects.  I think in theory it would even be possible to use it cross-os if the cache infastructure was more cooperative, but I'm not 100% sure on that one.

Komodo5197: Actually, I just realized that we currently just overwrite an isar binary with the arm64 version in the pub cache instead of doing something sensible, so the linux arm64 cache should remain separate.  The android one should be fine to merge with linux x64.

luigi311: Yea that makes sense, i added in the aarch tag to the key which will let us do shared cache between the linux x64 and android builds but keeps the linux arm64 builds separate. Also added a comment specifying that once isar is able to be rebuilt without the override then we can remove the aarch specifier.

Chaphasilor: @luigi311 is this thread resolved? if so, we can probably merge this!

luigi311: hi yes i do believe this thread is resolved

Chaphasilor:
@luigi311 thanks for the PR! really nice to have a more efficient CI!

PR #1562, opened 2026-03-22
add amoled theme

### LINKED ISSUE #1229: [FEATURE REQUEST] AMOLED Mode (Pure Black Background)
### Type:
- [x] New feature
- [ ] Improve or extend an existing feature
- [ ] Other: <describe here>

### Explanation 

It would be nice to have an AMOLED Mode toggle below the Theme Dropdown Selector.
I'd make it opt-in - disabled by default - so people don't see a difference except they want to.
I am aware of #220, but this issue seems to be quite old, meaning it was before the redesign.

### Reasoning

If someone has an (AM)OLED display, the experience is, in my opinion, way better with a pure black theme.

### Implementation Notes

https://github.com/jmshrv/finamp/commit/8c6139158f1367839e10c89ecb29de324401d100
I already tried to implement it as far as I could, but I couldn't test it.
Flutter just doesn't work for me...

Thank you in advance, have a great day!

SchweGELBin:
## Changes

| Disabled | Enabled |
| -------- | ------- |
| Screenshot_FinampDebug_Disabled | Screenshot_FinampDebug_Enabled |

Hey, I've added a toggle to enable an AMOLED (pure black) theme that replaces the dark schemes with a darker version:
The background and surface are set to `0xFF000000`.

This option might improve battery life and looks a lot better on AMOLED displays in my opinion.

I've tried to add this feature before, but wasn't able to build the app.
Thanks to the _newly_ provided flake.nix file the build process got a lot easier for NixOS.

## Todo before merging

- [x] Translations

- [x] Reset Settings

## Related Issues

Resolves #1229

### DIFF
--- a/lib/color_schemes.g.dart
+++ b/lib/color_schemes.g.dart
@@ -90,14 +90,20 @@ const darkColorScheme = ColorScheme(
 /// If [color] is provided -> returns a generated color scheme
 /// otherwise falls back to default color schemes
 /// [lightColorScheme] or [darkColorScheme]
-ColorScheme getColorScheme(Color? color, Brightness brightness) {
+ColorScheme getColorScheme(Color? color, Brightness brightness, bool amoledTheme) {
+  ColorScheme scheme = brightness == Brightness.dark ? darkColorScheme : lightColorScheme;
+
   if (color != null) {
-    return ColorScheme.fromSeed(
+    scheme = ColorScheme.fromSeed(
       seedColor: color,
       brightness: brightness,
       dynamicSchemeVariant: DynamicSchemeVariant.fidelity,
     );
   }
 
-  return brightness == Brightness.dark ? darkColorScheme : lightColorScheme;
+  if (amoledTheme && brightness == Brightness.dark) {
+    scheme = scheme.copyWith(background: Color(0xFF000000), surface: Color(0xFF000000));
+  }
+
+  return scheme;
 }
--- a/lib/components/LayoutSettingsScreen/accent_color_selector.dart
+++ b/lib/components/LayoutSettingsScreen/accent_color_selector.dart
@@ -78,6 +78,7 @@ class AccentColorPopup extends ConsumerStatefulWidget {
 
 class _AccentColorPopupState extends ConsumerState<AccentColorPopup> {
   Color? previewColor = FinampSettingsHelper.finampSettings.accentColor;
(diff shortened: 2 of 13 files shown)


Chaphasilor:
@SchweGELBin so this completely strips away any surface tinting, right? I'll have to test how well this works contrast-wise. Most menus use a blurred image as the background anyway, so I guess this only affects very few menus, but most other non-player surfaces.

Aside from that, the PR looks good, thanks!

SchweGELBin:
The TL;DR of this PR is:
It adds a new toggle to the settings that sets the `background` and `surface` values of the *dark theme* to `0xFF000000`.

It's not replacing anything else so there could be some weird looking contrast, but I didn't notice anything "very bad".
The main screens' (home and settings) background are affected and this is what I wanted to change.
In my opinion, this approach is best because of simplicity.

If you would prefer to set a full amoled theme, you can create the const `amoledColorScheme` and do something like this in getColorScheme(): `ColorScheme darkScheme = amoledTheme ? amoledColorScheme : darkColorScheme;`

Thank you for your reply, have a great day!

Chaphasilor:
@SchweGELBin so far the only issue I've found are two menus - "Share Server" and "Authorize QuickConnect Request", both at the bottom of the settings screen. Please either ensure that un-themed menus like these explicitly use a different color when AMOLED mode is enabled, or find a different theme token that works well for the models (contrasting slightly with the background and a bit more strongly with any content/controls within the menus).  
Also, please run `dart format . ` to fix one of the failed checks, you can ignore the other one.

Afterwards I'm happy to merge this! 😁

SchweGELBin:
Hey, I've realized that the Drawer in the Main Menu didn't use the tint and background of the colorscheme, so I've fixed that.

The "Share Server" and "Authorize Quick Connect" dialogs were matching the background color of the current theme.
These dialogs darken the background, while a dialog is visible, which isn't enough for a black backgrounds.
So I've updated these dialogs to use the same tinted color as the new Main Menu Drawer's.

Below, you can see the new Drawer colors.

| Before (Dark) | After (Dark) | After (Black) |
| ------------- | ------------ | ------------- |
| Screenshot_FinampDebug_Before | Screenshot_FinampDebug_Dark | Screenshot_FinampDebug_Black |

Chaphasilor:
I guess that's good enough, thanks!

PR #1563, opened 2026-03-23
feat(player): add swipe-to-skip animation using PageView

### LINKED ISSUE #469: Request: Add Animated Swipe-to-Skip Feature
I would like to request the addition of an animated swipe to skip feature for my media player on iOS. Swipe works for switching songs, but it would be nice to have an animation in the swipe such as spotify does. 

btw: I appreciate the excellent work being done on this project. The app looks fantastic! Although I am currently unable to contribute due to my limited experience, I am eager to learn and become an active member of this community. If there are any resources or guidance on how to contribute available for newcomers like me, I would greatly appreciate it. Sorry for having to deal with ios compatibility.

Thank you for all for creating this beautiful application.

tom2208:
## Changes

Replaces the static `SimpleGestureDetector` in `PlayerScreenAlbumImage` with a `PageView.builder`, giving the swipe-to-skip gesture a native sliding animation and bounce physics instead of an instant track switch.

**What changed:**
- `PlayerScreenAlbumImage` now renders a full-queue `PageView` — one page per track — so swiping produces a smooth slide transition between album covers.
- The current-track index drives the controller: adjacent track changes animate in (400 ms ease-in-out); jumps of more than one page (e.g. looping from end back to beginning) use `jumpToPage` to avoid a sluggish multi-page scroll.
- Queue state is cached locally (`_displayQueue` / `_displayIndex`) so the widget keeps rendering correctly during the brief moment when `currentTrack` is null during track transitions.
- All existing gestures (single tap → play/pause, double tap → favourite, right-click → context menu) are preserved per page.

## Todo before merging

- [x] Translations *(no new UI strings)*
- [x] Reset Settings *(no new settings)*
- [x] looked at CONTRIBUTING.md

## Related Issues

- closes #469

### DIFF
--- a/lib/components/PlayerScreen/player_screen_album_image.dart
+++ b/lib/components/PlayerScreen/player_screen_album_image.dart
@@ -70,6 +70,7 @@ class _PlayerScreenAlbumImageState extends ConsumerState<PlayerScreenAlbumImage>
       stream: queueService.getQueueStream(),
       builder: (context, snapshot) {
         if (!snapshot.hasData) {
+          // show loading indicator
           return const Center(child: CircularProgressIndicator());
         }
 
@@ -134,7 +135,7 @@ class _PlayerScreenAlbumImageState extends ConsumerState<PlayerScreenAlbumImage>
           // image widget's own semantics are suppressed via excludeSemantics.
           label: AppLocalizations.of(context)!.playerAlbumArtworkTooltip(
               queueInfo.currentTrack?.item.title ?? AppLocalizations.of(context)!.unknownName),
-          excludeSemantics: true,
+          excludeSemantics: true, // replace child semantics with custom semantics
           container: true,
 
           child: PageView.builder(
--- a/lib/components/PlayerScreen/player_screen_album_image.dart
+++ b/lib/components/PlayerScreen/player_screen_album_image.dart
@@ -12,97 +12,211 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:get_it/get_it.dart';
 import 'package:simple_gesture_detector/simple_gesture_detector.dart';
 
-import '../../services/current_album_image_provider.dart';
 import '../../services/favorite_provider.dart';
 import '../../menus/track_menu.dart';
 import '../album_image.dart';
 
-class PlayerScreenAlbumImage extends ConsumerWidget {
+/// The swipeable album artwork panel displayed on the player screen.
+///
+/// Renders a [PageView] of album images, one per track in the current queue,
+/// so the user can swipe left/right to skip forward/backward. The displayed
+/// page is kept in sync with [QueueService]: when the queue's current-track
+/// index changes externally (e.g. auto-advance, skip button) the controller
+/// animates or jumps to the correct page, and when the user swipes the page
+/// the offset is forwarded to [QueueService.skipByOffset].
+///
+/// Gesture summary:
+/// - **Swipe left/right** — skip to next/previous track (unless gestures are
+///   disabled in settings).
+/// - **Single tap** — toggle play/pause.
+/// - **Double tap** — toggle favorite for the current track (online mode only).
+/// - **Secondary tap / right-click** — open the track context menu.
+class PlayerScreenAlbumImage extends ConsumerStatefulWidget {
   const PlayerScreenAlbumImage({super.key});
 
   @override
-  Widget build(BuildContext context, WidgetRef ref) {
+  ConsumerState<PlayerScreenAlbumImage> createState() => _PlayerScreenAlbumImageState();
+}
+
+class _PlayerScreenAlbumImageState extends ConsumerState<PlayerScreenAlbumImage> {
+  /// Controller for the [PageView] that drives the swipe-to-skip animation.
+  /// Lazily initialised on the first build that has queue data.
+  PageController? _pageController;
+
+  /// The queue index that was active on the previous build, used to detect
+  /// when the current track has changed so the controller can be animated.
+  int? _lastIndex;
+
+  /// A snapshot of [FinampQueueInfo.fullQueue] captured on the last build.
+  /// Kept as state so the PageView can still render after the queue stream
+  /// emits a null current-track (e.g. briefly during track transitions).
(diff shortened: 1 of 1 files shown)


Chaphasilor:
@tom2208 thanks for the PR!  
Most of these changes appear to be fine, but since it's AI generated we'll have to take a closer look before merging. I saw some old comments being removed for some reason, that shouldn't happen. Also, the gesture summary as a comment isn't really helpful, it will just become outdated eventually.  

Finally, could you see if you can also implement this pageview solution for the `now_playing_bar.dart`, to make it consistent across the app?

In the meantime I'll give this a try to see how well it works and if the animations are sound.

Chaphasilor:
Okay, here's my initial impression about the interaction:

- In general it is what I had in mind, so good job there
- At the moment, if you drag to the sides and keep touching your finger down, you can "hold" the page to take a peek at the previous or next page, which is fine. However, if you slide more than half the cover width to the side, the pageview switches index, and Finamp skips to the previous/next track. I think this "commit" to the next track should only happen when letting go, i.e. I can peek, but nothing happens until I let go. If I drag the cover back towards the center and let go, nothing happens either. Only if I drag to the size (e.g. more than half the width) and then let go does the track change. There could be haptic feedback once past the center, indicating that letting go now will change tracks.  
  Of course, "flinging" the cover to the side quickly should still work - it's still swipe to skip after all
- Can we speed up the animation for `animateToPage`? I think it should be between 200 and 300 ms
- If the screen was off and the next track starts, the slide animation still happens. That should not be the case; animations should only run if the widget is visible.

Another thing about the code I noticed  - the comments mention introducing caching to avoid a `currentTrack` that's temporarily `null`. I think this doesn't actually happen, unless the queue is replaced fully. So I'm not sure if this is really needed?

tom2208:
I restored the old comments and removed the gesture comment. I generated the docs with Claude code and kept everything i deemed worthy. The rest of the code is frankensteined together using me Gemini and the internet ^^

I'll address the next comment soon.

tom2208:
Ok here is what I've been doing:
- I added your suggestion that it only skips if you lift your finger, but that lead me down a road :D You still get haptic feedback but just on releasing your finger, you can also skip ahead fast with swiping without every track being loaded. Instead of load track 1, 2, 3 it just skips right to track 3 that made my experience smoother, so just one `skipByOffset` call
- Cleaned up the currentTrack null comment, the caching just prevents a loading spinner flash during full queue replacement, not a real null-track case
- I formatted the code
- i did NOT implement the screen-off animation guard yet, i will follow up on this

I did quite a bit of testing on my phone and even though it was just the debug version it felt quite good.

tom2208:
Ok, the animation guard also implemented and all seems to work fine. I don't know how to test the "screen off" thing properly though. I would be happy if you could take a look.

Komodo5197:
So I've got a number of thoughts on this:
- I don't really like the slide animation firing on every track change.  It seems a bit overdone, and doesn't match all the rest of the player screen elements just cutting over immediately.  If you two prefer the animation, I want an option to disable it.
- The disable gestures toggle currently just stops switching tracks but still lets you scroll the image to a mismatching one.  You need to fully remove the pageview instead.
- You lost the check to prevent multi-page scrolls in the latest changes.  Also, the one frame delay in index updating allows an incorrect cover to be shown while toggling shuffle.
- The gesture doesn't work on desktop, because the default scrollBehavior ignores mouse drags.
- To prevent excess scrolls, you want something closer to widget visibility, not app visibility.  You can look at _ThemeTransitionCalculator to see how it does it.  I believe you're mostly missing the route check.
- If the track advances while the image is being dragged, things desync until an outside queue change fires.  You should probably be using skipToIndex or something.  And maybe we should be triggering a rebuild when we stop suppressing the index sync up?
- Any checks for a null current track or empty display queue should be unnecessary.  The player screen only builds if there is a current track.  Even the preexisting hasData check for the streambuilder should be unnessecarry, because it seems the StreamBuilder is just missing initialData: queueService.getQueue() to get the first frame build off.
- I'm not sure of a great way to fix this myself, but the album click to pause gesture seems to have gotten finicky due to the pageview stealing the input, at least on desktop.  Also, I noticed a prexisting issue where we already have both a GestureDetector and SimpleGestureDetector - I think these can be simplified?
- There's a matching swipe to skip gesture on the now playing bar that could also use an animation like this.  Of course, getting the full now playing bar to build for an alternate track seems a lot harder than just the album image, so maybe you want to just defer this.
- The comment around the debounce timer occasionally mention the server - this is incorrect.  The thing being communicated with is the local player backend.  Related to this, 400 milliseconds seems like way longer than should be needed to perform a track skip.  You probably don't even need any time at all in the suppress sync, because it seems highly unlikely anything else would cause a rebuild before the track skip comes through.
- I haven't fully thought through the exact logic on the debounce logic and variables, but it seems good at first glance?  Similarly, I think the change to player image loading should cooperate with the existing player image loading and precaching logic, but I haven't fully run through everything yet.

Chaphasilor:
Thanks for the review Komodo!  
Regarding the animation on skip, I just tried the new version with the shorter animation, and I think I agree. The animation is still too slow when skipping very quickly, and there is a short delay before the animation starts, which makes the app feel pretty sluggish (this only applies when skipping using the buttons, the swipe gesture feels really nice now!). So it's probably fine to only use the animation/PageView to support the gestures, but nothing more.  
The only potential benefit of showing the animation at other times would be to confirm the skip direction, but normally that should be apparent just from the locality of the buttons. A "slide and fade" animation that isn't a full side transition but more of a directional blend could work better for that, but is most likely overkill.

Regarding the now playing bar, I'd really like to see this be improved. It should be doable by passing all relevant information (`FinampQueueItem`, progress information, etc.) to the now playing bar widget directly, and then maybe adding very specific reactivity for animation the progress bar for just the current track, using a boolean flag?

Edit: I've now experienced hangs/freezes with this build twice, so something to keep an eye on. It happened after going back into the app or unlocking the screen with the app open, and then trying to unpause playback / interact with the player UI.

tom2208:
I will address all this if I find some time. Got some exams coming up in 2 weeks.

Chaphasilor:
@tom2208 hey, are you still interested in getting this up to speed? 😁

PR #1573, opened 2026-04-03
[Redesign] Add android gain effect disable

### LINKED ISSUE #1343: Unable to play any song on Lenovo tablet - Repeated

### Type of Report

- [x] The app is unusable
- [ ] Something isn't quite right
- [ ] Accessibility
- [ ] None of the above

0.9.19

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

Lenovo Legion Y700 2025 (ZUI 17 / Android 15), Lenovo Xiaoxin Pad Pro 2022 (TB132FU, ZUI 16 / Android 14)

Try to play any song.

### Logs

[finamp-logs-0.9.19.zip](https://github.com/user-attachments/files/21845575/finamp-logs-0.9.19.zip)

### Additional Information

This is a repeated issue of #1238, since for me new updates never fixed this bug and there was no reaction to my comment.

Komodo5197:
Adds a setting to force android devices to use the basic volume control for replay gain instead of the loudness enhancer effect, and a fallback to enable automatically on error  This is a workaround for #1343.  This is currently untested due to the hardware dependency, but I felt like throwing a potential fix together because it seems someone new has hit this error on discord.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -368,6 +368,21 @@ Future<void> _setupOSIntegration() async {
       albumBuffer.asUint8List(albumImageBytes.offsetInBytes, albumImageBytes.lengthInBytes),
     );
   }
+
+  if (Platform.isAndroid) {
+    var themeModeChannel = MethodChannel("com.unicornsonlsd.finamp/set_native_theme");
+    GetIt.instance<ProviderContainer>().listen(finampSettingsProvider.themeMode, (_, mode) {
+      _mainLog.info("Setting android native theme to $mode");
+      themeModeChannel.invokeMethod("setNativeThemeMode", {
+        "targetMode": switch (mode) {
+          ThemeMode.system => 0,
+          ThemeMode.light => 1,
+          ThemeMode.dark => 2,
+        },
+      });
+      // Fire on startup to correct desyncs and apply migration
+    }, fireImmediately: true);
+  }
 }
 
 Future<void> _setupPlaybackServices() async {
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -541,10 +541,10 @@ class _FinampState extends State<Finamp> with WindowListener {
     SchedulerBinding.instance.addPostFrameCallback((_) {
       _uriLinkSubscription = AppLinks().uriLinkStream.listen((uri) async {
         linkHandlingLogger.info("Received link: $uri");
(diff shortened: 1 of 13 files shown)


lib/main.dart
```diff
@@ -368,6 +368,21 @@ Future<void> _setupOSIntegration() async {
       albumBuffer.asUint8List(albumImageBytes.offsetInBytes, albumImageBytes.lengthInBytes),
     );
   }
+
+  if (Platform.isAndroid) {
+    var themeModeChannel = MethodChannel("com.unicornsonlsd.finamp/set_native_theme");
+    GetIt.instance<ProviderContainer>().listen(finampSettingsProvider.themeMode, (_, mode) {
+      _mainLog.info("Setting android native theme to $mode");
+      themeModeChannel.invokeMethod("setNativeThemeMode", {
```
Chaphasilor, line 376: won't this lead to a color flash if the initial system theme doesn't match the app theme?

Chaphasilor:
sorry about the merge conflict. I can resolve and merge later :)

lib/main.dart
Komodo5197: No, this setting is actually persisted by the OS.  I think it might be able to flash during the first app load with the new code, I guess.  But after that the Os will remember and use the correct splash screen matching the in-app setting through the whole load.

Chaphasilor: Well, until the next time the theme is changed, I guess?  
Maybe you could invoke the platform method whenever the theme setting in Finamp is changed, too?

Komodo5197: The call being commented on is already inside a theme setting listener.

Chaphasilor: ah, my bad. let me just build this and see what happens ^^

Chaphasilor: okay, so I tried it out, coming from redesign with dark theme in the app but system in light mode, and on first launch the splash screen was light, and stayed that way. only once the music screen was shown did things turn dark. Next launch was with the dark splash screen.
Not what I expected given our discussion, but I think it's absolutely find with that behavior.

Chaphasilor:
Thanks for these fixes!

PR #1575, opened 2026-04-03
Add support for mTLS client certificates

### LINKED ISSUE #1347: mTLS support to pick a certificate during a connection to the server

### Type of Feature

- [x] Something new
- [ ] Improvement of a feature
- [ ] None of the above

### Description

Would you be able to implement client certificate authentication in the Android app? this works flowlessly in accessing Jellyfinthrough a web browser on mobile but the app needs to code this feature to access the certificates.

This is a good security measure to expose public addresses ie. the Jellyfin Server address without having to rely on VPN and local connections. Very useful security feature that I would much appreciate on this platform.

Many other popular selfhosting apps like HA, Immich, Nextcloud are now supporting this and it should be a simple addition to a good project.

Thanks

P

### Reasoning

_No response_

### Implementation Notes

_No response_

### Additional Information

_No response_

Maxr1998:
## Changes
Adds a new setting that allows importing a `.p12` client certificate to be used for authenticating to a server/reverse proxy implementing mTLS.
The imported certificate and password are stored in Hive and applied to the Dart SecurityContext and Android/JVM SSLContext on startup, if available.
Linux/Windows are unsupported since we cannot supply a certificate to the MediaKit player, as this would require upstream library changes.

Tested on Android (fully functional) and Linux (library browsing works, downloads and media playback are broken).

## Todo before merging
- [ ] ~~Test on iOS/macOS~~ iOS support needs additional modifications, out of scope for this PR
- [x] Limit to supported platforms, e.g., Android + iOS + macOS
- [x] Allow to pick a certificate file during initial connection/login
- [ ] ~~Use separate SecurityContext and not the default one, so that we're able to reset it/clear the certificates (see TODO in `ClientCertificateInstaller`)~~ refactor moved to later PR
- [X] Translations
- [x] Reset Settings (should not be applicable for this)

## Related Issues
Fixes #1347

### DIFF
--- a/lib/components/LoginScreen/login_server_selection_page.dart
+++ b/lib/components/LoginScreen/login_server_selection_page.dart
@@ -4,15 +4,17 @@ import 'package:finamp/components/finamp_icon.dart';
 import 'package:finamp/l10n/app_localizations.dart';
 import 'package:finamp/menus/client_certificate_authentication_menu.dart';
 import 'package:finamp/models/jellyfin_models.dart';
+import 'package:finamp/services/finamp_settings_helper.dart';
 import 'package:finamp/services/jellyfin_api_helper.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
 import 'package:get_it/get_it.dart';
 import 'package:logging/logging.dart';
 
 import 'login_flow.dart';
 
-class LoginServerSelectionPage extends StatefulWidget {
+class LoginServerSelectionPage extends ConsumerStatefulWidget {
   static const routeName = "login/server-selection";
 
   final ServerState serverState;
@@ -21,10 +23,10 @@ class LoginServerSelectionPage extends StatefulWidget {
   const LoginServerSelectionPage({super.key, required this.serverState, this.onServerSelected});
 
   @override
-  State<LoginServerSelectionPage> createState() => _LoginServerSelectionPageState();
+  ConsumerState<LoginServerSelectionPage> createState() => _LoginServerSelectionPageState();
 }
 
-class _LoginServerSelectionPageState extends State<LoginServerSelectionPage> {
+class _LoginServerSelectionPageState extends ConsumerState<LoginServerSelectionPage> {
   static final _loginServerSelectionPageLogger = Logger("LoginServerSelectionPage");
 
   final jellyfinApiHelper = GetIt.instance<JellyfinApiHelper>();
@@ -69,6 +71,9 @@ class _LoginServerSelectionPageState extends State<LoginServerSelectionPage> {
 
   @override
   Widget build(BuildContext context) {
+    final clientCertificate = ref.watch(finampSettingsProvider.clientCertificate);
+    final isClientCertificateInstalled = clientCertificate != null;
+
     return SingleChildScrollView(
       padding: const EdgeInsets.symmetric(horizontal: 32.0),
       child: Center(
@@ -99,8 +104,10 @@ class _LoginServerSelectionPageState extends State<LoginServerSelectionPage> {
             _buildServerUrlInput(context),
             ConstrainedBox(
               constraints: const BoxConstraints(minHeight: 95.0),
-              child: widget.serverState.manualServer != null
-                  ? Padding(
+              child: Column(
+                children: [
+                  if (widget.serverState.manualServer != null)
+                    Padding(
                       padding: const EdgeInsets.only(top: 12.0),
                       child: JellyfinServerSelectionWidget(
                         baseUrl: widget.serverState.baseUrl,
@@ -109,9 +116,19 @@ class _LoginServerSelectionPageState extends State<LoginServerSelectionPage> {
                           widget.onServerSelected?.call(widget.serverState.manualServer!, widget.serverState.baseUrl!);
                         },
                       ),
+                    ),
+                  if (isClientCertificateInstalled)
+                    Padding(
+                      padding: const EdgeInsets.symmetric(vertical: 12.0),
+                      child: CTAMedium(
+                        text: AppLocalizations.of(context)!.clientCertificateInstalled,
+                        icon: TablerIcons.certificate,
+                        onPressed: () => showClientCertificateMenu(context: context),
+                        minWidth: 0,
+                      ),
                     )
-                  : widget.serverState.clientCertificateRequired
-                  ? Column(
+                  else if (widget.serverState.clientCertificateRequired)
+                    Column(
                       mainAxisAlignment: MainAxisAlignment.center,
                       crossAxisAlignment: CrossAxisAlignment.center,
                       children: [
@@ -134,9 +151,9 @@ class _LoginServerSelectionPageState extends State<LoginServerSelectionPage> {
                           ],
                         ),
                       ],
-                    )
-                  : widget.serverState.baseUrlToTest != null
-                  ? Padding(
+                    ),
+                  if (widget.serverState.baseUrlToTest != null && widget.serverState.manualServer == null)
+                    Padding(
                       padding: const EdgeInsets.only(top: 12.0),
                       child: Row(
                         mainAxisAlignment: MainAxisAlignment.center,
@@ -149,8 +166,9 @@ class _LoginServerSelectionPageState extends State<LoginServerSelectionPage> {
                           ),
                         ],
                       ),
-                    )
-                  : const SizedBox.shrink(),
+                    ),
+                ],
+              ),
             ),
             Padding(
               padding: const EdgeInsets.only(top: 20.0, bottom: 16.0),
--- a/lib/components/LoginScreen/login_server_selection_page.dart
+++ b/lib/components/LoginScreen/login_server_selection_page.dart
@@ -4,12 +4,15 @@ import 'package:finamp/components/finamp_icon.dart';
 import 'package:finamp/l10n/app_localizations.dart';
 import 'package:finamp/menus/client_certificate_authentication_menu.dart';
 import 'package:finamp/models/jellyfin_models.dart';
+import 'package:finamp/services/client_certificate_installer.dart';
 import 'package:finamp/services/finamp_settings_helper.dart';
 import 'package:finamp/services/jellyfin_api_helper.dart';
+import 'package:finamp/services/server_client_discovery_service.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
 import 'package:get_it/get_it.dart';
+import 'package:http/http.dart' show ClientException;
 import 'package:logging/logging.dart';
 
 import 'login_flow.dart';
@@ -42,12 +45,37 @@ class _LoginServerSelectionPageState extends ConsumerState<LoginServerSelectionP
       }
     };
 
+    _startDiscovery();
+  }
+
+  void _startDiscovery() {
     widget.serverState.clientDiscoveryHandler.discoverServers((ClientDiscoveryResponse response) async {
-      _loginServerSelectionPageLogger.finer("Found server: ${response.name} at ${response.address}");
+      _loginServerSelectionPageLogger.fine("Found server '${response.name}' at ${response.address}");
 
       final serverUrl = Uri.parse(response.address!);
-      PublicSystemInfoResult? serverInfo = await jellyfinApiHelper.loadCustomServerPublicInfo(serverUrl);
-      _loginServerSelectionPageLogger.finer("Server info: ${serverInfo?.toJson()}");
+      final PublicSystemInfoResult? serverInfo;
+      try {
+        serverInfo = await jellyfinApiHelper.loadCustomServerPublicInfo(serverUrl);
+      } catch (error) {
+        if (ClientCertificateInstaller.isSupported &&
+            error is ClientException &&
+            error.message.contains("TLSV1_ALERT_CERTIFICATE_REQUIRED")) {
+          _loginServerSelectionPageLogger.info(
+            "Discovered server '${response.name}' at ${response.address} requires an mTLS client certificate",
+          );
+          if (mounted && !widget.serverState.clientCertificateRequired) {
+            setState(() {
+              widget.serverState.clientCertificateRequired = true;
+            });
+          }
+        } else {
+          _loginServerSelectionPageLogger.severe(
+            "Failed to load public server info for discovered server '${response.name}' at ${response.address}: $error",
+          );
+        }
+        return;
+      }
+      _loginServerSelectionPageLogger.fine("Server info from ${response.address}: ${serverInfo?.toJson()}");
       if (serverInfo != null && mounted) {
         if (serverInfo.serverName == null) {
           serverInfo.serverName = response.name;
@@ -57,10 +85,19 @@ class _LoginServerSelectionPageState extends ConsumerState<LoginServerSelectionP
         }
         // no need to filter duplicates, we're using a map
         setState(() {
-          widget.serverState.discoveredServers[serverUrl] = serverInfo;
+          widget.serverState.discoveredServers[serverUrl] = serverInfo!;
         });
       }
     });
+  }
+
+  /// Discards the current discovery handler and starts over with a fresh socket.
+  /// The old socket can end up in a broken state after the app was backgrounded, e.g. while picking a certificate file,
+  /// and servers that responded before a certificate was installed need to be probed again either way.
+  void _restartDiscovery() {
+    widget.serverState.clientDiscoveryHandler.dispose();
+    widget.serverState.clientDiscoveryHandler = JellyfinServerClientDiscovery();
+    _startDiscovery();
   }
 
   @override
@@ -148,6 +185,7 @@ class _LoginServerSelectionPageState extends ConsumerState<LoginServerSelectionP
                               onPressed: () => showClientCertificateMenu(
                                 context: context,
                                 onImported: () {
+                                  _restartDiscovery();
                                   final baseUrl = widget.serverState.baseUrl;
                                   if (baseUrl != null) {
                                     widget.serverState.onBaseUrlChanged(baseUrl);
--- a/lib/components/LoginScreen/login_server_selection_page.dart
+++ b/lib/components/LoginScreen/login_server_selection_page.dart
@@ -145,7 +145,15 @@ class _LoginServerSelectionPageState extends ConsumerState<LoginServerSelectionP
                             CTAMedium(
                               text: AppLocalizations.of(context)!.importClientCertificate,
                               icon: TablerIcons.certificate,
-                              onPressed: () => showClientCertificateMenu(context: context),
+                              onPressed: () => showClientCertificateMenu(
+                                context: context,
+                                onImported: () {
+                                  final baseUrl = widget.serverState.baseUrl;
+                                  if (baseUrl != null) {
+                                    widget.serverState.onBaseUrlChanged(baseUrl);
+                                  }
+                                },
+                              ),
                               minWidth: 0,
                             ),
                           ],
--- a/lib/components/LoginScreen/login_server_selection_page.dart
+++ b/lib/components/LoginScreen/login_server_selection_page.dart
@@ -1,9 +1,11 @@
+import 'package:finamp/components/Buttons/cta_medium.dart';
 import 'package:finamp/components/Buttons/simple_button.dart';
 import 'package:finamp/components/finamp_icon.dart';
+import 'package:finamp/l10n/app_localizations.dart';
+import 'package:finamp/menus/client_certificate_authentication_menu.dart';
 import 'package:finamp/models/jellyfin_models.dart';
 import 'package:finamp/services/jellyfin_api_helper.dart';
 import 'package:flutter/material.dart';
-import 'package:finamp/l10n/app_localizations.dart';
 import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
 import 'package:get_it/get_it.dart';
 import 'package:logging/logging.dart';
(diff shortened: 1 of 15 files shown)


Maxr1998:
Ok, on the `Use separate SecurityContext and not the default one, so that we're able to reset it/clear the certificates` TODO:
The modified `SecurityContext` needs to be passed to each created `HttpClient`, so specifically to each `ChopperClient` that's (re-)created in the app, including in `JellyfinApi.create()`. Since a lot of those are currently final, this would require quite a few changes, which feel out-of-scope for this PR. If you're ok with it, I'd keep the current behavior of modifying the default context, which means that removing a certificate requires restarting the app to apply.

Chaphasilor:
@Maxr1998 restarting the app to remove it fine. It would just be annoying when applying a new certificate.  
Are mTLS certificates always applied on a per-app basis? Is there any way to apply them globally?  

---

Also, I currently have no idea how to test this. Guess I'd have to look into setting up mTLS myself for that. **If anyone else already has such a setup, please help with testing this.** You can download the APK from the Actions/CI page: https://github.com/UnicornsOnLSD/finamp/actions/runs/23960312203

Maxr1998:
> Are mTLS certificates always applied on a per-app basis? Is there any way to apply them globally?

They need to be stored per app, since we need direct access to the certificate. Certificates installed in the Android keystore unfortunately don't work with Flutter.

---

> Also, I currently have no idea how to test this. Guess I'd have to look into setting up mTLS myself for that.

You can try `https://finamp-test.maxr1998.de/stable` with [this certificate file](https://github.com/user-attachments/files/26662282/cert.zip) (import password can be left empty). It forwards requests (through my server) to the Jellyfin demo server, I set up the CA & my reverse proxy the other day to let someone at FFmpeg test it with my bug report.

(I'll rebase the PR and fix the conflicts after your review. I might be a bit too busy in the coming days.)

lib/menus/client_certificate_authentication_menu.dart
```diff
@@ -0,0 +1,197 @@
...
+
+class _ClientCertificateMenuContentState extends ConsumerState<_ClientCertificateMenuContent> {
+  final _clientCertificateInstaller = ClientCertificateInstaller();
+  bool _importing = false;
+
+  Future<void> _importCertificate() async {
+    final l10n = AppLocalizations.of(context)!;
+
+    final result = await FilePicker.platform.pickFiles(type: FileType.custom, allowedExtensions: ['p12', 'pfx']);
+    if (result == null || result.files.single.path == null) return;
```
cbueth: If the user cancels the file picker or no certificates exist on the device, the function silently returns with no feedback. This is not breaking the functionality, but a hint could be more helpful.
I suggest showing a snackbar with feedback when `result == null`.


```suggestion
    final result = await FilePicker.platform.pickFiles(type: FileType.custom, allowedExtensions: ['p12', 'pfx']);
    if (result == null) {
      if (mounted) ScaffoldMessenger.of(context).showSnackBar(
        SnackBar(content: Text(l10n.clientCertificateNoCertFound)),
      );
      return;
    }
    if (result.files.single.path == null) return;
```

lib/components/LoginScreen/login_server_selection_page.dart
```diff
@@ -98,7 +100,46 @@ class _LoginServerSelectionPageState extends State<LoginServerSelectionPage> {
...
+                              CTAMedium(
+                                text: AppLocalizations.of(context)!.importClientCertificate,
+                                icon: TablerIcons.certificate,
+                                onPressed: () => showClientCertificateMenu(context: context),
+                                minWidth: 0,
+                              ),
+                            ],
+                          ),
+                        ],
+                      )
```
cbueth: The certificate import option only appears after receiving `TLSV1_ALERT_CERTIFICATE_REQUIRED` error. This is a design choice, in the sense that this option only needs to be shown at the point we know a cert is required, correct?

There's no indicator during normal login flow showing whether a certificate is installed. Users can't know their cert setup before encountering a connection error.

There could be a certificate status indicator. I suggest adding after the server URL input (before the discovered servers section), a status row along the lines of this:

```java
if (ClientCertificateInstaller.isSupported) {
  final cert = ref.watch(finampSettingsProvider.clientCertificate);
  Padding(
    padding: const EdgeInsets.only(top: 8.0),
    child: Row(
      mainAxisAlignment: MainAxisAlignment.center,
      children: [
        Icon(
          cert != null ? TablerIcons.certificate : TablerIcons.certificate_off,
          size: 16,
          color: Theme.of(context).colorScheme.onSurface.withAlpha(153),
        ),
        const SizedBox(width: 8),
        Text(
          cert != null ? l10n.clientCertificateInstalled : l10n.clientCertificateNone,
          style: Theme.of(context).textTheme.bodySmall,
        ),
      ],
    ),
  );
}
```

lib/components/LoginScreen/login_server_selection_page.dart
```diff
@@ -98,7 +100,46 @@ class _LoginServerSelectionPageState extends State<LoginServerSelectionPage> {
...
+                              style: Theme.of(context).textTheme.bodyLarge,
+                            ),
+                          ),
+                          Row(
+                            mainAxisAlignment: MainAxisAlignment.center,
+                            children: [
+                              CTAMedium(
+                                text: AppLocalizations.of(context)!.importClientCertificate,
+                                icon: TablerIcons.certificate,
+                                onPressed: () => showClientCertificateMenu(context: context),
```
cbueth: Here thecertificate import could be hidden in an expandable section, like 'More options' or 'Advanced' that would reveal it.

cbueth (commented):
I'm interested in this feature as there are few music self-hosted apps that support mTLS (none that I know of). Since I was considering implementing this myself, I'm happy to see this progress.

Is there a technical reason you chose Hive over Android KeyChain? Using the KeyChain may be simpler as it hands off certificate management through system UI and is hardware-backed, which also improves security since the raw certificate bytes don't need to be stored locally, securing the certificate from device storage extraction. I have implemented mTLS in another Android app where I used the Android KeyChain (https://github.com/mtotschnig/MyExpenses/pull/1905).

There are three suggestions I have added in the changes. None of them are breaking or important for the functionality, but might be nice for the users.

Again, thank you, @Maxr1998, for working on this! I hope my unsolicited review helps in some way.

lib/menus/client_certificate_authentication_menu.dart
Maxr1998: Good point, I'll add that.

lib/components/LoginScreen/login_server_selection_page.dart
Maxr1998: It was indeed a design choice, since I didn't want to overwhelm users who don't have mTLS set up and never heard about it in the first place. Showing an indicator if a certificate already exists makes sense though.

Maxr1998:
> Is there a technical reason you chose Hive over Android KeyChain? Using the KeyChain may be simpler as it hands off certificate management through system UI and is hardware-backed, which also improves security since the raw certificate bytes don't need to be stored locally, securing the certificate from device storage extraction. I have implemented mTLS in another Android app where I used the Android KeyChain (https://github.com/mtotschnig/MyExpenses/pull/1905).

My main motivation behind this was retaining access to the raw certificate bytes, since we need to feed them into Flutter's SecurityContext and also mpv on Linux. Android's KeyChain doesn't return those, unfortunately. Ideally, we'd store these private keys in flutter_secure_storage, but for a first implementation I found the app's private storage good enough (especially since that's where we store the access token as well, please tell me if I'm wrong).

And thanks for the feedback & review, definitely helps to have more eyes on this!

lib/components/LoginScreen/login_server_selection_page.dart
cbueth: Thanks for the quick response and openness. After reviewing the code I just installed it on my device and I can fill in the server address, but for me no import option appears. Looking into the logs, the reply it sends does not seem to be `TLSV1_ALERT_CERTIFICATE_REQUIRED` but a "400 No required SSL certificate was sent". This is using nginx as a proxy with ssl_protocol TLS1.3 TLSv1.2. I assume this is the nginx default behaviour. So even when adding some second trigger the import action, I think there might always be user configurations that we cannot cover which need support sending the client certificate in the first place.

What proxy did you test with?

cbueth:
Ok, I was imagining there would be a reason along these lines. I think the level of security fine for this case and for now.

cbueth:
As I wanted to rebuild with the selection changes, I have added some changes on a fork in https://github.com/cbueth/finamp/commit/a76acf748b7c1cbb46d4de50562b992c54106f67 and https://github.com/cbueth/finamp/commit/01667507a1e596fe4cb45c5dc5b205994cfc6e69. The certificate selection looks quite nice, and I was able to connect to my server and play music. 👌 

With these changes I could add a `.pfx` certificate from local storage. After typing in the password and clicking 'ok', I get an error saying `package:flutter/src/widgets/framework.dart': Failed assertion: line 6271 pos 12 '_dependents.isEmpty': is not true.`. Restarting the app showed that the certificate was successfully added. I suppose the app tried to update the UI (or a provider) after the certificate was successfully processed, but the widget that started the process was no longer in a valid state to handle the update. Maybe this is some problem of the changes.

In case I notice anything not working in the app, I will report back.

lib/components/LoginScreen/login_server_selection_page.dart
Maxr1998: Thanks for testing with nginx! I've been using caddy in my test, I haven't yet researched how other proxies behave and if there's a standard for it.

thiagofigueiro:
## Hive schema compatibility issue on iOS

When installing this PR over an existing Finamp install (from a different branch), the app crashes on startup with:

```
type 'bool' is not a subtype of type 'FinampHomeScreenConfiguration' in type cast
```

The Hive field indices in this PR differ from redesign; field 151 is used for `ClientCertificate` here, but on the current redesign branch that index is assigned to a bool (`homeScreenImageSize?`). Installing over existing on-device data reads the wrong type.

Workaround: delete the app and reinstall fresh to clear the old Hive database.

Might be worth noting in the PR that this will affect existing users upgrading. They'd need a Hive migration or a type adapter version bump.

thiagofigueiro:
## Can't access server on iOS

After typing the server address, I get asked for the client certificate. This succeeds, and the `Connect to Jellyfin` screen shows `Certificate installed`.

However, the server doesn't appear nor we progress to the library. All I see is an infinite spinner for `Scanning for servers...` under `Servers on your local network:`.

lib/components/LoginScreen/login_server_selection_page.dart
Maxr1998: > Showing an indicator if a certificate already exists makes sense though.

Added that during my last rebase.

Maxr1998:
> ## Hive schema compatibility issue on iOS
> When installing this PR over an existing Finamp install (from a different branch), the app crashes on startup with:
> 
> ```
> type 'bool' is not a subtype of type 'FinampHomeScreenConfiguration' in type cast
> ```
> 
> The Hive field indices in this PR differ from redesign; field 151 is used for `ClientCertificate` here, but on the current redesign branch that index is assigned to a bool (`homeScreenImageSize?`). Installing over existing on-device data reads the wrong type.

Which branches did you test before? 151 doesn't seem to be assigned in FinampSettings on the current redesign.

Maxr1998:
> ## Can't access server on iOS
> After typing the server address, I get asked for the client certificate. This succeeds, and the `Connect to Jellyfin` screen shows `Certificate installed`.
> 
> However, the server doesn't appear nor we progress to the library. All I see is an infinite spinner for `Scanning for servers...` under `Servers on your local network:`.

I looked at your implementation on iOS, and it appears that a lot more is required for mTLS on iOS than this branch currently does. Specifically, splitting out the SecurityContext like I mentioned a few posts above ("Ok, on the `Use separate SecurityContext and not the default one, so that we're able to reset it/clear the certificates` TODO") might actually be a requirement, which is complex enough to warrant another PR after this one is merged. Thus, I'll only enable Android for now.

lib/menus/client_certificate_authentication_menu.dart
Maxr1998: Thinking about it again, I found that cancelling the selection (pressing back) is a deliberate action that doesn't warrant an extra message.

thiagofigueiro:
> Which branches did you test before? 151 doesn't seem to be assigned in FinampSettings on the current redesign.

@Maxr1998, my mistake. The collision is at field 146.

I was running my own branch (based on redesign, with the same FinampSettings Hive layout). Installing the PR over it hit this mismatch.

This PR introduces `FinampHomeScreenConfiguration homeScreenConfiguration` at `@HiveField(146)`, but on redesign that slot holds `bool amoledTheme`. 

When Hive reads the stored bool and tries to cast it to FinampHomeScreenConfiguration, you get the crash I saw.

thiagofigueiro:
@Maxr1998 

> I looked at your implementation on iOS, and it appears that a lot more is required for mTLS on iOS than this branch currently does. 

The iOS subsystems (AVPlayer, URLSession, WebSocket) all bypass Dart's `SecurityContext.defaultContext`, so a separate approach is needed there anyway.

Thanks for working on this feature. It's a step in the right direction, and it's reasonable that the iOS support will take longer. 🙏

Maxr1998:
> The collision is at field 146.
> 
> I was running my own branch (based on redesign, with the same FinampSettings Hive layout). Installing the PR over it hit this mismatch.
> 
> This PR introduces `FinampHomeScreenConfiguration homeScreenConfiguration` at `@HiveField(146)`, but on redesign that slot holds `bool amoledTheme`. 
> 
> When Hive reads the stored bool and tries to cast it to FinampHomeScreenConfiguration, you get the crash I saw.

I'm slightly confused, this branch is a clean rebase on top of the current redesign, where amoledTheme uses an ID of 148. Maybe you were running a version before the merge of the home screen PR?

thiagofigueiro:
>  Maybe you were running a version before the merge of the home screen PR?

That was it. 👍 

There should be no issues upgrading. Apologies for the noise.

cbueth:
Thanks, @Maxr1998 for adding the suggested change in bfed52e. I have been able to use the debug version for a month without noticing problems.

Chaphasilor:
@thiagofigueiro conflicting Hive IDs are pretty standard. We always correct them before/after merging a PR, so that updating from the `redesign` branch should always work (or if not, at least upgrading between releases should work). So if you ever get this when trying out something, just delete the Finamp settings and restart the app (of course, you shouldn't be using these versions as your main install, just for testing, since you could lose settings and downloads at any time).  

@Maxr1998 as for the PR, I'll take a look at the UI, since this is the only thing I can review. And since @cbueth and @thiagofigueiro were nice enough to try this out with a proper and tested setup, that's more than enough for me and better than what I could offer :)

Maxr1998:
I can also rebase it if necessary ^^

Chaphasilor:
merging is fine ^^

Chaphasilor (changes requested):
So at the moment this will require setting up a client certificate if *any* server, including auto-discovered ones, has mTLS set up. This is probably not what we want? Although I guess without the mTLS cert set up, FInamp won't be able to load the server info anyway. Maybe we should change the wording around this. 

I guess we'd also need a "re-fetch servers" button or something like that for server auto-discovery in combination with mTLS?
 
I also don't see why the button should be conditionally hidden. If someone knows they want to set up mTLS, they shouldn't have to make a failing request first ^^

Another unrelated thing we could do regarding the translations is extract the certificate states into an enum and then use a `select` command(?) in the translations instead of creating one translation token per state. But for the two states it's probably fine?

The final thing I noticed are that the logs and language buttons look a bit weird, since they are in a different scaffold than the bottom sheet? Maybe you can try to fix that, but it's nothing dramatic:

Image

Maxr1998:
> So at the moment this will require setting up a client certificate if _any_ server, including auto-discovered ones, has mTLS set up. This is probably not what we want? Although I guess without the mTLS cert set up, Finamp won't be able to load the server info anyway. Maybe we should change the wording around this.

Realistically, most people will have exactly one server on their network/entered in the URL bar. mTLS won't even work for most local servers since mTLS also needs TLS and that's probably rare for local-LAN-IP servers, and I'm not entirely sure if Jellyfin/Finamp discovery works that well with reverse proxies/when using TLS. And for servers that you enter yourself, the current behavior feels ok to me. Please clarify though if I misunderstand anything :sweat_smile:

> I guess we'd also need a "re-fetch servers" button or something like that for server auto-discovery in combination with mTLS?

It should be automatic, so I don't see the need for a refresh button :thinking:

> I also don't see why the button should be conditionally hidden. If someone knows they want to set up mTLS, they shouldn't have to make a failing request first ^^

That's a fair point, but as a counter-argument, should it always be visible if you don't care about mTLS? A failed requst isn't all too bad, it signals the app that it needs to authenticate — same thing in any browser before you install the certificate.

> Another unrelated thing we could do regarding the translations is extract the certificate states into an enum and then use a `select` command(?) in the translations instead of creating one translation token per state. But for the two states it's probably fine?

Imo, for just two strings, using select unnecessarily complicates things.

> The final thing I noticed are that the logs and language buttons look a bit weird, since they are in a different scaffold than the bottom sheet? Maybe you can try to fix that, but it's nothing dramatic:

Good catch, I didn't even notice that in dark mode. I'll fix it.

cbueth:
> > I guess we'd also need a "re-fetch servers" button or something like that for server auto-discovery in combination with mTLS?
> 
> It should be automatic, so I don't see the need for a refresh button 🤔

When testing in with another user of my setup, we failed setting it up. If I remember it right, the log seemed to point that it tried to connect before adding the user certificate and did not afterwards. In this case a re-fetch in set-up could be helpful.
Was there automatic refresh logic added or is it the same as from before?

Maxr1998:
> When testing in with another user of my setup, we failed setting it up. If I remember it right, the log seemed to point that it tried to connect before adding the user certificate and did not afterwards. In this case a re-fetch in set-up could be helpful.
> Was there automatic refresh logic added or is it the same as from before?

My memory failed me when I initially wrote my reply and it was indeed not implemented yet. I finally did that yesterday night though.

cbueth:
> My memory failed me when I initially wrote my reply and it was indeed not implemented yet. I finally did that yesterday night though.

Thanks for the confirmation @Maxr1998! I suppose it is in 5be6adb.

Chaphasilor:
1. The behavior is fine, although I would prefer more robust support for multiple servers. We should at least be showing that a server was discovered but can't be contacted due to a missing certificate. Discovery could work fine even with reverse proxies, as long as UDP access is allowed.
2. The refresh button was in the case of servers being discovered before the user sets up the certificate. In that case, any previously discovered servers couldn't be contacted for their version and name and such, and we should probably retry that request once the certificate has been installed. What you added now is just a retry for the manual server input I think, not for any auto-discovered servers.
3. I guess it's fine either way. How does Immich handle it? They've probably discussed this too, and I'm fine with taking the same approach

Maxr1998:
> 1. The behavior is fine, although I would prefer more robust support for multiple servers. We should at least be showing that a server was discovered but can't be contacted due to a missing certificate. Discovery could work fine even with reverse proxies, as long as UDP access is allowed.

I somehow completely missed the discovery path, probably because my personal server's auto-discovery was completely broken. You unknowingly forced me to finally fix that, so thanks for that :smile:
Previously, auto-discovered servers that could not be contacted would silently be ignored. I made some changes now that mirror the manual server entry by showing the client certificate prompt if any auto-discovered server requires mTLS. The UI isn't ideal yet, and showing the prompt next to the respective server would be nicer. However, that would require a bigger rewrite of the UI, which I'd prefer to avoid to not increase the size of this PR even further.

> 2. The refresh button was in the case of servers being discovered before the user sets up the certificate. In that case, any previously discovered servers couldn't be contacted for their version and name and such, and we should probably retry that request once the certificate has been installed. What you added now is just a retry for the manual server input I think, not for any auto-discovered servers.

I handled this now by restarting discovery completely after a certificate is installed. That way, the discovered server shows up immediately.

> 3. I guess it's fine either way. How does Immich handle it? They've probably discussed this too, and I'm fine with taking the same approach

Correct me if I'm wrong, but Immich doesn't have auto-discovery at all in the mobile app. Instead, what they do is have a settings button below the URL input, and from the settings, you can simply import one mTLS certificate to be used globally. Obviously, the certificate is only used for the servers that request it, but there's no way to import multiple certificates, for example. We could, of course, implement something to handle multiple certificates, but I'm pretty sure that would involve dynamically swapping the SecurityContext, which is a much bigger refactor, and considering the average user has exactly one Jellyfin server… I'd say it is absolutely not worth the effort.

Chaphasilor:
My third point was about showing the mTLS cert option before a request fails, sorry for not providing adequate context there 😅  
So if Immich exposes it in the settings, I guess technically users could set up the cert before. I'm a bit worries about people desperately searching for such a button because they *know* the request will fail, and therefore never triggering the button.  
So how about using a `SimpleButton` with a "Set up an mTLS certificate" prompt below the URL input, and upgrading that to the full CTA if a request fails due to a missing cert?

All other changes sound good to me, thanks!

Happy that you managed to fix your setup! 😁

Maxr1998:
> My third point was about showing the mTLS cert option before a request fails, sorry for not providing adequate context there 😅

Ah, got it.

> So if Immich exposes it in the settings, I guess technically users could set up the cert before. I'm a bit worries about people desperately searching for such a button because they *know* the request will fail, and therefore never triggering the button.  
> So how about using a `SimpleButton` with a "Set up an mTLS certificate" prompt below the URL input, and upgrading that to the full CTA if a request fails due to a missing cert?

I'm personally worried that always showing such a button would confuse the majority of users who don't run an mTLS setup, perhaps assuming that it would be required to use Finamp at all. And then they might not try to connect anyway, and instead stop using Finamp. Or they would already be confused about what mTLS even is supposed to be.

In Immich the mTLS option is one of multiple options in the settings, so more hidden away, which noticeably reduces the risk of such confusion. 

I also honestly think that people running an mTLS setup would be more likely to just attempt to connect (or ideally have their auto-discovery show the certificate request immediately). And even if not, the volume of those people would be much lower so the impact would also be lower.

Chaphasilor:
Okay, let's do it this way for now. If we ever get someone asking how to set up mTLS, we can discuss this again 😁  
Is there anything blocking a merge or in need of another review? @cbueth did you give this another test with the latest changes? :)

Maxr1998:
> Okay, let's do it this way for now. If we ever get someone asking how to set up mTLS, we can discuss this again 😁

Yes, I'll think about better approaches, but I'd still merge this as is for now.

> Is there anything blocking a merge or in need of another review?

Not from my side :slightly_smiling_face:

cbueth:
> Okay, let's do it this way for now. If we ever get someone asking how to set up mTLS, we can discuss this again 😁 Is there anything blocking a merge or in need of another review? @cbueth did you give this another test with the latest changes? :)

As of now, I did not test with the latest changes.

Chaphasilor:
Alright, I'll merge this, and if you get around to testing before the next update, just build the `redesign` branch directly :)

Thanks for the PR Max!

lib/models/finamp_models.dart
```diff
@@ -920,6 +921,9 @@ class FinampSettings {
   @HiveField(150, defaultValue: DefaultSettings.homeScreenImageSizeMobile)
   int homeScreenImageSize;
 
+  @HiveField(151, defaultValue: DefaultSettings.clientCertificate)
```
Komodo5197, line 924: Wouldn't it make more sense to put this into the FinampUser than the generic settings?  Especially if it's needed in the background isolate.

Maxr1998: Doesn't feel to me like it belongs there. Also, this would clash with the server selection where the credentials (and thus the FinampUser?) aren't available yet.

Komodo5197: I feel like it would make sense there, because it's server-specific and if you changed your login to a different server you would need data.  Of course, that would also force you to re-enter the data if you logged back in to the same server, which is a downside.  I guess it's fine where it is.

PR #1579, opened 2026-04-04
fix(android-auto): paginate large libraries and add Browse-by-Letter

### LINKED ISSUE #1576: Android Auto app doesn't render/return all Artists.

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.9.23

### Affected Devices/Platform

- [ ] Android
- [x] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

_No response_

When I try to select an artist from the A-Z menu option, it doesn't render/return all Artists. Its only returning some letters. Yes there are artists in Jellyfin for the other letters.

Image

### Logs

<details>

  <summary>Logs</summary>
  
</details> 

### Additional Information

_No response_

pdebuitlear:
## Summary

Fixes #1576

Returning entire album/artist collections in a single Android Auto `getChildren` response causes a `FAILED BINDER TRANSACTION` crash due to the ~1MB Binder IPC buffer limit. This PR fixes the crash and adds a Browse-by-Letter feature for navigating large libraries.

- **Pagination**: root collection responses are capped at 200 items per page; a non-playable "More… (N remaining)" node is appended when further pages exist
- **Browse by Letter**: a "Browse by Letter" node appears at the top of page 1 for Albums and Artists, opening an A–Z + # index; each letter page is independently paginated with its own "More…" node if needed
- **Jellyfin API**: added `NameStartsWith` query parameter to `getItems`, `getArtists`, and `getAlbumArtists` for server-side letter filtering (the `#` bucket falls back to client-side filtering)
- **MediaItemId**: added `nameFilter` (HiveField 4) and `pageStartIndex` (HiveField 5) to encode browse state inside Android Auto node IDs
- **Docs**: added `docs/android_auto_browsing.md` with implementation notes and Android Auto platform learnings

## Test plan

- [ ] Browse Albums/Artists on Android Auto with a large library (500+ items) — confirm no crash
- [ ] Scroll through paginated pages using "More…" nodes
- [ ] Open "Browse by Letter", select a letter, confirm correct items shown
- [ ] Confirm "More…" appears within a letter page when > 200 results
- [ ] Test offline mode — letter filtering and pagination should work from downloaded items
- [ ] Confirm Playlists, Genres, and Tracks are unaffected (no Browse by Letter node)


Chaphasilor:
@pdebuitlear thanks for the PR!  
The approach using special items is clever, although it's of course frustrating that we have to resort to this.

I'm guessing this doesn't play well with the built-in letter selector of AA? If so, can we disable it (maybe you did that already)?

pdebuitlear:
> @pdebuitlear thanks for the PR! The approach using special items is clever, although it's of course frustrating that we have to resort to this.
> 
> I'm guessing this doesn't play well with the built-in letter selector of AA? If so, can we disable it (maybe you did that already)?

The built-in letter selector is problematic. 

From Claude: 
"_**Can it be hidden or overridden?**
No on both counts. The native A-Z button is rendered entirely by the Android Auto host (Google's Android Auto app for phone projection, or the OEM car-ui-lib for AAOS). There is no documented MediaBrowserServiceCompat extra, MediaItem flag, or onGetRoot() hint that suppresses it. Its letter-tap fires an internal scroll command — no onLoadChildren callback is sent to the app, so it cannot be intercepted to call the Jellyfin API._"




An alternative is the following:

**Letter-first navigation**
Skip the flat list entirely for Artists and Albums. When the user taps Artists or Albums, immediately show the A–Z + # letter grid instead of a flat page of items:
```

Root → Artists → [A] [B] [C] ... [Z] [#]
                      ↓
               Artists starting with B (≤ 200 items)

```

**Advantages over current approach**:

- Eliminates the Binder transaction risk on the root list — no flat "all items" page is ever sent
- Eliminates the double navigation path (flat list + "Browse by Letter" both on the same page)
- Cleaner, more consistent UX — one clear entry point into letter browsing
- Aligns with how many OEM car music apps work (letter-first browsing)

**"More..." nodes within letter pages:**
The 200-item cap still applies inside each letter page. For Artists, >200 items under a single letter is unlikely in a library. For Albums it's possible (e.g. a large collection of recordings under "_T_"). The existing `_fetchLetterPage` pagination logic and "_More… (N remaining)_" node rendering is retained unchanged — it just fires far less often (only for unusually large single-letter result sets rather than for every root list open).

Let me know what you want to do

Komodo5197:
I skimmed over these, and there looks to be substantial duplication of changes/overlapping features between these four PRs.  If the changes can't be cleanly separated, I'd personally prefer if they were all in one PR so I can just review everything in one place without any duplication.  Also, the browse by letter and artist change from immediate playback to browsing seem like notable enough changes, with upsides and downsides for different library sizes and browsing habits, that I think keeping the original versions available behind a setting might be worthwhile.

pdebuitlear:
> I skimmed over these, and there looks to be substantial duplication of changes/overlapping features between these four PRs. If the changes can't be cleanly separated, I'd personally prefer if they were all in one PR so I can just review everything in one place without any duplication. Also, the browse by letter and artist change from immediate playback to browsing seem like notable enough changes, with upsides and downsides for different library sizes and browsing habits, that I think keeping the original versions available behind a setting might be worthwhile.

New consolidated PR [here](https://github.com/UnicornsOnLSD/finamp/pull/1601)

Komodo5197:
Consolidated into #1601.

PR #1585, opened 2026-04-06
feat(android-auto): letter-first navigation for Albums and Artists

pdebuitlear:
## Summary

Implements the alternative approach described in `docs/alternative.md` for #1576.

- When the user taps **Albums** or **Artists** in Android Auto, they go directly to the A–Z + `#` letter index — no flat paginated list is shown first
- The "Browse by Letter" node is removed (it was the only entry point into letter browsing; now it *is* the entry point)
- Letter nodes render as a **list** (`CONTENT_STYLE_BROWSABLE_HINT: 1` on the Albums/Artists root items)
- Items within a letter page continue to render as a **grid** (`CONTENT_STYLE_BROWSABLE_HINT: 2` on each letter node)
- `_fetchRootPage` and "More..." pagination are unchanged for Playlists, Genres, and Tracks

## Test plan

- [ ] Tap **Albums** → A–Z + `#` letter grid appears immediately (no flat list)
- [ ] Tap **Artists** → same behaviour
- [ ] Tap a letter → filtered results render as a grid
- [ ] Tap **Playlists** / **Genres** / **Tracks** → flat paginated list still works
- [ ] No `FAILED BINDER TRANSACTION` in logcat for any browse path
- [ ] Letter nodes render as a list (not a grid with warning icons)


Komodo5197:
Consolidated into https://github.com/finamp-app/finamp/pull/1601.

PR #1589, opened 2026-04-09
Add a transcoding switch to the playback page.

solitudechn:
## Changes

3f109ad7f0d2637a1ec73d122978779e
A transcoding button has been added to the playback page, and relevant Chinese translations have been added to the translation function.

## Todo before merging

<!-- Did you add UI text? 
Yes. A transcoding button has been added to the playback page, and relevant Chinese translations have been added to the translation function.
- [ ] Translations
A transcoding button has been added to the playback page, and relevant Chinese translations have been added to the translation function.

I apologize, I'm not a professional software developer. The reason I submitted this PR is that I frequently need to use Finamp at home or outside, but for security reasons, I haven't directly exposed the Jellyfin server to the internet. When I'm at home, I expect to use un-transcoded high-definition music from the server, but my phone doesn't support that much storage. Therefore, I submitted this PR. If this PR doesn't comply with relevant specifications, I'm happy to remove it, as it's simply to solve a transcoding toggle issue.

### DIFF
--- a/lib/components/PlayerScreen/transcode_toggle_button.dart
+++ b/lib/components/PlayerScreen/transcode_toggle_button.dart
@@ -1,0 +1,35 @@
+import 'package:finamp/l10n/app_localizations.dart';
+import 'package:finamp/services/finamp_settings_helper.dart';
+import 'package:finamp/services/feedback_helper.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_tabler_icons/flutter_tabler_icons.dart';
+
+import '../../components/Buttons/simple_button.dart';
+
+/// A toggle button shown on the player screen to quickly enable/disable transcoding.
+///
+/// Only meaningful for streamed (non-downloaded) tracks. The button visually
+/// indicates the current state and toggles [FinampSettings.shouldTranscode],
+/// which triggers a queue reload via [DataSourceService].
+class TranscodeToggleButton extends ConsumerWidget {
+  const TranscodeToggleButton({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final shouldTranscode = ref.watch(finampSettingsProvider.shouldTranscode);
+    final l10n = AppLocalizations.of(context)!;
+
+    return SimpleButton(
+      text: l10n.transcodeToggleButtonTitle,
+      icon: shouldTranscode ? TablerIcons.transform : TablerIcons.transform,
+      iconColor: shouldTranscode
+          ? Theme.of(context).colorScheme.primary
+          : null,
+      onPressed: () {
+        FeedbackHelper.feedback(FeedbackType.light);
+        FinampSetters.setShouldTranscode(!shouldTranscode);
+      },
+    );
+  }
+}
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -2622,6 +2622,18 @@
   "@outputMenuButtonTitle": {
     "description": "Button title for opening the output menu/panel"
   },
+  "transcodeToggleButtonTitle": "Transcode",
+  "@transcodeToggleButtonTitle": {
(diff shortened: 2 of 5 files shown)


solitudechn:
I apologize, I'm not a professional software developer. The reason I submitted this PR is that I frequently need to use Finamp at home or outside, but for security reasons, I haven't directly exposed the Jellyfin server to the internet. When I'm at home, I'd like to use the un-transcoded high-definition music on the server, but my phone doesn't support that much storage. Therefore, I submitted this PR. If this PR doesn't comply with relevant guidelines, I'm happy to delete it, as it's simply to solve a transcoding toggle issue.

I'm not very good at using GitHub, so please bear with me if there are any problems.

Chaphasilor:
Apologies, I did not mean to close this, it was a misclick!

@solitudechn hey, you're doing great! Don't worry.  
Regarding your PR, while this is definitely a useful feature, we should probably discuss how to best integrate it.  
Up until now our plan was to eventually implement "transcoding profiles", where you can configure a specific set of transcoding settings for specific connections. E.g. direct play when connected via your local network, and transcode to OPUS when connected remotely.  
At the same time, I'm definitely not opposed to having such a dedicated option for controlling the quality on the player screen. This could be similar to how Qobuz does it (see video below), as in showing the quality as a clickable chip (as opposed to the current text-only display in Finamp ("Direct Playing", "Transcoding OPUS@96", etc.), which can be clicked to open a menu.  

https://github.com/user-attachments/assets/59ba0039-6fc1-4494-bf76-f8e90f3267d0

We could also think about integrating the transcoding menu with the player screen menu, similar to the playback speed menu:
image

The main issue there is that we seem to be running out of menu space 😅

While your approach with the button at the bottom of the player screen might work in Chinese, it won't work for languages with longer translations (like German, or even English).

solitudechn:
Thank you for your response, and please accept my apologies for the delay in getting back to you. I have carefully reviewed your feedback and given it a lot of thought.

First of all, you are absolutely right about the text length issue. Being accustomed to the concise nature of Chinese characters, I overlooked how much more space other languages might require. That was a lapse in my design consideration. I’ve been brainstorming ways to make the UI both aesthetically pleasing and functional across different languages, but I must admit my "Chinese-centric" design habits have made it difficult to find a perfect solution. My current ideas—such as "hiding text to show icons only" or "moving text below the icons"—feel like temporary fixes rather than a fundamental solution, as you pointed out.

Regarding the suggestion to hide the toggle in a sub-menu, I have some reservations. In my opinion, if it’s placed in a sub-menu, it might be more intuitive to move it to the Sidebar instead. This would maintain consistency with existing items like "Downloads" or "Offline Mode" in both visual style and functional logic. I also considered placing it in the top-right corner (next to Sort/Search), but without a standardized "transcoding" icon, users might easily mistake it for a "Language Switch" button.

I would like to clarify why I am advocating for a manual toggle rather than the "rule-based automatic switching" you mentioned.

I have used similar apps like StreamMusic or Symfonium, which offer network-based auto-transcoding. However, my setup presents a unique challenge: for security reasons, I do not expose my Jellyfin instance directly to the internet. I previously tried Cloudflare Zero Trust, but due to my local network conditions, the latency was unbearable.

Currently, I access my home lab via a persistent VPN connection. Since my phone is always connected to the VPN to access other local services, any app relying on "network detection" will incorrectly assume I am on a local high-speed network and thus fail to trigger transcoding automatically. This is the primary reason why a manual shortcut is essential for my workflow.

Thank you for your patience and understanding. I'm open to further suggestions on how we can implement this effectively.

Chaphasilor:
@solitudechn we're currently working on some changes regarding transcoding in #1612. If I understand your constraints correctly, you'd need a way to disable any automatic network-based rules, and then control a "default" transcoding profile via the button on the player screen. I think this could work with the changes planned in the PR I linked.  
Additionally, if we can make the "cellular" connection mode have higher priority than the "local network" one, I think we could even get your setup to work with automatic transcoding switching. What do you think about that?

In either case, I'd probably close this PR in favor of the other one (only if we end up implementing a solution of course). Would you be okay with that?

PR #1595, opened 2026-04-10
Fix validation when configuring address of a server outside the local network

Beedit:
## Changes

Minor bug fix for checking if a URL is a jellyfin server. Previously, if you were not on the same network, the function would always false as if the server was not accessible even if it was. This just fixes that.

## Related Issues
Dicussed in the discord.


Chaphasilor:
@Beedit thanks for the PR! Seems all good :)  
Did you set up a dev environment to test this?

PR #1596, opened 2026-04-11
Add AlbumNormalizationGain field to BaseItemDto

HeroBrine1st:
## Changes

Following https://github.com/jellyfin/jellyfin/pull/14745, this PR uses new dto field to (hopefully) avoid using metadata provider somewhere in the future. But currently it at least allows to avoid waiting on response :-)

There's no Jellyfin release with that PR currently. Do not merge until version is updated in comment I added.

I used https://github.com/UnicornsOnLSD/finamp/commit/658adaeb74f85191e9f50e6b35628f23693a4773 as an example on how to add support for download repair.

## Other changes

- Renamed "parentNormalizationGain" to "albumNormalizationGain" in MetadataProvider. I was thinking Jellyfin uses normalization gain for everything (e.g. movies) and named it so because.. well, it's fixed now
- Removed outdated comment about secondary metadataProvider arguments while there's only one

No tests aside from basic building were done - waiting for Jellyfin release. Also there can be problems with `check-codegen` task (it does a lot of removals and reformatting on my system), fortunately it prints a diff so I can apply it.


lib/services/music_player_background_task.dart
```diff
@@ -1538,18 +1538,14 @@ double? getGainForCurrentPlayback(MediaItem currentTrack, jellyfin_models.BaseIt
...
-      // Otherwise, use code commented above
-      final parentNormalizationGain = providerContainer
-          .read(metadataProvider(baseItem))
-          .valueOrNull
-          ?.parentNormalizationGain;
+      // metadataProvider is still used for Jellyfin versions 10.11.8 and below
+      // TODO update version before merge
+      final albumNormalizationGain =
+          baseItem.albumNormalizationGain ??
+          providerContainer.read(metadataProvider(baseItem)).valueOrNull?.albumNormalizationGain;
```
HeroBrine1st, line 1553: Idk why formatter suddenly changed its mind :-(
Aside from outdated comment, generated code and rename, there could be exactly one line replacement and everything else is added. I really like non-invasive changes :-)

HeroBrine1st:
Oh wow there's no problems with `check-codegen` aside of messed up hive field number (I used duplicated one initially and IDK how it generated code on my end without errors)

lib/services/music_player_background_task.dart
Chaphasilor: it's probably because of the `??` operator and Dart trying to avoid excessive indentation

Chaphasilor:
@HeroBrine1st your upstream changes have landed in the Jellyfin 12.0 RC builds, so I think it's time to merge this in? Or does this break backwards compatibility in any way?

HeroBrine1st:
This does not break compatibility at all but I'm still considering upgrading to RC to test it

HeroBrine1st:
It required me to clean up app folder (probably just a conflict with #1630). Otherwise, this works on 10.11.x so we can merge it. I won't have enough time to update to RC due to plugin upgrades and I don't want to clone my instance.

Chaphasilor:
I can test this on RC-2. Any specific steps I should follow?

HeroBrine1st:
So, this is the part where it gets tricky.
Normally, you'd just set it to dynamic/album-based and test right away, looking into logs for absence of request, but it is also done by UI.

You need to make metadata fetcher slow using patch from [this message](https://github.com/UnicornsOnLSD/finamp/pull/1331#issuecomment-3193709947). It passes if it doesn't break after that :-)

Or you can simulate reproduction of that bug (in PR of patch) by keeping player in background (UI won't make request) and then looking into logs to see whether it did metadata requests or not.

Chaphasilor:
Okay, so:  
With dynamic mode enabled and the 15 second delay in the metadata provider, I started a playlist that didn't contain tracks from the album I was testing with.  
Then I added the first track from the album to Next Up, and skipped to it directly. It played at track normalization. Then I added the second track to next up (without changing tracks), and the volume of the current track (first in album) was adjusted to match the album normalization. That was before the metadata for the track had loaded.  

I guess this means your changes are working? I did test this on 10.10 too though, with the same behavior, so I'm not entirely sure.  

12.0 returns `AlbumNormalizationGain` by default though, whereas 10.10 doesn't return `ParentNormalizationGain` by default.

HeroBrine1st:
It's strange it works like that on 10.10 given you tested the same revision on both jellyfin versions. It should have waited 15 seconds before changing to album gain, but note that it will change to `null` gain because there's no metadata for **current track**, not the next one. It will then change again, to album gain this time.

Chaphasilor:
@HeroBrine1st I think what's happening is that [music_player_background_task.dart](https://github.com/komodo5197/finamp/blob/051b2c012b4718c323188d4f3e7ed5554d24352e/lib/services/music_player_background_task.dart#L1534-L1573) kicks in, and that somehow get the `parentNormalizationGain`?  
Can you confirm that you actually get the behavior you describe on 10.10/10.11? Can you maybe share a video of how you're testing it?

HeroBrine1st:
> It's strange it works like that on 10.10 given you tested the same revision on both jellyfin versions. It should have waited 15 seconds before changing to album gain, but note that it will change to `null` gain because there's no metadata for **current track**, not the next one. It will then change again, to album gain this time.

This message is completely wrong.

It should have waited 15 seconds *from adding the first track to queue*, since that's what defines the album gain.

Second, it literally falls back to track gain:

https://github.com/finamp-app/finamp/blob/47629a823c80d809ccffd1992ff80cb1efbb3e63/lib/services/music_player_background_task.dart#L1553-L1556

So no issue here, it can't fall back to null so it didn't

I was postponing this for two weeks lol

(btw I simply rebased it to redesign, I believe `git range-diff cc032e4~4..cc032e4 47629a8~4..47629a8` can confirm that)

HeroBrine1st:
So the correct sequence of events is:

- Start playing album with multiple tracks - hear track normalisation
- Wait 15 seconds - hear album normalisation

(that's the same as adding two tracks from one album to queue and skipping to first one of them, it's just simpler to start from album)

OR

- Start playing one track from album, wait 15 seconds - hear track normalisation still
- Add another track from album - hear album normalisatrion
- Optionally: skip to that track in under 15 seconds - hear track normalisation
- Wait 15 seconds in total from step 2 - hear album normalisation again

Chaphasilor:
@HeroBrine1st sorry, got sidetracked. So does this need another round of testing? Or is the behavior I described in https://github.com/finamp-app/finamp/pull/1596#issuecomment-4881762946 enough to confirm this?

HeroBrine1st:
No code changes so no testing required :-)

Chaphasilor:
Awesome. Thanks for the PR!

PR #1598, opened 2026-04-13
fixes #1020 Pressing "Next track" via headphones will skip to next track but pause

### LINKED ISSUE #1020: [Redesign] Pressing "Next track" via headphones will skip to next track but pause
Hoi, bit of an odd one.

Like most modern wireless headphones, they can send "next track" or "previous track" commands. 
When I do this Finamp 0.9.13 skips to the next songs as expected but never plays it. 
I can via heaphones send the "play/pause" command and it starts playing just fine.

Pressing play in the UI works fine.

Let me know if you need anything.

Edit: I just notice when the track is paused and I send "next" via headphones, finamp skips to the next tracks and starts playing.

### LINKED ISSUE #1220: Pixel Buds Issue
### Type: 

- [x] Bug
- [ ] Major Bug (makes the app unusable)
- [ ] Other: <describe here>

### What version of Finamp are you using?
- [x] Beta/Redesign: <replace this with the version, something like 0.9.x>
- [ ] Stable: <replace this with the version, something like 0.6.x>

### Affected Device/Platform

- [x] Android
- [-] Android Auto
- [-] iOS
- [-] Linux
- [-] Windows
- [-] Mac

When Pixel Buds are used, if you use the skip function. Finamp skips, but also toggles play/pause. 
Double pressing will skip song, but also pause, doing it again will skip again, but play.

### Further information

Other Android apps have had this issue before. I'm completely oblivious to Android Dev, but [here's an issue](https://github.com/Automattic/pocket-casts-android/issues/1624) from another app with the issue if you want to see how others fixed it.

root-intruder:
## Problem

On Bluetooth headsets (reproduced with Pixel Buds Pro 2 on a Pixel 9), single tap correctly pauses playback, but double tap (`next track`) had wrong behavior:

- the app moved to the next track
- playback then paused/stopped instead of continuing playback automatically

Expected behavior: The headset "next" action should skip and continue playback immediately.

This problem is also described in issue #1020 .

## What changed

Adjusted skip handling in `lib/services/music_player_background_task.dart`:

- Updated `skipToNext()` and `skipToPrevious()` to better handle cases where a pause fade-out is in progress.
- If a fade-out is active when skip is triggered, playback is resumed appropriately so skip does not end in a paused state.
- Added a safety check to resume playback after skip when playback should still be active.

This targets media-button event sequences from some Bluetooth headsets where pause/skip can arrive in quick succession.

## Verification

Verified manually on my own device setup:

- Phone: Pixel 9
- Headset: Pixel Buds Pro 2

Result: double tap now skips to the next track and playback continues as expected.

## Notes

I am not a Dart developer.  
This fix was done with vibe coding support, then tested on-device to confirm the behavior works in practice.

### DIFF
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -141,6 +141,13 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
   final _volumeNormalizationLogger = Logger("VolumeNormalization");
   final _outputLogger = Logger("Output");
 
+  /// Time window used to ignore spurious play/pause callbacks immediately
+  /// after a skip command from certain Bluetooth headsets. Behavior before this fix was that a double-tap on such headsets would trigger a skip followed by an unintended pause, because the headset sent play/pause events immediately after the skip event. With this guard, if a play/pause event is received within this window after a skip command, it will be ignored.
+  static const Duration _skipPlayPauseGuardWindow = Duration(milliseconds: 50);
+
+  /// Timestamp of the most recent explicit skip command (next/previous).
+  DateTime? _lastSkipCommandAt;
+
   // Init the new sleep timer with a length of 0
   // SleepTimer sleepTimer = SleepTimer(SleepTimerType.duration, 0);
 
@@ -163,6 +170,24 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
   late final BehaviorSubject<FadeState> fadeState;
 
   final outputSwitcherChannel = MethodChannel('com.unicornsonlsd.finamp/output_switcher');
+
+  /// Some Bluetooth headsets send skip and pause/play media button events in
+  /// very quick succession for a double-tap skip gesture. This guard ignores a
+  /// trailing play/pause event if it arrives right after skip, preventing an
+  /// unintended pause while still allowing normal controls outside the window.
+  bool get _shouldIgnorePlayPauseAfterRecentSkip {
+    final lastSkipCommandAt = _lastSkipCommandAt;
+    if (lastSkipCommandAt == null) return false;
+
+    final elapsed = DateTime.now().difference(lastSkipCommandAt);
+    if (elapsed <= _skipPlayPauseGuardWindow) {
+      _audioServiceBackgroundTaskLogger.fine(
+        "Ignoring play/pause because skip was ${elapsed.inMilliseconds}ms ago (threshold ${_skipPlayPauseGuardWindow.inMilliseconds}ms)",
+      );
+      return true;
+    }
+    return false;
+  }
 
   Future<void> showOutputSwitcherDialog() async {
     if (!Platform.isAndroid) {
@@ -583,6 +608,12 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
 
   @override
   Future<void> play({bool disableFade = false}) async {
+    _audioServiceBackgroundTaskLogger.fine(
+      "play() start: disableFade=$disableFade, playing=${_player.playing}, fadeDirection=${fadeState.value.fadeDirection}, currentIndex=${_player.currentIndex}, position=${_player.position}",
+    );
+    if (_shouldIgnorePlayPauseAfterRecentSkip) {
+      return;
+    }
     if (!disableFade && FinampSettingsHelper.finampSettings.audioFadeInDuration > Duration.zero) {
       return fadeInAndPlay();
     } else {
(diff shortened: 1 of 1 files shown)


Komodo5197:
It looks like this fix relies on the fade period to determine the double click happened, but fading is actually disabled by default, so I believe this fix will not work for most users.  I think you need to add a variable tracking the time of last play/pause and use that to catch the double click.  Also, do you have any data on how much time could maximally elapse between the double click?  I want to minimize the chances we accidentally revert a legitimate pause and skip input from a non-faulty interface.

root-intruder:
@Komodo5197 My fading is disabled in the UI (set to 0) and pause as well as skip works well. For the double click time i would assume the time span between the clicks is below 1sec.

Komodo5197:
Interesting.  If you have a 0 fade time and the fix is working, that presumably means that _player.playing is true when the skip is called but false when the actual seek completes.  So I guess it would be relying on the gap between the pause and skip inputs being smaller than the latency for us to receive the pause confirmation from the native player.  That does not feel like something we should be relying on, so I think explicit time-since-pause tracking worth implementing.

I think it might help me out if you stick a log statement at the very start of the play(), pause(), skipToNext(), and skipToPrevious() methods.  You can then do a couple skips around and upload the logs.  That way we can get the exact timings the app is receiving and how consistent they are.

Also, the linked issue confirms that the issue also manifests as playback beginning when skipping while paused, but I believe this PR currently only handles the case where you start while playing.  The paused case should also be addressed.

root-intruder:
Good point @Komodo5197 .
I added entry logging at the start of:
- play()
- pause()
- skipToNext()
- skipToPrevious()

The logs include playing state, fade direction, index/position, and relevant flags (disableFade/forceSkip, hasNext/hasPrevious, loopMode).  
In the logs there is first skip while playing with my fix and then also skip while paused which skips and automatically starts playing the next track, a behavior also common on other players. 
[log.txt](https://github.com/user-attachments/files/26719308/log.txt)

Komodo5197:
So the skip fires before the pause?  That does not match what I expected at all, but if it is consistent it does make our jobs easier, because it allows us to detect the false input immediately instead of retroactively when the true input fires.  You can just create a member variable which you record the current time to when calling skip in either directions, and then add code to play/pause to avoid taking action if that variable has a time less than 50 milliseconds ago.  This avoids any need to consider the fade state and completely prevents sending incorrect commands to the native player.

By the way, generally resuming playback when skipping may be the way some players operate, but it does not match how Finamp behaves on other devices or via non-bud interfaces, so that's not a good argument to leave the behavior in place.  If we want that behavior we can implement it properly, but having it only in certain places due to a bug is not good.

root-intruder:
@Komodo5197 OK changed it like you recommented. working also like a charm with pixel buds. Additionally, now if playback is paused and i skip to next playback will remain paused like you whished for.

lib/services/music_player_background_task.dart
```diff
@@ -141,6 +141,9 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
   final _volumeNormalizationLogger = Logger("VolumeNormalization");
   final _outputLogger = Logger("Output");
 
+  static const Duration _skipPlayPauseGuardWindow = Duration(milliseconds: 50);
+  DateTime? _lastSkipCommandAt;
```
Komodo5197, line 149: A doc ( /// ) comment explaining what this variable is here for would be nice.

lib/services/music_player_background_task.dart
```diff
@@ -164,6 +167,20 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
 
   final outputSwitcherChannel = MethodChannel('com.unicornsonlsd.finamp/output_switcher');
 
+  bool _shouldIgnorePlayPauseAfterRecentSkip() {
```
Komodo5197: I think the getter syntax would be a tad cleaner for this.

Komodo5197: Add a doc comment to this function explaining what the bud skip/pause issue is and how this fixes it.

Komodo5197 (commented):
The code looks good to me, so if the fix is working for you I think the functional changes are complete.  If you add the requested comments, this should be good to merge.

lib/services/music_player_background_task.dart
root-intruder: done

root-intruder: as a java developer, this looks less expressive, but since you are obviously the better dart dev I trust you ;-)

lib/services/music_player_background_task.dart
root-intruder: done

lib/services/music_player_background_task.dart
Komodo5197: I tend to prefer the getter syntax for basically any no-argument function without side effects, myself.

Komodo5197:
Okay, I've merged the changes.  Hopefully the other hardware mentioned in the issue thread presents similarly enough for this fix to apply.

root-intruder:
awesome, thanks!

Chaphasilor:
@Komodo5197 thanks for handling this one. From what I've seen the Pixel Buds are notorious for sending different commands than other bluetooth devices, and Finamp isn't the only app affected. If this PR really fixes it that's great!  

And thanks for the PR of course, @root-intruder !

PR #1600, opened 2026-04-15
Preserve server path for discovery info

StayBlue:
The manual `/System/Info/Public` probe discarded any path prefix from a discovered server URL, causing prefixed Jellyfin instances to be queried at the wrong endpoint.


Komodo5197:
Looks good to me.

PR #1601, opened 2026-04-15
feat(android-auto): consolidated browsing with user-selectable mode

pdebuitlear:
## Summary

Consolidates Android Auto browsing features with a user-configurable option to choose between:
- **Flat list** (default) — traditional paginated view with "Browse by Letter" node
- **Letter-first** — skip flat list, go straight to A–Z index

Addresses feedback from #1576 and #1579 to keep both approaches available behind a setting.

## Features

- **Recently Played Albums** — shows up to 20 recently played albums in root menu
- **Artist Albums Browsing** — tap an artist to browse their albums
- **Letter-first Navigation** — optional A–Z index for Albums & Artists

## Settings

New "Android Auto" section in Settings → Android Auto with dropdown to select browsing mode.

**Default:** Flat list (preserves existing behavior)

## Display Rendering

Items in letter-filtered pages render as a plain **text list** (not grid with icons), addressing the maintainer's feedback about long names being cut off.


lib/models/finamp_models.dart
```diff
@@ -2644,12 +2657,21 @@ enum MediaItemParentType {
   rootCollection,
   @HiveField(2)
   instantMix,
+  @HiveField(3)
+  recentlyPlayed,
 }
 
 @JsonSerializable(converters: [BaseItemIdConverter()])
```
Komodo5197, line 2656: It might be worth adding includeIfNull: false here with all these usually empty fields.

lib/models/finamp_models.dart
```diff
@@ -251,6 +251,15 @@ class DefaultSettings {
   static const forceAudioOffloadingOnAndroid = false;
   static const previousTracksPersistenceMode = PreviousTracksPersistenceMode.persistent;
   static const useAndroidGainEffect = true;
+  static const androidAutoBrowsingMode = AndroidAutoBrowsingMode.flat;
+}
+
+@HiveType(typeId: 79)
```
Komodo5197: This class should be at the end of the file with a higher typeId than the current highest.

lib/screens/android_auto_settings_screen.dart
```diff
@@ -0,0 +1,77 @@
...
+  Widget build(BuildContext context, WidgetRef ref) {
+    final currentMode = ref.watch(finampSettingsProvider.androidAutoBrowsingMode);
+
+    return Padding(
+      padding: const EdgeInsets.all(16.0),
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Text(
+            "Albums & Artists Browsing",
```
Komodo5197: All strings displayed to the user need to run through the localizations system.  This applies to all the other displayed strings in this PR.

lib/services/music_player_background_task.dart
```diff
@@ -926,18 +926,42 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
 
   /// Returns the top-level browsable categories for use in a media browser.
   List<MediaItem> _getRootMenu() {
+    // Choose browsing mode hints based on user settings.
+    // - flat: Grid for browsable items (traditional flat list with "Browse by Letter" node)
+    // - letterFirst: List for browsable items (letter nodes render as a list)
+    final isLetterFirst = FinampSettingsHelper.finampSettings.androidAutoBrowsingMode ==
+        AndroidAutoBrowsingMode.letterFirst;
+    final albumsBrowsableHint = isLetterFirst ? 1 : 2; // 1=list, 2=grid
```
Komodo5197: There's currently code in main which sets the browsing hint based on the app-wide list vs grid display setting, although I'm not 100% sure it works properly.  We should reproduce that here.

lib/services/android_auto_helper.dart
```diff
@@ -49,9 +53,218 @@ class AndroidAutoHelper {
...
+  /// Letters used for the "Browse by Letter" nodes. Empty string = '#' bucket.
+  static const List<String> _alphabet = [
+    'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M',
+    'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', '#',
+  ];
+
+  /// Returns the letter nodes A–Z plus a '#' bucket for the Browse-by-Letter view.
+  List<MediaItem> _getLetterNodes(MediaItemId itemId) {
+    return _alphabet.map((letter) {
+      final nameFilter = letter == '#' ? '#' : letter;
```
Komodo5197: This ternary is pointless.

lib/services/android_auto_helper.dart
```diff
@@ -49,9 +53,218 @@ class AndroidAutoHelper {
...
+          contentType: itemId.contentType,
+          parentType: MediaItemParentType.rootCollection,
+          nameFilter: nameFilter,
+          pageStartIndex: 0,
+        ).toString(),
+        title: letter,
+        playable: false,
+        extras: const {
+          // Items filtered by letter should render as a list (no grid).
+          "android.media.browse.CONTENT_STYLE_BROWSABLE_HINT": 1,
```
Komodo5197: We should define constants for these hint values instead of just using ints.

lib/services/android_auto_helper.dart
```diff
@@ -49,9 +53,218 @@ class AndroidAutoHelper {
...
+
+    try {
+      // Step 1: fetch recently played tracks — tracks reliably have DatePlayed set.
+      final tracksResult = await _jellyfinApiHelper.getItemsWithTotalRecordCount(
+        parentItem: parentItem,
+        includeItemTypes: BaseItemDtoType.track.jellyfinName,
+        sortBy: 'DatePlayed',
+        sortOrder: 'Descending',
+        filters: 'IsPlayed',
+        limit: 100,
```
Komodo5197: It might make sense so set this larger, to 300 or so.  I tend to play whole albums, so I'm only getting 9 results with this limit.

lib/services/android_auto_helper.dart
```diff
@@ -525,21 +747,124 @@ class AndroidAutoHelper {
...
+                AppLocalizations.of(GlobalSnackbar.materialAppScaffoldKey.currentContext!)?.shuffleAll ??
+                "Shuffle All Tracks",
+            playable: true,
+          ),
+        );
+      }
+
+      // For Albums and Artists, check the browsing mode setting.
+      // If letterFirst, return immediately; otherwise fall through to flat list.
+      final supportsLetterBrowse = itemId.contentType == TabContentType.albums ||
```
Komodo5197, line 832: I don't see why we shouldn't also allow per-letter browsing in the tracks tab.

lib/services/android_auto_helper.dart
```diff
@@ -525,21 +747,124 @@ class AndroidAutoHelper {
...
+      }
+
+      // For flat list mode: "Browse by Letter" node only on first page, if supported.
+      final pageStart = itemId.pageStartIndex ?? 0;
+      if (pageStart == 0 && supportsLetterBrowse && !isLetterFirst) {
+        mediaItems.add(MediaItem(
+          id: MediaItemId(
+            contentType: itemId.contentType,
+            parentType: MediaItemParentType.rootCollection,
+            nameFilter: '',
```
Komodo5197: This should be a meaningful string like letter_root or something, not ''.

lib/services/android_auto_helper.dart
```diff
@@ -525,21 +747,124 @@ class AndroidAutoHelper {
...
+        final mediaItem = await queueService.generateMediaItem(
+          item,
+          parentType: MediaItemParentType.collection,
+          parentId: item.parentId,
+          isPlayable: _isPlayable,
+        );
+        mediaItems.add(mediaItem);
+      }
+
+      if (pageStart + items.length < totalCount) {
```
Komodo5197, line 871: There's a bug in playlists where jellyfin gives us a slightly higher record count than the actual available items, and that causes this logic to erroneously report negative tracks remaining on the last page.

lib/services/android_auto_helper.dart
```diff
@@ -151,16 +364,25 @@ class AndroidAutoHelper {
...
-      parentItem: parentItem,
-      sortBy: includeItemTypes == BaseItemDtoType.track
-          ? "ParentIndexNumber,IndexNumber,${sortBy.jellyfinName(itemId.contentType)}"
-          : sortBy.jellyfinName(itemId.contentType),
-      sortOrder: includeItemTypes == BaseItemDtoType.track ? null : sortOrder.toString(),
-      includeItemTypes: includeItemTypes.jellyfinName,
-      limit: onlineModeLimit,
-    );
-    return items ?? [];
+    final List<BaseItemDto> allItems = [];
```
Komodo5197, line 376: I don't see a reason to paginate these server requests.  We're just outputting the whole list, and the main app doesn't currently paginate the item requests this covers.

lib/services/android_auto_helper.dart
```diff
@@ -49,9 +53,218 @@ class AndroidAutoHelper {
...
+    // Jellyfin's NameStartsWith doesn't support '#'; for that bucket we fetch
+    // the full first page without a name filter and discard alphabetic starters.
+    // For '#' we fall back to client-side filtering of the first _pageSize items.
+    if (nameFilter == '#') {
+      final result = await _jellyfinApiHelper.getItemsWithTotalRecordCount(
+        parentItem: parentItem,
+        sortBy: sortBy.jellyfinName(itemId.contentType),
+        sortOrder: sortOrder.toString(),
+        includeItemTypes: itemId.contentType.itemType.jellyfinName,
+        startIndex: 0,
```
Komodo5197: This # sort seems to be generally messy, but I think it makes sense to still use pageStart as the starting index to at least sort of get pagination.

lib/services/android_auto_helper.dart
```diff
@@ -49,9 +53,218 @@ class AndroidAutoHelper {
...
+      return mediaItems;
+    } catch (err, trace) {
+      _androidAutoHelperLogger.severe("Error loading recently played items", err, trace);
+      return [];
+    }
+  }
+
+  /// Fetches a single page of [_pageSize] items for a root collection browse,
+  /// returning the items and the server's total record count for pagination.
+  Future<(List<BaseItemDto>, int)> _fetchRootPage(MediaItemId itemId) async {
```
Komodo5197, line 238: I'm not sure splitting up the normal and per-letter pages really makes much sense.  But all this logic is probably getting refactored and simplified after the home screen PR lands, so whatever.

lib/models/finamp_models.dart
pdebuitlear: Done — added `includeIfNull: false` to the `@JsonSerializable` annotation on `MediaItemId`.

lib/screens/android_auto_settings_screen.dart
pdebuitlear: Done — all user-facing strings now go through `AppLocalizations`. Added keys to `app_en.arb` for the screen title, dropdown label/subtitle, option labels, 'Browse by Letter', 'Not available offline', and the 'More...' pagination node.

lib/services/android_auto_helper.dart
pdebuitlear: Done — defined `_contentStyleList = 1`, `_contentStyleGrid = 2`, and `_contentStyleCategory = 4` as static constants on `AndroidAutoHelper`.

lib/models/finamp_models.dart
pdebuitlear: Done — moved `AndroidAutoBrowsingMode` to the end of the file. Kept typeId 79 to avoid breaking existing Hive boxes on devices that already have the setting stored.

lib/services/android_auto_helper.dart
pdebuitlear: Done — replaced the empty string sentinel with `'letter_root'` (stored as `_letterRootNameFilter` constant).

lib/services/music_player_background_task.dart
pdebuitlear: Done — `_getRootMenu` now reads `contentViewType` from settings and uses grid hint for albums when the app is in grid view mode (flat browsing mode), falling back to list otherwise.

lib/services/android_auto_helper.dart
pdebuitlear: Done — removed the pointless ternary, `letter` is used directly as `nameFilter`.

lib/services/android_auto_helper.dart
pdebuitlear: Done — the 'More...' node is now only added when `remaining > 0`, guarding against Jellyfin's inflated record count for playlists.

lib/services/android_auto_helper.dart
pdebuitlear: Done — bumped the limit to 300.

lib/services/android_auto_helper.dart
pdebuitlear: Done — `supportsLetterBrowse` now includes `TabContentType.tracks` as well.

lib/services/android_auto_helper.dart
pdebuitlear: Agreed — left the server requests in `getBaseItems` unpaginated as before. The pagination only applies to the root collection browse pages.

lib/services/android_auto_helper.dart
pdebuitlear: Done — the '#' bucket fetch now uses `pageStart` as `startIndex` in the server request.

lib/services/android_auto_helper.dart
pdebuitlear: Noted — left the split as-is for now given the expected refactor after the home screen PR lands.

lib/services/android_auto_helper.dart
Komodo5197: I don't see this change.

lib/services/android_auto_helper.dart
Komodo5197: Reuse the constants in music_player_background task too.

lib/models/finamp_models.dart
Komodo5197: Change the ID.

lib/models/finamp_models.dart
Komodo5197: I don't see this change.

PR #1603, opened 2026-04-21
[Redesign] Add integration tests

Komodo5197:
Adds some basic integration tests.  These start the app, login to the jellyfin demo server, and start playback of a song.  They are configured to only run in CI on push the the primary repository, not in forks or on pull requests.  This is to help keep the CI fast in the less important cases.  Successful run can be found at https://github.com/Komodo5197/finamp/actions/runs/24701928893.

Additional fixes:
- Issue with app startup on windows after dependency upgrade
- Android app failing to start on older android versions due to call to uiManager.setApplicationNightMode
- Minor bugs in login flow
- Improved GlobalSnackbar context handling and fixed "More" button on error snackbars


Komodo5197:
@Chaphasilor Hey, any thoughts on this?  I'd like to get this merged up and into the home screen PR.

integration_test/load_test.dart
```diff
@@ -0,0 +1,83 @@
...
+import 'package:integration_test/integration_test.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  // These integration tests all rely on the previous one working.  Not good practice, but whatever.
+  group('Integration Tests', () {
+    testWidgets('Verify app loads without errors', (tester) async {
+      // Login testing flag redirects file accesses to integration_test folder and clears it on startup.
+      // Downloads are still left in original folder, but they shouldn't really affect testing.
```
Chaphasilor: I guess we can change that if at any point we add download tests?

integration_test/load_test.dart
```diff
@@ -0,0 +1,83 @@
...
+      await tester.pumpAndSettle();
+
+      final startButton = find.byType(CTALarge);
+      await tester.tap(startButton);
+      await tester.pump();
+
+      final urlEntry = find.byType(TextFormField);
+      await tester.enterText(urlEntry, "https://demo.jellyfin.org/stable");
+
+      final serverButton = find.text("Stable Demo");
```
Chaphasilor: Can we find the button differently? If they ever change the name of the server, this test breaks. Not a big deal of course, so it's not a pressing issue.

Also, how does this deal with asynchronous activity? From reading the code it seems we "construct" an element via `find.text()` and then `waitFor` is extended to wait until an element/widget with that text is discovered? Loading server info can sometimes take one or two seconds (for me at least).  
To be honest I think "find" is a terrible name in that case (I'd prefer something like "match"), but that's not your fault ^^

integration_test/load_test.dart
```diff
@@ -0,0 +1,83 @@
...
+      final customUserButton = find.text("Custom User");
+      await tester.tap(customUserButton);
+      await tester.pump(Duration(seconds: 1));
+
+      final userTextField = find.byType(TextFormField).first;
+      await tester.enterText(userTextField, "demo");
+      final loginButton = find.text("Log In");
+      await tester.tap(loginButton);
+
+      final musicScreen = find.byType(MusicScreen);
```
Chaphasilor: this depends on the test server always having a single library. I'm fine with verifying the behavior of skipping the library selection screen in that case, but we should keep it in mind if we ever switch the test server.

integration_test/load_test.dart
```diff
@@ -0,0 +1,83 @@
...
+      await tester.waitFor(musicScreen);
+      await tester.pump(Duration(seconds: 1));
+
+      // Verify Music screen loaded
+      // TODO any other verifications beyond reaching this point?
```
Chaphasilor: I guess we could ensure that the music screen contains at least one item? Although as of right now the following test (in the second commit) already does that, in a way.

lib/main.dart
```diff
@@ -143,14 +151,18 @@ void main() async {
     await _setupDiscordRpc();
     _mainLog.info("Setup Discord RPC");
   } catch (error, trace) {
-    hasFailed = true;
-    Logger("ErrorApp").severe(error, null, trace);
-    runApp(FinampErrorApp(error: error, trace: trace));
+    if (!integrationTesting) {
+      Logger("ErrorApp").severe(error, null, trace);
+      runApp(FinampErrorApp(error: error, trace: trace));
```
Chaphasilor, line 158: Wouldn't it make sense to be able to test the `FinampErrorApp` too? Ensuring people know what went wrong and have the means to report the issue is fairly important I think. It's not essential of course, and if this rethrow allows us to better extract the error message, then I guess we can stick with it for now and add a separate launch flag at a later point.

lib/main.dart
```diff
@@ -166,16 +178,18 @@ void main() async {
...
+  await findSystemLocale();
+  await initializeDateFormatting();
+  unawaited(fetchSystemPalette());
+  await initDBus();
 
-    _mainLog.info("Launching main app");
+  _mainLog.info("Launching main app");
 
+  if (!integrationTesting) {
     runApp(const Finamp());
```
Chaphasilor, line 196: This was unintuitive, but it's because we pump the `Finamp` widget inside the integration test, right? Maybe add a short comment here?

pubspec.yaml
```diff
@@ -56,6 +56,9 @@ dependencies:
       url: https://github.com/Komodo5197/frb_plugins.git
       ref: 7ece49cc16a459aaa0718ed23458802533daef37
       path: packages/smtc_windows/
+  # Despite having loose version constraints, smtc_windows and flutter_discord_rpc actually depend on the
+  # specific vaersion 2.11.1, so we must force it.
```
Chaphasilor: minor typo

lib/main.dart
```diff
@@ -943,3 +957,33 @@ class FinampProviderObserver extends ProviderObserver {
     GlobalSnackbar.error(error);
   }
 }
+
+class TestingPathProvider extends PathProviderPlatform {
```
Chaphasilor, line 966: You included a comment about what this is for in the integration test file, but I guess it would be useful here as well for reference?

integration_test/test_cases.dart
```diff
@@ -0,0 +1,141 @@
...
+
+  // These integration tests all rely on the previous one working.  Not good practice, but whatever.
+  group('Integration Tests', () {
+    testWidgets('Verify app loads without errors', (tester) async {
+      // Login testing flag redirects file accesses to testing folder and clears it on startup.
+      // Downloaded files are still left in original folder, but they shouldn't really affect testing.
+      await tester.runAsync(() => app.main(integrationTesting: true, loginTesting: true));
+
+      // Save off initialized provider container so tests can create and attach descendants
+      // TODO could we get errors from background tasks also attaching to this when they need persistence?
```
Chaphasilor: what kind of persistence are we talking about here? In any case, this might be hard to debug if it happens

integration_test/test_cases.dart
```diff
@@ -0,0 +1,141 @@
...
+      await tester.pumpWidget(app.Finamp());
+
+      await tester.pumpAndSettle();
+
+      // Verify FinampApp loaded.
+      expect(find.byType(LoginScreen), findsOneWidget);
+
+      container = GetIt.instance<ProviderContainer>();
+      GetIt.instance.unregister<ProviderContainer>();
+      GetIt.instance.registerSingleton(ProviderContainer(parent: container));
```
Chaphasilor: Is this repeated on purpose?

integration_test/test_cases.dart
```diff
@@ -0,0 +1,141 @@
...
+      await tester.pump();
+
+      // Progress into song
+      await Future.delayed(Duration(seconds: 30));
+      await tester.pump();
+      final playerService = GetIt.instance<MusicPlayerBackgroundTask>();
+      final playbackPosition = playerService.playbackPosition;
+
+      // Verify track has been playing for 30 seconds as expected.
+      expect(playbackPosition.inSeconds, inInclusiveRange(15, 40));
```
Chaphasilor, line 163: Is there a benefit of waiting 30 seconds? Shouldn't 10-15 work just as well? I'd argue for decreasing the range anyway, there should really be no more than +1/-5 seconds of variation.

lib/main.dart
```diff
@@ -108,7 +108,7 @@ late DateTime startTime;
 
 final providerScopeKey = GlobalKey();
 
-Future<void> main({bool integrationTesting = false, bool loginTesting = false}) async {
+Future<void> main({bool integrationTesting = false, bool loginTesting = true}) async {
```
Chaphasilor: maybe this should be called `temporaryLogin` or similar? Defaulting this to true implies we want it to run even if we're *not* specifically testing login.

lib/components/global_snackbar.dart
```diff
@@ -15,16 +15,31 @@ import 'package:flutter_sticky_header/flutter_sticky_header.dart';
 import 'package:http/http.dart' hide Response;
 import 'package:logging/logging.dart';
 
+typedef _QueueFunction = void Function(BuildContext);
+
 @Deprecated("Use GlobalSnackbar.error(dynamic error) instead")
 void errorSnackbar(dynamic error, BuildContext context) => GlobalSnackbar.error(error);
 
 class GlobalSnackbar {
+  /// Prefer using state and localization getters.  Avoid directly accessing context.
```
Chaphasilor, line 24: should be deprecate the property to properly enforce this?

lib/models/finamp_models.dart
```diff
@@ -2098,33 +2096,35 @@ class QueueItemSourceName {
   @HiveField(2)
   final String? localizationParameter;
 
-  String getLocalized(BuildContext context) {
+  String getLocalized(BuildContext context) => getLocalized2(AppLocalizations.of(context)!);
+
+  String getLocalized2(AppLocalizations localizations) {
```
Chaphasilor, line 2100: can we find a better name for this and `getName2`? maybe `getLocalizedFromLocalizations`? I know it sounds silly, but at least it has meaning.

.github/workflows/build.yml
```diff
@@ -98,31 +102,60 @@ jobs:
...
+      timeout-minutes: 20
+      # android emulator is flaky so we retry once
+      uses: Wandalen/wretry.action@v3.8.0
+      with:
+        action: reactivecircus/android-emulator-runner@v2
+        # Use the newest supported version of android
+        with: |
+          api-level: 36
+          arch: x86_64
+          profile: pixel
```
Chaphasilor, line 123: do we know which device the pixel profile corresponds to exactly?

lib/services/downloads_service.dart
```diff
@@ -220,15 +221,20 @@ class DownloadsService {
...
-                Future.sync(() async {
-                  assert(
-                    listener.file?.path == await event.task.filePath() ||
-                        (extension != null &&
-                            listener.file?.path.replaceFirst(RegExp(r'\.image$'), extension) ==
-                                await event.task.filePath()),
-                    "${listener.name} ${listener.path} ${listener.fileDownloadLocation?.baseDirectory} ${listener.file?.path} ${await event.task.filePath()} $extension",
-                  );
-                });
+                if (kDebugMode) {
```
Chaphasilor, line 224: I'm guessing there's no reason to ever run integration tests with a release build, right? It probably won't even work?

lib/models/jellyfin_models.dart
```diff
@@ -1809,9 +1808,10 @@ class BaseItemDto with RunTimeTickDuration implements PlayableItem {
   @HiveField(79)
   String? seasonName;
 
-  /// Gets or sets the media streams.
-  @HiveField(80)
-  List<MediaStream>? mediaStreams;
+  // Gets or sets the media streams.
+  // This field is removed in favor of mediaSources.first.mediaStreams.
```
Chaphasilor, line 1812: isn't it removed in favor of the new `mediaStreams` getter?

lib/builders/annotations.dart
```diff
@@ -10,11 +11,20 @@ class SettingsHelperIgnore {
   String toString() => "Excluded from automatic sub-provider/setter generation: $message";
 }
 
-/// Annotation for FinampSettings fields that should not have a setter and/or a
-/// finampSettingsProvider selector created
 @Target({TargetKind.setter, TargetKind.field})
 class SettingsHelperMap {
+  /// Annotation for FinampSettings Map fields, which must use per-field getters and setters.
+  /// If [keyGetter] is true, a getter for the keys is also created.  This compares by length, so
+  /// keys should never be updated in place, only added or removed.
```
Chaphasilor, line 18: so updating a value of a map setting would either require manually modifying FinampSettings or first removing the old value and then adding the new value for a given key?

Chaphasilor (approved):
thanks for getting this started, it's definitely been overdue.

I left some comments, but found no major issues. I went through the changes commit by commit, but I think that messed up the positioning of the comments in GitHub a bit, so you might have to check them commit by commit too 😅

The CI run looks good, although I'm wondering if this will be shown as a separate test or included in the build?

Also, I wasn't able to get the tests running on my Android phone (as in, I got a build error about unclosed caches), although I suspect the issue is on my side here, so I don't consider this blocking.  
Feel free to address and merge this as you see fit, especially if you need it for the home screen PR. I saw you added a ton of commits there, I'll have to check that out tomorrow!

integration_test/load_test.dart
Komodo5197: Maybe?  The problem is that background_dopwnloader resolves the base directories itself, so testing that would take more modification of our real downloads code than I'd really like.

lib/models/finamp_models.dart
Komodo5197: I'd prefer to just roll with 2.  The argument difference should be self explanatory, and I believe I've already seen the framework do this.  Or maybe I'll just delete the context version.

lib/services/downloads_service.dart
Komodo5197: Yeah.  This is basically just making sure the wrapping code only fires if we actually need it for the assert, though.

lib/models/jellyfin_models.dart
Komodo5197: From an api perspective, I guess?  But from a data structure perspective, this is where the actual data is stored, and I feel that's more relevant here.

lib/builders/annotations.dart
Komodo5197: Values associated with a specific key are watched independently.  This basically means if you had the downloads list watching the download location keys, but then you removed a location and added one with a different key at the exact same time, we wouldn't notice the keys we need to be watching had changed.  But we'd still notice the old keys value had gone to null.  I don't foresee see any real scenarios where this sort of thing would happen.

.github/workflows/build.yml
Komodo5197: Not really, but it probably doesn't matter.

integration_test/test_cases.dart
Komodo5197: This just seemed reasonable.  I'll note that the delay also includes playback startup time, so there's more variability then you might think.

integration_test/load_test.dart
Komodo5197: Yeah, waitFor basically just polls for a matching widget every second until it finds a match or it times out.  I figured 20 seconds per action should cover basically all cases where we'll actually get a response.

integration_test/load_test.dart
Komodo5197: Yeah, I feel any sort of dynamic behavior would be way to complex, so we should just tailor to the target.

lib/main.dart
Komodo5197: Maybe?  But we need to be throwing errors outwards for the tests to show them clearly.  And the test harness messes with the global error handling stack anyway, so doing anything to it ourselves isn't good.

integration_test/test_cases.dart
Komodo5197: Yeah, it's pretty nasty.  Basically, the testing harness tries to clear out all the async code between tests, and runs all the cases in individual async contexts as part of this.  I believe the expectation is that background services and realtime tasks will all be replaced with mockups, for more consistent and self-contained tests.  But this code doesn't do that and has real persistent background services, so I've had errors occasionally showing up in earlier tests occasionally, and a bunch of strange issues with providers were occurring.  Giving each test its own child providercontainer seems to have solved those, but it's all still a bit mysterious.

Komodo5197:
So I'll note this integration testing is actually kind of terrible.  All the tests depend on the previous ones for background service setup, and we also call out to a potentially flaky external server which could could failures.  In fact, the jellyfin demo server appears to be down right now as I write this.  But it's probably still better than nothing.  The CI is shown included in the build because it's the same workflow.  If I wanted it separate, I think I would have had to duplicate all the builds.  I have tested on my phone, so I'm not sure your build is failing.  Maybe you need to do flutter clean due to all the native plugin upgrades?

integration_test/load_test.dart
Chaphasilor: Well I guess some tests would be better than none, but for now it's definitely fine

integration_test/load_test.dart
Chaphasilor: If we don't support dynamic behavior, then I guess we'd need different test servers for different tests, e.g. a server with multiple libraries if we want to test the library picker (not that I feel like that picker is actually important, just as an example)?

lib/main.dart
Chaphasilor: Alright, let's keep it this way then. I guess any changes to the error screen should simply be properly vetted.

integration_test/load_test.dart
Komodo5197: That would probably be easiest, although I don't anticipate these tests getting majorly expanded anytime soon.

integration_test/test_cases.dart
Chaphasilor: Okay. I feel like this would be a great comment to add :)  
I guess the clean way to do this is to fully re-start the app for each test, but that's neither efficient nor does it work well with Flutter's `test` command. So if it works right now, let's keep it!

integration_test/load_test.dart
Komodo5197: I feel like that level of testing would be nice, but feels a bit unrealistic for a simple music playert with no actual funding.

integration_test/test_cases.dart
Chaphasilor: My 5 seconds included any load times, I feel like the connection to the test server should be speedy enough? But, let's keep this delay in mind, and if we ever run into issues with the test run taking too much time, we can decrease this.

lib/models/finamp_models.dart
Chaphasilor: Alright

integration_test/test_cases.dart
Komodo5197: It's usually fast, but sometimes it isn't and these tests are already flaky enough as is.  I think the total test runtime can't get over three minutes, anyway, and that's less than the builds.

lib/models/jellyfin_models.dart
Chaphasilor: Oh wait, this is jellyfin_models. Doesn't this mean that chopper will fail to parse the API response if a BaseItemDto contains a top-level `mediaStreams` property? Because I'm pretty sure I've seen that in the past, and that's why this is here?

Komodo5197: No, it will just ignore the value.

Chaphasilor:
Yeah, I saw your comments about the dependencies. It's definitely better than nothing. As for connecting to an external server, this is actually what I was hoping we'd eventually do. In the future this might allow us to do integration tests for playback reporting or Play On, where we can confirm changes directly on a server we control. That could identify regressions both in Finamp and Jellyfin. Of course that's still a long way off :)

I tried `flutter clean`, deleting `android/build`, and deleting the pub cache. Some open Fluter issues suggest it might be because of my setup (e.g., project not on C:). The home screen branch builds fine. I did notice you changing around some gradle config in earlier commits, maybe that was relate? Did you run into any issues back then?

One more thing, I now have a bunch of untracked files. Is there a reason you didn't add them to `.gitignore`?

Komodo5197:
I don't believe I was ever getting hard build failures on android, but I'm not sure.  There was definitely a lot of warnings about things, but a lot of the gradle changes where just me trying to bring everything to up to the latest standard flutter config now that I actually figured out how to get code analysis working on the android native files.  Relating to the drive comment, I believe moving the pub cache to the same drive as the project is actually one of the things I had to do to get that working, but I don't believe I was ever getting build failures from that.  I've just now reconfirmed that the android build is still going fine for me, so I'm not sure what's up.

As a more general note, I'll say that the primary motivation for me to do this was the discovery that a minor dependency upgrade to smtc_windows had actually completely broken the windows app by throwing during startup, but our CI wasn't showing anything because the build process was still going fine.  So the goal of this testing is to verify that all supported platforms can get in-app, and everything else is just a bonus.  This has actually already yielded fruit, as what I initially assumed to be CI bugs actually turned out to be a full app crash during startup on older android versions due to use of the unsupported setApplicationNightMode function.  As I mentioned, if we add more tests in the future I'd prioritize testing of settings migration using data from previous app versions, because that covers most of the remaining startup crash potential.

Anyway, I think I've covered all the comments, so if you don't have anymore thoughts I'll go ahead and merge this sometime tomorrow.

PR #1611, opened 2026-05-01
(discussion) I nailed why gapless playback is broken with transcoding

HeroBrine1st:
## Changes

- Add "isFragmented" field in transcoding profiles and add a transcoding profile which uses /Audio/{id}/Universal API like downloads service which is the one that doesn't have gapped playback issue.  
  Other profiles still use Audio/{id}/main.m3u8 API which is different to downloads service and that's, I think, is the cause of difference.
- Break seeking with new profile

It still has some gap but it is the same as in offline mode - most likely a mismatch of signal between tracks which is pretty normal and means transcoding should go through whole playback, not individual tracks, but this requires server change.

Also, the "gapped playback" in my case is outright skipping some frames and even MPV itself (standalone, on desktop) is confused with those .m3u8 files (opus+mp4) sooo I think the issue is related to that. If I transcode them to flac with ffmpeg, everything's okay and the "some gap" is present in concatenation, too.

## Todo before merging

- [ ] Make all transcoding profiles non-fragmented
- [ ] Fix seeking :-)

## Related Issues

Fixes https://github.com/UnicornsOnLSD/finamp/issues/861

## Discussion

~~I'll of course try to fix seeking right now - I guess there's some code that uses the fact transcoding is enabled to seek inside fragments and now there's only one~~ there's none, idk why it broke - but is this fix even appropriate given it drastically changes behaviour and, I guess, will do some breaks on iOS?

I'll also try to upload some audio files from github web

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -84,9 +84,11 @@
               exec "$@"
             elif [[ "$1" == "target" && "$2" == "list" ]]; then
               # This handles rustup "target" "list" "--toolchain" "stable-x86_64-unknown-linux-gnu" "--installed"
-              # if cross-compiling (e.g. for android), add all four targets here (idk separator though) and to fenix
-              # Currently looks like discord rpc is not enabled on mobile
               echo "${targetName}"
+              echo "armv7-linux-androideabi"
+              echo "aarch64-linux-android"
+              echo "x86_64-linux-android"
+              echo "i686-linux-android"
             else
               echo "Can't run $*"
               exit 2
@@ -117,6 +119,11 @@
                 stable.cargo
                 stable.rustc
                 # rust-src in case of any issues
+
+                targets."armv7-linux-androideabi".stable.rust-std
+                targets."aarch64-linux-android".stable.rust-std
+                targets."x86_64-linux-android".stable.rust-std
+                targets."i686-linux-android".stable.rust-std
               ])
             ] else [
               rustup
--- a/lib/models/finamp_models.dart
+++ b/lib/models/finamp_models.dart
@@ -2801,24 +2801,28 @@ enum KeepScreenOnOption {
 @HiveType(typeId: 73)
 enum FinampTranscodingStreamingFormat {
   @HiveField(0)
-  aacMpegTS("aac", "ts"),
+  aacMpegTS("aac", "ts", true),
   @HiveField(1)
-  aacFragmentedMp4("aac", "mp4"),
+  aacFragmentedMp4("aac", "mp4", true),
   @HiveField(2)
-  opusFragmentedMp4("opus", "mp4"),
+  opusFragmentedMp4("opus", "mp4", true),
   @HiveField(3)
-  flacFragmentedMp4("flac", "mp4"),
+  flacFragmentedMp4("flac", "mp4", true),
   @HiveField(4)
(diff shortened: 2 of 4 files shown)


HeroBrine1st:
[output.mp3](https://github.com/user-attachments/files/27285307/output.mp3) - with opus+ogg profile (new). It sounds like "kzzkt-bam-bam-bam-bam" and it is the same as if I use opus transcoding with offline mode
[output2.mp3](https://github.com/user-attachments/files/27285308/output2.mp3) - with opus+mp4 profile, and this sounds like "kzzkt-bam-bam" - which seems like a skipped frame, and it is also seen if I put the same m3u8 file (from transcodes-cache folder) into regular mpv on PC

kzzkt in both cases is an artifact of mismatch between different transcodes of consecutive tracks and that's a gap between 2 quants in different files (yes that's less than 0.02ms) which is very obvious due to speaker moving rapidly[^1] and doing a good air distortion which resonates with something idk. And this has a very good use in finding when tracks change. Of course original files don't have that gap but I did some manual changes to remove it, so this kzzkt could be problem on my end.

[^1]: Actually it could be due to the fact all speakers have some sort of RC filters which don't pass frequencies higher than e.g. 20 kHz (which is the case for e.g. meander but also for this kzzkt) so it's resonating in the circuit.

This might be placebo :P the difference is very subtle, also it's recorded on android debug build

felix920506:
I am going to chime in here as someone who has previously touched this part of the codebase and has made other Jellyfin music clients in the past.

The Jellyfin team (Niels specifically) said in no uncertain terms that the UniversalAudio endpoint should not be used in this message: https://discord.com/channels/772232779534172171/773650309163253760/1401275288960368661

HeroBrine1st:
Thanks for the info! It's not problematic to use /Audio/{itemId}/stream with segmentContainer unset, so I think that fix will still work, of course if it doesn't break offline transcoding[^1]. It's also possible that this API will fix seeking problems - it has `streamOptions` field which, I guess, is passed to ffmpeg? Idk, need to check

[^1]: I made this term up but I mean online is on-the-fly and offline is transcoded download

My current "idea" is to create some unified function that returns a streamable URL given transcoding parameters that is used by both online and offline transcoding, so this will be used in its implementation.

felix920506:
Another thing is that this is not going to go well with longer tracks that need to be transcoded, especially on servers that are slower. This needs the entire file to be done transcoding before the client can start playing, or else bad things will happen.

On my 5950X system the FLAC -> Opus transcodes can go at about 100x real time, for a 5 mins (300 sec) track it takes 3 seconds so probably no big deal.

On a Pi 4 it can only go about 30x according to Claude. For the same 5 mins track it is going to take 10 secs, not really ideal.

Longest single track I have on my server is 15 mins, which will take 30 secs on the Pi 4. This is simply unacceptable.

We haven't touched on more complex to decode codecs, downmixing, etc. which is only going to slow things down even further.

HeroBrine1st:
I thought the whole reason of broken seeking is the fact it streams transcodes, but... so this is the reason why APIs differ.

Okay, then it can only be fixed in just_audio or Jellyfin (whichever is source). I'll try to look why m3u8 has such strange behavior. My guess it's something with m3u8 file itself, not with individual chunks, since ffmpeg can concat it back into single file perfectly.

Issue #1609, opened 2026-04-30
Alphabetic sidebar weird when albums ordered by artist


n-buck:
### Are you using the newest Version?

- [x] Yes, the issue happens on the newest version
- [x] No, there aren't any open issues about this. I checked!

### Type of Report

- [ ] The app is unusable
- [x] Something isn't quite right
- [ ] Some functionality isn't accessible to me
- [ ] Accessibility
- [ ] None of the above

### The exact version of the app

0.9.23

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Description & Steps to Reproduce

- Chose Album overview
- Order by Artist
- Click on any letter in the side-bar on the left
The app scrolls on a random place, sometimes it takes a long time on larger libraries.

Jellyfin Version: 10.11.8

Expected behavior:
The app scrolls to the first artist starting with the letter I clicked on.
Alternatively, if the backed does not support that, the letter-sidebar should not be displayed.

### Logs

None

PR #1619, opened 2026-05-04
fix(api): request SortName field so fast scroll matches the sorted list

### LINKED ISSUE #978: Fast scroll letters should use the sort by field
There's two related things here, let me know if I should split the issue. (Also if this has been fixed for the redesign, I'm on stable)

First, the letters currently only consider the "real" title, which unfortunately isn't ideal when there's multiple languages. This means that if there's a letter ('K' in this example) that has no matching "real" titles, but matching "sort by" titles, those are skipped.

Also, if there's no matches, it awkwardly scrolls all the way to the end; it probably would make more sense to scroll to the next existing letter instead (so 'J', or if there's no 'J's, then 'K', etc). 

Attaching a screenshot of the library and a video of the scroll in action: 

image

https://github.com/user-attachments/assets/8eaf373d-4e7a-48ed-b153-76ab64037ccb

Thomas05000005:
## Summary

Fixes the fast scroll alphabet jumping to the wrong letter for items whose display name and sort name differ. The bug reported in the issue (and reproduced by `@DanielDeur` in the linked video) is "click `L` → jumps to *Les Bains Douches* even though the album is correctly sorted under `B` because its sort title is *Bains Douches, Les*".

Closes #978

## Root cause

`music_screen_tab_view.dart`'s `scrollToLetter` already uses `BaseItemDto.nameForSorting`, and `nameForSorting` already prefers `sortName` over `name` when available. The actual problem is that `sortName` was never being filled in by Jellyfin in the first place: the `defaultFields` constant in [`jellyfin_api.dart:21`](https://github.com/jmshrv/finamp/blob/redesign/lib/services/jellyfin_api.dart#L21) didn't list `SortName`, so the Items API never returned it. With `sortName == null`, `nameForSorting` falls back to `name` (and the English-only "the / a / an" stripping), which doesn't match the server-side sort order for non-English titles like the "Les …" example, items with `forcedSortName`, or any user-customized sort key.

## The fix

Add `SortName` to `defaultFields`. That's the only change. `nameForSorting` already does the right thing once `sortName` is populated.

```diff
-const String defaultFields = "…,ProviderIds,Tags";
+const String defaultFields = "…,ProviderIds,Tags,SortName";
```

`SortName` is documented as a valid `Fields` value directly inside this same file (the inline docstring on the `getItems` `@Query("Fields")` parameter lists it), so this isn't introducing anything that needs a Jellyfin server-version gate.

## Why this is safe

- The change is purely additive: existing callers that pass an explicit `fields` argument are unaffected, and no field is removed.
- Cost: one extra short string per item in the API response. Negligible vs. the existing payload (covers, tags, blurhash, etc.).
- No new code path: `nameForSorting` already handles `sortName` correctly; we're just feeding it data.
- Touches no Hive schema and adds no setting, so there's no migration concern.

`flutter analyze` returns 0 new issues. Code formatted with `dart format`.

## Test plan

> ⚠️ **Transparency:** Like #1615, #1616, #1617, #1618, this PR is opened as a draft because I have not been able to run the app on a physical device yet. Help testing it on a device would be very welcome.

Manual checks to perform on a device (a library with at least one album whose `name` starts with a different letter than its `sortName` is needed - French, Spanish, German articles, or any album with a custom `forcedSortName` works):

- [ ] Open the music tab, sort by "Sort name"
- [ ] Verify the album sits in the list at the letter of its sort name (e.g. *Les Bains Douches* under `B`)
- [ ] Tap the `B` letter on the right-hand fast scroller - the list should now jump to that section
- [ ] Tap the `L` letter - should NOT jump to *Les Bains Douches* anymore (it should land on actual `L*` items)
- [ ] Same checks while sorted by "Album artist"
- [ ] Reinstall: first-load albums also have `sortName` populated (this would catch any cache invalidation oversight, but I don't believe one is needed here)

### DIFF
--- a/lib/services/jellyfin_api.dart
+++ b/lib/services/jellyfin_api.dart
@@ -18,7 +18,7 @@ import 'jellyfin_api_helper.dart';
 
 part 'jellyfin_api.chopper.dart';
 
-const String defaultFields = "ChildCount,DateCreated,DateLastMediaAdded,Etag,Genres,ParentId,ProviderIds,Tags";
+const String defaultFields = "ChildCount,DateCreated,DateLastMediaAdded,Etag,Genres,ParentId,ProviderIds,Tags,SortName";
 
 @ChopperApi()
 abstract class JellyfinApi extends ChopperService {


Thomas05000005:
Hi,

Quick clarification, since #1623 was merged recently:

- **#1623 (merged)** resolves issue **#1609**: client-side sorting of multiple album-artists and stripping of diacritics for fast-scroll.
- **This PR (#1619)** resolves issue **#978**: asking the server to actually return the `sortName` field (one extra entry in `defaultFields`) so that `nameForSorting` can use it.

Without #1619, an album like "Les Bains Douches" keeps being sorted incorrectly because its server-side `sortName` ("Bains Douches, Les") is never included in the API response. `nameForSorting` then falls back to "name + the/a/an English stripping", which doesn't cover French/Spanish/German articles or custom `forcedSortName` values.

The two changes don't overlap (different files and code paths) and remain complementary.

Thanks!

Komodo5197:
Sorry for not getting this merged.  This exact change did end up getting included into e17b6c9 as part of a broader set of fast scroller fixes, so this should be issue should be resolved as of 0.9.25.

PR #1623, opened 2026-05-12
Fix(fast scroll)/jump to letter when ordered by album artist

### LINKED ISSUE #1609: Alphabetic sidebar weird when albums ordered by artist

n-buck:
#### fix: fix quick-scroll to letter for album artists
* Multiple album-artists are sorted, before taken into consideration for quick-scroll
   * Why is this an issue:
   I.e. Album-Artists `["Wolfgang Amadeus Mozart", "Academy of St Martin in the Fields"]` will be sorted by `A` for `"Academy..."`, but filterd by `"Wolfgang..."`, since it is the first element of the `albumArtists`. This causes all letters between `A` and (including) `W` to be jumping to the album of Mozart and  the Academy of  St Martin in the Fields. 
* Diacritic are removed for quick-scroll
## Related Issues
- fixes #1609

### DIFF
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -282,17 +282,13 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
     SortBy? tabSortBy = FinampSettingsHelper.finampSettings.tabSortBy[widget.tabContentType];
     bool reversed = FinampSettingsHelper.finampSettings.tabSortOrder[widget.tabContentType] == SortOrder.descending;
 
-    String getSortName(dynamic item) {
-      switch (tabSortBy) {
-        case SortBy.albumArtist:
-          final artists = item.albumArtists as List<NameIdPair>?;
-          return removeDiacritics(
-                artists?.sortedBy((e) => e.name ?? '').map((e) => e.name ?? '').join(", ") ?? item.albumArtist ?? "",
-              )
-              as String;
-        default:
-          return removeDiacritics(item.nameForSorting ?? "") as String;
+    String getSortName(BaseItemDto item) {
+      if (tabSortBy == SortBy.albumArtist) {
+        final artists = item.albumArtists;
+        return removeDiacritics(artists?.sortedBy((e) => e.name ?? '').map((e) => e.name ?? '').join(", ") ?? item.albumArtist ?? "");
       }
+
+      return removeDiacritics(item.nameForSorting ?? "");
     }
 
     var targetIndex = itemList.indexWhere((item) {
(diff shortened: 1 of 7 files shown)


lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -281,66 +281,72 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
...
         case SortBy.albumArtist:
-          sortName = itemList[i].albumArtist ?? "";
-          break;
+          final artists = item.albumArtists as List<NameIdPair>?;
+          return removeDiacritics(artists?.sortedBy((e) => e.name ?? '').map((e) => e.name ?? '').join(", ") ?? item.albumArtist ?? "") as String;
         default:
-          sortName = itemList[i].nameForSorting ?? "";
-          break;
+          return removeDiacritics(item.nameForSorting ?? "") as String;
       }
```
UnicornsOnLSD: This function can be simplified/be better typed:

```dart
    String getSortName(BaseItemDto item) {
      switch (tabSortBy) {
        case SortBy.albumArtist:
          final artists = item.albumArtists;
          return removeDiacritics(artists?.sortedBy((e) => e.name ?? '').map((e) => e.name ?? '').join(", ") ?? item.albumArtist ?? "");
        default:
          return removeDiacritics(item.nameForSorting ?? "");
      }
    }
```

This could also just be an if statement rather than a switch statement, but I can see that it was done as a switch statement before.

I tried removing this function and inlining it within the `itemList.indexWhere` below, but it was a bit ugly so I decided not to. I'm not a huge fan of functions-in-functions, so if this function could be useful generally it'd probably be better to add it to `BaseItemDto` so that it can be found later.

```dart
      final sortName = switch (tabSortBy) {
        SortBy.albumArtist => removeDiacritics(
          item.albumArtists
          ?.sortedBy((e) => e.name ?? '')
          .map((e) => e.name ?? '')
          .join(", ") ?? item.albumArtist ?? ""
        ),
        _ => removeDiacritics(item.nameForSorting ?? ""),
      }.toLowerCase();
```

I've made a commit with the changed function.

UnicornsOnLSD (approved):
Thanks for the fix, just noted this small improvement to the code :)

lib/components/MusicScreen/music_screen_tab_view.dart
n-buck: Thanks for the feedback, and fixing.

PR #1624, opened 2026-05-17
Add Sponsor Link

UnicornsOnLSD:
## Changes

This is just a small PR to add @Chaphasilor as Finamp's sponsor destination. I occasionally get emails asking how to sponsor Finamp, so having an obvious link on the repo should help with that.

`FUNDING.yml` allows listing multiple people, so if you think that any other people should be listed, feel free to add them :)

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+github: [Chaphasilor]


UnicornsOnLSD:
It also looks like the landscape has changed a bit on including donation links within apps. A few years ago they'd get mad at you and demand you do it via an in-app purchase, but that rule has been antitrusted to death in the case of pure donations

Chaphasilor:
I definitely don't want to be listed alone here, to be honest 😅  
Sure, I try to invest time into Finamp whenever I'm motivated enough and don't have other things to do, but I don't think I could keep the project going on my own. If @Komodo5197 would be listed too, I'd be much happier with that, since they are what actually keeps the development going.  
But overall donations aren't needed; I get enough to cover any Finamp-related expenses, even without my profile being linked anywhere :)

PR #1625, opened 2026-05-17
Feature: Cross-fade

medallyon:
## Changes
My attempt at adding an optional cross-fade feature to finamp. Adds a setting to Settings > Audio Service in the form of a slider, representing the number of seconds that tracks in a playlist should cross-fade into each other.

## Todo before merging
If you have any clue how to avoid the tiny gap that appears when the crossfade starts, feel free to spit out some ideas. I've tried buffering the secondary instance a little earlier but that doesn't seem to fix it.

- [x] Translations

- [x] Reset Settings

## Related Issues

Should address #237 if we can work out the final kinks.

Also needs more thorough testing on real devices.

### DIFF
--- a/lib/components/AudioServiceSettingsScreen/crossfade_duration_slider.dart
+++ b/lib/components/AudioServiceSettingsScreen/crossfade_duration_slider.dart
@@ -1,0 +1,36 @@
+import 'package:finamp/l10n/app_localizations.dart';
+import 'package:finamp/services/finamp_settings_helper.dart';
+import 'package:flutter/material.dart';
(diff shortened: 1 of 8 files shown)


Chaphasilor:
Hey @medallyon, thanks for whipping up this PR!  
I think the architecture with the outgoing crossfade player and skipping main player is a neat idea and should keep things reliable, if we can get rid of the gap.  
About that: maybe try setting `useLazyPreparation` to `false` when initiating the crossfade `AudioPlayer`? That should cause it to start loading up the audio sources immediately, afaik.

I haven't tested this yet and only skimmed the code, but maybe this pointer help you to finish this up? :)

PR #1630, opened 2026-05-27
Add setting to change time before a repeat is allowed in continuous radio mode

HeroBrine1st:
## Changes

<details>
<summary>A picture is worth a thousand words</summary>

</details>

Add "Allow repeats after" setting in radio menu, visible only when continuous mode is selected (currently appears without any animation), with 90 used as default but set to 5 on screenshot. If field is empty but submitted, it resets to 0, disabling repetition entirely and thus fixing https://github.com/UnicornsOnLSD/finamp/issues/1610.

I should say that, since I don't know flutter well, I decided to slow things down and run OpenCode+gemma4-26b/llama.cpp+[superpowers](https://github.com/obra/superpowers) to write UI code (the latter commit, positive branch of `if` statement only; that plugin was simply installed, listed for completeness). Since resulting UI was very poor, I added leading icon much like ChoiceMenuOption, fixed onSubmitted to reset to 0 on invalid submit and split title to title and subtitle. I say it's 50/50 and I checked it with other code so it looks good to me.

<sup>It might be a good idea to extract this option into reusable widget, but it's currently used in choice list so idk, it's definitely not a choice</sup>

It looks strange since there is some delay between mode selection and sheet closure (remove the delay?), however my idea, I think, has some point:

- Burying it in settings has no value and settings are already burdened
- Keeping it near radio menu allows to use *less* words to explain this setting
- This setting is shown briefly before the sheet is closed due to change in radio mode
  - This is a loss to UX, sure, but this setting is not to be changed often
  - User can open sheet again to change the setting
  - This conveys a clear message "you don't need to change it if you don't want"
  - This also clearly shows relationship between continuous mode and this option and actually guides user about its impact
  - But looks like a bug :-(

I almost said i can't test it but I finally found it (1 minute is used):

<details>
<summary>You just need an artist so unique it doesn't find anything closer than his other tracks</summary>
image

Great album btw. Calming.
</details>

## Alternatives

- Just made it up - somehow query jellyfin on total library size.
- 1 hour later - make it a toggle? 90 minutes or no repeats at all?

## Todo before merging

- [x] Translations
- [ ] Reset Settings - ???
  - This setting is clearly visible when "Continuous" mode is used and has direct and understandable impact, so I guess doesn't matter?

## Related Issues

Fixes https://github.com/UnicornsOnLSD/finamp/issues/1610


HeroBrine1st:
I will rebase later if this doesn't work, it anyway needed conflict resolution

HeroBrine1st:
idk why all of sudden it affected splash screen but anyway

HeroBrine1st:
Conflicted, will resolve conflicts later; I guess it needs reimplementation

Chaphasilor:
Just for reference, here's how this looks in Music Assistant: 
https://www.music-assistant.io/blog/2026/06/10/music-assistant-2-9/#a-kind-of-magic

One thing in general: we might need to change the position if it's not available for all radio modes. It could go right below the continuous mode? But maybe it would be useful to have this repetition stuff for all modes (although it goes somewhat against the "Random" mode)  
Also, UI-wise I think some kind of slider + buttons would be nice, like we have for the playback speed control or sleep timer. I think 5 minute increments (or even coarser) would be fine here. Music Assistant uses hours as an integer it seems.  
Do you think you could port that here?

Chaphasilor:
Either way, we should probably not use 0 for disabled here. Is someone has short tracks (e.g. 30s), that would prevent true randomness. Some kind of toggle or mode selector is probably clearer and less technical?

PR #1632, opened 2026-05-31
Add Android Widget

### LINKED ISSUE #156: [Feature Request] Android Widget
It would be nice to have a widget to control media from the home screen in Android. Not sure if its worth the hassle, though. Just a proposal.

Maybe display album art and such, similar to the way Spotify/Gelli do. Ideally clicking on the widget itself would either take you to the now playing screen, if media is playing/queued, or the home screen if nothing is queued or the status can't be fetched.

call-me-maple:
## Changes

Adds 2 android widgets using the home_widget plugin.

I added some strings defined as xml resources for the widget descriptions, titles, and accessibility descriptions.

### Todo before merging
The data save function is triggered multiple times by the stream listener unnecessarily and I'm not sure how to fix it. It works as it is but I think it might contribute to extra battery usage. 

The widget preview is just static images but they could be dynamic with user data like [this](https://developer.android.com/develop/ui/compose/glance/generated-previews). 

circular_widget_preview
Screenshot_20260602-112206

Closes #156


flloschy:
How do other app handle playback widget?
Just from the looks of things I feel like the wide widget is a bit too feature dense? I could imagine easily fat fingering a wrong button?
Not entirely sure how legit that concern is since I didn't test :D

A few questions:
- Are the widget colors synced with the app theme?
- Will the Finamp log be replaced with song cover (likely yes but just to be sure)
- On the wide widget the Padding to the favorites button looks to too small
- For the small widget will the buttons be configurable? I for example don't like songs that much that I need a button there, (besides that my homescreen doenst even support widgets) I personally I would prefer a shuffle button
- Would it look good if the wide widget has a blurred cover art? Could look bad honestly but worth testing, if thats even possible

Could have looked in the code but eh, no time :D

call-me-maple:
> How do other app handle playback widget? Just from the looks of things I feel like the wide widget is a bit too feature dense? I could imagine easily fat fingering a wrong button?

The width of the buttons meet the recommended size but the height is shrunk down some. I think media control height could be bigger if the font is smaller or we drop the album name since it's a little redundant with the album image. Here are the current Spotify and Youtube Music widgets for reference. 

Screenshot_20260603-135916
Finamp, Spotify, Spotify, YouTube Music

The top 2 widgets are matching the system theme where the other 2 are matching the album art. 


> Are the widget colors synced with the app theme?

They are synced with the system not app atm but I think this could be a configurable. Currently colors are provided from GlanceTheme which will change based on the users system theme. I think another option could be to match the colors to the current album art. 

> Will the Finamp log be replaced with song cover

Yeah 


> On the wide widget the Padding to the favorites button looks to too small

This is easy to change and I'm not opposed to removing it too. 

> For the small widget will the buttons be configurable?

I think this could totally be configured. Currently I haven't implemented any configuration options but the plugin supports it. I'm curious what options people would want besides buttons and colors. 

> Would it look good if the wide widget has a blurred cover art?

I think this is doable and if there's interest in it I don't mind trying it out.

flloschy:
> Finamp, Spotify, Spotify, YouTube Music

Oh ok, so the mainstream already has the buttons this small, then it'll probably be fine if finamp has a similar size :)

> They are synced with the system not app atm but I think this could be a configurable

Oh actually didnt think of that, yeah, probably best to be able to select between System, App and Cover colors huh. Forcing the app color scheme onto the widget (like I implied) might not look that good on some configurations.


> I'm curious what options people would want besides buttons and colors.

I didnt mean full configuration, just the upper right button. I can imagine multiple options being used by different user groups. Like some might want to have the favourite toggle, some want a shuffle, some maybe want a skip button. The lower play/pause button probably is fine actually.

Chaphasilor:
@call-me-maple hey, so firstly thanks for implementing this, seems like you really know your stuff and generally made sensible choices here!  
The only thing I'm wondering about is if we could re-use the existing playback state that is sent to Android, to power the widget, as opposed to manually syncing the state via an additional service? I think this would be sufficient for the current widget selection you added (other widgets might be a different story, although the media browser could potentially used for those.

And also, do you have time to fix the merge conflicts? I think we could include this in the next release, if all goes well.

call-me-maple:
I think I fixed the conflicts. Would changing how the data is saved complicate adding iOS support later on?

Chaphasilor:
Umm, something doesn't seem right with your latest changes 😅

Regarding iOS support, I don't know what the requirements for that will be. Worst case we'll have to add a dedicated message channel for the widgets then, but at this point we'll have a clearer picture of what data and functionality is actually needed. So I wouldn't worry about it too much for now.

PR #1657, opened 2026-06-20
Disable the volume slider when airplay is active

felix920506:
## Changes

Detects when airplay is active on iOS and disables the volume slider, as the in-app volume slider does not work when Airplay is active. Also specifically detects iOS app installed on macOS to avoid incorrectly disabling the slider. There is no reliable way of correctly doing it with iOS app installed on macOS from my research.

## Todo before merging

- [x] Translations

## Related Issues
None

## AI usage disclosure
Mostly written by Claude (Opus 4.8)
Manually tested on actual iOS (26.5) and macOS (15.7) devices, I don't have access to other platforms to test.

## Screenshots
圖片


Chaphasilor:
@felix920506 code looks good, although I'll probably cut down the comments by a good amount before merging. Could you ask Claude to rebase the commits to reduce the amount of commits? Having a few separate is fine, but 9 seems excessive. There are also 3 commits that seem to deal with a macOS-specific implementation that has been abandoned, those should probably be removed entirely and just mentioned in the other commits and a comment in the code.

felix920506:
@Chaphasilor done

Chaphasilor:
Thanks for the PR :)

PR #1662, opened 2026-06-24
[Redesign] Home Screen Followup Fixes

Komodo5197:
This is an ongoing collection of various fixes and improvements that have come up recently.  I expect more to roll in once the home screen finally rolls out to Android.


Chaphasilor:
Why did you decide to remove the section renaming for non-tab sections? I can see that keeping default names might be enough, but in this case preset tab sections should probably also be affected. Only taking this away for collections (and I guess recent queues?) seems a bit inconsistent?

The other changes sound good, haven't tested them yet. I think I might have opinions on the player minimizing feature, but the rest all seems reasonable, thanks!

lib/components/HomeScreen/home_screen_content.dart
```diff
@@ -489,7 +497,15 @@ class HomeScreenSectionContent extends ConsumerWidget {
                   padding: const EdgeInsets.symmetric(vertical: 2.0),
                   child: Container(
                     width: cardWidth * Random().nextDouble().clamp(0.2, 0.9),
-                    height: max(calculateTextHeight(style: TextTheme.of(context).bodySmall!, lines: 1) - 4, 0),
+                    height: max(
+                      calculateTextHeight(
+                            style: TextTheme.of(context).bodySmall!,
+                            lines: 1,
+                            scaling: MediaQuery.textScalerOf(context),
```
Chaphasilor, line 500: I think we also need to update for `track_list_tile.dart` too, while we're at it: https://github.com/UnicornsOnLSD/finamp/issues/1671

lib/components/MusicScreen/music_screen_tab_view.dart
```diff
@@ -126,8 +127,11 @@ class _MusicScreenTabViewState extends ConsumerState<MusicScreenTabView>
                   baseItem.albumArtists?.sortedBy((e) => e.name ?? '').map((e) => e.name ?? '').join(", ") ??
                   baseItem.albumArtist ??
                   "";
+              // TODO how does jellyfin sort handle this?  Do we match?
+              sortName = removeDiacritics(sortName).toLowerCase();
```
Chaphasilor, line 131: That's server logic, since they don't have a fast-scroller equivalent. So I don't think we need to match anything.  
But more relevant imo is if we should continue supporting the "scrolling" part at all, or if applying a simple filter wouldn't be a much better idea, getting rid of all this custom logic, becoming consistent with Jellyfin Web, and improving jump performance drastically. The new sort&filter stuff should handle that just fine, and so should the pagination. UI could also stay the way it is, for now at least.  
What do you think?

lib/components/MusicScreen/sort_and_filter_row.dart
```diff
@@ -293,8 +293,7 @@ class SortAndFilterRow extends ConsumerWidget {
                 SimpleButton(
                   icon: currentConfig.sortOrder.getIcon(),
                   text: currentConfig.sortBy.toLocalisedString(context.l10n),
-                  onPressed: showMenu,
-                  onIconPressed: () => controller._updateConfiguration(
```
Chaphasilor, line 298: Changing this makes it harder to open the filter menu, if you don't know that the center area is tappable. Maybe we should add some kind of placeholder/hint there, like "tap to configure sorting and filters"?  

Additionally, I think tapping the sortorder icon to toggle the order and tapping the sortby name to change the sortby is not a bad idea? I think you mainly noted that the hover effect looks off, right?

lib/l10n/app_en.arb
```diff
@@ -3190,7 +3190,7 @@
     "@favoriteGenres": {
         "description": "Name of a home screen section preset"
     },
-    "newlyAddedAlbums": "Newly Added Albums",
+    "newlyAddedAlbums": "Newly Added Albums (All Libraries)",
```
Chaphasilor: do we actually want to stick with this? I think Eric made a fair point about this being a bit unexpected, and I think *if* people use separate libraries, they'll most likely be aware of which one they added new files to.  
And if not, they are free to customize this section and just select "All libraries" if needed.

lib/menus/components/menuEntries/instant_mix_menu_entry.dart
```diff
@@ -18,18 +19,21 @@ class InstantMixMenuEntry extends ConsumerWidget implements HideableMenuEntry {
...
-      },
+          GlobalSnackbar.message((context) => AppLocalizations.of(context)!.startingInstantMix, isConfirmation: true);
+        },
+      ),
     );
   }
 
   @override
-  bool get isVisible => true;
+  bool get isVisible => !FinampSettingsHelper.finampSettings.isOffline;
```
Chaphasilor: can we just disable it instead? do we have a consistent policy for that now? 😅  
I'm not a fan of hiding options in certain conditions, if people might not be aware of their constraints (i.e. people probably don't know that Instant Mixes depend on the server).

lib/screens/player_screen.dart
```diff
@@ -199,7 +199,16 @@ class _PlayerScreenContent extends ConsumerWidget {
...
             toolbarHeight: toolbarHeight,
             title: PlayerScreenAppBarTitle(maxLines: maxToolbarLines),
-            leading: usingPlayerSplitScreen ? null : FinampAppBarBackButton(dismissDirection: AxisDirection.down),
+            leading: FinampAppBarBackButton(
+              dismissDirection: AxisDirection.down,
+              onPressed: () {
+                if (usingPlayerSplitScreen) {
+                  minimizeSplitScreen.value = true;
+                } else {
+                  Navigator.of(context).pop();
```
Chaphasilor: So one thing that I also saw getting requested (and am also a fan of) is a way to *maximize* the player screen (so essentially toggling the split screen setting from here. This would mean we have 3 states, minimized, split, and full screen. We could keep the existing appbar button for minimizing (like how it is on mobile), and then add a second button next to it for maximizing?

Komodo5197:
Only tab sections actually had retitling fully implemented, and I didn't feel like putting in the work to add it in when it isn't really that useful.

Any thoughts you have on the minimizing are welcome, I believe a corresponding maximizing feature was also requested, and I also considered something which is more like adding a toggle for the existing splitscreen setting to the player so that you need to deliberately toggle back instead of the automatic re-enabling of splitscreen the current implementation uses.

By the way, I belive you never answered in the Home Screen PR and I haven't properly tested yet - does sorting albums by playcount actually work?

lib/components/MusicScreen/music_screen_tab_view.dart
Komodo5197: If we don't match the sort form the server, the fast scroller gets stuck when it sees an item it thinks should be after the target before it.  Anyway, I think directly replacing the fast scroller with letter filters behind the same UI should be fine, although I don't actually use it, so I might not be the best person to ask from a UX perspective.

Chaphasilor:
Yeah I already added a review comment about maximizing :)

So what you're saying is that title editing only worked for tab sections anyway, and now it's just properly reflected in the UI?

About the playcount - I replied *somewhere*, but maybe not in that PR. The issue is that Jellyfin doesn't track play counts for albums at all, since clients can't report that they're playing an album. That's why Jellyfin Web doesn't give the option at all. *Maybe* there's some aggregation going on in the backend, but I'm not sure, and I don't think this could be relied on anyway, nor would it be semantically correct imo. So we should probably not give the option to sort albums by play count, at least until we've added support for the Playback Reporting plugin, where we could do our own aggregations.

lib/components/MusicScreen/sort_and_filter_row.dart
Komodo5197: So I'm pretty firmly against the icon vs name having different functions for two reasons.  First, it is highly inconsistent with the rest of the app, which makes heavy use of icon buttons, making it unexpected and confusing.  Secondly, even if you know the difference exists, the tap target is way too small.  My general thought is that either long-pressing the order button or clicking on the filter button are both fairly reasonable things to do if you want to mess with the order, and both will bring up the menu allowing you to do it.

lib/l10n/app_en.arb
Komodo5197: We can change the preset to be current library only if you'd prefer that.

lib/menus/components/menuEntries/instant_mix_menu_entry.dart
Komodo5197: Yeah, disabling is probably better.

lib/screens/player_screen.dart
Komodo5197: Do you think the splitscreen disable should be persistent?  That is, if we minimize and then click the now playing bar or maximize and then close the player screen, should we go back to splitscreen or go into the other minimized/maximized state?

Komodo5197:
Yeah, custom titles on non-tab sections will not actually save because you're specifically editing the tab title.

I wanted to double check with you on the album play count thing because you added an 'Often Played Albums' home section preset, and even added it to the defaults, but it obviously isn't going to work if play count sorting is broken.  You also added frequently played artists, and I'm not sure that would work properly either?  On a related note, does the unplayed filter work properly for non-tracks?  Because that also has a preset and can be applied to any section in the sort menu.

lib/components/MusicScreen/music_screen_tab_view.dart
Chaphasilor: I don't really use it either, but from the conversations I've seen and had, it seems like a better solution that waiting many seconds to scroll to the right page. Some people even specifically wanted to the Jellyfin Web behavior.

I was originally against it because we had no proper filtering capabilities and I didn't want to break the infinite scrolling, but with the filters this just seems better in essentially every way.

lib/components/MusicScreen/sort_and_filter_row.dart
Chaphasilor: Hmm. I personally more often change the sortby and not the sortorder, how's that for you?

And what about the hint in the center?

lib/screens/player_screen.dart
Chaphasilor: Yeah, it should. Minimize toggles between the now playing bar and the player screen being shown, without changing the split screen setting.

Chaphasilor:
Oh yeah, I guess I forgot to take those two out. I did add it based on the survey results, but didn't think of the API limitations in the moment.

I'm not even sure if the unplayed filter works properly for tracks 😅  
I can double-check for albums when I'm on the desktop.

Chaphasilor:
So I checked, and yeah albums and artists always have `isPlayed` set to false and plays at 0, but for tracks it seems to update fine.  
Tested on 10.10 and 12.0-rc2.

If I want to remove the affected fields (frequentlyPlayedAlbums, frequentlyPlayedArtists, neverPlayedAlbums) from the preset types, I guess I will have to add a migration? Should the old fields be deprecated or completely removed? If we can get play count tracking to work in the future (e.g. by tracking it client-side or using the Playback Reporting plugin), we could probably just recycle the old IDs.

I guess we should also stock up our presets then. The most highly-voted sections from the survey (that we could actually implement would be the following:
- random albums
- forgotten tracks
- recently released albums
- recently added (updated?) playlists
- random tracks
- recently released tracks
- recently added artists
- random artists

Should we just use the top 3 from that list as the replacement? Or do we want to reduce/increase the amount of presets we offer at this point?

lib/menus/components/menuEntries/menu_entry.dart
```diff
@@ -22,6 +30,12 @@ class MenuEntry extends ConsumerWidget {
         onTap: onTap,
       ),
     );
+
+    if (!enabled && disabledTooltip != null) {
+      return Tooltip(message: disabledTooltip, child: entry);
+    } else {
+      return entry;
+    }
```
Chaphasilor: Does anything speak against just adding a general tooltip option, and either only setting that in offline mode, or also providing an online tooltip?

lib/menus/track_menu.dart
```diff
@@ -253,6 +254,7 @@ class _TrackMenuState extends ConsumerState<TrackMenu> with TickerProviderStateM
       if (widget.showQueueActions) CreatePlaylistFromCurrentQueueMenuEntry(),
       if (widget.showQueueActions) ClearQueueMenuEntry(baseItem: widget.item),
       DeleteFromServerMenuEntry(baseItem: widget.item),
+      RemoveFromQueueMenuEntry(queueItem: widget.queueItem),
```
Chaphasilor, line 257: Should we be consistent about conditionally showing items either here (via the inline if) or within the menu item itself? Is there a clear semantic difference between these cases?

lib/main.dart
```diff
@@ -533,6 +533,19 @@ void _migrateHomescreen() {
...
   }
 
+  for (int i = 0; i < finampSettings.homeScreenConfiguration.sections.length; i++) {
+    final section = finampSettings.homeScreenConfiguration.sections[i];
+    if (section.presetType == HomeScreenSectionPresetType.recentlyAddedAlbums) {
+      if (section.base case TabsHomeSection base when base.libraryId == allLibraryPlaceholder) {
+        // We do not preserve the preset value on modified configs, so this section is still default and can be reset.
+        finampSettings.homeScreenConfiguration.sections[i] = HomeScreenSectionConfiguration.fromPreset(
+          HomeScreenSectionPresetType.recentlyAddedAlbums,
+        );
```
Chaphasilor, line 553: Wrote a comment about this before I saw your change here. I think it's possible for people to update the section name for presets without the preset value being removed (which is the whole reason why we don't just always rebuild from the current preset), so I think we should keep the section title here, if modified.

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -304,6 +304,7 @@ class QueueListTile extends StatelessWidget {
         TrackListItemFeatures.duration,
         TrackListItemFeatures.addToPlaylistOrFavorite,
         TrackListItemFeatures.swipeable,
+        TrackListItemFeatures.queueSwipe,
```
Chaphasilor: can we rename these to `swipeableInTrackList` and `swipeableInQueue`?

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -679,7 +683,10 @@ class TrackListItemTile extends ConsumerWidget {
                 height: 1.1,
               ),
               overflow: TextOverflow.ellipsis,
-              maxLines: 2,
+              // It would be better to increase tile height instead of clamping titles to one line and hoping things
+              // now fit, but getting the tile height scaling correct across all widgets is difficult.
+              // TODO properly scale track list tile height
+              maxLines: MediaQuery.textScalerOf(context).scale(15.5) > 15.5 * 1.11 ? 1 : 2,
```
Chaphasilor, line 746: So we trade vertical overflowing for horizontal overflowing?

lib/components/one_line_marquee_helper.dart
```diff
@@ -56,16 +55,7 @@ class OneLineMarqueeHelper extends ConsumerWidget {
...
-            width: constraints.maxWidth,
-            child: BalancedText(
-              text,
-              style: style,
-              overflow: TextOverflow.ellipsis,
-              maxLines: 2,
-              textAlign: TextAlign.start,
-            ),
-          );
+          return Text(text, style: style, overflow: TextOverflow.ellipsis, maxLines: 1);
```
Chaphasilor, line 58: This limits the track title on the player screen to a single line at all times? Why not use a text scaler condition for setting the max line count here too?

lib/models/finamp_models.dart
```diff
@@ -203,15 +203,12 @@ class DefaultSettings {
...
-      FinampFeatureChipType.additionalPeople,
       FinampFeatureChipType.playbackMode,
       FinampFeatureChipType.codec,
       FinampFeatureChipType.bitRate,
-      FinampFeatureChipType.bitDepth,
-      FinampFeatureChipType.sampleRate,
-      FinampFeatureChipType.size,
       FinampFeatureChipType.normalizationGain,
     ],
+    migrated: true,
```
Chaphasilor, line 212: what's the purpose of this? and do we only ever allow one migration through this? we could store the current version / build number instead?

lib/screens/player_screen.dart
```diff
@@ -183,126 +184,153 @@ class _PlayerScreenContent extends ConsumerWidget {
...
-                      );
+        child: ValueListenableBuilder(
+          valueListenable: splitScreenAvailable,
+          builder: (context, value, _) {
+            final appbarActions = [
+              if (!usingPlayerSplitScreen) FinampAppBarBackButton(dismissDirection: AxisDirection.down),
+              if (value)
+                FinampAppBarBackButton(
+                  dismissDirection: usingPlayerSplitScreen ? AxisDirection.left : AxisDirection.right,
+                  onPressed: () {
```
Chaphasilor: The label for the maximize button is still "back".  
But I'd probably avoid swapping the buttons entirely, and keep the left button for minimizing into the now playing bar, and then the right button with an icon like https://tabler.io/icons?icon=arrows-diagonal-2 (and https://tabler.io/icons?icon=layout-sidebar-right-collapse) to toggle between fullscreen and split screen.

Chaphasilor: Also, at the moment, the split screen setting is still toggled when toggling between player and now playing bar. Maybe the now playing bar is causing that?

lib/main.dart
```diff
@@ -377,6 +378,9 @@ Future<void> _setupOSIntegration() async {
         GetIt.instance<ProviderContainer>().listen(brightnessProvider, fireImmediately: true, (_, brightness) {
           windowManager.setBrightness(brightness);
         });
+        if (commandLineArgs.contains("--fullscreen")) {
+          await windowManager.setFullScreen(true);
+        }
```
Chaphasilor, line 389: The feature request for this was for the Flatpak version specifically, I'll have to test if that passes CLI args transparently or not.  
But it should be doable by using [`flatpak run --command`](https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-run:~:text=by%20the%20host.-,%2D%2Dcommand%3DCOMMAND,-The%20command%20to), even if that's not the most convenient.

lib/screens/player_settings_screen.dart
```diff
@@ -102,3 +103,32 @@ class ShowFeatureChipsToggle extends ConsumerWidget {
...
+  const FeatureChipToggle({super.key, required this.feature});
+
+  final FinampFeatureChipType feature;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return Padding(
+      padding: const EdgeInsets.only(left: 40.0),
+      child: SwitchListTile.adaptive(
+        title: Text(context.l10n.featureChipName(feature.name)),
```
Chaphasilor, line 151: Does anything speak against making this a reorderable list, like for tabs?

lib/main.dart
Komodo5197: No, the section editor will never include the preset enum in any sections it saves, so any presets are guaranteed to be completely unchanged.

lib/components/AlbumScreen/track_list_tile.dart
Komodo5197: Basically, except horizontal overflow is already handled with ellipses with vertical overflow just clips.  Scaling the tile height would probably be better, but it's hardwired into too many places and getting the proper scaling height without scaling all the padding is too messy, so I didn't bother.

lib/components/one_line_marquee_helper.dart
Komodo5197: The player screen handles things separately.  This class seems to be only used in the now playing bar and current track on the queue screen, and I don't believe either of them actually expects a multi-line title.  Note that the default setting is to use the fallback at line 15 which is always one line.

lib/screens/player_settings_screen.dart
Komodo5197: Mostly that I didn't feel like it, and it didn't seem that important.  There is the odd wrinkle that we only store the enabled entries, so we would have to figure out how to merge the disabled entries into the list in a hopefully sensible default order.  Do you think this is worth adding?

lib/models/finamp_models.dart
Komodo5197: It's the migration bool for switching to a different set of default features.  I put it here instead of directly into FinampSettings because it was continent.  We can always just add more bools if needed.  Storing the version number and comparing against that would probably be better in general, but I feel that ship has already sailed.

lib/screens/player_screen.dart
Komodo5197: Do you have more details?  The setting seems to be be acting as expected for me.

lib/menus/track_menu.dart
Komodo5197: I'd generally prefer to keep the logic internal to the entries, but I think the ones with the logic outside them currently do make sense to be exceptions to that.

Komodo5197:
I think we do need some sort of migration - just removing any configured invalid presets should be enough.  We need to keep the old enum preset values in as deprectated or the hive migration will not work, because it will fallback to the first enum value if invalid.

Regarding the amount of presets total and the amount of default sections, I don't really have any strong opinions.  The presets are currently presented as an unorganized list, so I wouldn't go too crazy though.

lib/components/MusicScreen/music_screen_tab_view.dart
Komodo5197: So I took a swing at this, and it seems like the server is just filtering by the start of the sortName.  This is an issue because the sortName for tracks starts with the track number for some reason, meaning if you try the filter on the tracks tab you generally get no results.  It looks like the web interface just disables this filter on the tracks page.  Anyway, I think the tracks tab is where this is most useful, so letter filtering might not actually be usable.

Chaphasilor:
One thing I noticed that seems to originate in your Playable changes: when tapping a track on the tracks tab, the queue source is "shuffle all" instead of the track itself (at least when searching). That breaks the radio feature, and is of course incorrect.

lib/screens/player_settings_screen.dart
Chaphasilor: I guess if you disable an entity, it could just be moved to the bottom of the list, instead of staying in place. Then we wouldn't need to worry about the order.  
I don't know if it's "worth" adding, but I mean if we make this customizable and already have a well-fitting UI pattern, why not? ^^  
I can do it too, if you don't want to handle it, that's fair :D

lib/screens/player_screen.dart
Chaphasilor: So when I last tried this, split screen was always disabled when I minimized the player and then opened it again. It never went back to the split player on the side. I think this might be related to how split screen worked before, where you couldn't have the side player and mini player without toggling split screen.

The behavior should be:
- split screen on (side player) -> minimize to now playing bar -> open back up to side player / split screen
- split screen off (maximized) -> minimize to now playing bar -> open back up to maximized player

Or do you disagree here?

lib/models/finamp_models.dart
Chaphasilor: Haha well you're right that we do have a nice assortment of migration bools already, but it's never too late to start ^^  
But I'll leave this up to you, adding more bools is indeed an option, if ever needed.

lib/components/one_line_marquee_helper.dart
Chaphasilor: ah, okay. that's fine then :)

lib/components/AlbumScreen/track_list_tile.dart
Chaphasilor: I guess using the track menu should be enough to see more of the title then. That also supports marquees, if enabled, iirc.

lib/components/MusicScreen/music_screen_tab_view.dart
Chaphasilor: whyyyyyyyyy  
I might bring this up with the server devs as a consideration for 13.0...

lib/screens/player_screen.dart
Komodo5197: So you want maximizing to persistently disable the sidebar until you hit the expand button but minimizing to put you into a temporary override state?  Isn't that inconsistent?  But I'm not sure what people would actually want if they minimize or maximize the player.

Komodo5197:
Should the source be the track itself?  You also get the surrounding tracks in the queue, so that would imply the source was the tracks list itself, or maybe the search you performed, to match the "Your Likes" source you get if you have the favorite filter.  But maybe the source should still have the track integrated somehow, even if it's not primary?  I'm not sure what is best here.

Chaphasilor:
Hmm, good point. We have a source for "<item> - Mix", maybe we could add something similar for "queue started from <item>"?  
Alternatively, I think when searching, using just the selected track as the source (and as the only queue content) would be fine, since I don't think playing all search result is very common. And then we could keep existing sources (like favorite tracks) for any track lists accessed via home screen sections, where there actually are related items to be played afterwards already.

lib/screens/player_screen.dart
Chaphasilor: I'd put it this way: I want the maximize / split screen setting to toggle if tapping the now playing bar opens the full screen player or the side player. It's not really a temporary override semantically (although I guess given the historic implementation it might be required here).

Or in other words, with the player screen open, you can toggle between full screen and split screen, and then they can stay in that state, only toggling between minimized and open.

lib/models/finamp_models.dart
```diff
@@ -4311,8 +4311,24 @@ enum FinampQuickActions {
   playRandomAlbum,
   @HiveField(4)
   playRandomTrack,
+  @HiveField(10)
```
Komodo5197, line 4391: I'd prioritize making this a configurable option now, because it really bulks up the action list and kind of obfuscates the actually unique actions.  I'm thinking we have a random item and random favorite option where you can check which item types you want.

Chaphasilor: I'll give it a try later

lib/services/jellyfin_api.dart
```diff
@@ -698,14 +699,13 @@ Future<DeviceInfo> getDeviceInfo() async {
       : kProfileMode
       ? "profile"
       : "";
+  final uniqueDeviceId = FinampSettingsHelper.finampSettings.deviceId;
   if (Platform.isAndroid) {
     AndroidDeviceInfo androidDeviceInfo = await deviceInfo.androidInfo;
-    final appSetId = await AppSetId().getIdentifier();
-    info = DeviceInfo(name: androidDeviceInfo.name, id: "$appSetId-$idExtension");
+    info = DeviceInfo(name: androidDeviceInfo.name, id: "uniqueDeviceId-$idExtension");
```
Komodo5197: You forgot to use the variable.  Also, this will not work properly for the background thread due to use of Hive, and I'm pretty sure it runs there.  Also, you forgot to set the uuid in the migration case.  Also, if the point of this is to remove app_set_id, you should probably completely remove it from pubspec.

Chaphasilor: ahh, the linter even told me but I just didn't see the missing $ and though it was lagging behind. oh man...  
about the background threads: can't we get Hive into those at all? should be be passing this when the API is initialized?  
And for the migration, you mean for when the settings already exist, right? Yeah, you're right of course, there's no default value for Hive to use...

Komodo5197: You get no hive access in the background threads.  If you want the data you need to pass it back manually.  I have considered making a service that watches all the settings and just passes them across threads for whoever wants them, but I never ended up actually doing that.  isar is accessible, but I don't think this makes sense in the useradata.

lib/components/HomeScreen/quick_action_editor.dart
```diff
@@ -116,7 +117,7 @@ class QuickActionConfigMenuState extends ConsumerState<QuickActionConfigMenu> {
               // if (preset == RadioMode.similar && radioModeOptionAvailabilityStatus.isAvailable)
               //   Icon(TablerIcons.star, size: 14.0),
             ],
-            enabled: true,
+            enabled: currentActions.where((action) => action.action == quickAction).isEmpty,
```
Komodo5197: It's a bit niche, but someone could conceivably want multiple copies of an quick action which is editable.

Chaphasilor: I can refactor that when adding the configurable quick actions for "Random xyz" 👍

lib/screens/home_screen_settings_screen.dart
```diff
@@ -525,8 +525,8 @@ class _ResponsiveListTile extends StatelessWidget {
         )..layout();
 
         // Measure action button text to calculate actual button widths
-        const minTextWidth = 200.0;
-        const horizontalGap = 48.0;
+        const minTextWidth = 150.0;
+        const horizontalGap = 30.0;
```
Chaphasilor: I chose these values to prevent some tiles being wrapped but not others on smaller screens. I know this "wastes" some space, but I believe it looks and behaves much more coherently?

lib/services/finamp_user_helper.dart
```diff
@@ -14,7 +14,7 @@ final finampUserHelperLogger = Logger("FinampUserHelper");
 /// Helper class for Finamp users. Note that this class does not talk to the
 /// Jellyfin server, so stuff like logging in/out is handled in JellyfinApiData.
 class FinampUserHelper {
-  FinampUserHelper() {
+  FinampUserHelper({required String? deviceId}) : _deviceId = deviceId {
```
Chaphasilor: Not sure if you're fine with doing it like this. I see no way to completely evade the user helper, since the device Id is needed for auth.  
But maybe it's better to store this in Isar then?

Komodo5197: Can't you just store the value given in jellyfinApi.create() and not need to pass it around any more than that?  Also, you put a hive access into _processRequestsBackground, which is a background function.

lib/screens/home_screen_settings_screen.dart
Komodo5197: I believe the previous values were large enough to force a guaranteed wrap on my phone, which makes the whole complex mode switching seem a bit pointless.  You can re-raise the min size if you want, but I think shrinking the gap, possibly even more than I already did, is the correct choice.  Also, I just noticed we seem to be lacking right padding in horizontal mode.

lib/services/finamp_user_helper.dart
Chaphasilor: Yeah I noticed that too now, I though it was just before the isolate is ran. For some reason it also worked in the Windows build, do isolates work differently on Desktop?  

The problem with `JellyfinApi.create()` is that the entire auth header generation happens outside that class; `getAuthHeader()` is a function, not a method. I could try refactoring this to be a method, but on startup the API is set up after the user helper, so that seems be problematic.

lib/screens/home_screen_settings_screen.dart
Chaphasilor: The mode switching is specifically for Desktop, were we want either all buttons to be inline with the title, or none (all in a separate line). Does that explain what I was trying to do?  
I'll fix the right padding

lib/services/finamp_user_helper.dart
Komodo5197: I think hive just fights itself over the files, so the behavior can be inconsistent, but I'm not sure.  If getAuthHeader is static, I guess all the passing around is fine for now.  Maybe I'll take a swing at cleaning it up myself.

lib/screens/home_screen_settings_screen.dart
Komodo5197: Okay, so this is primarily intended for desktop.  Honestly, if I were doing it I probably would have put in a basic switch on a conservative total width > 800 if that was the case instead of doing all the work calculating tile widths.  I'll note that if we do want all sections to use the same format, doing per-tile calculations on title length is counterproductive.

lib/components/MusicScreen/sort_and_filter_row.dart
Komodo5197: Honestly, I don't mess with it much at all.  I feel like putting a hint in the middle makes it music screen feel a bit inelegant, but I'm not sure that's meaningful.  I feel like the filter icon is the main hint to get the menu open.  I wouldn't be that opposed to just moving the toggle to long press if you feel this is all too messy.

lib/screens/home_screen_settings_screen.dart
Chaphasilor: I guess the threshold of 800 might work. Even the longest translations will probably not exceed that. I'll see if I can simplify things

Komodo5197:
@Chaphasilor I've been trying to come up with a way for the Finamp logo on the music screen to properly read as a button so that we don't need the redundant three dot button on the right.  However, I've been stuggling to find something that is clearly a button but still matches the rest of the design.  It seems like the standard button indication throughtout the app is just the uye of the standard icons, which leaves me without a lot to work with outside just reverting to the three bar menu icon, which is something I've also been considering.  The below are the best I've gotten so far, but I'm not particularly enamored with either.  Do you have any thoughts on this?

image
image

Chaphasilor:
So the only consistent pattern I know here is using the user avatar as a menu entry point. Spotify uses that instead of the app icon. YT Music shows the app icon, but it's not clickable (and instead also uses the avatar for the menu). So maybe we should be using that, although I feel like exposing the downloads screen via a "user" menu is semantically flawed (for playback history, recent queues, libraries, etc. it seems fine though).  

One thing that's currently not considered in your design is that tapping the library name also opens the menu/drawer. On discord @erikbucik suggested adding a chevron / down arrow as a kind of "library drop down", and then making the logo + library name appear to be one single unit. I was wondering if we could give the icon and library name some (subtle) elevation to help with that too? Something like this maybe:  
image


Also, on desktop I think we could even show the drawer at all times on the left size, we should have enough space for that. But I guess that's a separate discussion.

lib/components/MusicScreen/sort_and_filter_row.dart
Chaphasilor: yeah I think a long-press for toggling the sort order should be fine, and then we have the two buttons that bring up the full menu. That should be enough hints for now :)

lib/screens/player_screen.dart
Chaphasilor: perfect now!

lib/components/AlbumScreen/track_list_tile.dart
```diff
@@ -670,24 +671,46 @@ class TrackListItemTile extends ConsumerWidget {
...
-                overflow: TextOverflow.clip,
-                style: TextStyle(
-                  color: Theme.of(context).textTheme.bodyMedium?.color,
-                  fontSize: 16,
-                  fontWeight: FontWeight.w500,
-                ),
-              ),
+              child: isCurrentTrack && !features.contains(TrackListItemFeatures.cover)
+                  ? MiniMusicVisualizer(
+                      animate: true,
```
Komodo5197, line 684: We should stop the animation if the track is paused.

lib/components/now_playing_bar.dart
```diff
@@ -452,15 +436,42 @@ class NowPlayingBar extends ConsumerWidget {
...
-                                                  visualDensity: const VisualDensity(horizontal: -4),
-                                                ),
+                                              AddToPlaylistButton(
+                                                item: currentTrackBaseItem,
+                                                queueItem: currentTrack,
+                                                color: primaryTextColor,
+                                                size: 28,
+                                                visualDensity: const VisualDensity(horizontal: -4),
+                                              ),
+                                              Stack(
```
Komodo5197: Why are we moving this off the album cover?  I'd prefer keeping the actions minimal so we can see more of the song title.

Komodo5197:
Weren't you the one saying that the current drawer isn't the correct set of options to be showing on desktop?  So that would be a larger project.  I feel like making the library look like a dropdown does make some amount of sense, although that only displays the library switching, not the general menu functionality.  It might work if we elevate the logo+title without an additional icon, though?  I might try something out along those lines.  Regarding the user icon being the menu, not only do I agree it seems a bit flawed here, but I actually actively dislike when apps do this in general.  It feels very "mobile", in the sense of an app designed for users who will not want to or be allowed to express any actual preferences for how they use the program or go outside of the most common approved workflows.

lib/components/AlbumScreen/track_list_tile.dart
Chaphasilor: We don't have a provider for the play state yet, do we? Should I add one?

lib/components/now_playing_bar.dart
Chaphasilor: Two designers independently moved the button out of the cover, citing contrast issues and right-handed accessibility (https://discord.com/channels/1064474779043770408/1509905220879122482/1509905367780298823, https://discord.com/channels/1064474779043770408/1519082327244935198/1523379905033797823).  
Having the actions in a row also allows them to be configurable down the line, if we want that.

But this is not set in stone. If you feel like we could fix the contrast reliably, then that would be an option, too.

Komodo5197: I made some tweaks to the contrast a couple commits ago.
| Before | After |
| --------|------|
|image|image|
|image|image|
|image|image|

lib/components/AlbumScreen/track_list_tile.dart
Komodo5197: I don't think we do, so go ahead.

lib/components/now_playing_bar.dart
Komodo5197: Okay, I decided to revert it, but it should be pretty easy to add a setting for this if you'd like that.

Komodo5197:
I decided to just stick the logo in a basic rounded rectangle, but with less background contrast.  Tell me what you think, I guess?  Anyway, I'm looking to start move this closer to merging.  I think the last issue I was thinking about was potentially changing the behavior of the random radio mode.  The wording of the mode doesn't really clearly convey that's it's pulling from the queue, and I've seen some people confused by this, so I'm considering changing it so that if the initial source is shuffle all (and maybe also a genre), then it uses the full source, because both of those cut you off at the shuffle limit when starting playback.  Maybe if the source is a genre, we still combine the fetched genre tracks with any queue items that have a different track source.  I feel like this is a lot closer to what people will be expecting from the mode.

I've also been considering adding some sort of "Quick Settings" page that shows up once when you first boot the redesign.  It could have things like the grid mode toggle + grid size, dark mode setting, and the transcode setting with a link to the full transcode settings page.  Does any of this seem like something we should be doing?  If so, do you have any other settings you think are frequently used and worth advertising?

lib/components/now_playing_bar.dart
Chaphasilor: sounds good, the contrast really does look much better. Did you also apply this to the queue list?

Chaphasilor:
> Weren't you the one saying that the current drawer isn't the correct set of options to be showing on desktop?

Yeah, you're right. I do still think that things like the logs button doesn't really fit in a persistent side panel, but maybe that should be moved to the settings anyway. The library selector could work well though. I'm definitely not as opposed to it anymore than I was back then, I think.

About the logo, I'm not super happy with how it looks, so I'll try to tweak it further. But the general approach of adding elevation to signal clickability does make sense, I think.  

About the radio mode changes, if combined with the "Shuffle All Tracks" source, would this essentially serve as a full-library shuffle, but without de-duplication?  
I would like to keep the original use case of "just the queue tracks, but in a random order" intact, since this was contributed by someone looking for this feature specifically. But I think this is still given with your approach? For reference: https://github.com/finamp-app/finamp/pull/1313#issuecomment-3146741878  
Regarding the genre handling, I'm not sure what your reasoning is for that, why the different behavior?

Your suggestion about some sort of "onboarding" flow is valid. I think I'd prefer showing this as a kind of "banner" at the top of the home screen instead, so that people can start customizing on-demand, and don't have to make decisions before they even used the app at all. I'll think about good candidates for that (tabs come to mind), but also think this should be added in a separate PR. I'm also looking to merge this PR soon, and will try to get my own TODOs checked off asap. After this is merged, another update seems appropriate, possibly including #1504 and some smaller PRs that are ready to merge. Then, if there aren't any additional bugs popping up, we should start the preparations for the stable release (see https://github.com/orgs/finamp-app/projects/2/views/2 and feel free to add more things).

lib/components/now_playing_bar.dart
Komodo5197: Yeah, that's changed too.

Komodo5197:
I'm not particularly happy with the look either, but I think I got it good enough to be better than nothing.  If you can improve it, I'd be very appreciative.

It would essentially serve as a full library shuffle, yes.  I don't think shuffle all would ever have full deduplication, but this wouldn't even have the rolling deduplication that the similar modes have.  Still, with a full library that is fairly unlikely to matter.  I'm considering more advanced logic for genres because they get cut off by the shuffle limit, just like shuffle all does.  So in a similar fashion, if you play a genre and then select random, you probably want to select from the whole thing instead of what happened to come up.

Making the quick settings prompt available after the app is actually interactive does seem sensible.  Putting this in another PR does seem like the correct move.  I agree that #1504 is the last major feature that seems worth adding before stable, if it gets finished up in a reasonable time.

By the way, do you have any thoughts on the drawer open gesture?  I recently discovered that in addition to the custom swipe right on leftmost tab gesture, we also have the a flutter default swipe from left edge of screen gesture.  That default gesture seems very particular, which is presumably how I managed to miss that it existed for the last few years.  Does this gesture interfere with the OS back swipes in any way, or is it fine to keep around for quicker access from other tabs?  also, on iOS the custom gesture is disabled, forcing you to either use the flutter gesture or the top left button.  Even if it's bad, we definitely need the flutter gesture if it's the only one available, but I'm not actually sure why we don't let iOS use the custom one.

Chaphasilor:
The shuffle all random mode sounds reasonable, I agree that duplication is unlikely enough to not matter, and I would like to keep the random mode fully random. Extending that logic to genres as well makes sense too, although I'm wondering if artist tracks and playlists aren't similarly affected?

As for the gesture, I was never able to use the default Flutter gesture (which seems like a standard Android thing, I know other native apps like Solid Explorer support it too) because my current ROM completely overrides this with the system-level back gesture. But definitely no interference. I don't use our own drawer gesture at all, so I'm impartial here.

I have now added the customizable play item and play favorite item quick actions, although I had to add Hive IDs for the `BaseItemDtoType` enum in the process, which surprised me. Do you think I should be using different types here, or take another approach to begin with?

And I'll see what I can do about the icon, I'm sure we can figure out some improvement

lib/main.dart
```diff
@@ -369,7 +369,9 @@ Future<void> _setupOSIntegration(List<String> commandLineArgs) async {
       backgroundColor: Colors.transparent,
       skipTaskbar: false,
       titleBarStyle: TitleBarStyle.normal,
-      minimumSize: Size(336, 607),
+      minimumSize: Size(400, 250),
+      // This matches the size of the iPhone 5, which is probably the smallest screen size worth testing against
+      //minimumSize: Size(336, 607),
```
Chaphasilor, line 377: Is this in response to this message?  
https://discord.com/channels/1064474779043770408/1064571023212691577/1525949211319861349

If not, what do you think about reducing the minimum size further? I'd argue that people will notice if the UI breaks, and adjust the window size accordingly.

Komodo5197: I'm actually pretty sure this was changed by mistake, so I'm just reverting it.  400x250 is the original minimum size from a release or two ago, and 336x607 is a value that I use to mimic a small phone screen but have never intentionally committed.

Komodo5197:
Artists and playlists are not currently affected by the shuffle limit because we do not paginate them, so fetching from the server isn't needed.

I wish I knew why the drawer gesture was disabled on iOS, but I guess I'll just leave everything to be for now.

I think most of type storage we have is done via ContentType, so it might not be crazy to use that instead?  I'm not sure that really matters, though.

lib/main.dart
Chaphasilor: As I said, I'm happy to reduce the minimum further

Chaphasilor:
What kept me from using `ContentType` was that to me it seemed like it was meant for a different scope, with values like `home`,`mixed`, and `inPerformingArtistAlbums`. Item type seemed better suited, but not ideal either. If we were to use `ContentType`, how would you filter it down to the relevant options? Via `directlyDisplayable`, or something similar?  

I've made some changes to the Finamp icon button, does that work for you?

Edit: I forgot the quick action type migration, didn't I? 🙃

Komodo5197:
So I tried the quick action out, and it seems to behave a bit oddly if you select multiple item types at once.  First of all, you are far more likely to get items from types with more items overall - this does make sense, but I'm not sure it's something anyone would actually want?  Also, getItems redirects requests based on item type so artists and genres will not work properly in multi-type requests.  Combined with the title questions, I'm wondering if it's better to just limit users to one type?  Alternatively, we might want to choose a random type first, and then request just that type from the server to achieve more balanced results and fix the broken requests.

Komodo5197:
If we were to use contentType, you could just add another bool like isBasicJellyfinType or something, although you're right that it doesn't exactly match.  However, I've always wanted to refactor the BaseItemDtos to be aligned closer to Finamps usecases and not 1-to-1 with the jellyfin values, so I'm hesitant to start saving them.  Alternatively, we could save off the actual jellyfin type names as a comma separated string, although that feels a bit silly.

Regarding the Finamp button, I'm a bit concerned the dark-mode version might look a tiny bit too much like the OS app icon, fooling people into thinking we're just including the background in the logo.  But otherwise, I think it's better?  I guess we can just push it to the beta and see what everyone else thinks.

Chaphasilor:
I think randomly picking the type first and then making a request really is the better approach here. I'd like to keep the multi-item functionality, since that was requested multiple times.  

And I guess we'll just use the ContentType then with the new bool. Better keep Hive clean to avoid more migrations.

About the icon, yeah we'll see. I'm using the same color as the buttons in most scenarios, aside from when the full-color icon. Although maybe I can tweak this a bit further to only force the color if the accent color is dynamic, and otherwise use the default blue accent to keep the style consistent.

Chaphasilor:
Okay, is this ready to be merged from your side? There still are some build failures with GH Actions seeming to be the cause, maybe we should try to fix those before the next release, so the CI can upload the artifacts automatically?

Oh and have looked at https://github.com/finamp-app/finamp/pull/1673 yet? Can that be merged?

Komodo5197:
Regarding merging this, I think I've got pretty much everything I want in, although you might want to double-check that I didn't break things with the random radio change.  We do still need to remove and migrate the invalid frequently played album/artist presets and replace the default.  It looks like we also have a now-invalid randomFavorite quick action that needs a type in defaults.

Regarding the CI issue, It looks like the issue is an outdated java path occasionally getting preserved in the build cache?  But beyond that, we seem to have acquired flutter 3.44-specific code.  The crash doesn't replicate on my machine, so I don't know if it's safe to revert the version pinning on Linux.

Chaphasilor:
Will add the missing types, thanks for catching that. One thing I was wondering, right now the quick actions on the home screen don't reflect the configured item types. That's probably not ideal, but I see no nice way to indicate this space efficiently? Maybe we could use icons, and change the title to "Play random <track icon><artist icon><album icon>", for example? Although that might be hard to do with proper localization.

I can try to build the linux version locally, if that works I at least have a way to create the artifact myself. Although I probably won't be able to get the ARM artifact...  
We could also try to downgrade Flutter if you think it would help?

I'll test the radio changes and try to add the migrations.

Komodo5197:
I did change the actions to include the type in the name if there's only one configured, which is at least something.  Regarding Linux, the problem is that we pinned it to an older version of flutter because builds with 3.44 were crashing on some machines for an unknown reason.  But we ended up using some flutter 3.44 features in the latest code, so that doesn't build anymore.  The CI was one of the environments where it was crashing, but I ran the integration tests and they seem to work now, so it's possible the issue has been patched and we're fine to step Linux up to flutter 3.44.

lib/services/radio_service_helper.dart
```diff
@@ -347,32 +351,142 @@ Future<List<BaseItemDto>> generateRadioTracks(
...
-      return originalQueue[nextIndex];
-    });
+      int nextIndex = _radioRandom.nextInt(queueTracks.length + sourceTracksLength);
+      if (nextIndex < queueTracks.length) {
+        output.add(queueTracks[nextIndex]);
+      } else {
+        // Just go through source tracks in order, relying on server shuffle for randomness
+        // This avoids generating excessive duplicates by overweighing the returned values
+        if (sourceTracksIndex < sourceTracks.length) {
+          output.add(sourceTracks[sourceTracksIndex]);
```
Chaphasilor: so at the beginning we fetch child tracks with the default sort for that item (so not random), but then at the end we rely on randomness of server-fetched tracks. That seems like it won't behave as desired in those cases?

Chaphasilor:
Testing the random radio mode, I found the code issue above, and some other issues.
1. When starting random radio from a genre with several hundred tracks, the radio mode seems to fail? Initial generation works, but after that no other radio tracks are generated. Logs show that 10 tracks are needed, but 0 generated, based on the correct genre item.
2. For Shuffle All, the "Playing from Radio - Radio" source name seems a bit silly. Maybe we can use "Playing from All Tracks - Radio" instead?

The rest seems to be working well, with some duplicates even if the underlying source hasn't been exhausted yet.

lib/services/radio_service_helper.dart
Komodo5197: If tracks are loaded through loadChildTracksFromBaseItem, they will always fall into the case where we move them from sourcetracks into queueTracks, so they get the full random treatment from there.  The only way to potentially not hit that branch is to load from shuffleAll, which is obviously random.  So I believe the logic is sound, but I'll add a better comment.

lib/models/finamp_models.dart
```diff
@@ -292,8 +292,9 @@ class DefaultSettings {
     sections: [
       HomeScreenSectionConfiguration.fromPreset(HomeScreenSectionPresetType.recentlyAddedAlbums),
       HomeScreenSectionConfiguration.fromPreset(HomeScreenSectionPresetType.favoriteTracks),
-      HomeScreenSectionConfiguration.fromPreset(HomeScreenSectionPresetType.frequentlyPlayedAlbums),
+      HomeScreenSectionConfiguration.fromPreset(HomeScreenSectionPresetType.favoriteAlbums),
       HomeScreenSectionConfiguration.fromPreset(HomeScreenSectionPresetType.favoritePlaylists),
+      HomeScreenSectionConfiguration.fromPreset(HomeScreenSectionPresetType.favoriteArtists),
```
Komodo5197: This seems like a lot of favorite sections - we seem to be up to five by default?  I guess we don't have a crazy variety of presets, especially without a lot of the frequently played ones, but can we really not get more variety than this?

Chaphasilor: We could definitely get more variety. I honestly just went through the survey results and picked the sections with the most votes that were feasible to implement, making sure to cover all major item types aside from genres (sorry Lukas, but they just weren't super popular and are available on-demand).  

I do see that offering more ways to browser for things to play would be neat, but that seemed to be less in demand? We could use random albums instead of favorite ones, but that was much less popular that favorite albums (96 vs 64 selections).   
Theoretically we could also add both the favorite and random sections, but I don't know if a migration should increase the total number of sections? 

One opportunity to increase "browsability" down the road is to add more personalized sections based on the `/similar` endpoint or AudioMuse results, but that's not ready yet.

Komodo5197: I believe that variety is actually more important than aligning with the most common desires, primarily to aid discoverability.  The defaults should be a sort of sampler of everything the home screen can do, so that as many people as possible can see the potential value it has and establish an idea of what they might actually want out of it.  They can then double down on whatever they like best with the customization.

Chaphasilor:
Okay, tested your radio mode changes and that seems to be working now. So let's merge this :)

PR #1670, opened 2026-06-29
Fix empty Genres tab

### LINKED ISSUE #1653: Genres tab does not get populated

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.9.24

### Affected Devices/Platform

- [ ] Android
- [ ] Android Auto
- [x] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

_No response_

The `Genres` tab does not get populated on the latest Beta version even though the genres themselves do exist.

https://github.com/user-attachments/assets/c2c29050-eea0-4615-b04d-0bb07ed2cfff

### Logs

[finamp-logs-2026-06-16T15-21-06-220586.zip](https://github.com/user-attachments/files/28986692/finamp-logs-2026-06-16T15-21-06-220586.zip)

### Additional Information

Tested on both Jellyfin v10.11.10 and v10.11.11.

kontell:
## Problem
The Genres tab is empty when online. Other tabs work and offline mode lists genres fine.

I had Claude (Opus 4.8) debug the issue and it came up with the below.

## Cause
`getIsFavoriteFilter` returns `false` (instead of `null`) for the genres/mixed tabs when no favorite filter is active, so finamp sends `isFavorite=false` to the `/Genres` endpoint. Jellyfin 10.10/10.11 returns HTTP 500 on `/Genres` whenever `isFavorite` is present (`true` or `false`), so the request fails and the tab stays empty. `/Items` (used by the other tabs) accepts `isFavorite=false`, and offline never calls `/Genres` — which is why only the online Genres tab breaks.

## Fix
Return `null` when the favorite filter is inactive, so `isFavorite` is omitted from the `/Genres` request.

## Verification (Jellyfin 10.11.11)
- `/Genres?…&isFavorite=false` → 500
- `/Genres?…` (no `isFavorite`) → 200
- `/Items?…&isFavorite=false` → 200

## Out of scope
Filtering the Genres tab *by favorites* still won't work on 10.11 (that path sends `isFavorite=true`, which also 500s). That looks like a server-side bug in Jellyfin's `/Genres` endpoint and may be worth reporting upstream separately.

Fixes #1653

### DIFF
--- a/lib/services/jellyfin_api_helper.dart
+++ b/lib/services/jellyfin_api_helper.dart
@@ -1276,7 +1276,9 @@ class JellyfinApiHelper {
     // I guess part of the reason for this is that it's not possible to favorite a genre through the Jellyfin Web UI at all...
     if ([finamp_models.ContentType.genres, finamp_models.ContentType.mixed].contains(contentType)) {
       // Only send isFavorite when the filter is actually active. Passing isFavorite=false makes
-      // Jellyfin 10.10/10.11 return HTTP 500 on the /Genres endpoint, leaving the Genres tab empty (#1653).
+      // Jellyfin 10.11 return HTTP 500 on the /Genres endpoint, leaving the Genres tab empty (#1653).
+      // On Jellyfin 10.10 and 12.0, isFavorite=false returns only items that are *not* favorites,
+      // which is also not what we want here (we want all genres to be shown, unfiltered)
       return filters.any((filter) => filter.type == ItemFilterType.isFavorite) ? true : null;
     }
     return null;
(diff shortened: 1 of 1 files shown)


Chaphasilor:
@kontell can you share a video and logs of the issue please? I've tried this with 10.10.7, 10.11.11, and 12.0-rc2, and all of them work fine, showing all genres with no filter active, and only favorite ones when the favorite filter is applied.

Chaphasilor:
Whoops sorry, I got a bit confused about which branch and test server I was on 😅

So what I found is:
- Jellyfin 10.11 is completely broken (no genres shown at all when filter is set to `false`)
- Jellyfin 12.0-rc2 and 10.10 are "broken" in the same way (favorite genres are excluded when favorites filter is set to false, this seems like it's actually the desired behavior)

Your fix indeed works for all three versions, so thanks! It's what I would've done too, so props to Claude I guess ^^  
Speaking of, could you please remove Claude as a co-author in the commit? You can keep the attribution and which model you used in the PR description, but the contributor section in Finamp is meant for humans :P  
I can also do that myself, if you prefer.

kontell:
Claude's been nixed now, thanks for reviewing this.

Chaphasilor:
Awesome, thanks for whipping this up!

PR #1678, opened 2026-07-10
Fixed Images showing as html code in the end of README

helloitshecker:
Just fixed the images by using markdown instead of pure html. if you want, i can resize images to make them smaller

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ This app is still a work in progress, and has some bugs/issues that haven't been
 
 | | |
 |:-------------------------:|:-------------------------:|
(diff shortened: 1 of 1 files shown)


Chaphasilor:
Thanks! We'll have to go over the readme soon anyway when publishing the beta changes, but I guess fixing things now doesn't hurt. And yes, it might be a good idea to make the images a bit smaller ^^

helloitshecker (commented):
also, the style tag doesnt work for github readmes.

Chaphasilor:
Thanks a ton!

PR #1680, opened 2026-07-12
Use homescreen code paths for CarPlay where possible

### LINKED ISSUE #1586: Update CarPlay to use MusicScreenProviders
CarPlay has a lot of custom methods that could be standardised on the new homescreen work once both are merged. 

See: https://github.com/UnicornsOnLSD/finamp/pull/1516#discussion_r3040435362

APIUM:
## Changes
Replace as much of my custom CarPlay code for things like queue building with work from the new homescreen.

## Related Issues
- closes #1586

### DIFF
--- a/lib/services/carplay_helper.dart
+++ b/lib/services/carplay_helper.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 
 import 'package:finamp/components/MusicScreen/sort_and_filter_row.dart';
@@ -9,6 +10,7 @@ import 'package:finamp/services/music_providers.dart';
 import 'package:finamp/services/music_screen_provider.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter_carplay/flutter_carplay.dart';
+import 'package:infinite_scroll_pagination/infinite_scroll_pagination.dart';
 import 'package:audio_service/audio_service.dart';
 import 'package:finamp/models/finamp_models.dart';
 import 'package:finamp/models/jellyfin_models.dart';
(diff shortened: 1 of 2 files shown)


APIUM:
I also have in here something I used for debug "[Allow Jellyfin calls from CarPlay in debug builds](https://github.com/finamp-app/finamp/pull/1680/commits/52f3a2e144a6cf6251006400255f8d21392f6705)" I am happy to remove this, or submit separately.

Chaphasilor:
@APIUM would you mind posting some screenshots or a short video of how it looks/works now, just for reference?

APIUM:
> @APIUM would you mind posting some screenshots or a short video of how it looks/works now, just for reference?

Display is identical for this change, I have used these to make some UI changes to be submitted in a follow up PR

Chaphasilor:
Ah, so you just re-create the existing CarPlay content via the new home screen methods, got it. I just got confused by the mentions of home screen sections, and thought you already rendered the sections configured for the in-app home screen.

This looks mergeable from my side. @Komodo5197 what do you think? Is this something you'd like to postpone until after the stable release?

Komodo5197:
It doesn't look too crazy, but I haven't properly looked through everything yet.  I noticed the providers are used via read(), including using the slice method on the page provider, and I want to think about whether we might be able to do things in a more reactive way.  I'm also wondering about the changes to verifyCallable(), because I wouldn't have expected carplay to be acting during build().  The changes are very isolated from the main app though, which does increase my confidence for allowing this to be pushed before stable.  If this is just a pure code improvement I might wait on it, but if it also has bug fixes it's probably not too crazy to try to get it in.

APIUM:
> I noticed the providers are used via read(), including using the slice method on the page provider, and I want to think about whether we might be able to do things in a more reactive way. I'm also wondering about the changes to verifyCallable(), because I wouldn't have expected carplay to be acting during build().

Thanks @Komodo5197 I’ve played around this and made some changes, it is more reactive but it’s slightly different to the normal path now. Let me know what you think. 

There are a fair few bug fixes in this, but they are all pretty minor. Up to you on risk v reward. I want this in as I’ve got some CarPlay feature updates that I’d like to add such as allowing viewing far more of the library, a better homescreen, and now playing controls such as radio, shuffle and favourite. I’ll get that up when we’re ok with this one.

The fixes are (commit msg snippets):
- Offline only and favourites only browsing now transfers to carplay
- List data stays cached until carplay rebuilds, so tapping deep into a long list plays the tapped track instead of re-fetching
- A failed fetch cancels the navigation instead of showing an empty library, and the next tap retries cleanly
- Offline Recently Added is sorted by date added instead of random download order
- Recently Played shows your server-wide play history, and hides when offline since play dates only exist on the server
- Shuffle all respects the only show favourites setting.
- Start radio is the same Surprise Me as the rest of the app, still falling back to shuffle when offline.

Chaphasilor:
Those fixes sound valid to me, so lets wait what Komodo thinks :)

lib/services/carplay_helper.dart
```diff
@@ -136,146 +144,132 @@ class CarPlayHelper {
...
-      // Offline: instant mix not available, fallback to shuffle
-      await shuffleAllTracks();
-      return;
+    try {
+      final items = await completer.future;
+      // pagedContentProvider isn't generic enough to express that children here are always FinampPlayableDto.
+      return items.take(limit).map((x) => (x as FinampPlayableDto).item).toList();
+    } catch (_) {
+      // Reset the failed pages like the main UI retry button so the next tap refetches cleanly.
+      providerRef.read(provider.notifier).retry();
```
Komodo5197: I think this retry should be somewhere else, because I believe if the page provider is being listened to elsewhere, prephaps by the main app, this will lead to an aggressive retry loop.

lib/services/carplay_helper.dart
```diff
@@ -136,146 +144,132 @@ class CarPlayHelper {
...
+  Future<List<BaseItemDto>> _loadPagedItems(FinampPagedPlayable<FinampPlayableDto> request, int limit) async {
+    final provider = pagedContentProvider(request);
+    final completer = Completer<List<FinampDisplayableOrPlayable>>();
+
+    // Retain the paged data so it survives for later taps, until the next root
+    // rebuild releases it.
+    _templateSubscriptions.add(providerRef.listen(provider, (_, _) {}));
+
+    // Drive the load on a private subscription so a root rebuild closing the
+    // retained subscriptions cannot strand this pending load.
```
Komodo5197: I'm not sure why closing this subscription early would be a problem?  Wouldn't that be what we want because the results are no longer needed?  I do see that we might want to fully cancel this listener while keeping the page provider alive with a different subscription, though.

APIUM: Was because it lead to a deadlock, but your way is cleaner, updated to that.

lib/services/carplay_helper.dart
APIUM: Thanks yep good pick up, moved to next tap in the list / next load

Komodo5197:
Looks good to me.  I'm fine with merging this for the upcoming release.

Chaphasilor:
Great. Thanks a ton @APIUM for following up with this! Looking forward to the other planned improvements 😁

PR #1681, opened 2026-07-12
fix(android-auto): refresh playback state after toggling shuffle

Chachigo:
## Changes

Toggling shuffle from the Android Auto / now-playing controls didn't update the shuffle icon visually.

`togglePlaybackOrder()` wasn't awaited and `refreshPlaybackStateAndMediaNotification()` wasn't called afterwards — unlike `toggleFavoriteStatusOfCurrentTrack()`, which already does both. This aligns the shuffle action with that pattern.

1 file, +2/-1.

### DIFF
--- a/lib/services/music_player_background_task.dart
+++ b/lib/services/music_player_background_task.dart
@@ -1103,7 +1103,8 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
       switch (action) {
         case CustomPlaybackActions.shuffle:
           final queueService = GetIt.instance<QueueService>();
-          return queueService.togglePlaybackOrder();
+          await queueService.togglePlaybackOrder();
+          return refreshPlaybackStateAndMediaNotification();
         case CustomPlaybackActions.radio:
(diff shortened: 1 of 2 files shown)


Chaphasilor:
@Chachigo thanks for the fix. Could you please remove Claude as a coauthor? Feel free to move the model you used into the PR description instead, but we want to reserve the "Contributor" title for humans, not agents!

lib/services/music_player_background_task.dart
```diff
@@ -1103,7 +1103,8 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
       switch (action) {
         case CustomPlaybackActions.shuffle:
           final queueService = GetIt.instance<QueueService>();
-          return queueService.togglePlaybackOrder();
+          await queueService.togglePlaybackOrder();
+          return refreshPlaybackStateAndMediaNotification();
```
Komodo5197: I think this should be inside setPlaybackOrder so that toggling from the player screen also results in an update.

Komodo5197:
I tested it out, and this seems to work nicely.

Issue #377, opened 2023-01-06
Control server DLNA


provokateurin:
Split off from https://github.com/jmshrv/finamp/issues/50

provokateurin:
@jmshrv I just looked at what the web UI does and it only uses the https://api.jellyfin.org/#tag/Session APIs to control the playback. This way it should probably be even simpler.

jmshrv:
Ah yeah it should be easy if DLNA can be controlled there :)

Finamp should probably be controllable through the session API, but that'll be quite a niche and complex feature that's out of scope here anyway 🙃

Chaphasilor:
> Finamp should probably be controllable through the session API, but that'll be quite a niche and complex feature

I mean I honestly wouldn't mind if the next/prev track and play/pause buttons would be shown on the dashboard, but I'm guessing this requires a socket connection?

provokateurin:
That part can be implemented separately, but I agree it's probably a feature that's not going to be used much.

jmshrv:
I'd assume so, I'd need to look for some documentation. [Emby's doc on remote control](https://github.com/MediaBrowser/Emby/wiki/Remote-control) only talks about controlling other clients, not about how to become a controllable session.

provokateurin:
Re-architecting this is quite hard. It's basically impossible to just make two separate classes with the different implementations. It's not as clean, but I think it actually makes more sense to just have the two implementations in the existing MusicPlayerBackgroundTask. This also makes it a lot easier to transition between the two implementations when playing something, because they know of each other.

provokateurin:
And I think some kind of WS will be required, because the clients need to be informed about changes of the session (e.g. play state). Not sure where to find documentation about that. Maybe it makes sense to implement the being-remote-controlled part first, because then we can already listen for those changes.

Sadly this seems to be a lot of work and I don't have enough time to get into this in the next few weeks. I can take a look again (hopefully) around mid February, but please don't have any expectations. I hope someone else can pick this up in the mean time :)

Chaphasilor:
> And I think some kind of WS will be required, because the clients need to be informed about changes of the session (e.g. play state). 

Ahh right, because the playback happens server-side and could be influenced by the playback device itself (e.g. pressing a pause button on the speaker/tv).

As for documentation, if there really is nothing to be found, the source code of the Jellyfin app should offer some insights. That's one of the clients supporting remote control, afaik. The Jellycon Kodi addon as well, but that's not a mobile client.

[...]

PR #1683, opened 2026-07-14
Add DLNA/UPnP audio output support

jtabet:
## Summary

Adds DLNA/UPnP audio casting support, allowing users to play music on DLNA-compatible speakers and receivers on their local network.

The device selector is integrated into the existing **Output** menu (the "Sortie" button on the player screen), alongside the system output routes (speaker, Bluetooth, etc.).

## How it works

- **Discovery**: Combined SSDP (via `dart_cast`'s `DlnaDiscoveryProvider`) + mDNS (`_raop._tcp.local` via `multicast_dns`) for devices that don't respond to SSDP multicast. Manual IP entry is also available for devices not found via discovery.
- **Playback**: Audio is **proxied through the phone** — the DLNA device fetches from the app's local HTTP proxy, not directly from Jellyfin. This means the DLNA device only needs to reach the phone, not the Jellyfin server.
- **Protocol info**: Sends correct audio `protocolInfo` (e.g. `audio/mpeg:DLNA.ORG_PN=MP3`) and `object.item.audioItem.musicTrack` DIDL-Lite class, instead of the video-oriented metadata that `dart_cast` defaults to.
- **Format detection**: Maps Jellyfin container strings (mp3, flac, wav, m4a, etc.) to MIME types for correct DLNA protocol info.
- **Resume position**: When switching to DLNA output, the current playback position is passed to the DLNA device so playback resumes from where the user left off.
- **Auto-advance**: When the DLNA device reports a track has finished, the queue automatically advances to the next track.

## Files changed

- `lib/services/dlna_service.dart` (new) — DLNA service: device discovery (SSDP + mDNS), connect/disconnect, `loadMedia()` with audio protocolInfo, media proxy, position polling
- `lib/services/music_player_background_task.dart` — DLNA mode integration: `connectToDlna()`/`disconnectFromDlna()`, DLNA-delegated play/pause/stop/seek/skip, auto-advance, Jellyfin URL + MIME type helpers
- `lib/menus/output_menu.dart` — `DlnaDeviceList` widget in the existing output menu bottom sheet
- `lib/main.dart` — Register `DlnaService` in GetIt
- `lib/components/now_playing_bar.dart` — Cast indicator when DLNA mode is active
- `lib/l10n/app_en.arb` — New l10n strings
- `android/app/src/main/AndroidManifest.xml` — WiFi/network/multicast permissions
- `pubspec.yaml` — Add `dart_cast: ^0.6.0` and `multicast_dns: ^0.3.3`

## Tested against

A DLNA renderer that does not respond to SSDP multicast but advertises via mDNS as `_raop._tcp.local`. Both streaming from Jellyfin and playing downloaded files verified working.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -6,6 +6,9 @@
   <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
   <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
   <uses-permission android:name="android.permission.VIBRATE"/>
+  <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"/>
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+  <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE"/>
   <!-- io.flutter.app.FlutterApplication is an android.app.Application that
          calls FlutterMain.startInitialization(this); in its onCreate method.
          In most cases you can leave this as-is, but you if you want to provide
--- a/lib/components/now_playing_bar.dart
+++ b/lib/components/now_playing_bar.dart
@@ -453,6 +453,15 @@ class NowPlayingBar extends ConsumerWidget {
                                           Row(
                                             mainAxisAlignment: MainAxisAlignment.end,
                                             children: [
+                                              if (audioHandler.isDlnaMode)
+                                                Padding(
+                                                  padding: const EdgeInsets.only(top: 4.0, right: 4.0),
+                                                  child: Icon(
+                                                    Icons.cast_connected,
+                                                    size: 16,
+                                                    color: Theme.of(context).colorScheme.primary,
+                                                  ),
+                                                ),
                                               Padding(
                                                 padding: const EdgeInsets.only(top: 4.0, right: 4.0),
                                                 child: AddToPlaylistButton(
--- a/lib/l10n/app_ar.arb
+++ b/lib/l10n/app_ar.arb
@@ -2854,5 +2854,12 @@
     "previousTracksPersistenceModeExpanded": "موسع (ظاهر)",
     "@previousTracksPersistenceModeExpanded": {
         "description": "Overrides saved value to expand previous tracks header on queue list open"
-    }
+    },
+    "outputDevices": "أجهزة الإخراج",
+    "thisDevice": "هذا الجهاز",
+    "scanningForDevices": "البحث عن الأجهزة...",
+    "noDevicesFound": "لم يتم العثور على أجهزة",
+    "addDeviceManually": "إضافة جهاز يدوياً",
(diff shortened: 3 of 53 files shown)


jtabet:
## Refactored: switched from `dart_cast` to `upnp_client` + custom proxy

The branch has been force-pushed with a single clean commit replacing the previous 8 commits.

### What changed

**Before:** Used `dart_cast` (primarily a Chromecast library) for DLNA. This required workarounds:
- Manual raw SOAP XML building (dart_cast hardcodes `video/mp4` protocolInfo)
- Manual DIDL-Lite metadata (dart_cast uses `object.item.videoItem`)
- `DlnaSession` state machine hacks (`forceState()`, private polling)
- `IOClient` wrapper to work around connection keep-alive issues

**After:** Uses `upnp_client` (pure Dart UPnP client) for SOAP/UPnP control + our own `DlnaMediaProxy`:
- Clean typed API: `avTransport.play()`, `avTransport.pause()`, `renderingControl.setVolume()` instead of raw SOAP
- `upnp_client` correctly parses device description XML and extracts service URLs
- No session state machine to fight with — just call methods directly
- Our own `DlnaMediaProxy` adds `contentFeatures.dlna.org` and `transferMode.dlna.org` headers to media responses — these are required by some DLNA devices (notably Samsung TVs) for Pause/Stop/Seek to work

### Why the proxy matters

Some DLNA renderers (Samsung TVs in particular) play audio fine but return SOAP 501 "Action Failed" for Pause/Stop/SetVolume unless the media stream response includes DLNA content-feature headers. The `dart_cast` `MediaProxy` doesn't add these headers and its handlers are private (can't be extended). Our `DlnaMediaProxy` is a simple ~300 line class that adds them by default.

### Dependencies changed
- Removed: `dart_cast: ^0.6.0`
- Added: `upnp_client: ^1.0.4`, `xml: ^6.6.1`
- Kept: `multicast_dns: ^0.3.3` (for mDNS discovery of non-SSDP devices)

### Testing
- Build passes (`flutter build apk --debug`)
- Needs device testing on Rakoit Up2Stream and Samsung TV to verify the DLNA headers fix Samsung Pause/Stop/Volume

Chaphasilor:
@jtabet hey, I'll have to close this PR for several reasons. Client-side UPnP/DLNA support is not something we're currently interested in supporting in Finamp. We feel like this adds unnecessary complexity, and there are existing alternatives like Symfonium that support this. Additionally, there was no prior discussion about adding this feature before the PR was opened.  
If you look at existing features like #377 and #50, you'll see that the plan for that was to use the existing DLNA capabilities of Jellyfin to support this with minimal effort, instead of doing this client-side. However, this is already being worked on :)

jtabet:
Jellyfin's DLNA implementation is on the server side, which is fine if your goal is to stay on the same network as the server, but it doesn't really make sense to me. I'll keep my own fork then :shrug:

garrit-schroeder:
This feature would be much appreciated as my server is not running on the same network. Especially for users using the app but are not running the server. Friends and family members etc.

PR #1684, opened 2026-07-16
fix: detect mTLS client certificate requirement from HTTP 4xx responses

cbueth:
This PR follows up to the last merged PR adding mTLS support @Maxr1998 kindly worked on.

## Changes
Extend mTLS client certificate detection in the login flow to also handle HTTP 4xx responses, not just TLS-level alerts. PR #1575 added mTLS support, but the detection only handles `TLSV1_ALERT_CERTIFICATE_REQUIRED` during the TLS handshake. Some reverse proxy configurations (i.e. nginx with `ssl_verify_client optional` + application-level checking) reject requests without a valid client certificate at the HTTP level with 4xx status codes. When this happens, `clientCertificateRequired` is never set, the import prompt never appears, and the app falls through to HTTP on port 8096 where it gets a connection refused error.

There were two places of changes. Extended the catch block after `loadServerPublicInfo()` to check for any 4xx HTTP status code in addition to the existing TLS alert check in `login_flow.dart`. And the same logic in the auto-discovery catch block for `login_server_selection_page.dart`.

## Todo before merging
- [x] Reset Settings (no new settings)

## Related Issues
- #1575

### DIFF
--- a/lib/components/LoginScreen/login_flow.dart
+++ b/lib/components/LoginScreen/login_flow.dart
@@ -9,6 +9,7 @@ import 'package:finamp/services/server_client_discovery_service.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/widgets.dart';
 import 'package:get_it/get_it.dart';
+import 'package:chopper/chopper.dart' show ChopperHttpException;
 import 'package:http/http.dart';
 import 'package:logging/logging.dart';
 
@@ -263,15 +264,31 @@ class ServerState {
       try {
         publicServerInfo = await jellyfinApiHelper.loadServerPublicInfo();
       } catch (error) {
-        if (ClientCertificateInstaller.isSupported &&
-            error is ClientException &&
-            error.message.contains("TLSV1_ALERT_CERTIFICATE_REQUIRED")) {
-          clientCertificateRequired = true;
-          // Found server requiring mTLS certificate, no need to try other protocols/ports.
-          return;
-        } else {
-          serverStateLogger.severe("Error loading server info: $error");
+        if (ClientCertificateInstaller.isSupported) {
+          if (error is ClientException && error.message.contains("TLSV1_ALERT_CERTIFICATE_REQUIRED")) {
+            clientCertificateRequired = true;
+            // Found server requiring mTLS certificate, no need to try other protocols/ports.
+            return;
+          }
+          if (error is ChopperHttpException && error.response.statusCode >= 400 && error.response.statusCode < 500) {
+            clientCertificateRequired = true;
+            return;
+          }
+          // Some reverse proxies reject without client cert at HTTP level.
+          try {
+            final statusCode = (error as dynamic).statusCode as int?;
+            if (statusCode != null && statusCode >= 400 && statusCode < 500) {
+              clientCertificateRequired = true;
+              return;
+            }
+          } catch (_) {}
+          // Non-JSON response body (HTML error page).
+          if (error is FormatException) {
+            clientCertificateRequired = true;
+            return;
+          }
         }
+        serverStateLogger.severe("Error loading server info: $error");
       }
       if (this.baseUrlToTest != baseUrl) {
         throw Exception("Server URL changed while testing");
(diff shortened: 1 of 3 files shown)


Maxr1998:
Hm. I don't feel like treating all 4XX errors as potential certificate required markers is the right approach. Isn't there a different way to detect these alternative reverse proxies?

If we e.g. showed the certificate required message in case of a 404 (because of an incorrectly entered address), this would most likely just confuse most users not running a mTLS setup.

cbueth:
I see the problem. It is just that in the reverse nginx proxy case, which I think is not too uncommon, the current solution will never allow adding the certificate from the startup screen. While you can expect the json from `/System/Info/Public`, this is not the case in the nginx mtls case.
We could try to reduce false positives by narrowing it down with two conditions. To 400 (nginx ssl_verify_client optional + application-level checking) and 403 (forbidden/access denied). And checking if the content is text/html.

Chaphasilor:
@cbueth @Maxr1998 so yeah, this is why I suggested allowing people to proactively add mTLS certs :P  

But improving the auto-detection is good anyways. I'd probably err on the side of showing the mTLS prompt on errors, just to be sure. But maybe we can differentiate between invalid credentials, which are also a 4xx code?  

One approach here would be to keep the current `clientCertificateRequired` bool but also add a second one, `clientCertificateMaybeRequired`, where the latter shows an additional message in front of the button, like "If you have mTLS set up for your server, you can add a certificate here. Otherwise, something else went wrong."?

Maxr1998:
Maybe we should instead add the "advanced settings" at the bottom that I suggested once instead, to not make the connection logic itself any more complex. A simple, unconditionally shown button that opens a sheet, essentially.

cbueth:
I'm fine to make a commit with the solution you prefer. Do you say I should integrate what I last suggested, or remove this logic entirely and add some kind of dropdown or similar way one can always access this?

At setup the sidebar is not accessible, so one cannot add a cert there deliberately, but I think that way would also be least intuitive.

Chaphasilor:
@Maxr1998 so you're fully against detecting any errors during server discovery and showing the mTLS prompt in response? You just want to add the "Advanced Settings" button which opens a bottom sheet, which then contains just the mTLS button (for now)?

Chaphasilor:
@cbueth after some more discussion we decided on the following approach:
We add an "advanced settings" section at the bottom of the login screen (the server selection one). This is just an expandable section

<details>
<summary>think this</summary>
like the HTML details element
</details>

which can be tapped to reveal additional settings that might be needed, like the mTLS button. This should ensure that people always have the opportunity to configure mTLS, no prompt required.  
And the existing mTLS detection (the exact one, from Max) will just show a message like "mTLS setup detected. Configure it in the advanced settings below.", instead of showing the button directly.

As for implementation, you might want to use a "SimpleButton" for the advanced settings toggle, maybe with a chevron icon that reflects the current collapsed/expanded state.

What do you think about that? Can you handle the implementation? 😁

cbueth:
Thanks for the reply @Chaphasilor. I think this is a good solution meeting in the middle. I could make the changes. Should I open a follow-up PR for that and close this one?

Chaphasilor:
@cbueth feel free to reuse this one!

Chaphasilor:
@cbueth any updates on this? do you want us to handle this instead? I fear we need to make another release soon, because Google Play reviews are incredibly slow recently and the stable version "expires" at the end of August...

cbueth:
Hey @Chaphasilor, thank you for the reminder. It would be really helpful if you could handle this, my current schedule is too tight, thanks. You can close this issue when the follow-up is merged. Thank you for your work!!

Maxr1998:
I can take a look at this too.

Chaphasilor:
@Maxr1998 that would be really welcome! <3

Chaphasilor:
superseded by #1739. will give you a shout-out in the release notes though!

cbueth:
Thank you, @Maxr1998 and @Chaphasilor !!

Issue #1626, opened 2026-05-23
Quickly Maximize Player Panel on Dekstop/Tablet


SIRTACOOOOO:
### Are you using the newest Version?

- [x] Yes, the feature is still missing on the newest version
- [x] No, there aren't any open request about this. I checked!

### Type of Feature

- [ ] Something new
- [x] Improvement of a feature
- [ ] None of the above

### Description

Spotify on an ipad has this option to shrink the player if its on the right, or full screen the player. i would llike that to be apart of the finamp player ui

flloschy:
- wdym "if its on the right"? 
- we already have a full screen player?

Could you provide some screenshots?

SIRTACOOOOO:
Here is a screenshot I found on the Internet 

<img width="2160" height="1620" alt="Image" src="https://github.com/user-attachments/assets/ed28c296-cb11-4f0b-b24a-50ffd16189fc" />

If you screen is big enough on finamp the player is forced to the right side and you can only resize it. Spotify has an option with the button I circled to full screen the player, or you can shrink the player to be a tiny pill

flloschy:
Ah okay, that makes more sense to me now.
So if i understand correctly you want a quick-access button to "Settings>Layout&Theme>Allow SplitScreen Mode"?
basically momentarily disabling splitscreen without needing to go into the settings?

SIRTACOOOOO:
Exactly, or full screen the player.

flloschy:
Mkay
Im not accustomed to the Spoify UI so i was curious what you were referring to

dompetriella:
@SIRTACOOOOO hello, I'm actively working on this.  Just want to confirm functionality.

In Spotify, when the user presses the expand button it shows the full screen player.  Pressing that button again returns it to the split view (as pictured above)

As described by @flloschy would be something slightly different.  It would disable the split screen mode and vertically collapse the player (ie, not expand it across the screen like spotify's UI) and move the player to the small taskbar at the bottom of the screen.  Pressing the taskbar will make the player full screen.  Pressing the expand button again would return split screen (similar to the UI you've specified above).  Similar but with a couple extra steps, and a little clunky.

Are you looking to just replicate Spotify's functionality or a "quick access" for the split screen mode as described by @flloschy?

SIRTACOOOOO:
@dompetriella I think you have the same idea as me and I'm sure whatever you can make would be fine. Since I have zero coding experience in dart😅

[...]

PR #1701, opened 2026-07-30
Add split screen toggle for widescreen devices

dompetriella:
## Changes
- Added split screen toggle IconButton to AppBar for widescreen devices
- Refactored player_split_screen_scaffold.dart a bit for better separation and readability
- Added a SplitScreenHelper to more easily access split screen values
- Added the https://pub.dev/packages/material_symbols_icons package for "Spotify accurate" icons on the IconButton.  
(Can definitely remove if we want to avoid adding more packages to the yaml, just wanted it to be as accurate to the request as possible)

- [x] Translations

- [x] Reset Settings
(Settings toggles were not directly affected but settings are being manipulated, so it seems relevant)

## Related Issues

- addresses feature request #1626 

Notes:
This isn't the _best_ solution for this issue but I explored several other avenues with little success
- Forcing the split screen view to fully expand is the better / cleaner solution, but when the main screen is disposed of, the rebuild of the screen duplicates some widgets.  Couldn't figure out the root cause of this, and it would appear to not like being "squished"
- Pushing a new PlayerScreen widget to the top of the stack would smoothly show the screen as full screen, but there are some issues with .pop() in that pushed PlayerScreen() that could get messy.  This might be a decent long term solution if the above root cause isn't identified.
- The notification bar is blocked very briefly to prevent some of the jank of changing split screen settings.  While this could potentially cause the now player bar to not show, I don't see how without access to the split screen button, and would be fixed on app restart.
- Ended up going with a slightly modified version of what @flloschy suggested, thank you for that

### DIFF
--- a/lib/components/PlayerScreen/player_screen_appbar_title.dart
+++ b/lib/components/PlayerScreen/player_screen_appbar_title.dart
@@ -1,23 +1,28 @@
 import 'package:balanced_text/balanced_text.dart';
 import 'package:finamp/components/PlayerScreen/queue_source_helper.dart';
+import 'package:finamp/components/global_snackbar.dart';
 import 'package:finamp/l10n/app_localizations.dart';
 import 'package:finamp/models/finamp_models.dart';
+import 'package:finamp/screens/player_screen.dart';
+import 'package:finamp/services/finamp_settings_helper.dart';
 import 'package:finamp/services/queue_service.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:get_it/get_it.dart';
+import 'package:material_symbols_icons/symbols.dart';
 
 import '../../extensions/localizations.dart';
 
-class PlayerScreenAppBarTitle extends StatefulWidget {
+class PlayerScreenAppBarTitle extends ConsumerStatefulWidget {
   const PlayerScreenAppBarTitle({super.key, required this.maxLines});
 
   final int maxLines;
 
   @override
-  State<PlayerScreenAppBarTitle> createState() => _PlayerScreenAppBarTitleState();
+  ConsumerState<PlayerScreenAppBarTitle> createState() => _PlayerScreenAppBarTitleState();
 }
 
-class _PlayerScreenAppBarTitleState extends State<PlayerScreenAppBarTitle> {
+class _PlayerScreenAppBarTitleState extends ConsumerState<PlayerScreenAppBarTitle> {
   final QueueService _queueService = GetIt.instance<QueueService>();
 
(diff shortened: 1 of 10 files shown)


dompetriella:
This is my first time contributing to this repo, so if I royally messed something up please let me know :)

lib/components/PlayerScreen/player_screen_appbar_title.dart
```diff
@@ -35,39 +48,70 @@ class _PlayerScreenAppBarTitleState extends State<PlayerScreenAppBarTitle> {
...
+                child: IconButtonWithSemantics(
+                  label: splitScreenState
+                      ? AppLocalizations.of(context)!.switchToFullscreenViewTooltip
+                      : AppLocalizations.of(context)!.switchToSplitScreenViewTooltip,
+                  onPressed: () async {
+                    final splitScreenTransitionNotifier = ref.read(splitScreenTransitionProvider.notifier);
+                    if (splitScreenState) {
+                      await splitScreenTransitionNotifier.runTransitionActions(() async {
+                        FinampSetters.setAllowSplitScreen(false);
+                        await Future.delayed(Duration(milliseconds: 100));
```
dompetriella, line 100: Added a slight delay here (and below) to avoid some of the animation jank when changing modes.  Can remove, but this smooths things out

Chaphasilor:
hey @dompetriella thanks for creating the PR! However, we actually already implemented that in another PR that I just merged, #1662. I know the title doesn't make that obvious, so sorry about that.  
Can you build the `redesign` branch and check how that compares to your approach? Maybe we can make some small adjustments based on your ideas!

dompetriella:
> hey @dompetriella thanks for creating the PR! However, we actually already implemented that in another PR that I just merged, #1662. I know the title doesn't make that obvious, so sorry about that. Can you build the `redesign` branch and check how that compares to your approach? Maybe we can make some small adjustments based on your ideas!

@Chaphasilor Looked through the code and it's pretty much the same implementation.  I'll be honest, it's a better implementation ( PostFrameCallback instead of hard coding a duration is a MUCH better solution for the animation jank).  My PR includes a little housekeeping for the split_screen_scaffold but it's probably not really worth creating an entire PR for.  I'll just take this PR down and pick up another issue.  Thanks for bringing this to my attention instead of just sending the PR to the void haha

Komodo5197:
Sorry about this, I really need to be a lot more proactive about linking issues in my PRs.  Thanks for the work regardless.

dompetriella:
> Sorry about this, I really need to be a lot more proactive about linking issues in my PRs. Thanks for the work regardless.

No worries!  Thanks for all your work on this too.

PR #1705, opened 2026-08-03
[Redesign] Quick settings screen

Komodo5197:
This adds a "Quick Settings" screen with a couple of the most commonly used settings on it.  It also adds a one-time banner in the music screen header to direct new users towards it.

TODO:
- Setup translations
- Visual polish
- Collect additional feedback on commonly used settings


Chaphasilor:
@Komodo5197 is the plan to get this in pre-stable? Do you still need help with anything here or more ideas?

Komodo5197:
I do plan for this to go in before stable, and would consider this the last PR I would actively delay the release for, assuming we don't find some sort of crazy bug once the android play store beta actually comes out.  Any suggestions for additional common settings to add, or even remove, would of course be appreciated.  I do feel the total number of exposed settings is in a reasonable place, but the selection is probably still biased towards what I've personally tweaked.

Another big concern is the wording of all the new text.  I messed with a number of the settings to try to make the wording more clear to a new user, but there might be further possibilities on this front.  There's also the header paragraph I ended up adding, which feels a bit unusual and which I'd like your thoughts on.  Of special concern is the main message shown on the banner, which I'm currently not crazy about.  You can also feel free to tweak the order and grouping of setting if something seems out of place.

Chaphasilor:
Okay, here are some thoughs:
1. We can probably remove the fast scroller setting
2. Should we show the bitrate slider if transcoding settings are enabled?
3. Local network address settings might be useful and hard to discover, so they could go here
4. The "All settings" button should probably come at the end
5. What do you think about skipping the library selector during login (even if multiple libraries exist), and instead including that setting in quick actions?
6. Accessibility settings might also be something worth surfacing early
7. The system accent color might be something people want to pick at the start


I'll probably go over the wording a bit tonight, but most things definitely seem reasonable.

Komodo5197:
1. Sure.
2. Sure.
3. I guess that makes sense?
4. I feel like it makes more sense as part of the header?
5. I'd lean towards leaving it as-is, partially because I just don't want to mess with the login.  I would also expect people to already have the needed context immediately, but I would also expect most people to just want all libraries, so that does make the decision a bit redundant.  I do like it as a reminder that the other libraries exist for you to switch to, however.
6. Sure.  We don't actually have that many settings there, though.
7. Maybe?  Is this something people frequently use?  I'm also wondering about the relevance of the system vs custom accent color options.

Chaphasilor:
4. While I know what you're getting at, it's a bit weird that the first item in the "quick settings" is a link to the full settings. Especially if you go into quick settings via the regular settings screen, where this becomes cyclic. But we can keep it if you prefer it that way.  
5. We can play around with that after things go stable, right. But I would like to improve it, since as you said the default is "all libraries" and if there's just one library we currently have a brief "flash" of the library picker, too. I think if we show this screen, it should rather be aimed at selecting the current/active library, since otherwise it's just ordered alphabetically and people might have to open the drawer to change the library right-away, which is confusing and could be handled during onboarding directly.  
6. Yeah I'm aware. How about just putting a link to the accessibility settings page there? Since I'd consider all a11y settings worthwhile, but they would clutter up the quick settings for users that have no need for them.  
7. No idea, I can just guess. But I'd wager it's a nice way to customize the Finamp experience from the start. Custom accent color is probably a bit more niche, I'm guessing people either use default colors or try to have all apps use the same theme.

Chaphasilor:
Made some tweaks, If those look fine to you, please go ahead and merge!

PR #1714, opened 2026-08-06
Stop iOS counting each track as played twice

APIUM:
This fix addresses incorrect lastfm scrobbling doing double scrobbles for tracks as reported in discord and in several tickets if I understand them correctly. I was able to reproduce this quite clearly on the sim, so the fix was fairly easy to verify.

I think the commit message communicates this pretty clearly:

On iOS the queue index advances before the position does, so the outgoing
track's end position was measured against the incoming track's runtime and read
as a repeat. Jellyfin logged a second play for a track that had just started,
and last.fm scrobbled it.

- A repeat now also requires the track to have been current long enough to
  reach its end. The stale position survives an unpredictable number of events,
  so no position test alone can tell a repeat from a track change.
- That test also means seeking forward and then looping the same track no
  longer records a listen.
- A missing runtime made the loop check always true, so a track of unknown
  length counted as a repeat on any rewind.

Relates to https://github.com/finamp-app/finamp/issues/1172, https://github.com/finamp-app/finamp/issues/1604, https://github.com/finamp-app/finamp/issues/688.

### DIFF
--- a/lib/services/playback_history_service.dart
+++ b/lib/services/playback_history_service.dart
@@ -32,7 +32,7 @@ class PlaybackHistoryService {
   final List<FinampHistoryItem> _history = []; // contains **all** items that have been played, including "next up"
   FinampHistoryItem? _currentTrack; // the currently playing track
 
-  PlaybackState? _previousPlaybackState;
+  ({PlaybackState state, FinampQueueItem? item})? _previousPlayback;
   DateTime _lastPositionUpdate = DateTime.now();
 
   FinampQueueItem?
@@ -75,7 +75,8 @@ class PlaybackHistoryService {
     });
 
     _audioService.playbackState.listen((event) {
-      final prevState = _previousPlaybackState;
+      final prevPlayback = _previousPlayback;
+      final prevState = prevPlayback?.state;
       final prevItem = _currentTrack?.item;
       final currentState = event;
       final currentIndex = currentState.queueIndex;
@@ -136,20 +137,27 @@ class PlaybackHistoryService {
         // handle seeking (changes updateTime (= last abnormal position change))
         else if (currentState.playing &&
             prevState != null &&
+            // the previous event's item, not the current one: a jump across a track boundary is not a seek
+            prevPlayback?.item?.id == currentItem.id &&
             // comparing the updateTime timestamps directly is unreliable, as they might just have different microsecond values
             // instead, compare the difference in milliseconds, with a small margin of error
             (currentState.position.inMilliseconds - prevState.position.inMilliseconds).abs() > 1500) {
           bool isSeekEvent = true;
 
           // detect rewinding & looping a single track
-          if (
-          // same track
-          prevItem?.id == currentItem.id &&
-              // current position is close to the beginning of the track
-              currentState.position.inMilliseconds <= 1000 * 10) {
-            if (prevState.position.inMilliseconds >= (prevItem?.item.duration?.inMilliseconds ?? 0) - 1000 * 10) {
+          if (currentState.position.inMilliseconds <= 1000 * 10) {
+            final duration = currentItem.item.duration;
+            // a track cannot have looped before it has played through
+            // elapsed time is required as well as position, which is unreliable across a track change
+            // scale by speed so that faster playback still reaches the end of the track
+            final playedFor = _currentTrack == null
+                ? null
+                : DateTime.now().difference(_currentTrack!.startTime) * currentState.speed;
+            if (duration != null &&
+                playedFor != null &&
+                playedFor >= duration - const Duration(seconds: 10) &&
+                prevState.position >= duration - const Duration(seconds: 10)) {
               // looping a single track
-              // last position was close to the end of the track
               updateCurrentTrack(currentItem, forceNewTrack: true); // add to playback history
               //TODO handle reporting track changes based on history changes, as that is more reliable
               onTrackChanged(currentItem, currentState, prevItem, prevState);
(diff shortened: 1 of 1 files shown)


lib/services/playback_history_service.dart
```diff
@@ -136,20 +137,27 @@ class PlaybackHistoryService {
...
-              currentState.position.inMilliseconds <= 1000 * 10) {
-            if (prevState.position.inMilliseconds >= (prevItem?.item.duration?.inMilliseconds ?? 0) - 1000 * 10) {
+          if (currentState.position.inMilliseconds <= 1000 * 10) {
+            final duration = currentItem.item.duration;
+            // a track cannot have looped before it has played through
+            // elapsed time is required as well as position, which is unreliable across a track change
+            // scale by speed so that faster playback still reaches the end of the track
+            final playedFor = _currentTrack == null
+                ? null
+                : DateTime.now().difference(_currentTrack!.startTime) * currentState.speed;
```
Chaphasilor, line 158: if this is meant to compensate for shorter play duration if playback speed is increased, then I don't think the calculation is correct. Multiplying the shorter duration by the current speed doesn't add up to the original duration, right?

lib/services/playback_history_service.dart
```diff
@@ -136,20 +137,27 @@ class PlaybackHistoryService {
...
+            if (duration != null &&
+                playedFor != null &&
+                playedFor >= duration - const Duration(seconds: 10) &&
```
Chaphasilor: what do you think about making this less aggressive and allowing 30s?

Chaphasilor:
@APIUM thanks for the PR!  
The description of your fix seems sound, and although it is unfortunate that now seek+loop doesn't count as a play anymore (I think it should), it's probably better to ensure that regular play scrobbling works well, and any such edge cases aren't as important.  

On Android everything seems to be working, aside from the seek (more than 10s) + loop which you mentioned.  
I've left some general comments above :)

lib/services/playback_history_service.dart
APIUM: Doesn't it? My thinking was:

---

Track length 4 mins (240s)
Speed 2x

At the end
playedFor = 120 * 2 = 240

---

Track length 4 mins (240s)
Speed 0.5x

At the end
playedFor = 480 * 0.5 = 240

---

This of course assumes the user hasn't been changing the speed mid track. I wouldn't expect they'd be doing this on a track by track basis so it's likely to only miss a single track if there's an error though.

The main reason we have this is really the loop check, checking has the user actually listened to most of this for it to be on loop and should be counted again, as the lastfm plugin does the actual check for if it should be counted

lib/services/playback_history_service.dart
APIUM: Fine with that, done. I also wasn't happy with it being a magic number in retrospect, so I've named them

lib/services/playback_history_service.dart
Chaphasilor: Okay yeah seems like I got that wrong. Confused it with discount calculation where "10% off" can't be inverted so easily.  
That adjusting speed midway through doesn't work is fine, but because of that I'd err on the side of counting a play rather than filtering it, so just increasing the tolerance for deviations.

lib/services/playback_history_service.dart
```diff
@@ -155,8 +159,8 @@ class PlaybackHistoryService {
                 : DateTime.now().difference(_currentTrack!.startTime) * currentState.speed;
             if (duration != null &&
                 playedFor != null &&
-                playedFor >= duration - const Duration(seconds: 10) &&
-                prevState.position >= duration - const Duration(seconds: 10)) {
+                playedFor >= duration - playedThroughSlack &&
```
Chaphasilor: we could calculate this as a percentage? 10% deviation of the total duration might make more sense than fixing this at 30s. We could keep the 30s as a minimum for shorter tracks.  
But just imagine a podcast episode or audio book where you skip ahead over some section, that would quickly not count as a play anymore.

But it's not super important since this is only about looping.

APIUM: Easy change, done

lib/services/playback_history_service.dart
```diff
@@ -302,6 +314,13 @@ class PlaybackHistoryService {
     return groupedHistory;
   }
 
+  /// How far short of the end a track can stop and still count as having played through.
+  static Duration _playedThroughSlack(Duration duration) {
+    const minimumSlack = Duration(seconds: 30);
+    final proportionalSlack = duration * 0.1;
+    return proportionalSlack > minimumSlack ? proportionalSlack : minimumSlack;
```
Chaphasilor, line 321: this could be `return max(proportionalSlack, minimumSlack);`

Or did you put it this way on purpose?

Chaphasilor (approved):

lib/services/playback_history_service.dart
APIUM: It's because it's a `Duration`. I could do `return Duration(microseconds: max(proportionalSlack.inMicroseconds, minimumSlack.inMicroseconds));` instead, would have to import `dart:math`. I do find the max() slightly more readable personally, but also don't like imports. Will leave for now but happy to change it if you feel any stronger

Chaphasilor: ah, max() doesn't support durations? that's a bummer. but yeah, let's keep it this way then :)

Chaphasilor:
Thanks for the PR! From the feedback I've read this should make a bunch of people happy 😁

PR #1715, opened 2026-08-06
Add FlatHub link to Readme

Narfinger:
## Changes
Add flathub to the official links.


Narfinger:
I think it is important to have all official distributions recognized in the readme. The flathub version seems to be an official distribution (according to the username/package name).
Otherwise people might wonder if it is a separate fork or not.

Chaphasilor:
@Narfinger thanks, yes we can link this (although the Readme is due for a refresh anyway). But what's with the removed line breaks? Could you add those back in please?

Narfinger:
Sorry, editor did that. Should be fixed now.

Chaphasilor:
Great! Thanks a ton!

PR #1717, opened 2026-08-07
Add universal search screen

obiwantoby:
Replace the in-tab classic search with a global search screen that queries the Jellyfin server across tracks, albums, artists, playlists and genres with a 300ms debounce.

  Because a plain search term only matches item names, matched artists are expanded into their albums and tracks (fetched separately so albums aren't buried behind tracks), so e.g. searching an artist's name also lists their discography. Results are de-duplicated and ordered by type.

  Tracks reuse TrackListTile; other item types use ItemWrapper so tapping opens the artist/album/playlist/genre screen. The music screen search icon now routes to this screen.

  Changes
  
  - New lib/screens/universal_search_screen.dart — a dedicated /search screen replacing the old in-tab search:
    - 300ms debounced, server-side search across Audio, MusicAlbum, MusicArtist, Playlist, MusicGenre
    - Matched artists are expanded into their full discography,  albums and tracks are fetched as separate requests, because a combined recursive MusicAlbum,Audio query buries the handful of albums behind hundreds of tracks (so they can fall outside the limit and never appear)
    - Results de-duplicated by id and ordered by type, so artists and albums surface above the long tail of tracks
    - Tracks reuse TrackListTile and play in place; artists/albums/playlists/genres use ItemWrapper for tap navigation and the long-press context menu
    - Offline guard plus loading and empty states
  - lib/main.dart to register the /search route
  - lib/screens/music_screen.dart , the music-screen search icon now pushes the universal search screen

  No new setting or Hive field: universal search fully replaces classic search, so there's nothing to toggle or maintain.

  Built and tested on the iOS simulator against a live Jellyfin server.

  Related Issues

  Split out from #1533 as requested — this PR contains only the universal search feature; the iOS 26 crash / Podfile / file_picker fixes will follow in separate PRs.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -30,6 +30,7 @@ import 'package:finamp/screens/playback_reporting_settings_screen.dart';
 import 'package:finamp/screens/player_settings_screen.dart';
 import 'package:finamp/screens/playlist_edit_screen.dart';
 import 'package:finamp/screens/queue_restore_screen.dart';
+import 'package:finamp/screens/universal_search_screen.dart';
 import 'package:finamp/services/album_image_provider.dart';
 import 'package:finamp/services/android_auto_helper.dart';
 import 'package:finamp/services/audio_service_smtc.dart';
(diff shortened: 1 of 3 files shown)


Chaphasilor:
@obiwantoby thanks for the PR!  
Please have another look at my comment in the old PR you linked, where I provided a mockup of which elements the search screen should contain and how it should be structured. Right now this PR just has a blank screen with no prompts and no filters, and the results are also really hard to differentiate and in no particular order. I think this should be extended to either group items by type (different sections for tracks, albums, artists, etc., possibly with a button to "see all" like on the artist/genre screen), or with a plaintext label indicating the type (e.g. how Spotify does it).  

Most components needed (filter chips, headers) should already exist in the code base, so please try to reuse those.

PR #1719, opened 2026-08-07
Add frosted glass to the now playing bar and drawer

obiwantoby:
**What this does**
  
  Gives the now playing bar and the navigation drawer a frosted glass treatment that fits the redesign.

  The now playing bar now renders the current track's album art, blurred and glazed behind frosted glass, so the cover's own colour shines through the panel. This replaces the previous single averaged accent colour, which tended to read muddy and made text contrast unpredictable. A theme aware glass base sits under the artwork so the bar stays a distinct, legible panel in both light and dark modes, and even over an all black cover it still reads as a lit pane rather than melting into the page.

  Text uses a fixed light or dark colour with a tight shadow so the title and artist stay crisp over any artwork, and the progress fill is now translucent so it layers over the glass as a subtle played tint instead of blocking the text where it overlaps.

  The navigation drawer gets a matching frosted panel that adapts to light and dark, so the two surfaces feel consistent.

  **Notes**

  Built entirely on Flutter's BackdropFilter and ImageFiltered, so it renders the same on Android, iOS, and desktop. There are no platform specific paths. Performance is a single small blur over the bar plus the drawer blur only while the drawer is open, so there is no meaningful battery or frame cost.
  
  **Design**

Built to feel native in both worlds from one code path: on iOS it reads as system glass, a live blur with the album art bleeding through like Apple's own now playing surfaces; on Android it lands as a Material translucent surface, its tint and album derived accent driven by the color scheme the way Material expects. All Flutter BackdropFilter and ImageFiltered, so there is no platform branch and the look stays consistent everywhere Finamp ships.

  Tested live on the iOS simulator across bright, pale, and pure black covers in both light and dark mode.

  **Screenshots**

Screenshot 2026-08-07 at 10 48 37 AM
Screenshot 2026-08-07 at 10 50 06 AM
Screenshot 2026-08-07 at 10 48 18 AM
Screenshot 2026-08-07 at 10 48 13 AM
Screenshot 2026-08-07 at 10 49 16 AM

### DIFF
--- a/lib/components/now_playing_bar.dart
+++ b/lib/components/now_playing_bar.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:io';
 import 'dart:math';
+import 'dart:ui' show ImageFilter;
 
 import 'package:audio_service/audio_service.dart';
 import 'package:finamp/color_schemes.g.dart';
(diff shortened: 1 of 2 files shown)


Chaphasilor:
@obiwantoby thanks for the PR, but I think this design change is currently out-of-scope for Finamp. My main gripe is that is seems to increase code complexity for those widgets further, just to deviate from the current aesthetic :)  
Maybe we can revisit this at a later point!

PR #1720, opened 2026-08-07
Fetch track credits on demand instead of in every item query

obiwantoby:
### Problem

On Jellyfin 12 (reported on RC4 https://github.com/finamp-app/finamp/issues/1707), opening playlists and browsing the library became several times slower than on 0.9.24. Downgrading Finamp restored the speed.

The cause is in the default item fields Finamp requests. `defaultFields` includes `People` (the cast/credits list), which is applied to every bulk item query. On the server, People live in a separate table from the item, so requesting the field forces a per item relational join across the whole result. Jellyfin 12 handles that join much more slowly than earlier servers, so the previously tolerable cost became the bottleneck. 0.9.24 did not request `People` in these queries, which is why the downgrade "fixes" it.

### Measurements

Same query against a live Jellyfin 12.0.0 server, only difference is whether `People` is requested:

| Request | Without People | With People |
| --- | --- | --- |
| 53 track playlist | 41 ms | 258 ms (~6x) |
| 500 track library page | 179 ms | 2361 ms (~13x) |

The cost scales with track count, so large playlists are the worst affected.

### Fix

Stop requesting `People` in bulk queries, and fetch it on demand only for the currently playing track, through the metadata provider. That single item request measured ~18 ms and still returns the credits, so the additional people chip on the player screen is unchanged. Playlists load at 0.9.24 speeds again.

Only the current track (and the few tracks the player precaches) trigger the extra request, and only when the additional people chip is enabled and the app is online.

### Notes

The server returns an empty `People` list rather than omitting the field, so the item alone can't tell us whether credits were loaded; the provider always fetches when online and the chip is enabled.

There is likely also a server side regression worth reporting upstream (RC4 handling this join notably slower than older servers), but the client should not be over fetching credits for a whole playlist to display them for one track regardless, so this change is worthwhile independent of any server fix.

### DIFF
--- a/lib/services/downloads_service_backend.dart
+++ b/lib/services/downloads_service_backend.dart
@@ -1244,7 +1244,7 @@ class DownloadsSyncService {
       }
       _metadataCache[id] = itemFetch.future;
       item = await _jellyfinApiData
-          .getItemByIdBatched(id, "${_jellyfinApiData.defaultFields},sortName,MediaSources")
+          .getItemByIdBatched(id, "${_jellyfinApiData.defaultFields},sortName,MediaSources,People")
           .then((value) => value == null ? null : DownloadStub.fromItem(item: value, type: type));
       _downloadsService.resetConnectionErrors();
       itemFetch.complete(item);
@@ -1275,7 +1275,7 @@ class DownloadsSyncService {
       case BaseItemDtoType.playlist || BaseItemDtoType.album:
         childType = DownloadItemType.track;
         childFilter = BaseItemDtoType.track;
-        fields = "${_jellyfinApiData.defaultFields},MediaSources,SortName";
+        fields = "${_jellyfinApiData.defaultFields},MediaSources,SortName,People";
         sortOrder = "ParentIndexNumber,IndexNumber,SortName";
       case BaseItemDtoType.artist || BaseItemDtoType.genre || BaseItemDtoType.library:
         childType = DownloadItemType.collection;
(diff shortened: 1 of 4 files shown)


Komodo5197:
The metadata provider changes look good, but I think we need some additional tweaks so that we still get populated People fields on downloaded items.

obiwantoby:
Ok minor change here. Tested with offline enabled and the people chip on. It should be exactly as it was before for downloaded items.

lib/services/downloads_service_backend.dart
```diff
@@ -1244,7 +1244,7 @@ class DownloadsSyncService {
       }
       _metadataCache[id] = itemFetch.future;
       item = await _jellyfinApiData
-          .getItemByIdBatched(id, "${_jellyfinApiData.defaultFields},sortName,MediaSources")
+          .getItemByIdBatched(id, "${_jellyfinApiData.defaultFields},sortName,MediaSources,People")
```
Chaphasilor, line 1247: I believe this change is not needed, as the call is only used to get collection items (but not their children)?

obiwantoby: Good catch on the naming, but _getCollectionInfo isn't collection-only. In syncDownload (line 929) it's called for any node where type.requiresItem, which  includes standalone track downloads (offline favorites,individually downloaded tracks). Those aren't fetched through getCollectionChildren, so without People here they'd lose credits offline. Album/playlist children still get People from getCollectionChildren. Maybe this should be renamed to  getItemInfo ?

Komodo5197: Nice catch.  I think I'll rename this on my own time, so no change should be needed here.

Komodo5197:
Okay, looks good to me.

PR #1723, opened 2026-08-09
Add macOS AirPlay Selector

obiwantoby:
**Summary**
The AirPlay route picker only worked on iOS because flutter_to_airplay has no macOS implementation, so the desktop app showed an “Only iOS is supported” placeholder.

This change adds a native AVRoutePickerView platform view for macOS and a shared Dart wrapper (AirPlayRoutePicker) so the output menu and lyrics screen show a real picker on both platforms. The macOS deployment target is raised to 10.15 for AVRoutePickerView.

Startup queue restore is also bounded with a timeout so a stalled fetch can no longer leave the player stuck on “Restoring queue” forever; on timeout it falls into the existing tap-to-retry state.

**Test plan**
macOS: Picker lists AirPlay outputs and switches correctly

iOS: Behavior unchanged

Non-Apple platforms: Renders nothing (no placeholder/crash)

Queue restore: Stalled fetch times out and shows tap-to-retry


Komodo5197:
I'm wondering if most of these changes would make more sense as a pull request/fork of flutter_to_airplay instead of directly embedding into finamp.

obiwantoby:
> I'm wondering if most of these changes would make more sense as a pull request/fork of flutter_to_airplay instead of directly embedding into finamp.

Good question, I did think of this, does that look actively maintained, some pull requests are from 2023? If upstream is stale, a PR there sits forever and Finamp can't ship macOS AirPlay. 

Perhaps a pragmatic middle path:
  
  1. Fork MrJai/flutter_to_airplay, add the macOS platform (Swift factory + macos/ dir + macos in the pubspec platforms + register in the plugin).
  2. Open a PR upstream.
  3. In Finamp, point the dependency at the fork via git until/if it merges — then switch to the released version.

It does seem to focus on mobile platforms, but if Finamp is to be a desktop app as well, (which is great) it is an awesome feature.

Let me know on that middle ground, I can attempt a PR there in the meantime, I had already worked on some of that before this.

I quite enjoy it.

image

obiwantoby:
Yeah, thought about it and you're right, it needed doing.

Moved the macOS picker into the package instead of embedding it here. The native AVRoutePickerView now lives in flutter_to_airplay, so this branch just points the dependency at a fork that adds macOS support and deletes the embedded factory from MainFlutterWindow.swift. Net change is about +14 / -109 lines.

Upstream PR: https://github.com/MrJai/flutter_to_airplay/pull/40

Pointed at the fork via a git ref for now since that repo has PRs open since 2023, so this can ship without waiting on a release. Can switch back to the hosted version once it merges.

lib/menus/output_menu.dart
```diff
@@ -165,15 +165,15 @@ class OutputMenuHeader extends ConsumerWidget {
...
+          if (AirPlayRoutePicker.isSupported)
             Padding(
               padding: const EdgeInsets.symmetric(horizontal: 8.0),
               child: AnimatedSwitcher(
                 duration: MediaQuery.disableAnimationsOf(context) ? Duration.zero : const Duration(milliseconds: 1000),
                 switchOutCurve: const Threshold(0.0),
                 child: Consumer(
                   builder: (context, ref, child) {
-                    return AirPlayRoutePickerView(
+                    return AirPlayRoutePicker(
```
Komodo5197: Shouldn't we be able to just use AirPlayRoutePickerView instead of building our own widget now?

pubspec.yaml
```diff
@@ -126,7 +126,10 @@ dependencies:
     git:
       url: https://github.com/Komodo5197/flutter-balanced-text.git
       ref: a6e93dad04f650734eaa6fa4d332800c7523b2a7
-  flutter_to_airplay: ^2.0.5
+  flutter_to_airplay:
+    git:
+      url: https://github.com/obiwantoby/flutter_to_airplay.git
+      ref: feat/macos-support
```
Komodo5197: I prefer using full commit hashes instead of branches for all forks we point at.

lib/menus/output_menu.dart
obiwantoby: Yep, done. Both call sites now use the package's AirPlayRoutePickerView directly and I removed the custom wrapper.

pubspec.yaml
obiwantoby: Done, pinned to the full commit hash (2d09182f60fe95529446ee50cc54340827895af9).

Chaphasilor:
@obiwantoby are you fine with us moving your repo into the finamp-app org when we merge this? Just to keep all forks in one place.

obiwantoby:
> @obiwantoby are you fine with us moving your repo into the finamp-app org when we merge this? Just to keep all forks in one place.

No issue there at all.

Chaphasilor:
We'll save this for the 1.1 release as a nice feature-drop, so that we can get 1.0 out in time :)

PR #1724, opened 2026-08-09
Cap release log level to reduce hitching

obiwantoby:
## What

Release builds set the root logger to `Level.ALL`, so every FINE, FINER, and FINEST record is emitted in production. Each record runs message censoring plus a synchronous file write on the main. Profiling a short session showed roughly 74 percent of log lines were this verbose trace (SyncBuffer alone emitted over 1400 lines), and that main isolate work shows up as UI or playback hitching.

This keeps `Level.ALL` in debug and caps release at INFO, so the trace never reaches the record listener in production. A Verbose Logging toggle on the logs screen restores full detail on demand for bug reports. It is persisted and applied at startup after settings load, so it survives a restart (needed to reproduce startup issues) and re-applies immediately when toggled.

Also moves the "Trusted Android user certs" log inside the Android only branch so it no longer prints on other platforms.

I am not entirely sure on the logging history here in general, but this seems like a nice low risk change here? 

## Changes

- `setup_logging.dart`: `_levelFor({verbose})` returns ALL in debug or when verbose is on, else INFO. `applyLogLevel()` applies the persisted preference from `main` after Hive is ready.
- New `verboseLogging` setting (default off) with a `VerboseLoggingSwitch` on the logs screen.
- Moved the Android cert log into the Android only path.

## Testing

- Built and launched on the iOS simulator and a signed macOS build; the toggle renders and persists across restarts. Debug builds still log everything.

Follows Flutter's [build modes guidance](https://docs.flutter.dev/testing/build-modes): keep expensive diagnostics in debug, not release.

Screenshot 2026-08-08 at 11 58 40 PM

### DIFF
--- a/lib/setup_logging.dart
+++ b/lib/setup_logging.dart
@@ -1,5 +1,6 @@
 import 'package:finamp/components/global_snackbar.dart';
 import 'package:finamp/services/censored_log.dart';
+import 'package:finamp/services/finamp_settings_helper.dart';
 import 'package:flutter/foundation.dart';
 import 'package:get_it/get_it.dart';
 import 'package:logging/logging.dart';
@@ -12,6 +13,8 @@ Future<void> setupLogging() async {
   final finampLogsHelper = GetIt.instance<FinampLogsHelper>();
   await finampLogsHelper.openLog();
 
+  // Hive isn't open yet, so keep everything for now. applyLogLevel picks up the
+  // persisted verboseLogging toggle once settings are available.
   Logger.root.level = Level.ALL;
   Logger.root.onRecord.listen((event) {
     // We don't want to print log messages from the Flutter logger since Flutter prints logs by itself
@@ -36,4 +39,15 @@ Future<void> setupLogging() async {
         false, // we can't fetch server info yet, because the user helper isn't set up yet. it's also faster to skip this here
   );
   startupLogger.info("\n${metadata.pretty}");
+}
+
+/// Applies the persisted verboseLogging preference. Call once settings are
+/// available, since [setupLogging] runs before the Hive box is open.
+///
+/// Debug builds keep everything. Release builds cap at INFO unless verbose
+/// logging is enabled, so the FINE/FINER/FINEST trace doesn't run censoring and
+/// a synchronous file write on the main isolate for every line in production.
+void applyLogLevel() {
+  final verbose = FinampSettingsHelper.finampSettings.verboseLogging;
+  Logger.root.level = (kDebugMode || verbose) ? Level.ALL : Level.INFO;
 }
--- a/lib/components/LogsScreen/verbose_logging_switch.dart
+++ b/lib/components/LogsScreen/verbose_logging_switch.dart
@@ -1,0 +1,24 @@
+import 'package:finamp/l10n/app_localizations.dart';
+import 'package:finamp/setup_logging.dart';
+import 'package:flutter/material.dart';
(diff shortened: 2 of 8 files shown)


lib/setup_logging.dart
```diff
@@ -12,7 +13,9 @@ Future<void> setupLogging() async {
   final finampLogsHelper = GetIt.instance<FinampLogsHelper>();
   await finampLogsHelper.openLog();
 
-  Logger.root.level = Level.ALL;
+  // Hive isn't open yet, so start from the build default. applyLogLevel picks
+  // up the persisted verboseLogging toggle once settings are available.
+  Logger.root.level = _levelFor(verbose: false);
```
Komodo5197: I'd just keep this at Level.all for all builds until we get hive opened.  But leave the comment about this being updated later.

lib/setup_logging.dart
```diff
@@ -37,3 +40,20 @@ Future<void> setupLogging() async {
   );
   startupLogger.info("\n${metadata.pretty}");
 }
+
+/// Debug builds keep everything. Release builds cap at INFO unless verbose
+/// logging is enabled, so the FINE/FINER/FINEST trace doesn't run censoring and
+/// a synchronous file write on the main isolate for every line in production.
+Level _levelFor({required bool verbose}) {
```
Komodo5197: I don't think this needs a helper function - we can just bake the logic into applyLogLevel.

Komodo5197:
Look good to me.

Chaphasilor:
I think this is fine to add, yes. Especially if it improves UI performance.  
We just need to keep it in mind for bug repros, since Just INFO-level traces probably won't be as useful.

Would it be feasible to keep the full logs around *without* sanitizing and persisting them to "disk" immediately? Maybe we could batch this to improve performance?  
This would allow us to keep the current workflow of people noticing a bug and being able to share it right-away. For that it might even be enough to just keep the current session logs in ram, and then sanitize them on-demand when they get shared?

Komodo5197:
I think the most common sort of issue where the logs are really helpful is actual exceptions, which this doesn't affect.  A lot of what's being cut is all the sync logger output, which I honestly feel isn't that great - I think I need to redo the logging there to record a different set of information, because the current stuff seems to be frequently useless.  I am currently considering a quick pass through everything to check that we're logging messages at an appropriate level.

obiwantoby:
Two main thoughts on this:

**Log Utility:** While debug logs are definitely valuable, how often are the current verbose logs actually helping us resolve edge cases versus just adding overhead? 

**First Impressions:** Finamp and Jellyfin are gaining a lot of momentum right now. Ensuring a smooth, performant first impression for new users should be a priority. Experience is critical for retaining new users.

As a returning user, I’m focusing on cleaning up these smaller items, optimizing logging feels like a low hanging win for overall responsiveness.

Maxr1998:
> I think the most common sort of issue where the logs are really helpful is actual exceptions, which this doesn't affect.  A lot of what's being cut is all the sync logger output, which I honestly feel isn't that great - I think I need to redo the logging there to record a different set of information, because the current stuff seems to be frequently useless.  I am currently considering a quick pass through everything to check that we're logging messages at an appropriate level.

Yeah, this has been bugging me as well for a while already, but I haven't had the chance to properly look into it. I frequently check the logs to check playback events and for failed network requests, and the huge number of sync traces makes the logs much harder to read. It might make sense to only log these events if there are problems, and not at all for successful syncs.

Thanks for looking into it!

PR #1726, opened 2026-08-10
Fix contributor role fallback in feature chip + more fixes

Chaphasilor:
should probably be merged after #1720


lib/components/PlayerScreen/album_chip.dart
```diff
@@ -141,11 +148,20 @@ class _AlbumChipContent extends StatelessWidget {
...
-              albumName,
-              overflow: TextOverflow.ellipsis,
-              softWrap: false,
-              style: TextStyle(color: color ?? Theme.of(context).textTheme.bodySmall!.color ?? Colors.white),
+            child: Row(
+              spacing: 2.0,
+              crossAxisAlignment: CrossAxisAlignment.center,
+              children: [
+                Text(
+                  "$albumName and me too",
```
Komodo5197: Did you mean to push this?

Komodo5197:
Hey, a bit unrelated to this, but if you're just throwing random fixes into this I was wondering - do you think it makes sense to change the default grid mode from all lists to the new grid for albums, list otherwise preset now that it's an option?  I feel like for albums specifically, grid mode does really make a lot of sense, and it also makes sure that both modes are shown off by default.

lib/components/MusicScreen/sort_and_filter_row.dart
```diff
@@ -136,10 +136,12 @@ class StaticSortAndFilterController extends SortAndFilterController {
 }
 
 class TrackingSortAndFilterController extends SortAndFilterController {
+  ContentType contentType;
```
Komodo5197: Why did you create this variable when the super class already stores this info?

lib/components/MusicScreen/sort_and_filter_row.dart
```diff
@@ -549,10 +559,11 @@ mixin _SortAndFilterMenuEntriesMixin<T extends ConsumerStatefulWidget> on Consum
     ];
   }
 
-  Widget _makeFilterTile(ItemFilterType option) {
+  Widget _makeFilterTile(ItemFilterType option, ContentType contentType) {
     return ToggleableListTile(
       title: ItemFilter(type: option).getName(context.l10n),
       leading: Padding(padding: const EdgeInsets.only(left: 16.0), child: Icon(option.icon)),
+      enabled: contentType != ContentType.inPlaylistOrAlbum,
```
Komodo5197: I think we should still allow the filters in these contexts, they should just shouldn't track/update the global setting.

lib/components/MusicScreen/sort_and_filter_row.dart
```diff
@@ -176,8 +181,13 @@ class TrackingSortAndFilterController extends SortAndFilterController {
...
       sortOrder: ref.watch(finampSettingsProvider.tabSortOrder(_type)),
-      favoriteFilter: ref.watch(finampSettingsProvider.onlyShowFavorites),
-      onlyShowFullyDownloadedFilter: ref.watch(finampSettingsProvider.onlyShowFullyDownloaded),
+      favoriteFilter: contentType == ContentType.inPlaylistOrAlbum
+          ? false
+          : ref.watch(finampSettingsProvider.onlyShowFavorites),
+      onlyShowFullyDownloadedFilter: contentType == ContentType.inPlaylistOrAlbum
+          ? false
+          : ref.watch(finampSettingsProvider.onlyShowFullyDownloaded),
+      filters: contentType == ContentType.inPlaylistOrAlbum ? {} : null,
```
Komodo5197: I don't think we should be suppressing all filters like this.

lib/components/MusicScreen/sort_and_filter_row.dart
```diff
@@ -62,8 +62,8 @@ abstract class SortAndFilterController {
   static ResolvedSortConfig resolveOffline(Ref ref, ContentType type, SortAndFilterConfiguration config) {
     final output = resolveOfflineWithoutFallback(ref, type, config);
     if (output != null) return output;
-    if (type == ContentType.inPlaylist) {
-      return ResolvedSortConfig._(config.copyWith(sortBy: SortBy.defaultOrder));
+    if (type == ContentType.inPlaylistOrAlbum) {
+      return ResolvedSortConfig._(config.copyWith(sortBy: SortBy.defaultOrder, filters: {}));
```
Komodo5197: I don't think we should be suppressing all filters like this.

lib/components/PlayerScreen/album_chip.dart
Chaphasilor: ah, yeah no. that was just an overflow test 😅

lib/components/MusicScreen/sort_and_filter_row.dart
Chaphasilor: ah right I can just access `super.contentType`, right?

Edit: found it, it's `_type`. Not super obvious where that's coming from. Thoughts on prefixing such things with `super` for clarify? The linter doesn't seem to mind.

Chaphasilor:
@Komodo5197 re: hybrid grid mode, I don't have a strong opinion about this. No idea how many people prefer grid vs. list, so I'm fine with giving that a try! I personally also like grid mode for playlists, since those are very similar to albums, but that can be tweaked later if needed.

I'm honestly a bit hazy on how the sort and filter classes and controllers work exactly, so I was happy to get this far. I agree allowing filters on a local level would be fine. So we should strip any filters coming from the global context, and then if filters are re-applied, we only apply them temporarily in the controller? I've tried to do this now.

lib/components/MusicScreen/sort_and_filter_row.dart
Komodo5197: I don't really have an opinion either way.

Komodo5197:
I though about adding grid playlists to album settings, but none of my playlists have curated covers like the albums, just the jellyfin auto-generated ones which I wouldn't consider particularly great, so I decided they could be left out.  Of course, none of this matters too much since you can just choose a fully custom setup[ now.  Anyway, for the filters I think your changes look good. My expectation was that we simply prevent reading/writing the setting in the relevant functions, so that the filters section continues operates as a normal local controller which only persists as long as you're on the relevant screen via the widget state.

Chaphasilor:
Okay great. Yeah the per-tab customization is really good stuff, thanks for that!  
The only user-facing change I'd still like to get into this PR is improving the weighting of the random item quick actions as discussed in https://discord.com/channels/1064474779043770408/1535363689736183901. Any thoughts on how best to do that?

The other planned changes (https://github.com/orgs/finamp-app/projects/3/views/1?filterQuery=-status%3ADone&sortedBy%5Bdirection%5D=desc&sortedBy%5BcolumnId%5D=Status) are just some more logging and fixing the release CI (although the latter might already be fixed and can be done after the release artifacts have been created).

If there's any change you want to get in before the stable release, now would be the time.

Komodo5197:
Regarding the random items quick action, we should probably be fine just applying a static weighing for the normal case, because I feel like basically all libraries are guaranteed to have reasonably balanced numbers of the types.  For the favorite case, we do have the 'sync all favorites statuses' setting enabled by default, so we might be able to leverage the download info from that to determine the item balances and generate reasonable weights before making the request.

I already put pretty much everything I wanted to get into the initial stable release into the quick actions PR.  At this point the thing I'm most concerned about are the occasional reports of hangs/crashes on startup with the iOS build that we never resolved, but there isn't really anything that I personally can do in that regard.  By the way, are you planning on getting all the changes from that list into stable?  If so, I'd like to note that the unknown artist chip was added intentionally because I felt just having it missing made the player screen look strange, and I'd prefer not to mess with android auto at all for this release because it's less predictable than the main app.

Chaphasilor:
I'll see if I can hook this up to the synced favorite data, that's a good idea. And yeah this was mainly about the random favorite button, since I think this has a high chance of being unbalanced.

About the list, only the things tagged as "Iteration 1" (until August 31st) is what I want to add. The rest was just random notes I took or features that were requested on Discord, for me to keep track of. The concrete implementation could definitely use some discussion first.  
The iOS hangs/crashes aren't ideal, but I also have no way of debugging that. Maybe the stable release will attract new contributors that are able to debug and fix it? :)

Chaphasilor:
@Komodo5197 took a shot at this now, but ran into the problem that the favorite items aren't actually downloaded from what I can tell, so we can only get the IDs, not the item types (unless the items are actually downloaded). So I ended up making requests to the server to get the amount of favorites per item type. Using the `GetItemCounts` API method also won't work since that can't filter by favorite status.  
Worst this should do is slow things down a bit, but I hope that's more tolerable than a ton of duplication.

Komodo5197:
Even with six requests, I'm still getting results within about two seconds, so I guess jellyfin handles the limit well enough and this method should be fine.  I'm still using my pi 3B server, so I can't imagine anyone else will be getting much slower results.

Komodo5197:
Alternatively, I could be getting good performance due to the low amount of favorites on my test account.  Regardless, this is probably our best option.

Chaphasilor:
Alright then, should be fine. Thanks for the cleanup. On Jellyfin 12 the performance shouldn't be any more of an issue than on 10.10, so I'm optimistic. And as long as things don't break fully, it should be fine.

PR #1730, opened 2026-08-15
Fix music search only returning results for the first typed letter

### LINKED ISSUE #1729: Search shows outdated query results when typing over slow network

### Type of Report

- [ ] The app is unusable
- [ ] Accessibility
- [ ] None of the above

0.6.27

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

_No response_

When typing "Foo" by hand in the screen keyboard, Finamp will trigger three queries to jellyfin: "F", "Fo", and "Foo". Then depending on the order in which the replies come from the server, the results displayed might not be the ones expected (for "Foo"). Since "F" is a much bigger result and it might take longer to arrive, it is likely to be the last one, so that one will overwrite any results that arrive earlier, showing all matching items for "F" instead of "Foo".

### Logs

I am skipping logs because I don't think this is needed for this bug.

### Additional Information

_No response_

csaavedra:
Typing a query like "Nirvana" in the music search bar returned results for just "N". Every keystroke called PagingController.refresh(), but infinite_scroll_pagination coalesces refreshes for the same page key: while a "page 0" request is in flight it issues no new request. As a result only the first keystroke's term was ever fetched, and its (first-letter) results were all that showed once the slow response arrived.

Debounce the search field so a single query is issued for the final term once the user pauses typing, instead of one request per keystroke.

As a safety net for burst typing (a pause mid-word can still start a request for a partial term while a later refresh gets coalesced away), _getPage now appends the response before checking staleness and, if the term changed while the request was in flight, triggers a refresh so the current term actually gets fetched. Appending first is required: the controller won't issue a new request until the in-flight one completes, so dropping the result would leave the list stuck on a spinner.

Fixes finamp-app/finamp#1729

### DIFF
--- a/lib/components/MusicScreen/music_screen_tab_view.dart
+++ b/lib/components/MusicScreen/music_screen_tab_view.dart
@@ -77,6 +77,10 @@ class _MusicScreenTabViewState extends State<MusicScreenTabView>
 
   // This function just lets us easily set stuff to the getItems call we want.
   Future<void> _getPage(int pageKey) async {
+    // Capture the term this request is being issued for, so that once the
+    // (slow) response comes back we can tell whether the user has since typed
+    // a different search.
+    final searchTerm = widget.searchTerm?.trim();
     try {
       final sortOrder =
           widget.sortOrder?.toString() ?? SortOrder.ascending.toString();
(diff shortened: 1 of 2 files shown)


Chaphasilor:
@csaavedra hey, thanks for the fix but this should already be fixed on the `redesign` branch / the beta version. Have you look at that yet? :)  
We're planning to release that to stable by the end of the month

csaavedra:
No, I haven't. I am reluctant to use "beta"-tagged software. I guess I will wait until the end of the month; I am using my own build with this fix in the meantime.

PR #1739, opened 2026-08-22
Add "Advanced" login options screen and move mTLS setup there

Maxr1998:
## Changes
The previous "Client certificate required" alert & setup button is replaced with a new "Advanced" screen, where you can also set up mTLS if Finamp doesn't detect the server using it.
The alert remains, but opens the advanced settings instead of directly prompting you to pick a certificate.

image image

This also tweaks the strings a bit to clarify that mTLS is optional and not required to have a secure connection ("authentication" → "additional authentication").
I also fixed the bottom controls getting overlapped by the hardware keyboard pop-up, please let me know if I should move that to a separate PR.

## Related Issues
Related to and supersedes #1684.

### DIFF
--- a/lib/screens/advanced_login_options_screen.dart
+++ b/lib/screens/advanced_login_options_screen.dart
@@ -21,17 +21,19 @@ class AdvancedLoginOptionsScreen extends ConsumerWidget {
       body: ListView(
         padding: const EdgeInsets.only(bottom: 200.0),
         children: [
-          if (ClientCertificateInstaller.isSupported)
-            ListTile(
-              leading: Icon(TablerIcons.certificate),
-              title: Text(AppLocalizations.of(context)!.clientCertificate),
-              subtitle: Text(
-                ref.watch(finampSettingsProvider.clientCertificate) != null
-                    ? AppLocalizations.of(context)!.clientCertificateInstalled
-                    : AppLocalizations.of(context)!.clientCertificateUnavailable,
-              ),
-              onTap: () => showClientCertificateMenu(context: context),
+          ListTile(
+            leading: Icon(TablerIcons.certificate),
+            title: Text(AppLocalizations.of(context)!.clientCertificate),
+            subtitle: Text(
+              !ClientCertificateInstaller.isSupported
+                  ? AppLocalizations.of(context)!.clientCertificatesUnsupported
+                  : ref.watch(finampSettingsProvider.clientCertificate) != null
+                  ? AppLocalizations.of(context)!.clientCertificateInstalled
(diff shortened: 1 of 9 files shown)


lib/screens/advanced_login_options_screen.dart
```diff
@@ -0,0 +1,49 @@
...
+  static const routeName = "/login/advanced";
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return Scaffold(
+      appBar: AppBar(title: Text(AppLocalizations.of(context)!.advanced), leading: FinampAppBarBackButton()),
+      body: ListView(
+        padding: const EdgeInsets.only(bottom: 200.0),
+        children: [
+          if (ClientCertificateInstaller.isSupported)
```
Chaphasilor: we show the warning regardless of if installing the cert is supported or not, so we should probably show some kind of message here if someone gets the warning and opens the screen but the setting is not available?  
Disabling the setting with some message about why might be a better approach then fully hiding it?

lib/screens/advanced_login_options_screen.dart
```diff
@@ -0,0 +1,49 @@
...
+            ListTile(
+              leading: Icon(TablerIcons.certificate),
+              title: Text(AppLocalizations.of(context)!.clientCertificate),
+              subtitle: Text(
+                ref.watch(finampSettingsProvider.clientCertificate) != null
+                    ? AppLocalizations.of(context)!.clientCertificateInstalled
+                    : AppLocalizations.of(context)!.clientCertificateUnavailable,
+              ),
+              onTap: () => showClientCertificateMenu(context: context),
```
Chaphasilor: I think you're missing the "onImported" callback to re-scan for local servers

lib/screens/advanced_login_options_screen.dart
Maxr1998: Good point, this is what shows on unsupported platforms now:

lib/screens/advanced_login_options_screen.dart
Maxr1998: Good catch, I fixed that and also cleaned up some now unused code.

Maxr1998: (I also tested the new approach locally)

lib/screens/advanced_login_options_screen.dart
Maxr1998: Oh, the warning only shows on supported platforms already. Maybe that check should be removed then?

Maxr1998: Removed the check now, made it more robust and also tweaked the warning message:

PR #1740, opened 2026-08-22
Raise iOS deployment target to 15.0

martin-77:
## Summary

Raise Finamp's minimum iOS deployment target from 14.0 to 15.0.

Current Flutter/Xcode tooling automatically upgrades the iOS project to a minimum deployment target of 15.0. Finamp's existing iOS configuration still pins the Runner and all CocoaPods targets to 14.0, which causes builds with current Xcode versions to fail.

## Changes

- update `ios/Podfile` platform from iOS 14.0 to 15.0
- update the CocoaPods `IPHONEOS_DEPLOYMENT_TARGET` override to 15.0
- update Runner project deployment targets to 15.0
- update `ios/Flutter/AppFrameworkInfo.plist` minimum OS version to 15.0

## Reproduction

With current Flutter/Xcode tooling, the existing configuration fails with:

```text
Target Integrity (Xcode): The iOS deployment target
'IPHONEOS_DEPLOYMENT_TARGET' is set to 14.0, but the range of supported
deployment target versions is 15.0 to ...


UnicornsOnLSD (approved):
Changes look good to me, it's a shame that we have to do this but oh well

PR #1741, opened 2026-08-22
Finish analyzer and Dart API cleanup

martin-77:
## Summary

This PR continues the analyzer cleanup after #1738 and groups the remaining follow-up work into one reviewable change set.

It includes:
- additional analyzer warning and strict-inference cleanup
- straightforward Dart/API deprecation migrations
- `RadioGroup`, `onReorderItem`, and Navigator pop API migrations
- safer async / `BuildContext` handling where required by the analyzer
- typed Chopper responses and callback/error handlers
- removal of remaining avoidable private package imports
- a source-level Russian localization fix for the duplicated `genre` select case (`disc` was intended)
- analyzer exclusions for generated `.g.dart` and generated localization output instead of modifying generated files
- file-name / identifier cleanup for analyzer naming diagnostics

Generated `.g.dart` files are intentionally not modified directly.

## Relation to #1738

This branch contains the follow-up analyzer cleanup and is based on the same original `redesign` state as #1738 rather than being stacked on top of that PR.

Because of that, the branch by itself still contains the warnings already fixed in #1738.

After #1738 is merged, this branch can be rebased onto the updated `redesign` branch.

## Validation

The combined cleanup state including #1738 was tested locally:

- `flutter analyze`: no issues found
- `flutter gen-l10n`: succeeds
- Russian `confirmAddToNextUp` generates separate `genre` and `disc` cases
- changed handwritten Dart files are formatted
- `git diff --check`: clean

Generated code is excluded from analyzer lint diagnostics rather than being edited manually.

The repository currently has no `test/` directory, so `flutter test` is not applicable.


lib/menus/components/playbackActions/playback_actions.dart
```diff
@@ -599,7 +598,7 @@ class ShuffleAlbumsPlaybackAction extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final queueService = GetIt.instance<QueueService>();
-    assert(item is Genre || item is Artist || (item is FinampSortable<Album> && item is FinampPlayable));
```
Komodo5197, line 602: I'd prefer to keep these checks in place, even if they're guaranteed with the currently implemented types.

lib/components/PlayerScreen/queue_list.dart
```diff
@@ -431,15 +431,16 @@ class _PreviousTracksListState extends State<PreviousTracksList> with TickerProv
             return SliverReorderableList(
               proxyDecorator: (widget, _, _) => Material(type: MaterialType.transparency, child: widget),
               autoScrollerVelocityScalar: 20.0,
-              onReorder: (oldIndex, newIndex) {
+              onReorderItem: (oldIndex, newIndex) {
+                final serviceNewIndex = oldIndex < newIndex ? newIndex + 1 : newIndex;
```
Komodo5197: Has this been tested?  I thought onReorderItem was expected to generally remove this sort of offset.

lib/services/jellyfin_api_helper.dart
```diff
@@ -1171,7 +1171,7 @@ class JellyfinApiHelper {
     List<String> builtPath = List<String>.from(parsedBaseUrl.pathSegments);
     builtPath.addAll(["Items", item.imageId!, "Images", "Primary"]);
     final Map<String, dynamic> queryParams = {
-      if (format != null) "format": format,
+      "format": ?format,
```
Komodo5197: Have you run build-runner with these changes?  I believe last time I saw, it couldn't handle files with this syntax.

lib/services/music_player_background_task.dart
```diff
@@ -933,19 +939,25 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
...
     } catch (e) {
       _audioServiceBackgroundTaskLogger.severe(e);
-      return Future.error(e);
+      rethrow;
     }
   }
 
   @override
   Future<void> setRepeatMode(AudioServiceRepeatMode repeatMode) async {
+    if (repeatMode != AudioServiceRepeatMode.all &&
```
Komodo5197: I feel like pulling the errors out in these blocks just makes the logic more redundant.

lib/services/music_player_background_task.dart
```diff
@@ -1119,6 +1131,10 @@ class MusicPlayerBackgroundTask extends BaseAudioHandler with SeekHandler, Queue
       _audioServiceBackgroundTaskLogger.severe("Custom action '$name' not found.", e);
     }
 
+    if (actionFuture != null) {
```
Komodo5197: I'm not sure actionFuture really makes this more understandable?  I did notice a minor logic error in the existing code where actions like toggleRadio seem to be falling through super.customAction even though we're parsing them ourselves.

lib/menus/components/playbackActions/playback_actions.dart
martin-77: Makes sense. The FinampPlayable check is currently redundant, but keeping it as an explicit invariant is reasonable. I’ll restore it.

lib/components/PlayerScreen/queue_list.dart
martin-77: Good catch. onReorderItem already provides the corrected insertion index, so my additional +1 adjustment is wrong. I’ll remove it from all three handlers.

lib/services/jellyfin_api_helper.dart
martin-77: Not on the current branch. CI is currently failing the code-generation verification. I’ll revert this to the existing syntax.

lib/services/music_player_background_task.dart
martin-77: Agreed. Moving the checks out of the switch makes the logic more redundant. I’ll move them back, but keep rethrow so the original stack trace is preserved.

lib/services/music_player_background_task.dart
martin-77: Agreed. actionFuture doesn’t really improve this. You’re also right about the existing fall-through to super.customAction(). I’ll use explicit returns for handled actions instead.

Issue #1759, opened 2026-09-03
Finamp not installable on Android 17 through Play Store


MorphBonehunter:
### Are you using the newest Version?

- [x] Yes, the issue happens on the newest version
- [x] No, there aren't any open issues about this. I checked!

### Type of Report

- [x] The app is unusable
- [ ] Something isn't quite right
- [ ] Some functionality isn't accessible to me
- [ ] Accessibility
- [ ] None of the above

### The exact version of the app

unrelated 

### Affected Devices/Platform

- [x] Android
- [ ] Android Auto
- [ ] iOS
- [ ] Linux
- [ ] MacOS (iOS version)
- [ ] MacOS (native)
- [ ] Windows

### Device

Google Pixel 7a

### Description & Steps to Reproduce

At the moment it's not possible to install Finamp trough the Google Play Store on Android 17.
This is also true after joining the Beta Program.
(Screenshot is.only in German)

<img width="1080" height="1022" alt="Image" src="https://github.com/user-attachments/assets/b17f8f52-66bf-4dea-98b7-a5a51028a947" />

### Logs

<details>

  <summary>Logs</summary>
  
  PLACE YOUR *FULL* LOGS HERE, or attach them as a file
  you may censor any private information (although Finamp tries to remove that automatically), just please don't edit out larger sections.
  
</details> 

### Additional Information

I'm still able to install Finamp trough side loading and app runs without problems.

Chaphasilor:
@MorphBonehunter signing up for the beta can take 30 minutes or even longer sometimes. Have you waited that long?  
We're aware the old stable version currently can't be installed, but expect to release the current beta version to stable in the coming days

MorphBonehunter:
Hey @Chaphasilor, yes I've waited and as the screenshot shows the App Title changes from Finamp -> Finamp (Beta) and also the part below the exclamation mark shows "you are Betatester now.".

I'm do not know how the target android sdk stuff works but I read that it could be that the lowest version in any channel defines the compatible.
Ok, so I will wait for the stable release and will report back. 👍

paulstimpson:
Pixel 8 Pro here. Play store says I am a member of the beta but that I can't install it on my phone because Finamp was built for an old version of Android. It appears the Pixels with Android 17 aren't marked as a valid install combo

anto294:
Same warning on S26U.
OneUI 8.5 Android 16

YawningLyon:
How many days are we talking here?

[...]

PR #1769, opened 2026-09-13
Automated signed CI builds

UnicornsOnLSD:
This PR adds a new GH Actions step to create signed binaries on a tag. The end goal here is to be able to make Finamp releases autonomously without relying on me to make a signed build manually (to those snooping, this has delayed releases by a staggering amount of time :melting_face:)

# iOS

For iOS, I've mostly copied from [the GitHub docs](https://docs.github.com/en/actions/how-tos/deploy/deploy-to-third-party-platforms/sign-xcode-applications), which conveniently has a section on importing provisioning profiles/certificates for use in runners. I've made a distribution profile for CI, which is all pulled from GH Actions secrets. There were some shenanigans required to use this profile in the build (`flutter build ipa --release --export-method app-store` doesn't work >:( ), but the result is clean enough and allows the Xcode project to stay in "automatic" mode by default (rather than having its signing settings forced to CI all the time (not that it really matters, you have to mess with it anyway unless you're me)).

# Android

Still TODO, initially I was just going to bother with iOS since it's usually the most painful roadblock in getting releases out but we may as well get it done too. Theoretically this'll involve putting my Android signing key into GH Actions, will have a proper look on how to do it properly. From here we could both upload a bundle to Google Play and have a signed APK for those who get Finamp via GitHub Releases.

# Future Scope

I've got this set to run on tags for now, but we might be able to set things up to run on every commit to `redesign` (or the main branch in the future). We'll probably run into issues with build numbers there though, so for now it makes sense to only make signed releases when actually making a release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,8 +16,7 @@ on:
       - redesign
 
 env:
-  if_condition: &integration_if github.event_name == 'workflow_dispatch' ||
-    (github.event_name == 'push' && github.repository == 'finamp-app/finamp' )
+  if_condition: &integration_if false
 
 # Due to a flutter bug, flutter build is not generating localizations
 # workaround by always running flutter pub get immediately before building,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -199,6 +199,109 @@ jobs:
       if: *integration_if
       run: flutter test integration_test/test_cases.dart -d 'iPhone 17'
 
+  release-ios:
+    name: Release iOS to App Store Connect
+    if: >-
+      github.event_name == 'push' &&
+      startsWith(github.ref, 'refs/tags/') &&
+      github.repository == 'finamp-app/finamp'
+    needs: build-ios
+    runs-on: macos-latest
+    environment: app-store-connect
+    concurrency:
+      group: app-store-connect-${{ github.ref }}
+      cancel-in-progress: false
+    steps:
+    - uses: actions/checkout@v6
+
+    - uses: maxim-lobanov/setup-xcode@v1
+      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -15,13 +15,8 @@ on:
...
       - redesign
 
-# set permission so that actions can modify repo contents, which is needed to add release assets
-permissions:
-  contents: write
-
 env:
-  if_condition: &integration_if github.event_name == 'workflow_dispatch' ||
-    (github.event_name == 'push' && github.repository == 'finamp-app/finamp' )
+  if_condition: &integration_if false
```
UnicornsOnLSD, line 19: TODO: revert this

ciklois-neo:
I'd like to help with the Android part of this work. I'm preparing a small follow-up against `redesign` that builds signed APK/AAB artifacts from release tags, verifies their signatures, and documents the required environment secrets. Publishing to Google Play would remain a separate maintainer-controlled step initially, so this can be reviewed without anyone sharing production credentials.

Before wiring in publication, could you confirm whether the GitHub APK signing key and the Play upload key are the same, and whether someone is already working on the Android portion? I'll adjust the scope if this overlaps existing work.

The installability reports in #1759 prompted this: the official 1.0.1-beta GitHub APK has versionCode 201 and targetSdkVersion 36 (its SHA-256 matches the release asset digest). I haven't inspected the Play-delivered artifact or Play Console settings, so I don't assume that build automation alone resolves the store availability issue.

The signing helper's tests already round-trip special-character passwords through Java's actual Properties parser; I'm now validating a full build with a disposable key. Implementation and review are AI-assisted; I'll include the actual validation results and limitations with the contribution.

